PR #30, opened 2014-08-08
simplify base64 with inline lookup tables

technosaurus:
changes compile to a single mov instruction
the '~' in the decode helper is just filler everything else maps to octal escaped numeric value
the base64 string in the encoder has a useless \0 terminator, but to get the string without it looks really ugly
note: to save code size vs speed, the decode helper could be replace with  `strchr(base64string,x) - base64string` where base64string is the same one used in the decoder (should make it a static const char[] then)
Also, it looks like no error checking is done to ensure that the decode input is actually base64, I can trivially add this if that is the case.

No need to add me to the Authors list, consider the patch in its entirety as Public Domain so there is no need to contact me for any license changes - use it however you wish.
- Brad Conroy.

### DIFF
--- a/src/duk_api_codec.c
+++ b/src/duk_api_codec.c
@@ -44,19 +44,10 @@ static void duk__base64_encode_helper(const duk_uint8_t *src, const duk_uint8_t 
 			/* A straightforward 64-byte lookup would be faster
 			 * and cleaner, but this is shorter.
 			 */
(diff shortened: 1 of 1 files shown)


svaarala:
#465 now has a table-based base-64 encoder/decoder which has no safety issues and is equivalent to the slow path.

PR #278, opened 2015-08-13
Add Cmake Builder

kphillisjr:
This includes the basic implementation for CMake to build duktape from git master using Windows and MSVC 2013 Community edition.

Also, some examples are modified to improve windows support.


kphillisjr:
I probably should mention the build requirements when building directly from the git tree when on windows.
* Visual Studio ( https://www.visualstudio.com/ ) - I tested against the 2013 Community Edition, but the 2015 Community Edition should work just fine.
* CMAKE version 3.x ( http://www.cmake.org/ )
* Python ( https://www.python.org/ ) - I used Python v2.7 for this branch.
* PyYAML ( http://pyyaml.org/ ) - I used version 3.11.
* TortoiseGit ( https://tortoisegit.org/ )
* Git for Windows ( https://msysgit.github.io/ )

svaarala:
This is very out of date - there's now support for doing dist on Windows with quite minimal dependencies (Python 2 and YAML) so I'm closing this.

Issue #265, opened 2015-08-08
Use direct bytecode PC pointer rather than PC index for faster execution


svaarala:
Current bytecode execution loop maintains current PC as an index in the current activation. Compared to storing a direct bytecode pointer in the current activation there's one extra lookup for each opcode dispatch. Since the bytecode has a stable pointer, change to use a direct bytecode pointer for better performance.

Preliminary testing indicates the performance benefit is around 20% for code doing e.g. looping and arithmetic (heavy on dispatches).

- [x] Check normal test runs (API, Ecmascript) and test262 with assertions
- [x] Check debugger functionality, especially: line numbers in Status, call stack line numbers

svaarala:
Rough comparison of the draft over master:

```
test-array-read.js            : duk-O2-pc  4.75 duk-O2-master  6.26
test-array-write.js           : duk-O2-pc  5.78 duk-O2-master  6.82
test-bitwise-ops.js           : duk-O2-pc  1.02 duk-O2-master  1.25
test-buffer-nodejs-read.js    : duk-O2-pc  5.53 duk-O2-master  6.76
test-buffer-nodejs-write.js   : duk-O2-pc  6.55 duk-O2-master  7.64
test-buffer-object-read.js    : duk-O2-pc  5.44 duk-O2-master  6.73
test-buffer-object-write.js   : duk-O2-pc  6.54 duk-O2-master  7.74
test-buffer-plain-read.js     : duk-O2-pc  4.77 duk-O2-master  5.98
test-buffer-plain-write.js    : duk-O2-pc  4.97 duk-O2-master  5.96
test-call-basic.js            : duk-O2-pc 19.10 duk-O2-master 18.26
test-compile-mandel-nofrac.js : duk-O2-pc 15.28 duk-O2-master 15.05
test-compile-mandel.js        : duk-O2-pc 18.58 duk-O2-master 19.11
test-compile-short.js         : duk-O2-pc 11.65 duk-O2-master 11.69
test-empty-loop.js            : duk-O2-pc  2.74 duk-O2-master  3.40
test-fib.js                   : duk-O2-pc 10.63 duk-O2-master 10.43
test-hello-world.js           : duk-O2-pc  0.00 duk-O2-master  0.00
test-hex-decode.js            : duk-O2-pc  6.01 duk-O2-master  6.02
test-hex-encode.js            : duk-O2-pc 30.54 duk-O2-master 31.08
[...]

PR #286, opened 2015-08-20
Implement debugger halt-on-throw

fatcerberus:
When the debugger is attached and an error is thrown with no active catch site, either a runtime error thrown by Duktape or directly by Ecmascript code, the debug client is notified and execution is paused at the error site to allow the developer to view program state.  This is useful, as this information will be lost once the stack is unwound.

Implementation checklist:
- [x] Implement the basic logic for debugger error intercept.
- [x] Add a Throw notification to alert the client to the error.
- [x] Ensure yield/resume awareness to avoid spurious pauses.
- [x] Fix C stack overflow when an error is thrown at max recursion (double error pileup).
- [x] Fix crashes when intercepting errors originating from the compiler (fixed by not intercepting them).
- [x] Fix an assert failure when execution is stopped inside a Duktape/C call.
- [x] Finalize changes and ensure consistency with Duktape codebase.

### DIFF
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -1292,6 +1292,31 @@ Example::
 
 Logger output redirected from Duktape logger calls.
 
+Throw notification (0x05)
+-------------------------
+
+Format::
+
+    NFY <int: 5> <int: fatal> <str: msg> <str: filename> <int: linenumber> EOM
+
+Example::
+
+    NFY 5 1 "ReferenceError: identifier not defined" "pig.js" 812 EOM
+
+Fatal is one of:
+
+* 0x00: caught
+* 0x01: fatal (uncaught)
+
+Duktape sends a Throw notification whenever an error is thrown, either by
+Duktape due to a runtime error or directly by Ecmascript code.
+
+msg is the string-coerced value being thrown. Filename and line number are
+taken directly from the thrown object if it is an Error instance (after
+augmentation), otherwise these values are calculated from the bytecode
+executor state.
+
+
 Commands sent by debug client
 =============================
 
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -1316,7 +1316,6 @@ taken directly from the thrown object if it is an Error instance (after
 augmentation), otherwise these values are calculated from the bytecode
 executor state.
 
-
 Commands sent by debug client
 =============================
 
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -847,6 +847,51 @@ DUK_INTERNAL void duk_debug_send_status(duk_hthread *thr) {
 	duk_debug_write_eom(thr);
 }
 
+DUK_INTERNAL void duk_debug_send_throw(duk_hthread *thr, duk_bool_t fatal) {
+	/*
+	 * NFY <int: 5> <int: fatal> <str: msg> <str: filename> <int: linenumber> EOM
+	 */
+
+	duk_context *ctx = (duk_context *)thr;
+	duk_activation *act;
+	duk_uint32_t pc;
+
+	DUK_ASSERT(thr->valstack_top > thr->valstack);  /* ... [err] */
+
+	duk_debug_write_notify(thr, DUK_DBG_CMD_THROW);
+	duk_debug_write_int(thr, fatal);
+
+	/* Report thrown value to client coerced to string */
+	duk_dup(ctx, -1);
+	duk_safe_to_string(ctx, -1);
+	duk_debug_write_hstring(thr, duk_require_hstring(ctx, -1));
+	duk_pop(ctx);
+
+	if (duk_is_error(ctx, -1)) {
+		/* Error instance, use augmented error data directly */
+		duk_get_prop_stridx(ctx, -1, DUK_STRIDX_FILE_NAME);
+		duk_safe_to_string(ctx, -1);
+		duk_debug_write_hstring(thr, duk_require_hstring(ctx, -1));
+		duk_get_prop_stridx(ctx, -2, DUK_STRIDX_LINE_NUMBER);
+		duk_debug_write_uint(thr, duk_get_uint(ctx, -1));
+		duk_pop_2(ctx);
+	} else {
+		/* For anything other than an Error instance, we calculate the error
+		 * location directly from the current activation.
+		 */
+		act = thr->callstack + thr->callstack_top - 1;
+		duk_push_hobject(ctx, act->func);
+		duk_get_prop_string(ctx, -1, "fileName");
+		duk_safe_to_string(ctx, -1);
+		duk_debug_write_hstring(thr, duk_require_hstring(ctx, -1));
+		pc = duk_hthread_get_act_prev_pc(thr, act);
+		duk_debug_write_uint(thr, (duk_uint32_t) duk_hobject_pc2line_query(ctx, -2, pc));
+		duk_pop_2(ctx);
+	}
+
+	duk_debug_write_eom(thr);
+}
+
 /*
  *  Debug message processing
  */
@@ -1637,6 +1682,55 @@ DUK_LOCAL void duk__debug_process_message(duk_hthread *thr) {
  fail:
 	DUK__SET_CONN_BROKEN(thr);
 	return;
+}
+
+/* Halt execution and enter a debugger message loop until execution is resumed
+ * by the client. PC for the current activation may be temporarily decremented
+ * so that the "current" instruction will be shown by the client.
+ */
+
+DUK_INTERNAL void duk_debug_halt_execution(duk_hthread *thr, duk_bool_t use_prev_pc) {
+	duk_activation *act;
+	duk_hcompiledfunction *fun;
+	duk_instr_t *old_pc;
+
+	DUK_HEAP_SET_PAUSED(thr->heap);
+
+	act = duk_hthread_get_current_activation(thr);
+
+	/* NOTE: act may be NULL if an error is thrown outside of any activation, which may
+	 * happen in the case of, e.g. syntax errors.
+	 */
+
+	/* Decrement PC if that was requested, this requires a PC sync */
+	if (act != NULL) {
+		duk_hthread_sync_currpc(thr);
+		old_pc = act->curr_pc;
+		fun = (duk_hcompiledfunction *)DUK_ACT_GET_FUNC(act);
+		if (use_prev_pc && act->curr_pc != NULL &&
+			act->curr_pc > DUK_HCOMPILEDFUNCTION_GET_CODE_BASE(thr->heap, fun)) {
+			act->curr_pc--;
+		}
+	}
+
+	/* Process debug messages until we are no longer paused. */
+
+	/* NOTE: This is a bit fragile. It's important to ensure that neither
+	 * duk_debug_send_status() or duk_debug_process_messages() throws an error
+	 * or act->curr_pc will never be reset.
+	 */
+
+	thr->heap->dbg_processing = 1;
+	duk_debug_send_status(thr);
+	while (thr->heap->dbg_paused) {
+		DUK_ASSERT(thr->heap->dbg_processing);
+		duk_debug_process_messages(thr, 0 /* no_block */);
+	}
+	thr->heap->dbg_processing = 0;
+
+	if (act != NULL) {
+		act->curr_pc = old_pc;  /* restore PC */
+	}
 }
 
 DUK_INTERNAL duk_bool_t duk_debug_process_messages(duk_hthread *thr, duk_bool_t no_block) {
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -880,7 +880,7 @@ DUK_INTERNAL void duk_debug_send_throw(duk_hthread *thr, duk_bool_t fatal) {
 		 * location directly from the current activation.
 		 */
 		act = thr->callstack + thr->callstack_top - 1;
-		duk_push_hobject(ctx, act->func);
+		duk_push_tval(ctx, &act->tv_func);
 		duk_get_prop_string(ctx, -1, "fileName");
 		duk_safe_to_string(ctx, -1);
 		duk_debug_write_hstring(thr, duk_require_hstring(ctx, -1));
--- a/src/duk_error_misc.c
+++ b/src/duk_error_misc.c
@@ -5,6 +5,35 @@
 #include "duk_internal.h"
 
 /*
+ * Helper to walk the thread chain and see if there is an active error
+ * catcher. Protected calls aren't considered.
+ */
+
+DUK_LOCAL duk_bool_t duk__have_active_catcher(duk_hthread *thr) {
+	/*
+	 * XXX: As noted above, a protected API call won't be counted as a
+	 * catcher. This is usually convenient, e.g. in the case of a top-
+	 * level duk_pcall(), but may not always be desirable. Perhaps add an
+	 * argument to treat them as catchers?
+	 */
+
+	duk_size_t i;
+
+	DUK_ASSERT(thr != NULL);
+
+	while (thr != NULL) {
+		for (i = 0; i < thr->catchstack_top; i++) {
+			duk_catcher *cat = thr->catchstack + i;
+			if (DUK_CAT_HAS_CATCH_ENABLED(cat)) {
+				return 1;  /* all we need to know */
+			}
+		}
+		thr = thr->resumer;
+	}
+	return 0;
+}
+
+/*
  *  Get prototype object for an integer error code.
  */
 
@@ -41,6 +70,49 @@ DUK_INTERNAL duk_hobject *duk_error_prototype_from_code(duk_hthread *thr, duk_er
 
 DUK_INTERNAL void duk_err_setup_heap_ljstate(duk_hthread *thr, duk_small_int_t lj_type) {
 	duk_tval tv_tmp;
+
+#if defined(DUK_USE_DEBUGGER_SUPPORT)
+
+	/* If something is thrown with the debugger attached and nobody will
+	 * catch it, execution is paused before the longjmp, turning over
+	 * control to the debug client. This allows local state to be examined
+	 * before the stack is unwound.
+	 */
+
+	/* TODO: Allow customizing the pause and notify behavior */
+
+	if (DUK_HEAP_IS_DEBUGGER_ATTACHED(thr->heap) && lj_type == DUK_LJ_TYPE_THROW) {
+		duk_context *ctx = (duk_context *) thr;
+		duk_bool_t fatal;
+		duk_bool_t is_syntax_err;
+		duk_hobject *hobj;
+
+		/* Don't intercept a DoubleError, we may have caused the initial double
+		 * fault and attempting to intercept it will cause us to be called
+		 * recursively and exhaust the C stack.
+		 */
+		hobj = duk_get_hobject(ctx, -1);
+		if (hobj == thr->builtins[DUK_BIDX_DOUBLE_ERROR]) {
+			DUK_D(DUK_DPRINT("built-in DoubleError instance thrown, not intercepting"));
+			goto skip_throw_intercept;
+		}
+
+		DUK_D(DUK_DPRINT("throw with debugger attached, report to client"));
+
+		fatal = !duk__have_active_catcher(thr);
+
+		/* Report it to the debug client */
+		duk_debug_send_throw(thr, fatal);
+
+		if (fatal && !is_syntax_err) {
+			DUK_D(DUK_DPRINT("throw will be fatal, halt before longjmp"));
+			duk_debug_halt_execution(thr, 1 /* use_prev_pc */);
+		}
+	}
+
+skip_throw_intercept:
+
+#endif
 
 	thr->heap->lj.type = lj_type;
 
--- a/src/duk_js_executor.c
+++ b/src/duk_js_executor.c
@@ -1726,8 +1726,10 @@ DUK_LOCAL duk_small_uint_t duk__executor_interrupt(duk_hthread *thr) {
 	DUK_HEAP_SET_INTERRUPT_RUNNING(thr->heap);
 
 	act = thr->callstack + thr->callstack_top - 1;
+
 	fun = (duk_hcompiledfunction *) DUK_ACT_GET_FUNC(act);
 	DUK_ASSERT(DUK_HOBJECT_HAS_COMPILEDFUNCTION((duk_hobject *) fun));
+
 	DUK_UNREF(fun);
 
 #if defined(DUK_USE_EXEC_TIMEOUT_CHECK)
@@ -2250,6 +2252,17 @@ DUK_INTERNAL void duk_js_execute_bytecode(duk_hthread *exec_thr) {
 		 * whenever a thread switch occurs by the DUK_HEAP_SWITCH_THREAD() macro.
 		 */
 #ifdef DUK_USE_INTERRUPT_COUNTER
+
+		if (thr->heap->dbg_processing) {
+			/* Don't interrupt if we're currently doing debug processing as
+			 * this may cause the debugger to be called recursively. Check required
+			 * for correct operation of throw intercept and other "exotic" halting
+			 * scenarios.
+			 */
+
+			goto skip_interrupt;
+		}
+
 		int_ctr = thr->interrupt_counter;
 		if (DUK_LIKELY(int_ctr > 0)) {
 			thr->interrupt_counter = int_ctr - 1;
@@ -2284,6 +2297,9 @@ DUK_INTERNAL void duk_js_execute_bytecode(duk_hthread *exec_thr) {
 				goto restart_execution;
 			}
 		}
+
+skip_interrupt:
+
 #endif
 #if defined(DUK_USE_INTERRUPT_COUNTER) && defined(DUK_USE_DEBUG)
 		thr->heap->inst_count_exec++;
--- a/src/duk_debugger.h
+++ b/src/duk_debugger.h
@@ -19,6 +19,7 @@
 #define DUK_DBG_CMD_PRINT         0x02
 #define DUK_DBG_CMD_ALERT         0x03
 #define DUK_DBG_CMD_LOG           0x04
+#define DUK_DBG_CMD_THROW         0x05
 
 /* Initiated by debug client */
 #define DUK_DBG_CMD_BASICINFO     0x10
@@ -86,7 +87,9 @@ DUK_INTERNAL_DECL void duk_debug_write_eom(duk_hthread *thr);
 
 DUK_INTERNAL duk_uint_fast32_t duk_debug_curr_line(duk_hthread *thr);
 DUK_INTERNAL void duk_debug_send_status(duk_hthread *thr);
+DUK_INTERNAL_DECL void duk_debug_send_throw(duk_hthread *thr, duk_bool_t fatal);
 
+DUK_INTERNAL_DECL void duk_debug_halt_execution(duk_hthread *thr, duk_bool_t use_prev_pc);
 DUK_INTERNAL_DECL duk_bool_t duk_debug_process_messages(duk_hthread *thr, duk_bool_t no_block);
 
 DUK_INTERNAL_DECL duk_small_int_t duk_debug_add_breakpoint(duk_hthread *thr, duk_hstring *filename, duk_uint32_t line);


svaarala:
I'd like to find a solution which avoids changing the bytecode being executed for two reasons:

- Performance impact when debugging support is enabled but a debugger is not attached.
- Bytecode dumped from another Duktape environment with no debugger support cannot be properly debugged in another instance.

Neither is a blocker if there is no way around adding an opcode, of course.

An easy solution would probably be to check for debugger status in the THROW opcode; if the debugger is enabled and suitable pausing flags are set, pause and enter debugger loop *before* executing the longjmp. We'd then be running debugger commands while still in the THROW opcode case clause. Once the debugger resumes, the THROW gets executed.

In general I'd like to ensure we can enter the debugger message loop at any point inside Duktape. This should more or less be already the case, but if not, it should be fixable.

fatcerberus:
On the first point, performance: STHROW acts as a THROW if the debugger is not attached, so there should be no noticeable impact for the extra instruction.

That said, adding an additional opcode does feel tacky to me, but I didn't see another alternative.  My first attempt was to pause in THROW and decrement PC, but that prevents resuming (it will just keep pausing).

What I could have done to avoid the additional opcode was to augment the thrown object so that, the second time the THROW is executed (on resume) it sees the additional flag and passes it through.  But that only works for objects, if a primitive (or null/undefined) is thrown that approach doesn't work.

So this is the solution I came up with, an STHROW which pauses if the debugger is attached, followed by a THROW to unwind on resume.

Is it possible currently to enter a debugger message loop at any arbitrary point?

svaarala:
Ah, you're right about the performance of course. The bytecode dump/load issue remains though.

What I was suggesting is something like:

```
case DUK_OP_THROW:
    if (debugger attached && pause on throw) {
        set paused, enter debugger message loop
    }

    handle throw normally
```

This would have the wrong PC though so maybe:

```
case DUK_OP_THROW:
    if (debugger attached && pause on throw) {
        pc--
        set paused, enter debugger message loop
        pc++
    }

    handle throw normally
```

I'm not 100% sure if you can currently call the message loop anywhere - probably yes, because e.g. duk_debugger_cooperate() does it. That would need to be checked carefully and fixed if not. Ensuring that the message loop works from any place would provide a lot of flexibility.

fatcerberus:
Wait a minute... If it's possible to enter the debugger loop on the fly like that... I might be able to modify it to intercept errors thrown in Duktape/C functions as well.

I'll have to revisit this later this evening.

svaarala:
Somewhat related, the debugger will need some behavior flags like "pause on all throws" or "pause on uncaught throw". Perhaps a SetFlags command with a bitmask would suffice for basic boolean flags (and would be easy to use from C).

svaarala:
Yeah, there's no fundamental reason why you can't enter the debugger message loop from a Duktape/C call - but there's currently no public API for that yet. I'm hesitant to add one until there's a clean idea of how these things best work together.

fatcerberus:
The behavior flags idea is a good one by the way, not just for this but for other uses as well.  For example, there could be a flag to disable tailcalls while debugging.

fatcerberus:
How's this?  No more STHROW, handled directly in THROW:

```c
#if defined(DUK_USE_DEBUGGER_SUPPORT)
				/* If an error is thrown with the debugger attached and nobody will
				 * catch it, execution is paused before the longjmp to allow local state
				 * to be examined.
				 */

				if (DUK_HEAP_IS_DEBUGGER_ATTACHED(thr->heap)) {
					duk_bool_t have_catch = 0;
					int i;
					
					/* check the catchstack for active catch clauses */

					for (i = 0; i < thr->catchstack_top; i++) {
						if (thr->catchstack[i].flags & DUK_CAT_FLAG_CATCH_ENABLED) {
							have_catch = 1;
						}
					}
					
					if (!have_catch) {
						DUK_D(DUK_DPRINT("Unhandled THROW, debugger attached, pause before longjmp"));
						--act->pc;  /* back up to the offending throw */
						DUK_HEAP_SET_PAUSED(thr->heap);
						duk__executor_interrupt(thr);
						++act->pc;  /* restore PC after resume */
					}
				}
#endif
```

I tested that, it *seems* to work, but I don't know if calling `duk__executor_interrupt()` mid-instruction like that will cause any problems (it's an "interrupt" though so I think I'm using it for its intended purpose...).  I couldn't figure any other way to get into a debugger message loop.  I'll force-push the change to the PR branch if this is acceptable.

svaarala:
It might be enough to call `duk_debug_process_messages()` there. There might be some trivia preventing that now though.

Also, there are some corner case issues like what if the debug messages changed breakpoints; do they need to be rechecked (goto restart_execution)? Presumably there's little that would affect us reaching the error throw - and an error throw will cause a restart_execution anyway.

Regarding force pushing, go right ahead. I actually prefer branches to be cleaned up before merge. It's only a practical problem if several people are working on the same branch.

svaarala:
About the debugger flags, I've been trying to figure out what the best "format" for that is, e.g.:

- `SetFlags bitmask`: easy to use from C, not all that extensible
- Individual setting commands like `SetPauseOnThrow true`: cleaner, generates a lot of commands, extends well
- Key/value settings like `SetSettings "pauseOnThrow" true "fooCallStackSize" 10`: extends well, but there's a second level of dispatch in some sense

Right now I'm leaning towards the second choice, with command numbers intentionally from relatively high numbers (which encode to more bytes which is OK for settings).

Apart from this cosmetic choice, there's a clear need to add the ability for settings and extending that set as we go along.

fatcerberus:
Regarding `duk_debug_process_messages()`: I tried that.  It only blocks until a message is received, then returns.  I suppose I could call it in a `while` loop though ("while paused, process messages")...

Something interesting I discovered though: The bit of code I posted above?  Works great when performed from a THROW instruction.  However when it is moved into `duk_err_longjmp()` or `duk_err_setup_heap_ljstate()` (both of which I tried in order to catch runtime errors), it leads to an internal longjmp error.  I see no reason why that should be the case?

fatcerberus:
So I tested: `duk_debug_process_messages()` in a while(paused) loop works. :)

fatcerberus:
Follow-up: `duk_debug_process_messages()` *doesn't* work.  The full interrupt seems to be necessary, otherwise, e.g. Eval commands cause the debugger to abruptly detach.

svaarala:
Sorry, I'm a bit busy with actual work (:-) so I can't lookup the code right now.

But yeah, I was thinking along the lines of "process messages until no longer paused". Probably there should be an internal helper for that which would be shared by the interrupt handler and whatever other call sites there are.

I wonder what the problem is with Eval. The only real blocker to calling the debugger from anywhere is that you must *never* do it when a debug message is being written out (otherwise two writers would write to the same byte stream). But that shouldn't be the case with Eval. I'll have a look hopefully later tonight.

svaarala:
Actually I think I know what the problem might be.

When Duktape is processing debug messages, this is flagged into the heap state so that we won't enter recursive debugger message handling when e.g. doing Eval.

If you just do "while (!paused) { process messages; }" this flag won't be set.

So what's needed is an internal helper which sets the flag and processes messages until no longer paused.

svaarala:
Anyway, this is what I meant by "fixing" the issue so that we can call debugger message loop from anywhere - it'd be good to figure out what the minimum set of internal helpers is, so that the structure is cleaned up a bit in the process.

fatcerberus:
Here's a better implementation which also intercepts runtime errors.  No extra opcodes are needed, so this should work with existing bytecode.

fatcerberus:
I'm wondering if the "Throw" debug notification should also include the filename and line number where the error was thrown.  This would allow the client to log all exceptions and their locations even if execution isn't halted.

fatcerberus:
Alright, the Throw notification is more informative, it now includes all the same information as a Status notify along with a flag specifying whether or not the error can be caught.  I also documented the new notification in `debugger.rst`.

Let me know if you want me to do anything else before this can be merged.

svaarala:
There are two issues that come to mind:

- It'd be nice to add a control command to set pause behavior
- Somewhat related, a debugger attach should reset debug related state and flags (perhaps even breakpoints, not sure if it's logical that they "carry over"

These don't necessarily have to be done in this branch though.

fatcerberus:
I'm hesitant to implement debug state resetting at the moment; I don't know if you remember from my other issue that I had to implement debugger reconnect in terms of an attach/detach (to ensure I didn't reconnect mid-message), and thus keeping breakpoints, etc. was desirable.

I have no problem adding a command to set the pause behavior though, that should be an easy addition.

svaarala:
Yep, there are pros and cons - on the other hand it's quite confusing if previous flags are preserved forever even when a session is broken and restarted a long time afterwards.

Anyway, given the pending 1.3 release, it's probably best to merge this afterwards. We can then do some other debugger work in 1.4.

fatcerberus:
That's probably for the best.  The error intercept code is in a somewhat sensitive location (longjmp setup), so it'd definitely be better to ensure it receives thorough testing to ensure there are no unforeseen issues.

fatcerberus:
So you mentioned that option commands should start at a high value, how high are we talking? 0x100, 0x1000?

fatcerberus:
I also wanted to add an option to disable tail calls during debug, but I think it'd be better to do that in a separate branch as it's not related to the changes here.

svaarala:
I was just thinking about the command list and how to keep the numbering reasonably organized. At the moment:

- Notify/command numbers initiated by Duktape are 0x00-0x0f so that they won't overlap with commands initiated by the debug client. There's no technical problem if they do, but it's confusing to read/dump.
- Command numbers initiated by debug client are 0x10 and above, with relatively little structure.

The main reason I was thinking about using high numbers was to ensure the option commands wouldn't "pollute" the main list if there turned out to be a great many of them; also lower numbers encode to a single byte which is nice for frequently used commands while config commands don't really matter.

But then again, maybe it'd just be cleanest to keep the commands in sequence. They're just arbitrary numbers after all.

svaarala:
One option for handling config would be to have a single SetConfig command which provided all the config values in one go (pause behavior, tail call behavior, etc).

This would be nice in that there wouldn't be a bunch of separate commands, and it would also ensure the debugger config state (not including breakpoints) would be set atomically in one go. We'd then just extend the contents of that command by appending new values (ignored by older Duktape versions).

I know you're eager to get started on this, but when new mechanisms are created, I try to ensure we get off on the right track, so to speak, so that things are easy to maintain later on. It takes some time to go through the options and prototype them though (in this case figuring out what and how large the potential config option set is, is it likely there will be structured values or lists, etc).

fatcerberus:
Sorry, I might get ahead of myself sometimes because I'm not much for planning things out--my typical process is to implement things first and then constant refactoring going forward until a natural equilibrium is arrived at.  It might sound inefficient, but it's actually very productive--minisphere was put together from the ground up To its present state in under a year.  You'd think this would lead to a mess of spaghetti code, but I enjoy refactoring so much that the "natural equilibrium" tends to include pristine code. :)

I totally understand the need for more stability than that with Duktape, though.  Compatibility is always an important factor for middleware so that upgrades can be painless.

svaarala:
I work mostly the same way by the way - most of my branches take weeks or months to be merge-ready. I prototype a lot of stuff, and also throw away many branches. Merging to master is a checkpoint for things being in a reasonably good shape (not 100% certainly) because some dependency on the merged stuff will start to build.

fatcerberus:
Anyway, to get back on topic, I want to do a bit of cleanup to this branch first, get some more real-world testing in (I'm also a big fan of "dogfooding", hehe), but ultimately I will probably push for this to be merged with only the basic pause-on-throw functionality and Throw notifications.  Since debug options are new, it's probably best to work on that independently.

fatcerberus:
@svaarala Alright, I cleaned up the branch, rebased to latest master and force-pushed the new commits.  Your call when you want to merge this.

fatcerberus:
Oh crud, #265 broke the build. :(

fatcerberus:
It's always something...

But yeah, this branch won't compile in its current state, which unfortunately `make dist` didn't tell me.  I'll have to look over the changes for #265 and adjust accordingly.

svaarala:
It's probably just that your code accesses `act->pc` or something like that. It should be trivial to fix.

"make dist" doesn't compile anything so it won't catch even the most basic syntax errors.

svaarala:
I'll be happy to look into what broke - mostly the changes in affected places are changing `act->pc` to `duk_hthread_get_act_curr_pc(thr, act)` and similar. I changed the name of the field (to `act->curr_pc`) so that it won't be accidentally used with the wrong type.

svaarala:
Regarding the Throw notify, there's a similar need for breakpoints: currently the debug client doesn't know *why* execution became paused. Perhaps that should be a field in status, so that if status is paused, there'd be another integer explaining why. Current values include at least: initial paused state, breakpoint triggered, debugger statement, throw.

svaarala:
Or another alternative would be just to make the "state" so that 0x00 is running, and >0x00 are the various pause reasons.

fatcerberus:
Weird, not sure why GitHub closed this, me force pushing must have confused it.  I HATE working with Git on the commandline.  In any case, I figured it out, the branch compiles now and seems to work correctly. :)

The status "pause reason" is a good idea, though not sure if that should go in this branch or not.

Note that a Throw notify is sent even if the error is caught.  This allows a client to display everything that was thrown, which is often useful.  Sometimes you need to see if an exception occurred, even if it was properly handled.

svaarala:
Hmm. There are some programs where try-catch is used for "normal" control transfer so these would also generate Throw notifys.

fatcerberus:
Hence why the default is not to pause execution unless the error is fatal. :)

svaarala:
No I mean it might generate a lot of Throw notifys. Status notifys are delivered a few times per second to ensure there's not a lot of traffic going on in normal execution state.

But anyway, we can make this behavior configurable on-the-fly or using a config option (similarly to what's already done for print, alert, and logging).

svaarala:
An example where a throw would be normal operation (instead of an actual error) is something like trial parsing.

fatcerberus:
That's true.

I think this represents the most useful defaults, in any case.  Notify all, halt on fatal.  That tends to be how a debugger is used in the majority of cases, in my experience.

I agree it should be customizable in the future though.

fatcerberus:
Will this be affected by #294?

See:
https://github.com/fatcerberus/duktape/blob/326ddd2bcd1cee229946ce810c12dde5ddf91c91/src/duk_js_executor.c#L4436-L4440

svaarala:
Most likely not, but if it does, it'll be easy to fix.

On a more general level, there's no regression coverage for debugger features so it's easy to break something when things are changed. I hope I'll have the time to address that soon, because having a regression harness is quite important. The JSON proxy should make it reasonably easy to write debug client test cases (though it's not as exact a process as Ecmascript expect output).

svaarala:
Agree about doing the pause status change separately, added #299.

fatcerberus:
Rebased to latest master and fixed a signed/unsigned comparison warning.

fatcerberus:
This branch currently segfaults if the recursion limit is reached due to the Throw notify using duk_safe_to_string(); this leads to a double error since the callstack can't grow anymore and in turn calls `duk_debug_send_throw()` again... and so on until the C stack topples over.

I haven't come up with a clean solution yet.  Perhaps we could allow the double error to be fatal and don't send a notify for it?

svaarala:
I think the recursion limits should be ideally applied to "user calls" only, and error handling, errCreate() etc would be exempt. Or there could be two limits, one for user calls and another for error handling. There's probably not enough state to distinguish the two reliably now.

fatcerberus:
I discovered syntax errors will lead to a crash as well: My working theory is that this is a result of calling PC-related functions when no bytecode is being executed or even exists.

fatcerberus:
Huh, apparently syntax errors are rethrown internally multiple times?  Once I fixed the crash and got it working properly I noticed several entries appeared in my IDE's error view for the same error, all listed as fatal.  I never saw this behavior for runtime errors.

svaarala:
The compiler itself will catch a SyntaxError and rethrow it, after altering the message. Rethrowing is pretty normal behavior also for various frameworks, which is why I mentioned it above as being an issue :) It'd be nice to distinguish the "original" throw from rethrows but it's not always possible because there isn't necessarily a logical connection between the two.

fatcerberus:
Yeah, I don't really mind, just wanted to verify it was normal behavior.  Reporting an error multiple times is better than not at all. :)

svaarala:
If I checked the diff correctly, the caught/uncaught check doesn't handle thread yield/resume correctly now. Errors that are not caught within a thread are propagated to a resuming thread (`thr->resumer`) if one exists (i.e. `thr->resumer != NULL`). So to check for a catcher, one also needs to follow the thread chain until `thr->resumer == NULL`.

fatcerberus:
Ah, good catch, thanks.  I don't use coroutines myself, I try to stick to ES standard functionality for game code to ensure I don't lock myself into Duktape permanently.  If I have to use the `Duktape` object at all I prefer to do it from the native side.

That said, the error propagation behavior intrigues me, as that's essentially what C# `await` does.  And I love that feature!

I'll make a note to fix this, thanks for the heads-up!

svaarala:
Just as a small stylistic annoyance, the check requires a two-level break:

```
thr_check = thr;
while (thr_check) {
    for (catchstack levels) {
        if (has catcher) {
            /* break past both loops */
        }
    }
    thr_check = thr_check->resumer;
}
```

I've mostly used goto for multi level breaks; I don't usually like tweaking the conditions or using flag variables myself. They also make the resulting assembly worse usually. But since inlining of small functions is more or less the default it's probably cleanest to use an external helper:

```
DUK_LOCAL duk_bool_t duk__have_active_catcher(duk_hthread *thr) {
    while (thr) {
        for (catchstack levels) {
            if (has catcher) {
                return 1;
            }
        }
        thr = thr->resumer;
    }
    return 0;
}
```

Old compilers won't necessarily inline these calls but it doesn't matter that much, especially outside of hot code paths like this.

fatcerberus:
I like the helper function.  The goto version would be awkward I think because the debugger can do different things depending on whether the error is caught or not (and in the future, depending on debug options).  That's why I used the "fatal" variable--it's, IMO, easier to reason about what the snippet's purpose is compared to a goto which requires you to look at the code after the label to be sure what's going on.

svaarala:
Ah, I didn't write it out but I meant that the goto part would set the "fatal = 1;" flag and goto out of the two loops:

```
while (thr) {
    for (...) {
        if (...) {
            fatal = 1;
            goto fatal_check_done:
        }
    }
}
fatal_check_done:

if (fatal) {
    ...
}
```

In other words, simulate a two-level break.

fatcerberus:
Oh, yeah, I'm fine with that too.  Unlike most developers I don't subscribe to the idea that goto is inherently "evil", in fact I think Duktape employs it very deftly.

Still a helper function might not be a bad idea anyway.

svaarala:
There may be a few too many, but they're hard to give up ;). For example I've used gotos to reduce indent in places. This could be replaced with local helper functions (which get inlined because they only have one call site). But the trouble is that sometimes the helpers need a lot of context of the parent function, which leads to a lot of arguments being passed, perhaps multiple output pointers, etc.

Multi-level breaks are another. One can always work around them by adding more control flags and tweaking the loop checks, but IMO is makes the loops harder to read than a goto to break out multiple levels.

But I realize I'm probably in the minority ;)

svaarala:
In this case though the helper has a very clear meaning and a single argument with no other dependencies on the calling context, so it works well.

fatcerberus:
What I think I'll do in a bit is update the OP of the pull to include a checklist of things that need to be done for this, so it's clearer how close it is to being merge-ready.  I like the checkbox feature, it's nifty.

fatcerberus:
@svaarala There, it's done.  Anything else you want me to implement in this branch, let me know.

fatcerberus:
@svaarala So... how often do you blow the stack working with error handling code? :)

Needless to say this is a very fragile bit of code.  I've found I have to be very careful to avoid throwing during error intercept because doing so will almost invariably cause a recursive DoubleError pileup.

fatcerberus:
I wonder, should the Throw notify include the stack trace for Error instances? Hm...

fatcerberus:
So it turns out there's not a whole lot of headroom when handling errors once the recursion limit has been hit.  Even something as seemingly benign as reading `err.fileName` for the throw notify will blow the native stack because there's no more Ecmascript callstack space left to call a property getter.

What I will probably have to do (for now) is to detect whether we're in a double fault state and, if so, don't attempt to intercept or notify at all.  That's preferable, actually, since if you've caused a double fault then something already went horribly wrong and abandoning normal operation is completely acceptable.

fatcerberus:
I think I got all the bugs out now. :)

Here's a snapshot:
https://drive.google.com/open?id=0BxPKLRqQOUSNNGU0TTB0dlpiYlk

Still needs a bit more cleanup to the code, but all the functionality should be there.

fatcerberus:
That's promising, I didn't break any tests. :-)

fatcerberus:
Found an interesting bug.  During an Eval on a cyclic object, I get the following corrupt Throw notify:
```
		[0]	NFY	dynamic {minisphere.Remote.Duktape.DValue}
		[1]	5	dynamic {int}
		[2]	0	dynamic {int}
		[3]	"TypeError: cyclic input"	dynamic {string}
		[4]	REP	dynamic {minisphere.Remote.Duktape.DValue}
		[5]	0	dynamic {int}
		[6]	"DoubleError: error in error handling"	dynamic {string}
		[7]	EOM	dynamic {minisphere.Remote.Duktape.DValue}
```

Which is very weird.  The double fault hits before the EOM of the first message is written, which makes no sense.  Should be fun to diagnose.

svaarala:
The general challenge in writing out debug messages is that one must not write a single byte unless there's a guarantee that you'll be able to write the rest -- errors and other side effects included. This is one reason why it's not trivial to add debugger stuff anywhere.

fatcerberus:
I figured out the bug in this case at least--`filename` is set to undefined (not "input" or an empty string as expected) for the Eval error, so the `duk_require_hstring()` to get the filename from the error caused another error.  Fixed by adding an explicit string coercion.

I'll reorganize the throw notify code to avoid corrupt messages in the future, though.

svaarala:
I guess one simple protection would be to flag a "message in progress" situation and at least assert for that so it's easier to catch these cases.

A simple way to do that would be to add a debug flag to heap (`heap->dbg_writing_message`) when assertions are enabled. That flag could then be managed and asserted for in the REQ, REP, NFY, ERR, and EOM writers. For example:

- REQ: assert for dbg_writing_message == 0, then set to 1 (same for REP, NFY, ERR)
- EOM: assert for dbg_writing_message == 1, then set to 0

No other code would need changes; in particular, the individual message write call sites didn't need a tedious assert wrapping.

If there's some case that can't be solved, the flag could then be included in a normal build too, and a debug message could be avoided if `heap->dbg_writing_message != 0`.

svaarala:
If you implement the above assert change or something generic, we can merge it to master ahead to the throw-on-debug too - but in this branch is fine as well.

fatcerberus:
I like that idea, it appeals to my love of simplicity. :)

I've found that more often than not, rather than piling on asserts everywhere, bugs can be avoided altogether by simply changing your assumptions.  And then if they ARE violated, make sure the resulting failure is catastrophic. ;-)

fatcerberus:
By the by, with "catastrophic" I didn't mean "security breach" level failure, just in the sense that it's better to crash early rather than having the program hobble along in an inconsistent state until finally falling over a while down the road.  Those are the worst kind of issues to deal with.

svaarala:
I find that to be unachievable in practice, so asserts are the only practical way of keeping track of assumptions that are *rarely* violated in practice (e.g. depend on timing, triggering of finalization or garbage collection in just the right place etc).

svaarala:
A "crash" failure, unless caused by an intentional segfault/abort/exit, is almost always at risk of being an exploitable security failure. That's why I prefer using a lot of asserts, and being as sure as possible that they are never triggered in full test runs.

fatcerberus:
In fairness of course, a program in an inconsistent state is itself at risk of exploitation.  One only has to look at video game glitches and gamers' abuse thereof to see evidence of this. :)

svaarala:
Yes, and assertions are intended to document and check for what is meant by a "consistent" state, so that such errors are caught as early as possible (during development).

If you don't do that, and get a bug that *doesn't* cause a crash in most cases, you don't even know you're running in an inconsistent state.

svaarala:
By the way, what I mean by assertions is the "original" meaning of "this is necessarily true here in all cases, unless the program is broken by design or some documented assumptions have been violated". Sometimes people use assertions in code to mean "if this isn't true, we should abort" and use it e.g. for argument checking, even when the caller is external code which might actually trigger the assert.

Duktape assertions are the first kind: they're documenting what guarantees should be in place, by the way Duktape works. That's why they're not needed in release builds, as they're never intended to be triggered. (That's the general goal anyway; API call `ctx` argument is an exception to this, and is explicitly documented as being an exception.)

fatcerberus:
Indeed, and I have nothing against asserts in general (I suggested the assert() function after all ;), I just think that a lot of developers go overboard with them.  At some point you have to step back and realize you're making way too many assumptions for the code to stay stable and take time to refactor it.

Do you want me to add the debug message asserts to this branch or do you want that in a separate merge?

svaarala:
Ideally I'd like to work with a language that supported design-by-contract natively (preconditions, postconditions, state invariants, all understood and verified by the compiler statically and/or dynamically). Unfortunately out language environment doesn't really allow that if you also want portability to a wide range of targets, memory management tricks, etc. Invariants actually also enable compiler optimizations because the compiler can just skip checks that are verified/assumed. Sigh, back to real world :)

Adding asserts for debug messages in this branch or separately are both fine - it should be easy to do separately but I don't mind it coming with this pull either.

fatcerberus:
On the other hand, Java has checked exceptions, which forces you to either catch an exception you have no means to handle or declare it and break encapsulation.  There's no perfect solution. :-(

Anyway, this still has tons of bugs.  I've noticed it won't send a pause notify unless I issue a manual Pause request from the client afterwards (this isn't 100% reproducible).  Eval commands will also sometimes detach the debugger.

I'm able to fix most of the issues by calling `duk__executor_interrupt()` to pause but this asserts if the interrupt is generated during a C call.

fatcerberus:
This is a weird issue.

Debugger message loop to pause on throw:
```
	thr->heap->dbg_processing = 1;
	while (thr->heap->dbg_paused) {
		duk_debug_process_messages(thr, 0 /* no_block */);
	}
	thr->heap->dbg_processing = 0;
```

Debug output from Duktape following an Eval command during this pause:

```
[D] duk_debugger.c:1289 (unknown): debug command eval
[D] duk_debugger.c:1724 (unknown): paused, process debug message, blocking if necessary
[D] duk_debugger.c:1680 (unknown): invalid initial byte, drop connection: 0
[D] duk_debugger.c:1715 (unknown): debug connection broken, stop processing messages
[D] duk_debugger.c:94 (unknown): attempt to read flush in detached state, ignore
[D] duk_js_executor.c:1683 (unknown): debugger became detached, resume normal execution
[D] duk_debugger.c:512 (unknown): attempt to write 1 bytes in detached state, ignore
[D] duk_debugger.c:471 (unknown): attempt to write 1 bytes in detached state, ignore
[D] duk_debugger.c:471 (unknown): attempt to write 1 bytes in detached state, ignore
[D] duk_debugger.c:471 (unknown): attempt to write 9 bytes in detached state, ignore
[D] duk_debugger.c:512 (unknown): attempt to write 1 bytes in detached state, ignore
[D] duk_debugger.c:113 (unknown): attempt to write flush in detached state, ignore
[D] duk_debugger.c:203 (unknown): attempt to read 1 bytes in detached state, return zero data
[D] duk_debugger.c:1715 (unknown): debug connection broken, stop processing messages
[D] duk_debugger.c:94 (unknown): attempt to read flush in detached state, ignore
```

`invalid initial byte: 0` - which means for some reason, it's misinterpreting the EOM for the Eval request as a new message.  I have a feeling Duktape might be entering debug processing recursively even though `heap->dbg_processing = 1` _should_ have prevented it.

svaarala:
Yeah, extending the debugger support so that it can be called from anywhere has a lot of small details that need to be right. That's why I restricted the debugger to the executor loop originally to make it reasonably easy to manage. The challenge is to find easy-to-understand mechanisms for preventing parsing or writing out messages-within-messages.

I'll be able to help you out with this once 1.3.0 is out :)

fatcerberus:
I was right.  What it boils down to is that Eval assumes that:

1. It is handled during an executor interrupt.
2. The interrupt handler is not reentrant.

The message loop shown above breaks (1), which in turn implicitly breaks (2).  Which means the only reliable way to enable this for all errors right now is:

```c
DUK_HEAP_SET_PAUSED(thr->heap);
duk__executor_interrupt(thr);
```

So I'll do that for now in minisphere, but this is definitely something to improve on in the future, particularly if we want to implement more debugger "traps" in the future (for example, as part of #301).

fatcerberus:
> preventing parsing or writing out messages-within-messages.

Duktapeception!

Sorry, it was obligatory. :)

svaarala:
I'm not much help until I have time to look at the code more closely, but I guess the re-entrancy check should be in the debugger message loop rather than in the interrupt mechanism (which may need its own re-entrancy prevention for other reasons). This would prevent recursive parsing of inbound messages. For the writing side one would need to ensure that a full message can be written without debug-code-triggering side effects - or at least if there's such a risk, ensuring that it's possible to e.g. drop a debug message (rarely) if that side effect is triggered.

Anyway, I'd prefer trying to figure out good mechanisms for these in general so that it's not a "patchwork" of individual checks here and there.

fatcerberus:
Actually I'd imagine there are good reasons not to have the interrupt handler called during a debugger eval (which it currently is not, but only as long as the two assumptions above aren't violated), so forcing an interrupt in this case actually seems like the best solution.

svaarala:
Yeah, it is intentional that an interrupt doesn't happen within interrupt.

But in addition to that, debug message loop handling shouldn't happen inside debug message handling. In master this is automatically the case because of the interrupt handling checks in place. But if debug message handling is called outside, a separate check is needed to prevent that recursion.

svaarala:
Note that entering the interrupt handler without reason will lead to interrupt counters being off sync with the number of instructions executed -- the idea is that the interrupt handler will be able to count how many instructions have been executed. This will allow e.g. sampling based profiling in the future. The interrupt is also integrated with the executor in other ways, so just calling it from another place won't automatically be correct, all the state changes need to be accounted for.

My hunch is that the debugger message loop needs to be made callable from any place, which means it will need it's own recursive calling prevention mechanisms. The interrupt handler has its own mechanisms and they're needed for things other than debugging too.

fatcerberus:
Indeed, I figured it wasn't really a clean solution, but at least for my use case right now in minisphere it's sufficient (and won't really throw anything off sync since the only thing you can do with it at present is resume and unwind).  There's a reason I haven't actually pushed anything to this branch recently. :wink:

As mentioned way above in this conversation, interrupting a Duktape/C function additionally triggers an assert, so I know what I'm doing isn't technically kosher. :)

svaarala:
Hmm, if debugger Eval happens when the debugger message loop is called from outside the executor, the executor interrupt would probably still need to be disabled in this case. Otherwise that Eval might trigger breakpoints which is intentionally prevented when it is called from inside the executor.

It's not a trivial thing because the mechanisms are related. I can draft a solution in master once I get 1.3.0 out of the way and you can then rebase this branch. Or we can develop it in this branch, though it's more of a prerequisite for adding the throw support than part of it.

fatcerberus:
I think I'd have to agree with developing the solution separately.  That tends to be my workflow at least--if I can't develop a feature due to limitations in the implementation, I try to get the necessary refactoring out of the way first.

fatcerberus:
Just so you know, there's no rush at all on this--I'm resourceful enough to work around issues in the meantime for my own use.  Honestly, posting the pull was probably premature.

The WIP Duktape 1.3 has been great, by the way.  I think this is the biggest upgrade yet. :)

svaarala:
Thanks, that's nice to hear :) It should really have been two minor releases around 2-3 months each, but it's nice to finally get it finalized.

fatcerberus:
A bit later today I'll rebase this against 1.3.0 and then try to come up with a better solution that doesn't involve creating extra interrupts.

fatcerberus:
Okay, rebased against 1.3.0 and updated to no longer generate spurious interrupts.  The solution turned out to be easier than I thought, I just had to check whether `heap->dbg_processing == 1`, and if so, skip interrupt processing.  This avoids interrupts during Eval and also avoids debug processing being enter recursively.

I tested it out in minisphere, there are no obvious bugs.  Automated testing would be better, but we're not quite there yet. :)

Check it out and let me know if I should do anything else with this.

fatcerberus:
Oh, forgot to mention, this latest iteration includes a new helper, `duk_debug_halt_execution()`, which can be used to do what we discussed above, to pause and enter debug processing at will from anywhere in Duktape.

src/duk_debugger.c
```diff
@@ -847,6 +847,57 @@ DUK_INTERNAL void duk_debug_send_status(duk_hthread *thr) {
...
+	duk_activation *act;
+	duk_uint32_t pc;
+
+	DUK_ASSERT(thr->valstack_top > thr->valstack);  /* ... [err] */
+
+	duk_debug_write_notify(thr, DUK_DBG_CMD_THROW);
+	duk_debug_write_int(thr, fatal);
+
+	duk_get_global_string(ctx, "Error");
+	is_err_inst = duk_instanceof(ctx, -2, -1);
```
svaarala: This should probably just be `duk_is_error()`, I'll comment a bit more on the pull.

src/duk_debugger.c
```diff
@@ -847,6 +847,57 @@ DUK_INTERNAL void duk_debug_send_status(duk_hthread *thr) {
...
+	if (is_err_inst) {
+		/* Error instance, use augmented error data directly */
+		duk_get_prop_stridx(ctx, -1, DUK_STRIDX_FILE_NAME);
+		duk_safe_to_string(ctx, -1);
+		duk_debug_write_hstring(thr, duk_require_hstring(ctx, -1));
+		duk_get_prop_stridx(ctx, -2, DUK_STRIDX_LINE_NUMBER);
+		duk_debug_write_uint(thr, duk_require_uint(ctx, -1));
+		duk_pop_2(ctx);
+	}
+	else {
```
svaarala: Style trivia: `} else {` on the same line.

src/duk_error_misc.c
```diff
@@ -42,6 +71,60 @@ DUK_INTERNAL duk_hobject *duk_error_prototype_from_code(duk_hthread *thr, duk_er
...
+		duk_context *ctx = (duk_context *) thr;
+		duk_bool_t fatal;
+		duk_bool_t is_syntax_err;
+		duk_hobject *hobj;
+
+		/* Avoid intercepting syntax errors. Doing so will usually cause a
+		 * segfault since the JS executor isn't guaranteed to be active.
+		 */
+		duk_get_global_string(ctx, "SyntaxError");
+		is_syntax_err = duk_instanceof(ctx, -2, -1);
```
svaarala: This should probably be `duk_get_error_code(ctx, -1) == DUK_ERR_SYNTAX_ERROR`.

doc/debugger.rst
```diff
@@ -1292,6 +1292,31 @@ Example::
...
+
+Duktape sends a Throw notification whenever an error is thrown, either by
+Duktape due to a runtime error or directly by Ecmascript code.
+
+msg is the string-coerced value being thrown. Filename and line number are
+taken directly from the thrown object if it is an Error instance (after
+augmentation), otherwise these values are calculated from the bytecode
+executor state.
+
+
```
svaarala: Extra newline.

svaarala:
Added a few style comments in commits. One issue is checking for whether a value is an Error or a SyntaxError.

I'd suggest using `duk_is_error()` and `duk_get_error_code() == DUK_ERR_SYNTAX_ERROR` (which is basically a synonym for duk_is_syntax_error(); perhaps there should be a macro for it - #340).

This needs a bit of background. When defining semantics, the E5 specification almost never looks up a global value like `Error` but rather refers to "the original value of Error". This concretely means that if a user replaces the Error object from ordinary Ecmascript code, that replacement doesn't affect whatever Error related semantics are defined in the specification.

`duk_is_error()` and `duk_get_error_code()` follow this approach: they compare the inheritance chain against the original error objects: https://github.com/svaarala/duktape/blob/master/src/duk_api_stack.c#L2856-L2879.

The upside of this approach is that it's consistent with other E5 semantics, and the user cannot mess up semantics very easily. The downside is that to truly replace `Error` with another object can only be done in an engine specific way. Duktape sandboxing code does not yet allow that, for example: the `thr->builtins[]` entries cannot yet be replaced, though that will be easy (and necessary) to fix. User code should simpy modify the existing `Error` object instead of replacing it.

Anyway, that's the background of avoiding `duk_get_global_string()` and using `duk_instanceof()` against values looked up from there :)

src/duk_debugger.c
```diff
@@ -1639,6 +1690,37 @@ DUK_LOCAL void duk__debug_process_message(duk_hthread *thr) {
...
+	/* Process debug messages until we are no longer paused */
+	thr->heap->dbg_processing = 1;
+	duk_debug_send_status(thr);
+	while (thr->heap->dbg_paused) {
+		DUK_ASSERT(thr->heap->dbg_processing);
+		duk_debug_process_messages(thr, 0 /* no_block */);
+	}
+	thr->heap->dbg_processing = 0;
+
+	act->curr_pc = old_pc;  /* restore PC */
```
svaarala: Assumption is that we'll always come to restore the `act->curr_pc` values, i.e. there is no code path which can avoid this even in error situations etc. This assumption should probably be mentioned shortly. Concretely we rely on `duk_debug_send_status()` and `duk_debug_process_messages()` never throwing.

svaarala:
I'll review the control flow, interrupt executor, and debug message loop control flags etc tomorrow when I have more time.

src/duk_debugger.c
fatcerberus: Hehe, I must have lapsed into my own coding style there, I usually try very hard to follow Duktape style (old-style block comments are painful for me to write!).  Although it's also possible MSVC auto-formatted it.  I'll fix it. :)

src/duk_error_misc.c
fatcerberus: Ah, I knew there must have been a simpler way to do that!

doc/debugger.rst
fatcerberus: Huh, not sure where that came from, I specifically remember deleting that when I reviewed the diff pre-commit...

src/duk_debugger.c
fatcerberus: I addressed the style issues and error semantics in the latest commit.

I completely understand the thing about not looking up globals to enforce semantics, by the way--I had to do a similar thing with minisphere  a while back.  My engine is meant to be compatible with the old Sphere engine, which didn't have any constructors for built-in objects, and I ran into an issue where legacy games were clobbering the constructors, causing mysterious errors when something like `LoadImage()` was called.  I had to switch to putting the built-in prototypes in the global stash so that the engine could still construct builtins without having to look up the class name directly.

fatcerberus:
I addressed the style issues and error semantics in the latest commit.

I completely understand the thing about not looking up globals to enforce semantics, by the way--I had to do a similar thing with minisphere a while back. My engine is meant to be compatible with the old Sphere engine, which didn't have any constructors for built-in objects, and I ran into an issue where legacy games were clobbering the constructors, causing mysterious errors when something like LoadImage() was called. I had to switch to putting the built-in prototypes in the global stash so that the engine could still construct builtins without having to look up the class name directly.

src/duk_debugger.c
fatcerberus: I agree, it's just it's not entirely clear what, concretely, can be done.  It would be nice to make the assumption explicit with an assert, but where to put it? If an error is thrown, we lose all information that the PC was ever changed, so it's hard to verify that it was indeed changed back.

svaarala: Oh, I just meant that a short comment might be in order - it's maybe not obvious catastrophic things will happen if that assumption is violated :) The catastrophy here would be executing a bytecode instruction twice which might be very interesting to diagnose :)

fatcerberus: Ah, I see, I'll add a quick comment to that effect when I'm near a computer again later today.

src/duk_debugger.c
```diff
@@ -1639,6 +1684,37 @@ DUK_LOCAL void duk__debug_process_message(duk_hthread *thr) {
...
+	duk_activation *act;
+	duk_instr_t *old_pc;
+
+	DUK_HEAP_SET_PAUSED(thr->heap);
+
+	/* Decrement PC if that was requested, this requires a PC sync */
+	act = duk_hthread_get_current_activation(thr);
+	duk_hthread_sync_currpc(thr);
+	old_pc = act->curr_pc;
+	if (use_prev_pc && act->curr_pc > 0) {
```
fatcerberus: I just noticed this isn't actually correct anymore as curr_pc is a pointer now.  Should probably be `act->curr_pc != NULL`, yes?

svaarala: The `act->curr_pc > 0` is now incorrect when `act->curr_pc` is a pointer and not a number. There's a helper to get the previous PC so perhaps this should be:

```
...
duk_hthread_sync_currpc(thr);
if (use_prev_pc) {
    old_pc = duk_hthread_act_get_prev_pc(thr, act);
} else {
    old_pc = duk_hthread_act_get_curr_pc(thr, act);
}
act->curr_pc = old_pc;
```

Not very clean I agree but should be correct.

svaarala: `!= NULL` would not be correct when `curr_pc == first_bytecode_instruction`: doing `curr_pc--` would then cause `curr_pc` to be outside valid bytecode range.

fatcerberus: That won't work, `duk_hthread_get_act_curr_pc(thr, act);` (and the `prev_pc` variant) returns a numeric PC, not a pointer.

fatcerberus:
There, I think I dealt with the curr_pc handling adequately.  I also added a note to the effect of "throwing errors out of debug handling here is bad" :)

src/duk_debugger.c
```diff
@@ -847,6 +847,51 @@ DUK_INTERNAL void duk_debug_send_status(duk_hthread *thr) {
...
+	duk_debug_write_hstring(thr, duk_require_hstring(ctx, -1));
+	duk_pop(ctx);
+
+	if (duk_is_error(ctx, -1)) {
+		/* Error instance, use augmented error data directly */
+		duk_get_prop_stridx(ctx, -1, DUK_STRIDX_FILE_NAME);
+		duk_safe_to_string(ctx, -1);
+		duk_debug_write_hstring(thr, duk_require_hstring(ctx, -1));
+		duk_get_prop_stridx(ctx, -2, DUK_STRIDX_LINE_NUMBER);
+		duk_debug_write_uint(thr, duk_require_uint(ctx, -1));
```
svaarala: The "require" call here breaks the promise that if we start writing a debug message we're guaranteed to finish it.

src/duk_error_misc.c
```diff
@@ -5,6 +5,35 @@
...
+
+	duk_size_t i;
+
+	DUK_ASSERT(thr != NULL);
+
+	while (thr != NULL) {
+		for (i = 0; i < thr->catchstack_top; i++) {
+			duk_catcher *cat = thr->catchstack + i;
+			if (DUK_CAT_HAS_CATCH_ENABLED(cat)) {
+				return 1;  /* all we need to know */
```
svaarala: Hmm, what about "finally" enabled? A finally clause can catch and handle (neutralize) an error, e.g. by turning it into a return.

src/duk_error_misc.c
```diff
@@ -42,6 +71,58 @@ DUK_INTERNAL duk_hobject *duk_error_prototype_from_code(duk_hthread *thr, duk_er
...
+	if (DUK_HEAP_IS_DEBUGGER_ATTACHED(thr->heap) && lj_type == DUK_LJ_TYPE_THROW) {
+		duk_context *ctx = (duk_context *) thr;
+		duk_bool_t fatal;
+		duk_bool_t is_syntax_err;
+		duk_hobject *hobj;
+
+		/* Avoid intercepting syntax errors. Doing so will usually cause a
+		 * segfault since the JS executor isn't guaranteed to be active.
+		 */
+		is_syntax_err = duk_get_error_code(ctx, -1) == DUK_ERR_SYNTAX_ERROR;
```
svaarala: This comment seems a bit vague, "usually" and segfault don't mix well ;)

Anyway, there's never a guarantee of the executor being active: the callstack may consist entirely of native Duktape/C calls, regardless of what error is being thrown?

src/duk_error_misc.c
```diff
@@ -42,6 +71,58 @@ DUK_INTERNAL duk_hobject *duk_error_prototype_from_code(duk_hthread *thr, duk_er
...
+		if (is_syntax_err) {
+			DUK_D(DUK_DPRINT("SyntaxError thrown, not intercepting"));
+			goto skip_throw_intercept;
+		}
+
+		/* Don't intercept double errors either, something's not right
+		 * and if we generate more double errors it will cause a recursive
+		 * pileup and blow the stack.
+		 */
```
svaarala: "blow the stack" doesn't really say what happens - what stack is affected, is there an infinite loop, memory unsafe behavior etc?

src/duk_js_executor.c
```diff
@@ -2250,6 +2252,17 @@ DUK_INTERNAL void duk_js_execute_bytecode(duk_hthread *exec_thr) {
...
 #ifdef DUK_USE_INTERRUPT_COUNTER
+
+		if (thr->heap->dbg_processing) {
+			/* Don't interrupt if we're currently doing debug processing as
+			 * this may cause the debugger to be called recursively. Check required
+			 * for correct operation of throw intercept and other "exotic" halting
+			 * scenarios.
+			 */
+
+			goto skip_interrupt;
```
svaarala: Skipping the interrupt handler entirely will cause instruction counts based on cumulative interrupt counter values to be off sync and should be debug logged e.g. here: https://github.com/svaarala/duktape/blob/master/src/duk_js_executor.c#L1700-L1706

The guarantee that interrupt counter matches exactly to opcodes executed is a goal but it's not yet needed for anything specific (like sampling based profiling). So, this can be addressed in a cleanup branch too.

The fix would probably be to move the check where the recursive interrupt check is: https://github.com/svaarala/duktape/blob/master/src/duk_js_executor.c#L1716

Not sure if that causes other issues.

src/duk_error_misc.c
fatcerberus: I was going to do that, but then information on the original throw site is lost if it IS rethrown (execution pauses at ENDFIN), which is the common case in a finally with no catch.  By that point the stack is already unwound so that pause is useless.

src/duk_debugger.c
fatcerberus: Hm, I was using send_status as a template, and you did the same thing there... :)

src/duk_error_misc.c
fatcerberus: "Blow the stack" is a common programming idiom for a stack overflow, in context it seemed clear to me at least that it would be the C stack affected (recursive pileup).

I'll reword it anyway. :)

src/duk_error_misc.c
svaarala: Ah, I see. So the downside will be that if a finally block catches and neutralizes an error, it will be incorrectly flagged as fatal and causes an unnecessary pause. Not a big deal, so we can leave it as is.

src/duk_debugger.c
svaarala: The `duk_require_hstring()` is OK because there's a `duk_safe_to_string()` (which is guaranteed not to throw and guarantees a string result).

The `duk_require_uint()` is the potential offender here.

src/duk_js_executor.c
fatcerberus: In practice it should only be skipped for Eval command code though only, where we don't want the interrupt counter to run anyway?  That's what I thought, at least.

src/duk_debugger.c
svaarala: `duk_get_uint()` would be fine because it won't (a) throw on type mismatch and (b) will supply a 0 if the type is invalid.

This would only happen if someone intentionally changed the error `.lineNumber` to a bogus value like `["foo"]`.

src/duk_error_misc.c
svaarala: Yeah, I've been programming for 20+ years, but I still don't know what it means *in this context* :)

In other words, is there an infinite loop, memory unsafe behavior, or an ugly error, etc.

svaarala: Here I guess there would be a potential infinite loop with C stack being exhausted?

fatcerberus: A crash, but not due to memory unsafe behavior, merely because too many functions were called recursively.  Specifically, if we attempt to intercept double errors and screw up, the error handling process throws another one, which is then intercepted, throwing another one, etc.  The throw intercept code keeps being called recursively and we run out of stack space.

svaarala: Ok, so saying that C stack would be exhausted would be pretty clear IMO. That's memory unsafe on some platforms but leads to a controlled process exit in other platforms, etc.

src/duk_error_misc.c
fatcerberus: This comment was based on experience--catching a SyntaxError _usually_ caused a segfault, but not always (i.e. there are certain scenarios where it's actually completely safe, so saying it WILL segfault is incorrect).

Regardless, I agree the wording is very awkward.

src/duk_debugger.c
svaarala: Hmm, regarding `duk_require_hstring()` here, it'd be cleaner to use `duk_get_hstring()` (here and in existing debugger codebase) and ensure `duk_debug_write_hstring()` wrote an empty string on a NULL input.

This is the case because throwing an error is never correct behavior here. So the best choices would be IMO: (1) `duk_get_hstring()` + assert, or (2) `duk_get_hstring()` and silently tolerating a NULL.

No need to fix these in this pull - but I'll add a note to clean this up later on.

The `duk_require_uint()` issue still remains, and it's an actual potential problem. You should be able to reproduce the issue by constructing an error like:

```
duk> e = new Error('foo'); Object.defineProperty(e, 'lineNumber', {value:"foo"})
= Error: foo
duk> e.lineNumber
= foo
```

src/duk_error_misc.c
fatcerberus: Yeah, it's not an exact science, without actually throwing the error the closest determination we can make is "This error is probably fatal" and pause to allow locals to be viewed.  A catch clause always neutralizes, a finally may not, and in practice, usually doesn't.

I'm actually curious now how, e.g. the MSVC# debugger handles this scenario.  I'll have to try that later.

src/duk_error_misc.c
svaarala: I guess the main issue for me is that I don't understand what happens when there's no executor running - and because that can happen with or without a SyntaxError, is there a concrete problem here?

src/duk_error_misc.c
fatcerberus: Yes, that's exactly what it means to "blow the stack". :wink:  But yeah, I'll fix the comment.

src/duk_js_executor.c
svaarala: I think that right now Eval also consumes the interrupt counter. It just doesn't execute the interrupt handler: https://github.com/svaarala/duktape/blob/master/src/duk_js_executor.c#L1709-L1733

As you see, the interrupt counter is updated during recursive interrupts but no actions are taken. The debug code that tries to validate that interrupt counter/init counter match bytecode execution count exactly is updated.

If Eval opcodes were not to be counted, they should also be ignored for the purposes of updating `thr->heap->inst_count_interrupt` in the main dispatch loop. Otherwise the two values won't match - that's not a concrete issue now, but the attempted guarantee is for them to match. I agree it might actually be preferable not to count Eval and other debugger induced bytecode execution towards a sampling profiler. But that's a separate change, and it'd be nice if there was one consistent policy in place at a time.

Anyway, it doesn't have to be fixed with this pull, we can handle these cleanups separately.

src/duk_error_misc.c
fatcerberus: Yes, pausing without a running executor WILL segfault (or at least, catching a syntax error without it running will, I haven't tested throwing other types of errors in this scenario).

...which means this check actually does have to be updated, is there a way to check whether bytecode is being executed?

svaarala: Hmm - what happens if the error-about-to-be-thrown is thrown from C code instead of Ecmascript code? Does that work, as long as there's *some* Ecmascript activation in the call stack?

I think it'd be best to fix this so that it's not an issue that no bytecode is executing. After all, there's no reason why the debugger loop shouldn't be runnable outside of any execution: `duk_debugger_cooperate()` already does it.

src/duk_debugger.c
```diff
@@ -1639,6 +1684,47 @@ DUK_LOCAL void duk__debug_process_message(duk_hthread *thr) {
...
+
+DUK_INTERNAL void duk_debug_halt_execution(duk_hthread *thr, duk_bool_t use_prev_pc) {
+	duk_activation *act;
+	duk_hcompiledfunction *fun;
+	duk_instr_t *old_pc;
+
+	DUK_HEAP_SET_PAUSED(thr->heap);
+
+	act = duk_hthread_get_current_activation(thr);
+	fun = (duk_hcompiledfunction *) DUK_ACT_GET_FUNC(act);
```
svaarala: Related to the SyntaxError issue: I guess there's actually no guarantee that `fun` will be a `duk_hcompiledfunction` here if the error is thrown from C code? That doesn't seem to cause issues below: the `act->curr_pc` check checks for NULL first, and a non-NULL `act->curr_pc` implies `fun` is a compiled function. But reading this one might expect `fun` to be a compiledfunction.

fatcerberus: I have to do a bit more digging to see what, concretely, causes the segfault and then I'll get back to you. :)

svaarala: Ok, thanks -- sorry for being a bit pedantic here, I really don't want to merge memory unsafe behavior into master. Other misbehavior is more or less fine as long the impacts are understood :)

fatcerberus: Results of testing:
* There is no issue when an error is thrown from a Duktape/C function.  There is no actual compiledfunction in that case, but as there is a check for that, it's completely safe.
* The actual, concrete issue is this: If an error is thrown with _nothing_ on the Duktape callstack (as is the case for, say, a top-level `duk_call()` throwing a SyntaxError), it will segfault because `duk_hthread_get_current_activation(thr);` returns NULL.

svaarala: Ok, that should be easy to fix - it would probably then make sense to allow SyntaxErrors to be handled without exception?

svaarala:
So in summary I think there are two actual issues before merge:

- The `duk_require_uint()` case which should actually be triggerable and cause a debugger connection to fail.
- The memory safety issue with an empty call stack.

After that, remaining issues can be addressed in a separate cleanup branch before 1.4 release.

fatcerberus:
Okay, I reworded a couple comments and addressed the two issues above.  I dealt with the memory safety issue by adding an explicit `act != NULL` check before touching curr_pc, and changed `duk_require_uint` to `duk_get_uint`.

fatcerberus:
Oh, and SyntaxErrors can now be handled, as well.

fatcerberus:
Should I squash down some of the commits before merge?

svaarala:
You can squash if you prefer, I can merge either way :)

fatcerberus:
Alright, done.  How does it look now?

svaarala:
Looks good - but I think I found one more small issue with pushing `act->func` in `duk_debug_send_throw()`: `act->func` is NULL for lightfuncs.

fatcerberus:
Ah, didn't realize that, thanks.  What should I send for the filename and line number in that case? `undefined`?

fatcerberus:
Done.

svaarala:
If you don't change anything except use `duk_push_tval(ctx, act->tv_func)` instead of `duk_push_hobject(ctx, act->func)`:

- `.fileName` will come out as undefined and result in the string "undefined".
- Line number will come out as zero: _Pc2line will be NULL and `duk_hobject_pc2line_query()` will return zero.

I think this is OK at this stage, lightfuncs are not that user friendly anyway because they can't really hold much metadata.

A slight improvement over this baseline would be to read `.name` instead of `.fileName` (or perhaps try `.fileName` and fall back to `.name`): lightfuncs have a virtual `.name` like `lfunc_XXX`. This identifies the function to some extent: if you have a symbol map you can figure it out.

Anyway, for this merge, as long as there's no segfault it's good to merge. The minimum change is the `duk_push_tval()` so that there's no NULL pointer crash.

svaarala:
Oops, that needs to be `duk_push_tval(ctx, &act->tv_func)`.

fatcerberus:
Let's try that again.  I was wondering why the test failed...

svaarala:
Ok, I'll merge this now. I'll test the lightfunc handling in practice in master and fix if necessary.

fatcerberus:
There are a few warnings related to uninitialized variables, I didn't get a chance to fix them before you merged.  Just a heads-up. :)

svaarala:
Ok, I'll fix them.

svaarala:
Hmm, I got the fatal notify but the debugger didn't pause in my test.

svaarala:
Well, I'll figure it out later tonight and test the lightfunc thing.

fatcerberus:
I have a feeling the cause might be:
```c
if (fatal && !is_syntax_err) {
```

This is the cause of the uninitialized variable warning mentioned above, and depending on what garbage it contains (i.e. anything other than 0), will prevent the pause.

svaarala:
I replaced that with `if (fatal) {` though.

fatcerberus:
Huh, just tested it with the latest master in minisphere, it did pause for me.  Do you have a testcase?

fatcerberus:
This is why we really need the debugger regression testing. :)

svaarala:
Ok, it was caused by `duk_debug.js` not having support for the new notify. I'll fix the web UI to match the change in a separate cleanup pull.

fatcerberus:
It should still have sent a "paused" status notify though?

svaarala:
Yeah but `duk_debug.js` is strict and dropped the connection on an unknown notify. I changed it to ignore them instead. It kind of makes sense for it to be strict because it's assumed to be in lock step with the current debug protocol, but maybe it's better to allow unknown notifys.

fatcerberus:
Sorry about that, I would have updated that myself, but I'm not familiar with that codebase.  I can't get Node.js working under Cygwin. :(

svaarala:
Not a problem, I'll fix in a separate pull.

svaarala:
It'd be much preferable to have a headless regression client which was strict towards unknown notifys etc, though. The web UI didn't then have to serve that purpose. We'll see if there's time to work on that for 1.4 :-)

svaarala:
Now that this works in the debugger it's pretty nice to break on uncaught error :-)

fatcerberus:
Yes, the only thing that would make it better is if you could walk the callstack and view locals, etc. at each level.

Issue #314, opened 2015-08-30
Protect call handling and longjmp handling from finalizer side effects


svaarala:
The issue in #287 is for a finalizer side effect happening during TRYCATCH longjmp handling, and the fix for that was to prune TRYCATCH catch registers early so that they don't have finalizable values during longjmp handling.

On a more general level preventing finalizer execution during longjmp handling and call handling would be a good idea. In either case a finalizer call may change something critical in the execution state (even with call handling mostly storing and restoring relevant values), leading to very difficult bugs like #287.

There's now a torture test option for both refcount and mark-and-sweep finalizer execution, which triggers a few more issues than #287. The triggered cases are quite difficult to trigger in practice and may require the user to e.g. add a finalizer to `Object.prototype`. Even if a case triggered by a torture test cannot be triggered in practice it's best to fix it if possible.

As a quick test, all finalizer torture test errors were fixed by preventing finalizers during longjmp handling (using `heap->lj.type != 0` as the proxy for "longjmp handling active"). It's better to use an explicit state flag for this, and to also add finalizer side effect protection for call handling (although there are no known issues with that, even with torture options).

PR #315, opened 2015-08-31
Error handler callstack bump to fix GH-191

### LINKED ISSUE #191: Infinitely recursive function causes double error
The following:

```javascript
function f() { f(); }
f();
```

Causes Duktape to throw "DoubleError: Error in error handling".  It would be more informative if this could be fixed to throw a "Too much recursion" error like SpiderMonkey does.

fatcerberus:
This fixes #191.  Note that I couldn't figure out how to run the tests, so I can't see whether the test actually passes or not (it did when I tried the test code in minisphere).

### DIFF
--- a/src/duk_error_throw.c
+++ b/src/duk_error_throw.c
@@ -120,7 +120,7 @@ DUK_INTERNAL void duk_err_create_and_throw(duk_hthread *thr, duk_errcode_t code)
 
 	duk_err_setup_heap_ljstate(thr, DUK_LJ_TYPE_THROW);
 
-	thr->callstack_max = DUK_CALLSTACK_DEFAULT_MAX;  /* reset reclimit */
+	thr->callstack_max = DUK_CALLSTACK_DEFAULT_MAX;  /* reset callstack limit */
 	thr->heap->handling_error = 0;
 
 	DUK_DDD(DUK_DDDPRINT("THROW ERROR (INTERNAL): %!iT, %!iT (after throw augment)",
--- a/src/duk_error_throw.c
+++ b/src/duk_error_throw.c
@@ -47,10 +47,10 @@ DUK_INTERNAL void duk_err_create_and_throw(duk_hthread *thr, duk_errcode_t code)
 		 * for, e.g. a print() call at the deepest level.
 		 */
 		DUK_ASSERT(thr->callstack_max == DUK_CALLSTACK_DEFAULT_MAX);
-		thr->callstack_max = DUK_CALLSTACK_DEFAULT_MAX + 11;
+		thr->callstack_max = DUK_CALLSTACK_DEFAULT_MAX + DUK_CALLSTACK_GROW_STEP + 11;
 	}
 
-	DUK_ASSERT(thr->callstack_max == DUK_CALLSTACK_DEFAULT_MAX + 11);  /* just making sure */
+	DUK_ASSERT(thr->callstack_max == DUK_CALLSTACK_DEFAULT_MAX + DUK_CALLSTACK_GROW_STEP + 11);  /* just making sure */
 
 	/* Sync so that augmentation sees up-to-date activations, NULL
 	 * thr->ptr_curr_pc so that it's not used if side effects occur
--- a/src/duk_error_throw.c
+++ b/src/duk_error_throw.c
@@ -41,6 +41,17 @@ DUK_INTERNAL void duk_err_create_and_throw(duk_hthread *thr, duk_errcode_t code)
 
 	thr->heap->handling_error = 1;
 
+	if (!double_error) {
+		/* Allow headroom for calls during error handling (see GH-191).
+		 * We allow space for 10 additional recursions, with one extra
+		 * for, e.g. a print() call at the deepest level.
+		 */
+		DUK_ASSERT(thr->callstack_max == DUK_CALLSTACK_DEFAULT_MAX);
+		thr->callstack_max = DUK_CALLSTACK_DEFAULT_MAX + 11;
+	}
+
+	DUK_ASSERT(thr->callstack_max == DUK_CALLSTACK_DEFAULT_MAX + 11);  /* just making sure */
+
 	/* Sync so that augmentation sees up-to-date activations, NULL
 	 * thr->ptr_curr_pc so that it's not used if side effects occur
 	 * in augmentation or longjmp handling.
@@ -107,9 +118,10 @@ DUK_INTERNAL void duk_err_create_and_throw(duk_hthread *thr, duk_errcode_t code)
 	 *  Finally, longjmp
 	 */
 
-	thr->heap->handling_error = 0;
-
 	duk_err_setup_heap_ljstate(thr, DUK_LJ_TYPE_THROW);
+
+	thr->callstack_max = DUK_CALLSTACK_DEFAULT_MAX;  /* reset reclimit */
+	thr->heap->handling_error = 0;
 
 	DUK_DDD(DUK_DDDPRINT("THROW ERROR (INTERNAL): %!iT, %!iT (after throw augment)",
 	                     (duk_tval *) &thr->heap->lj.value1, (duk_tval *) &thr->heap->lj.value2));
--- a/tests/ecmascript/test-err-callstack-headroom.js
+++ b/tests/ecmascript/test-err-callstack-headroom.js
@@ -1,0 +1,47 @@
+/*
+ *  Error handling callstack headroom (GH-191)
+ *
+ *  When a RangeError is thrown due to the Ecmascript callstack limit being
+ *  reached, there should be enough headroom for at least 10 further recursions.
+ *  Prior to Duktape 1.3, this wasn't the case and a DoubleError would be
+ *  generated if the callstack limit was reached and, e.g. Duktape.errCreate
+ *  is set.
+ */
+
+/*---
+{
+    "custom": true
+}
+---*/
+
+/*===
+10
+9
+8
+7
+6
+5
+4
+3
+2
+1
+RangeError: callstack limit
+===*/
+
+try {
+	function recurse(n) {
+		print(n);
+		if (--n > 0) recurse(n);
+	}
+
+	Duktape.errCreate = function(e) {
+		recurse(10);
+		return e;
+	};
+
+	function f() { f(); }
+	f();
+}
+catch(e) {
+	print(e);
+}
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -24,6 +24,7 @@ their contributions under the Duktape ``LICENSE.txt`` (in order of appearance):
 * László Langó <llango.u-szeged@partner.samsung.com>
 * Legimet <legimet.calc@gmail.com>
 * Karl Skomski <karl@skomski.com>
+* Bruce Pascoe <fatcerberus1@gmail.com>
 
 Other contributions
 ===================


svaarala:
The usual tests I do before a merge is `make apitest` and `make qecmatest` which has decent coverage, especially if you run with assertions enabled (I just edit the Makefile locally to uncomment `-DDUK_OPT_ASSERTIONS`).

`make codepolicycheck` is also useful because it catches various nits; in particular it reports `FIXME` tags in code which I use as a mini task list in branches: as a first step I often just add FIXMEs for things to figure out. `make codepolicycheck` will then remind me what's still to be done. Finally, I rebase to make the commits clean for merge.

For 1.4 I'm going to add an automatic test run for pull requests, probably Travis at least initially, so that it'd be easier to see pull request status.

src/duk_error_throw.c
```diff
@@ -107,10 +113,11 @@ DUK_INTERNAL void duk_err_create_and_throw(duk_hthread *thr, duk_errcode_t code)
 	 *  Finally, longjmp
 	 */
 
-	thr->heap->handling_error = 0;
-
 	duk_err_setup_heap_ljstate(thr, DUK_LJ_TYPE_THROW);
 
+	thr->callstack_max -= DUK_CALLSTACK_DEFAULT_MAX * 2;  /* reset reclimit */
+	thr->heap->handling_error = 0;
```
svaarala: An assert here for the result being == DUK_DEFAULT_CALLSTACK_MAX would be useful to catch any issues where the restore path was somehow skipped, run twice, or whatever.

tests/ecmascript/test-err-callstack-headroom.js
```diff
@@ -0,0 +1,37 @@
+/*
+ *  Error handling callstack headroom (GH-191)
+ *
+ *  When a RangeError is thrown due to the Ecmascript callstack limit being
+ *  reached, there should be enough headroom for at least 10 further recursions.
+ */
+
```
svaarala: Could mark the testcase as being custom as it requires Duktape.errCreate:

```
/*---
{
    "custom": true
}
---*/
```

src/duk_error_throw.c
fatcerberus: Technically it might be enough just to set it back to DUK_CALLSTACK_DEFAULT_MAX directly and forgo the assert entirely (the entry assert would catch it on the second go-around if it were skipped).

tests/ecmascript/test-err-callstack-headroom.js
fatcerberus: Ah.  Will do.

fatcerberus:
I use Travis for minisphere, it's very nice.  Since I develop on Windows, it's useful as a way to ensure me I didn't break the build on Linux (if it builds on Linux it will usually build on OSX as well).

src/duk_error_throw.c
svaarala: Asserting and setting back to DUK_CALLSTACK_DEFAULT_MAX might be the best. It's good for the assert to fail early rather than possibly quite a long time before the next error occurs. Setting it back is probably better than staying off-sync when asserts are disabled.

svaarala:
I've also used Travis in other projects and it mostly works well. However, there are Duktape test runs that take hours and hours so it won't be enough by itself. Also it'd be nice to have test coverage for all three major platforms and one or two exotic ones (including a "mixed endian" ARM). But Travis is an easy first step.

src/duk_error_throw.c
fatcerberus: But keep in mind if the restore path is skipped, the assert will be also. :)

svaarala: Hmm, that's true - unless there are two successive exits for some reason. Doesn't sound all that likely, but I kind of prefer a "balanced" assert in these critical sections. Doesn't really matter I guess.

fatcerberus: Yeah, I'll add it, just wanted to make sure that caveat was on the table.

svaarala:
Ok, I can run the tests too as soon as I get off work.

fatcerberus:
Re: That ARM platform: There really are mixed endian platforms?  I thought those were a myth! ;-). Mixed endian never made sense to me, how do you make the hardware efficient with the bytes shuffled around like that?

svaarala:
I'm working with several of them. CPUs fetch data into cache lines anyway and shuffling is more or less cost free. That's why there are also configurable endianness processors.

fatcerberus:
Yeah, you'll have to run the tests, I can't get them going on Cygwin, there's a long string of these before it finally fails:
```
{ [Error: ENOENT, no such file or directory 'D:\tmp\runtests-7220-1']
  errno: -4058,
  code: 'ENOENT',
  path: 'D:\\tmp\\runtests-7220-1',
  syscall: 'open' }
```

Followed by:
```
Testcase execution done, running diffs
D:\src\duktape\runtests\runtests.js:571
            if (res.stdout !== res.testcase.expect) {
                   ^
TypeError: Cannot read property 'stdout' of undefined
    at <node stacktrace>
Makefile:523: recipe for target 'apitest' failed
make: *** [apitest] Error 1
```

svaarala:
Some API tests currently write to `/tmp` which is not nice. `make qecmatest` doesn't do that though.

Anyway, I'll run the tests.

svaarala:
Btw, add yourself to AUTHORS.rst. It's both a nice list of contributing people but also relevant to licensing.

fatcerberus:
Oh, no problem.  I should probably rebase this to master while I'm at it.

svaarala:
I'm running the tests now; it'll take a while for the assertions enabled test.

svaarala:
I think there are some obscure corner cases that will break with finalizer torture here: when the longjmp state is set up, there may be a finalizer (at least from the point of view of torture testing which sometimes generates cases that are not actually triggerable) which would cause another longjmp while preparing for this one.

But if that's the case I'll fix the issue as I deal with the other torture test issues in #314.

svaarala:
It might be safer to use `callstack_max = DUK_DEFAULT_CALLSTACK_MAX + 10` rather than `+=` so that if assertions are disabled and there'd be some control flow bug, recursive error calls wouldn't keep bumping the limit upwards (potentially without limit). Using a straight assignment would eventually break out of such a process.

fatcerberus:
Good point.

On a sidenote, I actually found I had to add a little more than 10, as Duktape apparently sometimes adds calls in-between.  For example the recursion limit is 10000 but the recursive `f()` function throws at ~9990.

svaarala:
You have Windows newlines in the test case at least - could you check that both source and testcase are clean of them? CR LF pairs cause the testcases to break now.

(I need to get that git attributes thing in place.)

fatcerberus:
Oh, hm.  I know Windows Git converts them to LF before commit, Cygwin Git apparently does not.  I use Notepad2 for editing, which preserves Unix newlines for existing files but creates new files as CR LF.

I'll fix that too. :)

svaarala:
I think that's related to the callstack growth - it throws when the callstack size would go over the maximum, e.g. if it were to resize from 9990 to 10020, that would trigger the callstack limit. It's intentionally tied to the resize so that it's only checked from time to time.

svaarala:
Codepolicycheck detects CR FL newlines, and covers API tests but not yet Ecmascript tests because the FIXMEs haven't been cleaned up yet. Maybe I could finally clean them up for 1.3 release so that newline errors would be easy to detect in pull requests.

svaarala:
This is the full whine BTW, there's also a non-leading tab on line 25:

```
vim +1     tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - /*
vim +2     tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - *  Error handling callstack headroom (GH-191)
vim +3     tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - *
vim +4     tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - *  When a RangeError is thrown due to the Ecmascript callstack limit being
vim +5     tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - *  reached, there should be enough headroom for at least 10 further recursions.
vim +6     tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - */
vim +7     tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - 
vim +20    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - ===*/
vim +21    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - 
vim +22    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - try {
vim +23    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - function recurse(n) {
vim +24    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - print(n);
vim +25    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - if (--n > 0) recurse(n);
vim +26    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - }
vim +26    tests/ecmascript/test-err-callstack-headroom.js : non-leading tab (idx 25) - }
vim +27    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - 
vim +28    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - Duktape.errCreate = function(e) {
vim +29    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - recurse(10);
vim +30    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - return e;
vim +31    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - };
vim +32    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - function f() { f(); }
vim +33    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - f();
vim +34    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - }
vim +35    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - catch(e) {
vim +36    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - print(e);
vim +37    tests/ecmascript/test-err-callstack-headroom.js : trailing whitespace - }
*** Total: 26 problems
```

fatcerberus:
I actually fixed all that (I saw the code policy update when I went to rebase), and also added an additional assert to verify the bump.  I didn't add an assert before the longjmp because it wasn't really clear what to do there.

Rebased and force pushed the branch.

fatcerberus:
Oops, just realized I forgot to make the test custom. :(

svaarala:
You can fix that or not -- the idea is that if comparison tests are executed for other engines, custom test errors won't cause warnings (because those engines won't be able to run the test). But the testcases are not "pure" in this respect now. So feel free to fix you need to touch something or leave as is if not.

svaarala:
There's now a basic Travis setup in master. For now it's just codepolicycheck, but should still be useful to catch trivia in pull requests. (If you rebase there should be a Travis run for the pull.)

svaarala:
Ok, apitest and qecmatest pass with assertions enabled (two tests time out with current timeout limits but that's normal).

fatcerberus:
Including the new test for this fix?

svaarala:
Yeah, I checked out the branch from your repo.

svaarala:
Hmm, you're still missing from AUTHORS?

fatcerberus:
Yeah, sorry, haven't been near a computer for the past few hours.  I'll fix it in a bit, along with adding a custom tag to the testcase.

svaarala:
Didn't mean to pester you - I'm going to sleep myself so no hurry from my side :) I added `apitest` and `qecmatest` to master Travis tests too, they take about 5 minutes in total. I'd like to run the tests with assertions enabled but that'd probably take too long to be the default for every push.

fatcerberus:
That must be why my debug builds got so slow recently - I enabled assertions.  They really do slow everything way down!

svaarala:
Yes, assertions are slow because they apply to many low level operations like refcounts. Also some pre- and post-checks are quite expensive but important for asserting state.

fatcerberus:
Okay, just rebased again and added myself to AUTHORS.rst.  Looks like all tests check out, except for one:

```
test-to-number.c: fail; 15 diff lines; known issue: number parsing bug for strings containing NUL characters (e.g. '\u0000')
```

Since it says "known issue" though, I assume this one isn't my fault. :-)

svaarala:
Yes, known issues are known issues and don't cause a test failure. I don't want to hide them from the test output though, because it'd be easy to forget they exist :)

fatcerberus:
That should do it.  The comment after the assert bugs me because it's so far past the 80-column mark but it didn't feel right adding a whole new line just for the short comment.

svaarala:
Looks merge ready except for one minuscule nit: "reset reclimit" should probably be "reset callstack limit", reclimit refers to the heap native call recursion limit.

fatcerberus:
Okay, that was easy enough. :)

fatcerberus:
Sorry if I kept you from sleep. ;-)

svaarala:
I kept myself from sleep improving the code policy checks and Travis stuff. Also webhooks in preparation for a repository-based Wiki setup :)

fatcerberus:
It occurs to me that the bug testcase for this is not strictly correct--the bump only guarantees 10 additional entries while the test actually requires 12--1 for the errCreate, 10 for the recurse() call and another for the print().

PR #316, opened 2015-08-31
Minor Fixes to Duktape - August 2015

kphillisjr:
Here is a few of the fixes for CMake that are highly useful. This does not include the main development tree changes for cmake.


svaarala:
The Windows debug socket is covered by #579, which uses a separate .c file to minimize the `#ifdef` sections.

The MSVC inline attributes are covered by #580:

```
#elif defined(DUK_F_MSVC)
#if defined(DUK_F_MSVC_FULL_VER)
#define DUK_NOINLINE        __declspec(noinline)
#define DUK_INLINE          __inline
#define DUK_ALWAYS_INLINE   __forceinline
#endif
```

svaarala:
I'm closing this as this can't be easily merged. As said above Windows debug socket support was added in #579 and MSVC attributes are tracked in #580. The CMake changes are no longer applicable because the dist process is now entirely Python2 based (util/make_dist.py).

PR #319, opened 2015-09-02
Updates to references

kphillisjr:
This is only minor changes to the references folder.


svaarala:
The Ecmascript 6 PDF is huge by the way (12MB) which is why I didn't add it earlier (on purpose).

If I was starting from scratch I'd probably leave all references out of the repo to keep it smaller. But there's already ES5.1 and draft ES6 in there so maybe merging this one more specification would be useful.

fatcerberus:
Honestly I think it would be better to remove the PDFs from the repo entirely and just add links where they can be read online.  If you link directly to the Ecma page such a link should be quite stable.

svaarala:
I added the specs originally because I work a lot offline (and with multiple laptops) and it's easy when the references are always there if you have the repo.

One compromise would be to add a Makefile target to fetch the relevant references automatically (e.g. "make references"). (Without rewriting history, the previous binaries will still be in the repo, unless you do a shallow clone.)

fatcerberus:
I can understand that, in a similar vein I develop across several different machines, but cloud sync software tends not to like Git repos very much so I end up generating way more commits than I probably should, essentially using GitHub as cloud storage. :)

svaarala:
Duh, at least ES5.1 link in http://www.ecma-international.org/publications/standards/Ecma-262-arch.htm is broken. This gives me 404: http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262%205.1%20edition%20June%202011.pdf

kphillisjr:
I believe the best solution is to see about having a secondary git repository to handle documentation requests like this one. This should help keep the primary git repository mostly clean of binary blobs ( like pdf files ).

fatcerberus:
Huh, that's weird.  All the other links are fine, but the ES5.1 PDF is nowhere to be found.  Someone really dropped the ball on that one. :)

fatcerberus:
The ES5.1 HTML link does work though.

svaarala:
@fatcerberus I was thinking about adding a `make references` target which would download the references for offline use. It could download the HTML version but the PDF version would be easier because it doesn't have stylesheets etc.

svaarala:
Makefile refs download tracked in #631.

PR #320, opened 2015-09-02
Utilility changes for Duktape ( version 2 )

kphillisjr:
This is a few of the utility changes that I had placed in my cmake branch that are fairly useful.


svaarala:
This can't be merged as is unfortunately (some of the changes are no longer applicable) so I'll close this. Please open separate pulls for individual features if that seems useful.

PR #326, opened 2015-09-04
Packed tvals are not possible on ppc64

andoma:
commit 422f66fb739734bd712de987ad9478c321e7b774 caused a regression for me using duktape on ps3 (ppc64). This patch fixes the problem.

### DIFF
--- a/config/header-snippets/packed_tval.h.in
+++ b/config/header-snippets/packed_tval.h.in
@@ -34,8 +34,8 @@
 #define DUK_USE_PACKED_TVAL_POSSIBLE
 #endif
 
-/* PPC: packed always possible */
-#if !defined(DUK_USE_PACKED_TVAL_POSSIBLE) && defined(DUK_F_PPC)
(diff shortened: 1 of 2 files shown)


andoma:
Even though the DUK_F_PPC64 define is unused now I leave it there as a reminder for you so you don't forget about us ppc64 users :-)

svaarala:
Thanks and sorry for the fallout :)

fatcerberus:
What is the difference between packed and unpacked tval?  I could never figure that out.

svaarala:
Packed duk_tval is 8 bytes long and can represent any IEEE double except that NaNs are replaced by a single "representative" NaN.  Tagged values other than numbers are represented as IEEE double NaN values so that 32-bit pointers and such fit into the same 8 byte value.

Unpacked duk_tval is usually 16 bytes long and is a straightforward type + value union.

Packed values are better on 32-bit targets and especially low memory devices.

PR #342, opened 2015-09-13
Add minimal fast return support

svaarala:
Add minimal fast return handling to avoid longjmp() calls for ordinary return cases.  While longjmp() is not terribly expensive on normal desktop hosts, it dominates execution when Duktape is Emscripten compiled and executed in a browser.

The condition for fast returns in this pull is:

* The return is not inside any try statements.  The particular problem with try-catch-finally is that the finally part may intercept (and even cancel) the return statement.

* The call is returning from an Ecmascript function to a calling Ecmascript function.  If the call returns to C code, a slow return is taken.

Label sites are allowed, so that fast return works from inside iteration statements which is important.

The fast return requirements can be relaxed further (in separate pulls):

* Try-catch statements can be allowed as long as there isn't an active finally part.

* Returning from Ecmascript code to C code could be allowed, if that was handled directly outside the longjmp handler.

It may also be that longjmp() will be removed later on from all RETURN opcode handling.  There's no real reason why it shouldn't be possible to handle all RETURN statements directly without resorting to a longjmp().  This will be done in separate pulls though.  The intent of this pull is to improve the call handling situation for Emscripten-compiled Duktape.

Tasks:

- [ ] Verify unwind handling so that it matches the longjmp path
- [ ] Full testcase runs with assertions and torture tests
- [ ] Resolve FIXMEs, ensure Travis run passes
- [ ] Release note entry


svaarala:
This pull request is still a work in progress, there are FIXMEs, etc. But, testing `tests/perf/test-call-basic.js` with and without fast returns has about 10x performance difference.

@JoshEngebretson: If you want to try this branch with Atomic, it would give a basic idea of how much this matters in practice.

fatcerberus:
@svaarala
You opened a pull request against your own repo?

svaarala:
Yes, pull requests make merges easier to reference, and it also facilitates discussion before a pending merge.

fatcerberus:
Incidentally, I noticed there was some "fast return" code already in Duktape, but `#ifdef`'d out.  Was this actually enabled at some point in the past?

JoshEngebretson:
I use PR's against repo for discussion as well as it runs CI on the branch before merging to master.  

@svaarala I should be able to test this week, Chrome is where the main issue is if you're running any kind of benchmarks yourself.  This was both on Windows and OSX.  Firefox also slows on the longjmp, though not nearly like Chrome.

svaarala:
In this repo Travis actually runs for all pushes (pull request or not), but for pull requests it's easy to see what their status is. What I don't like about pull requests is that none of their contents make it into the actual repo as part of a merge. But that's the case with issue discussion too, so it's no better when using branches only.

svaarala:
@fatcerberus Yes, I commented it out because I didn't have time to properly finish it earlier.

The earlier approach was to use a "fast return" when there's no catch stack at all to unwind. That would save a check at runtime altogether. But because the compiler now generates a label site for every iteration statement (in case they contain a break/continue) that's often not a good assumption. The fast path code in this branch allows label sites so it only falls back to slow return when a return happens from inside a function local `try` statement.

(One way to improve the current unwind code is to specify the catch stack unwind count directly in the RETURN opcode. It's known statically by the compiler so the executor could unwind the catch stack without first looking for the proper unwind target.)

The performance difference for ordinary targets was also not that significant, seems to about 15%, so it wasn't critical to finish the fast return. But for Emscripten the performance difference is huge because `longjmp()` needs to be emulated in apparently a pretty expensive manner.

svaarala:
@JoshEngebretson I did test the difference quickly using Node.js/V8 but not in a browser; we'll probably get the best application performance impact from your test.

Here's a little browser test using Dukweb (after disabling assertions).  Test code is:

```js
function test() {
    function f() { return; }

    for (var i = 0; i < 1e5; i++) {
        f();
    }
}

var t1 = Date.now();
test();
var t2 = Date.now();
print(t2-t1);
```

I'm running the code 5 times for each case to account for possible JIT effects (seems to affect Chrome a bit, Firefox not measurably).

Results with fast return:

- Firefox: 64, 63, 60, 60, 59
- Chrome: 280, 60, 66, 61, 70

Results without fast return:

- Firefox: 8943, 8951, 8936, 8933, 8939
- Chrome: 10055, 9581, 9584, 10177, 9597

There's a clear improvement in both cases, but these Emscripten results are quite sensitive to options it seems. I'm just using `-O2 -std=c99 --memory-init-file 0` at the moment.

So if these numbers are correct (it's quite easy to miss a hard refresh etc) and you can replicate them, there's hope it might improve your situation quite a bit :-)

JoshEngebretson:
Wow! Wow, wow, wow, wow!  Now I am excited to try this, did I say wow?

svaarala:
Let's hope you see a major improvement :)

The numbers above are of course measuring call handling only, so the improvement won't be as great for ordinary application code. There's also a chance I made some mistake in the test -- at one point I had assertions unintentionally enabled so I had to rerun the numbers. I did double check the assertion situation for the final run though.

By the way, increasing the loop count to `1e7` and comparing to x64 `gcc -O2` gives:

- Native x64: 1240
- Firefox with fast return: 5946

It's slower than native but still somewhat reasonable considering what's going on :-)

JoshEngebretson:
It does seem perhaps a bit smoother, but longjmp is still dominating the profiler in the #1 position:

screen shot 2015-09-13 at 5 03 11 pm

svaarala:
There are two probably causes for that I think - any idea which one is more likely:

- The fast return is now limited to returning from an Ecmascript callee to an Ecmascript caller. All function calls originated from C code will go through the slow path.
- Any return statements from inside a try statement won't use the fast path.

Both limitations will probably go away before 1.4 release but I'm not sure yet at what point, as it depends on the order of call handling reworks.

I can prototype the first case relatively quickly I think. It'd be interesting to see where the Emscripten bottleneck is so that we can be sure it's addressed in the reworks.

I'll push some more code for you to test later in the coming week :-)

JoshEngebretson:
I will revise that to say it is definitely improved, it was *really* hitching before on Chrome.  

This could be a factor: "The fast return is now limited to returning from an Ecmascript callee to an Ecmascript caller. All function calls originated from C code will go through the slow path."

Thanks Sami! :+1:

svaarala:
@JoshEngebretson We'll get to the bottom of this, no worries :-)

PR #345, opened 2015-09-14
Rework RETURN opcode handling to avoid longjmp() entirely

svaarala:
- [x] Merge current longjmp type `DUK_LJ_TYPE_RETURN` handler and fast return handler
- [x] Use the new return handler from RETURN opcode handling directly
- [x] Use the new return handler from ENDFIN too, to handle the case where the finally statement finished without affecting the completion being propagated
- [x] Check debug log levels, replacing `DUK_DPRINT` with less verbose calls, resolve FIXMEs
- [x] Documentation updates
- [x] Test case updates
- [x] Check Github issues for fast return references and update
- [x] Performance measurement
- [x] Assert and torture runs - initial assert-enabled apitest, qecmatest, test262 test, and refzero torture+assert qecmatest ok
- [x] test-conv-samevalue assert failure
- [x] Release notes


svaarala:
The commits in the branch are a very rough initial draft of the changes. There are a few issues to fix (e.g. a finally statement that captures a return and finishes without throwing doesn't work right yet). But since this is performance work, it might work well enough for @JoshEngebretson to try out and see if this matters in Atomic.

Testing this quickly using Dukweb again, test case:

```js
function test() {
    var arr = [];
    for (var i = 0; i < 1e3; i++) {
        arr[i] = i;
    }
    for (var i = 0; i < 1e3; i++) {
        // Each .forEach() call is a native call so that return 123 causes a
        // return-to-native situation which was previously a longjmp.

        arr.forEach(function () { return 123; });
    }
}

var t1 = Date.now();
test();
var t2 = Date.now();
print(t2 - t1);
```

Results without any fast return changes (= roughly 1.3.0):

- Firefox: 2703, 2682, 2716, 2664, 2717
- Chrome: 6409, 2838, 2806, 2829, 2828

Results from master (previous Ecma-to-Ecma return fast path included):

- Firefox: 2595, 2585, 2569, 2559, 2600
- Chrome: 6329, 2866, 2874, 2844, 2778

Results from this branch:

- Firefox: 987, 991, 992, 992, 994
- Chrome: 4746, 1550, 1500, 1494, 1519

The improvement is not as dramatic as for the initial fast return handling.

@JoshEngebretson It'd be interesting to hear how this affects your situation. In particular, does `_longjmp()` still dominate the profiler list? Does setjmp() have significance?

Longjmps are still used after this for ordinary errors, but also to handle yield, resume, break, and continue. I'll most likely rework break and continue to be longjmp-free too before 1.4 release. Yield and resume possibly too, although in most code they're quite rare in comparison to ordinary returns etc. I'll also see what we can do about reducing setjmp()s, as they're also relatively expensive.

svaarala:
Had to reintroduce the `DUK_LJ_TYPE_RETURN` as a pseudotype (not actually used for longjmp but to denote a continuation type) so that ENDFIN+RETURN case works correctly. In concrete terms, when we enter a finally part because of a RETURN and the finally finishes without errors (or an overriding return) we must resume propagating the previous return. The pseudotype is needed so that ENDFIN knows what it is propagating (error, normal value, return, break/continue, etc). There's already a pseudotype `DUK_LJ_TYPE_NORMAL` to denote a "normal" (non-abrupt) completion.

svaarala:
I'll let this one sit for a while until @JoshEngebretson has time to indicate if this helped with Atomic (which is a great test case for longjmp handling ;).

svaarala:
To figure out what the remaining longjmp() calls are with Emscripten, I did a small test with logging longjmp types when compiling Duktape with Emscripten and then running the result with Duktape. The "outer" Duktape log will then provide some information why Emscripten-generated code longjmps.

- With 1.3.0 there were 37977 longjmps.
- With master there were 3290 longjmps.
- With this branch there were 2903 longjmps.

The longjmps that remain with this branch are:

- Longjmp type 1 (actual error throw): 15
- Longjmp type 2 (slow break): 2816
- Longjmp type 3 (slow continue): 72

So, fast/slow break/continue handling is probably something that will be needed to make Emscripten-compiled Duktape to work well. The compiler emits a "fast break/continue" (which is just a JUMP) in situations where it can be sure that no label sites are crossed so that catch stack unwind is not necessary. It's not very intelligent about it, and will e.g. generate a slow break for a multi-level break/continue.

Here's an example case which generates a slow break now:

```js
function test() {
        var i;

 foo:
        if (1 > 0) {
                for (i = 0; i < 10; i++) {
                        break foo;
                }
        }
}
```

I can address this by either making the slow/fast break/continue detection more intelligent in the compiler - or simply by converting break/continue handling to avoid longjmp() entirely, as is done for RETURN in this branch. The latter approach seems best to me, because the compiler state needed doesn't (yet) exist and it will be useful for performance in any case.

Having eliminated longjmp() from break/continue handling, there'd only be 15 longjmps for running Duktape-with-Duktape.

svaarala:
Native (non-Emscripten) call handling difference to master is roughly:

- `test-call-basic.js`: 15.8 seconds -> 12.9 seconds
- `test-fib.js`: 9.1 seconds -> 8.5 seconds

The Fib test is actually dominated by a slow path function name lookup, so it doesn't benefit as much as the basic call handling test.

svaarala:
Humm, actually the break/continue counts are probably not an issue as such: they're reported by the outer Duktape so that when the same code is executed using e.g. V8 the break/continues in question are probably handled efficiently by V8. I don't know how Emscripten concretely emulates longjmp, but I'd probably do it by throwing an error to simulate the longjmp. Those error throws would then correspond to these (and other errors thrown/caught when setting up execution).

Anyway - I'll wait for Josh to report :)

JoshEngebretson:
I am working on getting some numbers, but currently am just too blown away playing with profiling C/C++ on the web :)

emscriptencppprofile

JoshEngebretson:
Ok, on this branch on Chrome, longjmp isn't showing up, and physics is top, also have a lot of idle time which may mean need to do something with the update loop as it stutters. On Firefox, is very smooth.

9651.8 ms	9651.8 ms	(idle)	
1504.4 ms13.96 %	1504.4 ms13.96 %	AtomicPlayer.js:241941__ZN15b2ContactSolver24SolveVelocityConstraintsEv	
870.2 ms8.08 %	870.2 ms8.08 %	(program)	
408.2 ms3.79 %	628.0 ms5.83 %	AtomicPlayer.js:195671_inverse_mdct	
368.3 ms3.42 %	427.0 ms3.96 %	AtomicPlayer.js:365548__ZN15b2ContactSolver24SolvePositionConstraintsEv	
314.6 ms2.92 %	3191.0 ms29.62 %	AtomicPlayer.js:245213__ZN8b2Island5SolveEP9b2ProfileRK10b2TimeStepRK6b2Vec2b	
265.9 ms2.47 %	308.4 ms2.86 %	AtomicPlayer.js:314508__ZN15b2ContactSolver29InitializeVelocityConstraintsEv	
236.0 ms2.19 %	1662.9 ms15.43 %	AtomicPlayer.js:26120_duk_js_execute_bytecode	
236.0 ms2.19 %	1662.9 ms15.43 %	AtomicPlayer.js:685605dynCall_vi	
236.0 ms2.19 %	1662.9 ms15.43 %	AtomicPlayer.js:11308invoke_vi	
229.7 ms2.13 %	229.7 ms2.13 %	AtomicPlayer.js:462658__ZNK14b2PolygonShape11ComputeAABBEP6b2AABBRK11b2Transformi	
212.2 ms1.97 %	259.7 ms2.41 %	AtomicPlayer.js:283419__ZN15b2RevoluteJoint24SolveVelocityConstraintsERK12b2SolverData	
196.0 ms1.82 %	485.6 ms4.51 %	AtomicPlayer.js:62112_start_decoder	
182.3 ms1.69 %	182.3 ms1.69 %	AtomicPlayer.js:210052_duk__get_own_property_desc_raw	
174.8 ms1.62 %	2087.4 ms19.37 %	AtomicPlayer.js:70525_duk_handle_call	
64.9 ms0.60 %	1789.0 ms16.60 %	AtomicPlayer.js:512750_duk_pcall	
43.7 ms0.41 %	204.7 ms1.90 %	AtomicPlayer.js:116983_duk_hobject_getprop	
21.2 ms0.20 %	141.1 ms1.31 %	AtomicPlayer.js:677428dynCall_iiii	
13.7 ms0.13 %	11.2 ms0.10 %	AtomicPlayer.js:648517_duk_call_method	
12.5 ms0.12 %	56.2 ms0.52 %	AtomicPlayer.js:85212_duk_hobject_putprop	
8.7 ms0.08 %	20.0 ms0.19 %	AtomicPlayer.js:515327_duk_call	
6.2 ms0.06 %	147.3 ms1.37 %	AtomicPlayer.js:232654_duk_new	
1.2 ms0.01 %	1.2 ms0.01 %	AtomicPlayer.js:568442__ZN6Atomic11JSComponent16CallScriptMethodERKNS_6StringEbf	
1.2 ms0.01 %	149.8 ms1.39 %	AtomicPlayer.js:257387_duk_hobject_delprop	
1.2 ms0.01 %	1.2 ms0.01 %	AtomicPlayer.js:465350_duk__finalize_helper	
166.0 ms1.54 %	166.0 ms1.54 %	AtomicPlayer.js:452698_imdct_step3_inner_r_loop	
154.8 ms1.44 %	154.8 ms1.44 %	AtomicPlayer.js:377031__ZN15b2RevoluteJoint24SolvePositionConstraintsERK12b2SolverData	
152.3 ms1.41 %	746.6 ms6.93 %	AtomicPlayer.js:116983_duk_hobject_getprop	
134.8 ms1.25 %	551.8 ms5.12 %	AtomicPlayer.js:596927__ZN6b2Body19SynchronizeFixturesEv	
126.1 ms1.17 %	397.0 ms3.68 %	AtomicPlayer.js:10392_usleep

JoshEngebretson:
Sorry for the horrible formatting on that profiler data...  Uff da!

svaarala:
@JoshEngebretson So once this branch makes it to master, would Emscripten performance be workable for you?

JoshEngebretson:
Here is profiler data for a more script heavy example, duktape is on the top:

2504.6 ms18.19 %	8812.2 ms63.99 %	AtomicPlayer.js:26120_duk_js_execute_bytecode	
1052.3 ms7.64 %	1571.1 ms11.41 %	AtomicPlayer.js:116983_duk_hobject_getprop	
644.8 ms4.68 %	2705.3 ms19.64 %	AtomicPlayer.js:85212_duk_hobject_putprop	
625.2 ms4.54 %	627.7 ms4.56 %	AtomicPlayer.js:178305_duk__get_identifier_reference

I am sure there is more to the "Duktape on Emscripten" story, though getting the longjmp out of there would be huge, yes.  Great work!

I need to try and figure out what is going on with Chrome being so stuttery, which may be something to do with the idle time.

svaarala:
Any idea how that would compare to a native example? For example, would the relationship between physics and Duktape be similar?

Emscripten performance is definitely an interesting issue: it's running VM on top of VM, so the performance characteristics are quite special. The first step will be to somehow automate Duktape-on-Emscripten performance test runs so that there's a better picture of the ballpark we're in.

JoshEngebretson:
I think that particular demo would be good to optimize in native VM before even looking at Emscripten's performance on it.  It hits property lookups pretty hard, as you can see.  I am thinking it is "similar" on Emscripten, would be good to test.  

You can see the loop here, nothing fancy: https://github.com/AtomicGameEngine/AtomicExamples/blob/master/BunnyMark/Resources/Scripts/main.js#L93

I have a good feeling about "current Duktape on Emscripten" performance, just working on smoothing out Chrome atm, which seems to be something with the main loop.

One thing that would be awesome is perhaps some optional "extra" api for interacting with the browser JS, this could be used to optimize hotspots, etc.  Just like C/C++ is used when embedding duktape for native apps.  Web developers are going to want access to the browser here, so this is something we need to address.

svaarala:
Regarding that automated test ability, the `emduk` target works and provides a command line Duktape-on-Emscripten utility but with two limitations:

- It cannot read files
- It can only read up to 256 bytes from stdin

These together mean it can't run existing testcases as is. Access to files should be fixable, but I'll need to dive more into Emscripten to see how that needs to be done. Ideally it'd just be able to read the test cases from the host filesystem using `fopen()` and friends. With that change in place I'll be able to follow what impact changes have on Emscripten microbenchmarks.

svaarala:
@JoshEngebretson Calling into the native VM for hotspots sounds interesting. There'd be several ways to do that, and one concrete issue would be figuring out how to handle typing.

If there was such a mechanism, it would also be very useful to replace the pretty hacky bridge between Duktape and the native JS environment used by Dukweb.js.

svaarala:
Added #354 to track that "call host JS directly" feature.

svaarala:
Property access performance is another area with little performance optimization yet. I'll work on call handling in the 1.4 release and look at the property handling in the next release.

JoshEngebretson:
Property access optimization I think would be a significant win as this is a very common operation :)  Glad that is on the roadmap. It should help some of the "benchmark examples" be a tad more impressive.

fatcerberus:
Nice, the longjmp-free return handling is in master now. :)  I'll have to try this out, see if it improves performance any on x64.  I know this was more an optimization for Emscripten, but I'd imagine it's at least a small improvement elsewhere too.

svaarala:
There's a small improvement for call handling but actually some performance loss for raw opcode dispatch. There's no logical reason for that, i.e. there's no diff there, but the executor is quite sensitive to even small changes in code alignment, case block sizes, etc. I'll do a final optimization/tweaking pass on that later in the release.

svaarala:
Looking at the diff to 1.3.0 more carefully, there is actually one performance relevant change in opcode dispatch, introduced by the break-on-uncaught debugger feature:

- https://github.com/svaarala/duktape/blob/master/src/duk_js_executor.c#L2261-L2271

This flag check executes for every dispatch even if a debugger is not attached so it roughly doubles the interrupt counter performance hit when debugger support is enabled. A better approach would be to let the interrupt counter run and move the check to when the interrupt counter reaches zero.

I'll fix this in a branch after the BREAK/CONTINUE longjmp changes.

When you disable interrupt counter and debugging, base dispatch performance matches 1.3.0.

fatcerberus:
Ah, that makes sense.  I should have realized that, the executor is the primary hotpath, even a seemingly trivial `if` check there will hit hard.

svaarala:
It did come up during the branch review but we postponed the cleanup to later :)

PR #347, opened 2015-09-14
Debugger throw notify cleanups

svaarala:
Cleanups and additions to #286:

- [x] Add support for Throw notify to debugger web UI
- [x] Minor fixes to compilation warnings -> separate small pull #351, to make master clean
- [x] Add missing `DUK_USE_DEBUGGER_SUPPORT` guards where missing
- [x] DUK_USE_xxx flags for supporting Throw notify, and for default automatic pause-on-uncaught-throw behavior (alternatively, add control flags to debug protocol)
- [x] Test caught and uncaught throw from lightfunc
- [x] Change lightfunc handling to so that lightfunc `.name` is reported in Throw => This might not actually make sense: .fileName is used to locate sources in the web UI (and other debug clients) and lightfunc virtual name is not very useful for that. Callstack shows lightfunc virtual name anyway.
- [x] Fix paused-in-lightfunc assertion failure
- [x] Change DEBUGGER opcode to use `duk_debug_halt_execution()` to fix off-by-one PC
- [x] Add `duk_safe_to_hstring()` helper and make debugger code use it instead of `duk_safe_to_string()` + `duk_require_hstring()`
- [x] Fix other instances of `duk_write_xxx(..., duk_require_xxx(...))` to use assert + `duk_get_xxx()` instead (throwing an error is not OK)
- [x] Review `duk_debugger.c` with respect to the topmost activation being a non-Ecmascript function
- [x] Test all debugger commands while paused in a native function, e.g. StepOver/StepInfo?
- [x] Releases entry


svaarala:
I tested the lightfunc handling quickly by enabling `-DDUK_OPT_LIGHTFUNC_BUILTINS` and using the following test:

```js
function test() {
    // error throw from native lightfunc
    decodeURIComponent('%ff%ff');
}
test();
```

This causes the following assert now:

```
[...]
[D] <20260673918647> duk_error_misc.c:99 (duk_err_setup_heap_ljstate): throw with debugger attached, report to client
[D] <20260674346011> duk_error_misc.c:107 (duk_err_setup_heap_ljstate): throw will be fatal, halt before longjmp
PANIC 54: assertion failed: h != NULL (duk_api_stack.c:4135) (segfaulting on purpose)
```

svaarala:
Using valgrind the culprit is here:

```
PANIC 54: assertion failed: h != NULL (duk_api_stack.c:4135) (segfaulting on purpose)
==12573== Invalid write of size 4
==12573==    at 0x402108: duk_default_panic_handler (duk_error_macros.c:120)
==12573==    by 0x4227EC: duk_push_hobject (duk_api_stack.c:4135)
==12573==    by 0x4469B7: duk_debug_send_status (duk_debugger.c:834)
==12573==    by 0x4488B1: duk_debug_halt_execution (duk_debugger.c:1725)
==12573==    by 0x4498A3: duk_err_setup_heap_ljstate (duk_error_misc.c:108)
==12573==    by 0x449C68: duk_err_create_and_throw (duk_error_throw.c:121)
```

So, this is caused by the debugger being paused in a lightfunc, and the existing Status notify code doesn't handle that correctly. This wasn't previously possible because we could only pause inside an Ecmascript function. I'll fix this and review any other potential gotchas.

svaarala:
Ok, the issue was the same `duk_push_hobject(ctx, act->func)` as in #286. After fixing that (happened in two places) seems to work. This is what the paused state looks like in the web UI:

lightfunc_paused

So, `.fileName` is undefined (which is actually more or less correct for lightfuncs - we have no idea). The lightfunc virtual name is shown in the callstack which is nice.

svaarala:
Hmm. Stepping while inside a native function doesn't seem to do much good. Not sure what it really should do. StepOut is clear but Step Into / Step Over less so.

fatcerberus:
Personally I would probably expect any of the step operations to work the same as a StepOut in that scenario, since a C call is logically one indivisible Ecmascript instruction.  Although wouldn't any of the stepping operations just result in longjmp() being called anyway?

fatcerberus:
Hm, although if the native function then calls back into Duktape... Hm.  You're right, it's a bit of a puzzler.

svaarala:
This is what happens now with e.g. StepOver:

```
[D] <23902390637678> duk_debugger.c:1043 (duk__debug_handle_step): debug command stepinto/stepover/stepout: 21
[D] <23902391040086> duk_debugger.c:1053 (duk__debug_handle_step): cannot determine current line, stepinto/stepover/stepout ignored
[D] <23902391454860> duk_debugger.c:1768 (duk_debug_process_messages): paused, process debug message, blocking if necessary
```

svaarala:
This behavior may need fixing to match principle of least surprise. OTOH not doing anything (and especially nothing unexpected) is not a bad default.

svaarala:
Well, there's time to mull it over before 1.4 release :)

fatcerberus:
Hm, just musing here but perhaps the DEBUGGER opcode could be reworked now to use `duk_debug_halt_execution()`?  This would allow it to visually stop on the debugger statement itself instead of the next statement.

svaarala:
That's definitely worth trying :) I wonder how Step commands work with that - I'll check it out.

svaarala:
@fatcerberus I'm prototyping using `duk_debug_halt_execution()` for the debugger statement.

I'm having a bit of second thoughts about the `use_prev_pc` parameter and `act->curr_pc` tweaking in the halt handler. It will be a problem if the debugger gains the ability to force a jump to a certain PC where to resume execution: the restoring code will then break that behavior.

Maybe it would be better to just use a flag (`thr->heap->dbg_xxx`) for indicating what to do with the topmost activation's PC within the debugger code: use as is, or subtract one.

svaarala:
After a pc2line attribution issue for the compilation of debugger statement, it seems `duk_debug_halt_execution()` would make debugger statement work as expected :)

fatcerberus:
That's good. :)

Agreed that a flag would be more flexible, yes.  Either way we do still need the ability to report "previous PC" in the status/throw notifys.

svaarala:
Yes, I agree - the debugger statement needs the same ability so it's certainly necessary.

I probably won't rework that curr_pc handling in this branch, but I'll add a note and maybe open a separate pull for that. Just quickly looking at how to do that it'd be a bit messy.

svaarala:
Stepping seems to work correctly when debugger statement is implemented using the halt primitive also.

svaarala:
Tested all debugger commands from inside a native function:

- Nothing breaks anymore
- Step commands are no-ops, including StepOut, because they all conclude that "cannot determine current line, stepinto/stepover/stepout ignored" -> #357

This branch is now merge ready and I'll merge it in a while unless something is raised.

@fatcerberus Please note that Throw notify is enabled by default but "pause on uncaught" behavior is not because it's a change to previous behavior. Use `DUK_OPT_DEBUGGER_PAUSE_UNCAUGHT` (`DUK_USE_DEBUGGER_PAUSE_UNCAUGHT` if you're modifying `duk_config.h` directly) to get that behavior.

fatcerberus:
No problem, I can easily add another #define to my build. :). Actually I need to come up with a better system than editing duk_config.h directly because it keeps getting overwritten every time I upgrade Duktape (sometimes you add things to it so I can't skip it)

fatcerberus:
In this commit:
https://github.com/svaarala/duktape/commit/0419666511772fdcaff017ebd85531114d2cba87

The throw notify defines don't seem to be consistent between the makefile and matrix script.

svaarala:
@fatcerberus If you're managing `duk_config.h` directly, I'd suggest the second method described here:

- https://github.com/svaarala/duktape-wiki/blob/master/Configuring.md#duktape-13

In essence, genconfig will automatically use defaults for any `DUK_USE_xxx` options that you don't specifically override. So you'll get the correct defaults for all new options, and only need to add them to your overrides if you don't like the default.

svaarala:
Good catch on the matrix test option, fixed.

fatcerberus:
Incidentally, using the current set of debugger commands is there a way to determine whenever you're stopped in a native or ecma call, since that is now possible?

svaarala:
Hmm, I don't think so.

svaarala:
It would of course be possible to add flags and additional information to the callstack frames to e.g. distinguish native and Ecmascript functions. But I'm afraid that will be a pretty awkward process: since the internals change, the flags and fields would change and all debugger command formats involved would need constant updating.

I think the best way to address this (and similar issues) is to figure out the safe heap walking process, and then just provide a function reference in the callstack information. The debug client can then inspect the function, read its properties, etc, so that all that stuff doesn't need to be effectively replicated in several debugger command formats.

With this approach the protocol remains pretty easy to maintain (just ensure relevant references are in place), and the rest would be up to the debug client. For example, a debug client might have a "value browser" so that you could click on a callstack function reference and it would be browsable just like any other value. The same would be true for any value/object reference anywhere in the protocol.

svaarala:
Added a Github issue for heap walking (#358). It's mentioned in the future work but it's a pretty important feature to track.

fatcerberus:
A hacky way to do it now might be to request bytecode for the current function (GetBytecode request) but of course for a large Ecmascript that would send the entire thing over the wire which might be inefficient...

svaarala:
Yeah. It would also be trivial to add a new command or some new fields to existing commands. But there'd a dozen similar additions down that road, which I'd like to avoid, and focus effort on a more general improvement :)

Implementing a minimal set of object walk commands would probably be 2-3 days work, so it should be doable within the 1.4 release. The first step would be to figure out the design approach and implement something very minimal (perhaps just property enumeration and internal prototype access for objects). The second step would be extending/changing the other debugger commands to take advantage of the walk mechanism, e.g. add function references to callstack frames etc.

svaarala:
Config metadata versions are wrong (1.3.0, should be 1.4.0).

fatcerberus:
So... now I want to know how you did that--it correctly pauses on the debugger statement itself, which I knew using `halt_execution()` would allow.  But to my surprise StepOver/StepInto *also* works properly, stepping through the debugger statement to stop on the next line.  When I tried to do this fix myself, I got it to pause at the debugger statement, but it would then step _over_ the line immediately after it.  So clearly I was doing something wrong.

By the way, I was almost going to report a bad regression in this ("uncaught error" when attaching the debugger), but then realized I overwrote duk_config.h with a fresh copy, losing all my custom options in the process, including the debugger support.  Silly me!

So anyway, everything checks out on my end, so feel free to merge this at your leisure. :-)

svaarala:
It might be related to the pc2line information being inaccurate for the debugger statement which I fixed here: https://github.com/svaarala/duktape/commit/398b1b0c9eafcde42a503513273cf9bb92e7617a

Btw, what you described should actually have been the behavior before any changes: because execution stopped after DEBUGGER was executed, a step command would use the current line information (pointing to the next statement) when setting up step state - and a StepOver would then only be triggered once *that* incorrect line was finished.

The PC/line inaccuracy assigned the DEBUGGER statement with the *previous* opcode (previous line number usually), so that once I used `use_prev_pc==1`, the line *preceding* the debugger statement would be highlighted (not the one after it) :) I don't see why this would cause the problem you describe though, but maybe it's related.

svaarala:
Added a debugger support config example (with pause uncaught enabled) so one can simply:

```
$ python config/genconfig.py \
      --metadata config \
      --output duk_config.h \
      --option-file config/examples/debugger_support.yaml \
      autodetect-header
```

Once Travis finishes I'll merge this.

fatcerberus:
I think the difference might be that you issue a restart_execution after DEBUGGER opcode processing.  My attempt didn't do that, simply continuing on with the next instruction post-pause, which is probably what caused the misbehavior.

PR #348, opened 2015-09-14
Rework BREAK and CONTINUE opcode handling to avoid longjmp entirely

svaarala:
The compiler emits a "fast break/continue", which is just a plain JUMP, in common situations but falls back to a "slow break/continue" in other situations. A slow break/continue means emitting a BREAK or CONTINUE opcode, and the executor then using a longjmp() to propagate the break/continue outwards. Note that a finally block may capture and even cancel a break/continue - uncommon, but required functionality.

At the moment a slow break is emitted when there are any try-catch sites between the break/continue source and target, but also if the break/continue breaks out more than one level. See the compiler source for the exact conditions.

Implement changes to avoid longjmp() entirely in BREAK and CONTINUE (= slow break/continue) handling. While longjmp() is slower than ordinary code in desktop platforms, there's an enormous different for Emscripten-compiled Duktape executed using e.g. Chrome or Firefox.

- [x] Refactor current `DUK_LJ_TYPE_BREAK` and `DUK_LJ_TYPE_CONTINUE` handler into a non-longjmp-bound handler.
- [x] Use the new handler from BREAK/CONTINUE directly.
- [x] Also use the new handler from ENDFIN, in case a break/continue will propagate outwards after a finally block finishes without affecting the completion being propagated.
- [x] Check debug log levels, resolve FIXMEs
- [x] Cleanup up both RETURN, BREAK/CONTINUE handling
- [x] Documentation updates
- [x] Test case updates, add a test case covering all direct and ENDFIN control paths
- [x] Add perf testcase for fast and slow break/continue
- [x] Check Github issues for fast/slow break/continue references and update
- [x] Performance measurement, both desktop and Emscripten
- [x] Assert and torture runs
- [x] Test without fastints
- [x] Release notes

This will be merged after #345 (rebase after merging).

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1151,6 +1151,10 @@ Planned
   longjmp() calls, improving performance slightly on ordinary platforms and
   significantly on Emscripten (GH-342, GH-345)
 
+* Internal performance improvement: rework BREAK and CONTINUE opcode handling
+  to avoid longjmp() calls, improving performance slightly on ordinary
+  platforms and significantly on Emscripten (GH-348)
+
 2.0.0 (XXXX-XX-XX)
 ------------------
 
--- a/src/duk_heap.h
+++ b/src/duk_heap.h
@@ -52,12 +52,12 @@
 
 #define DUK_LJ_TYPE_UNKNOWN      0    /* unused */
 #define DUK_LJ_TYPE_THROW        1    /* value1 -> error object */
-#define DUK_LJ_TYPE_BREAK        2    /* value1 -> label number */
-#define DUK_LJ_TYPE_CONTINUE     3    /* value1 -> label number */
-#define DUK_LJ_TYPE_YIELD        4    /* value1 -> yield value, iserror -> error / normal */
-#define DUK_LJ_TYPE_RESUME       5    /* value1 -> resume value, value2 -> resumee thread, iserror -> error/normal */
-#define DUK_LJ_TYPE_NORMAL       6    /* pseudo-type to indicate a normal continuation (for 'finally') */
-#define DUK_LJ_TYPE_RETURN       7    /* pseudo-type to indicate a return continuation (for 'finally') */
+#define DUK_LJ_TYPE_YIELD        2    /* value1 -> yield value, iserror -> error / normal */
+#define DUK_LJ_TYPE_RESUME       3    /* value1 -> resume value, value2 -> resumee thread, iserror -> error/normal */
+#define DUK_LJ_TYPE_BREAK        4    /* value1 -> label number, pseudo-type to indicate a break continuation (for ENDFIN) */
+#define DUK_LJ_TYPE_CONTINUE     5    /* value1 -> label number, pseudo-type to indicate a continue continuation (for ENDFIN) */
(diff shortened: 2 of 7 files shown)


svaarala:
Nothing here yet except FIXME notes.

svaarala:
Draft of implementation now in branch (on top of unfinished RETURN rework branch). API and Ecmascript test cases pass without assertions at least.

JoshEngebretson:
Ok, we really don't have much usage of break/continue in any current examples, this is probably a result of the component model :)

svaarala:
A quick performance test for `tests/perf/test-break-slow.js`:

- Native: master 15.8 sec, this branch 11.0 sec
- Emduk (Emscripten+Node.js): master ~49900 sec, this branch ~330 sec

The value for emduk+master is a rough estimate: had to drop loopcount to 1/1000 and multiply result by 1000 to run the test reasonably quickly.

fatcerberus:
Wow, that's... quite a massive improvement!  longjmp() on Emscripten is **extremely** inefficient, it seems.

JoshEngebretson:
Very nice!  Glad to see it speeds native up too, would be interesting to see how longjmp affects Android/iOS

fatcerberus:
I would love to try to compile minisphere for Android, it already compiles on Linux and I have a GS5 laying around somewhere, but everything I've read the build process is so involved...

svaarala:
The improvement in this branch is limited to "slow" break/continue which are not that common in practical code, so this is probably not a significant change for usual applications.  But it does reduce longjmp usage for Emscripten (which is critical for Emscripten performance) and overall it's better to rely less on setjmp/longjmp because it has a lot of downsides.

After this only actual throws, yield, and resume will use longjmps. It should be possible to convert yield/resume to be longjmp free too (this might actually matter to @JoshEngebretson?) but since they're currently ordinary function calls it needs a bit of thought.

fatcerberus:
What exactly is the criteria for a "slow" break/continue?

svaarala:
It's decided by the compiler: a "fast" break/continue is just a JUMP, a "slow" break/continue needs to check if there are catchers to unwind, finally blocks to execute, etc.

Here's the current check: https://github.com/svaarala/duktape/blob/master/src/duk_js_compiler.c#L5547-L5566

svaarala:
@JoshEngebretson The difference between an ordinary unwind and a longjmp-based unwind is measurable but not dramatic in native platforms, so I'd guess it'd be around 10-30% for a microbenchmark exercising that specifically.

Easiest improvements to call handling will probably come from reworking value stack handling, refcount handling for value stack "wipes", etc; these are quite inefficient now. I'll be looking into these shortly.

JoshEngebretson:
Right, guess I was thinking of the MacOSX  _setjmp and _longjmp kerfuffle which yielded nearly a 2x speedup

svaarala:
I probably won't do more longjmp rework in 1.4 release, remaining possible future work is in #373 and #374.

fatcerberus:
After all this is through I'm going to have to reacquaint myself with the Duktape codebase, I probably won't recognize it! :)

PR #349, opened 2015-09-14
Add mergeready tag to CONTRIBUTING

svaarala:
Using a mergeready tag tells others (repo owner in particular) that the changes are finished and are ready to be merged to master.  Without this indication it may be unclear whether changes should be merged or if more work or fixes are coming.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,6 +51,10 @@ To make a code contribution to Duktape
   - If test case status changes (tests are broken / fixed, test cases
     themselves needed fixing, test cases were added, etc), mention that.
 
+* A pull request can be created before you think your changes are finished.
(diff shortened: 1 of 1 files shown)


svaarala:
I'm working with some other repos with a similar convention and using a label works pretty well in explicitly communicating the merge-readiness of branches. It's better than just a comment because the label is visible when browsing pull requests.

The label can also be removed: if there's stuff to fix, the label can be removed until the fixes are done. Otherwise there's a risk I get trigger happy and merge before all the fixes are in (as happened with the debugger halt-on-throw changes).

svaarala:
The reverse approach would be to label pulls `inprogress` and then remove that label when work was ready. But I think it works better the other way, labeling when ready.

PR #360, opened 2015-09-17
Add hack "const" keyword support

svaarala:
Add a hack to support "const" so that it is treated exactly like "var". This is mainly useful for existing codebases, and a holdover until ES6 "const" is implemented.

- [x] Best solution for token numbers: `const` is an E5.1 reserved future keyword
- [x] Reject const without assignment (`const c;`)
- [x] Optional compilation => not implemented because const is part of ES6, and ES5.1 allows custom const semantics
- [x] Testcases
- [x] Documentation - website custom features / ES6 features, known limitations
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1141,6 +1141,10 @@ Planned
 1.4.0 (XXXX-XX-XX)
 ------------------
 
+* Add minimal support for "const" declarations with non-standard semantics,
+  intended mainly for minimal compatibility with existing code using "const"
+  (GH-360)
+
 * Add a debugger Throw notify for errors about to be thrown, and an option
   to automatically pause before an uncaught error is thrown (GH-286, GH-347)
 
--- a/src/duk_js_compiler.c
+++ b/src/duk_js_compiler.c
@@ -204,7 +204,7 @@ DUK_LOCAL_DECL duk_bool_t duk__nud_object_literal_key_check(duk_compiler_ctx *co
 
 /* statement parsing */
 DUK_LOCAL_DECL void duk__parse_var_decl(duk_compiler_ctx *comp_ctx, duk_ivalue *res, duk_small_uint_t expr_flags, duk_reg_t *out_reg_varbind, duk_regconst_t *out_rc_varname);
-DUK_LOCAL_DECL void duk__parse_var_stmt(duk_compiler_ctx *comp_ctx, duk_ivalue *res);
+DUK_LOCAL_DECL void duk__parse_var_stmt(duk_compiler_ctx *comp_ctx, duk_ivalue *res, duk_small_uint_t expr_flags);
 DUK_LOCAL_DECL void duk__parse_for_stmt(duk_compiler_ctx *comp_ctx, duk_ivalue *res, duk_int_t pc_label_site);
 DUK_LOCAL_DECL void duk__parse_switch_stmt(duk_compiler_ctx *comp_ctx, duk_ivalue *res, duk_int_t pc_label_site);
 DUK_LOCAL_DECL void duk__parse_if_stmt(duk_compiler_ctx *comp_ctx, duk_ivalue *res);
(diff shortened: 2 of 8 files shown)


svaarala:
Built-in string and token handling is quite awkward as is: a change needs to be applied to multiple places. Especially define renumbering is error prone. As a minimum measure the token defines and sanity checks should be autogenerated during build.

fatcerberus:
I'll try this out right away, thanks for the quick response!

How come there are two Travis checks for pull requests now?

svaarala:
One run tests the branch tip while the another tests what would happen if merged.

svaarala:
No worries, it's easy to test out hacks. It takes a bit more time to figure out if they can or should be merged :) No promises on that part yet :)

fatcerberus:
Just tried this out, `const` no longer generates a parse error and games using it run successfully.  This will be enough to allow me to close the const issue for minisphere 2.0, thanks.

svaarala:
Awesome :) Now we just have to figure out how to get this into master in a clean way.

fatcerberus:
Based on the diff the main difficulty looks like it will be the feature option: the keyword has to be classified differently depending whether const support is enabled or not, and the lexer doesn't seem to have any provisions in place to allow such special-casing.

fatcerberus:
Is `const` currently, in this hack, treated *exactly* like var?  i.e. Is it legal to do:

`const c;`

Because perhaps that shouldn't be the case. :)

svaarala:
That would work right now, yes. It's certainly not correct, but neither is:

```js
const c = 123.0;
c = 'foo';
```

What's the difference between the two, really? Both are incorrect, and not something one would normally use. :-)

fatcerberus:
Hm, I don't know, I guess I just see one as a necessary limitation of the underlying architecture (E5.1 has no concept of a read-only variable) whereas the other is more of a parser issue.

Ultimately the whole thing is a hack so anything we do short of implementing ES6 `const` properly is going to be incorrect, but for some reason the naked const declaration bothers me more than the writability.  *shrug*

fatcerberus:
I guess another thing is that it might be easy to accidentally declare a const with no value (in which case the lack of a syntax error could be surprising), whereas attempting to modify it is more a conscious act.

svaarala:
It should be easy to reject a const without an assignment - I'll try to fit that in. It's just there's little difference in principle IMO.

fatcerberus:
The only difference as I see it is that one is easier to do by accident than the other. :)

svaarala:
Ok, now the branch has the necessary smarts to reject const without initializer. The testcase is also updated to demonstrate how this currently works.

fatcerberus:
Sorry for giving you extra work, it's just that it's a mistake I could actually see myself making, as silly as that sounds. :)

svaarala:
No worries, I do this to make users happy (well, as happy as I can make them because they sometimes have conflicting needs :-).

svaarala:
This is still not merge ready, of course, but just trying to figure out a reasonable scope for the feature. 

I don't want to go into implementing the ES6 semantics in this branch because it'll affect the compiler internal metadata in several places - so the choice IMO is to either (a) merge a hacky const support at this point, or (b) postpone the whole const support until it can be implemented better.

fatcerberus:
For what it's worth, the semantics as implemented here match how SpiderMonkey (and possibly other engines) treated `const` prior to ES6 standardizing it, so there shouldn't be any surprises on that front so long as the writability is documented.

svaarala:
Well it's good if the behavior would match some existing engine(s) (even older ones) if the assumption is that this will be useful for legacy code.

Do you happen to have a link to something documenting the legacy behavior (it'd be good to link to).

svaarala:
I guess the main difference (apart from writability) is that this hack will "hoist" the declaration like `var` whereas ES6 `const` is scoped like `let`.

fatcerberus:
Here you go:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#Compatibility_notes

Writable "constants" are actually pretty common in JS-land, it seems.

svaarala:
Thanks - it seems indeed that this branch provides the previous Spidermonkey behavior :)

fatcerberus:
Another benefit to having this now: I can declare consts in code and Duktape will accept them, so that when real ES6 `const` is implemented, I automatically get the write-protection for free (potentially uncovering bugs that might have gone unnoticed with all constants declared `var`).

fatcerberus:
Oh, and the new branch works as expected.

fatcerberus:
Re: Hoisting: I believe SpiderMonkey did the same as well.  

Hoisting is one of those JS features I never got the point of.  I guess it makes the interpreter's job easier if all local variables are defined before any bytecode runs, but it can cause counterintuitive behavior in the case of, e.g. closures.

fatcerberus:
Can this be merged in its present state or is there still more to figure out?

Issue #362, opened 2015-09-17
Why is duktape.c distributed with '#line' cpp entries?


glima:
If it's purposed to be embedded in other projects at arbitrary locations? Generating debug packages spanning all the code for such derivative projects is then made clumsy: either you have to patch the fetched duktape release, removing those lines, or you get

`cpio: <arbitrary_path>/duk_util_hashbytes.c: Cannot stat: No such file or directory`

messages when rpmbuild is trying to generate debug packages.

svaarala:
One reason is to allow error messages to use `__LINE__` and still refer to the original source lines. For example:

```
((o) Duktape 1.3.99 (v1.3.0-44-ga6c3826)
duk> decodeURIComponent('%ff')
URIError: invalid input
	duk_bi_global.c:343
	decodeURIComponent  native strict preventsyield
	global input:1 preventsyield
```

Without the #line directives, that `duk_bi_global.c:343` would refer to the combined source and be less informative. It would be trivial to remove the #line directives, but it'd have some downsides like this.

For debugging one can provide the uncombined sources in e.g. gdb search path, so that you're actually debugging with the uncombined sources rather than the large single source.

Another option would be to build from the separate source files.

svaarala:
It would also be straightforward to create two different combined source files - one with #line directives and another without. The downside with that would be that it'd need to be explained and it gets somewhat technical so I'd prefer having either or.

Anyway, we can remove the #line directives if the benefits are worth it. I remember I didn't have them originally and some issue prompted me to add them, probably that traceback issue.

fatcerberus:
The #line directives are useful for reporting compiler errors and such... This way they can be traced back to the proper source file and line.  Otherwise the information is meaningless.

svaarala:
In this case there are two relevant source files though: the combined source and the individual source.

glima:
Yep, the reason for the individual source having this kind of info is void to me.

fatcerberus:
As mentioned in #363, removing the #line directives breaks debugging under MSVC because the combined source is >64k lines.  So no matter what is done here, something is going to be broken, unfortunately. :(

svaarala:
Ah, I was more or less clear that both are needed - but I'm not sure which should be the default.

svaarala:
This was fixed in 1.4.0 so closing the issue.

PR #363, opened 2015-09-17
Add a combined source without #line directives into the dist package

svaarala:
Related to #362. Prototype how duktape.c would work without #line directives:

- [x] Add optparse to combine_src.py
- [x] Add a `--line-directives` option to combine_src.py
- [x] Emit a file/line metadata file into dist/src/metadata.json
- [x] Disable line directives in dist combine_src.py step
- [x] Add an utility to resolve a combined source line into original file/line using dist/src/metadata.json
- [x] Drop #line directives, or provide both combined source files? => include both, as both variants have relevant use cases
- [x] Dist documentation changes
- [x] Other documentation changes; website note and point to wiki?

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1155,6 +1155,9 @@ Planned
 * Allow debugger detached callback to call duk_debugger_attach(), previously
   this clobbered some internal state (GH-399)
 
+* Add a combined duktape.c without #line directives into the dist package,
+  as it is a useful alternative in some environments (GH-363)
+
 * Fix "debugger" statement line number off-by-one so that the debugger now
   correctly pauses on the debugger statement rather than after it (GH-347)
 
--- a/dist-files/README.rst
+++ b/dist-files/README.rst
@@ -61,6 +61,10 @@ This distributable contains:
 * ``src/``: main Duktape library in a "single source file" format (duktape.c,
   duktape.h, and duk_config.h).
 
+* ``src-noline/``: contains a variant of ``src/duktape.c`` with no ``#line``
+  directives which is preferable for some users.  See discussion in
+  https://github.com/svaarala/duktape/pull/363.
+
 * ``src-separate/``: main Duktape library in multiple files format.
 
 * ``config/``: genconfig utility for creating duk_config.h configuration
--- a/util/combine_src.py
+++ b/util/combine_src.py
@@ -47,6 +47,8 @@
 import os
 import sys
 import re
+import json
+import optparse
 
 re_extinc = re.compile(r'^#include <(.*?)>.*$')
 re_intinc = re.compile(r'^#include \"(duk.*?)\".*$')  # accept duktape.h too
@@ -131,8 +133,13 @@ def processDeclarations(f):
 		elif line.data.startswith('extern int') or line.data.startswith('extern void '):
 			line.data = 'static ' + line.data[7:]  # replace extern with static
 
-def createCombined(files, extinc, intinc, duk_version, git_commit, git_describe, git_branch, license_file, authors_file):
+def createCombined(files, extinc, intinc, duk_version, git_commit, git_describe, git_branch, license_file, authors_file, line_directives):
 	res = []
+	line_map = []   # indicate combined source lines where uncombined file/line would change
+	metadata = {
+		'line_map': line_map
+	}
+
 
 	emit_state = [ None, None ]  # curr_filename, curr_lineno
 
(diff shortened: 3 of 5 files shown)


svaarala:
The file/line metadata map is useful for Duktape development: with a combined source traceback line it can resolved to an original source file/line, e.g.:

```
$ python util/resolve_combined_lineno.py dist/src/metadata.json 12345
duktape.c:12345 -> duk_builtins.c:793
```

svaarala:
I'll give a few days before merging this so that if anyone has a strong opinion one way or another they'll have some time to comment.

The concrete impacts of this change are:

- Debuggers will be less confused about file/line and source file mappings. The debugger will step through the combined source and not the individual sources as e.g. gdb did before.
- Some packaging tools (see #362) should work better without changes.
- Error tracebacks will contain `duktape.c` and a combined line number rather than uncombined source file/line.

To illustrate the traceback issue, before the change:

```
$ ./duk
((o) Duktape 1.3.99 (v1.3.0-45-g22a33b2-dirty)
duk> decodeURIComponent('%ff')
URIError: invalid input
	duk_bi_global.c:343
	decodeURIComponent  native strict preventsyield
	global input:1 preventsyield
```

After the change:

```
$ ./duk
((o) Duktape 1.3.99 (v1.3.0-42-ge19dec5-dirty)
duk> decodeURIComponent('%ff')
URIError: invalid input
	dist/src/duktape.c:29866
	decodeURIComponent  native strict preventsyield
	global input:1 preventsyield

$ python util/resolve_combined_lineno.py dist/src/metadata.json 29866
duktape.c:29866 -> duk_bi_global.c:343
```

The impact for troubleshooting is that if the Duktape source file/line matters, you'll need to figure out the uncombined file/line from the combined line somehow. The pull adds a utility to do that, and adds the necessary metadata into the dist, so that this will be relatively easy at least for official releases.

fatcerberus:
By the way MSVC behaves similarly to gdb, which threw me off at first: it won't step into duktape.c but instead prompts for the location of the original source files.  Normally I don't mind but it can become an issue when prototyping changes directly in the combined source as then the two copies don't match.

On the other hand, it adds an extra level of indirection when deciphering debug prints.  So there are pros and cons, and I'm not really sure which way is better.

svaarala:
For application programming the impact in tracebacks should be mostly minimal: tracebacks have a function `.name` too (`decodeURIComponent()` above) which usually identifies the relevant internal function quite well for troubleshooting.

With lightfuncs it's a bit different because the virtual lightfunc `.name` just contains a raw pointer and flags but no function name. This mostly matters when using `DUK_OPT_LIGHTFUNC_BUILTINS` because then almost all built-ins will be lightfuncs.

fatcerberus:
Hm, debugging may actually be hindered by this: It seems MSVC doesn't like source files with more than 64k lines:
https://connect.microsoft.com/VisualStudio/feedback/details/728010/visual-studio-10-c-debugger-doesnt-work-on-files-with-64k-lines

svaarala:
That sounds vaguely familiar. So, without adding both combined source variants we'll have to choose between either:

- Causing issues with some packaging tools
- Causing issues with MSVC debugger

We can add both file variants into the distributable if that's the only workable option though. It's just a bit awkward to explain. Concretely this would mean either adding a third source directory (so e.g. `src/`, `src-no-line-directive/`, and `src-separate/`, maybe abbreviated) or add a second file into `src/` (e.g. `src/duktape.c-no-line-directive` or `src/duktape-no-line-directive.c`).

glima:
No consensus here yet? :)

svaarala:
Any choice will lead to grumbling so I've yet to determine the path of least grumbling ;)

I'll try to get this merged soonish, I've been busy with some other stuff.

svaarala:
Ok, so as a compromise the pull now provides the following dist source structure:

- `src/duktape.c`: current single source version with `#line` directives
- `src-noline/duktape.c`: variant without `#line` directives, directory also contains headers etc
- `src-separate/`: current separate source version

The dist README has this note:

```
* ``src-noline/``: contains a variant of ``src/duktape.c`` with no ``#line``
  directives which is preferable for some users.  See discussion in
  https://github.com/svaarala/duktape/pull/363.
```

Hopefully this is a reasonable compromise :)

fatcerberus:
This seems reasonable.  I like that you chose to put it in a separate directory, keeps things neat.

svaarala:
Yeah, I thought it best not to change the `src/` directory; someone may be building with the wildcard `src/*.c` so having two `.c` files would have been an unwelcome surprise.

glima:
Thanks, everyone. I'll be updating for the next release on Soletta soon.

glima:
I'm a bit confused -- are there any releases contemplating this change yet?

fatcerberus:
It's in master, which means it will be in the next Duktape release (1.4.0).

glima:
Thanks a lot :)

PR #365, opened 2015-09-18
Migrate project list to Wiki

svaarala:


fatcerberus:
This is probably a good change, this way the projects can have more details added to them later without cluttering the main website.

PR #371, opened 2015-09-22
Fix debugger "lockout" in interrupt handling for better performance

svaarala:
The executor must prevent interrupt handling when (1) interrupt handling is *not* active but (2) debug handling is active. This may now happen at any time because we can call into the debugger message loop from anywhere.

The current check is in the executor hot path and is executed for every instruction dispatched. This shows up as a performance degradation from 1.3.0. It also makes the "opcodes executed" and "cumulative interrupt counter" numbers diverge in these special cases.

Fix so that the debugger-message-loop-in-progress check happens inside the interrupt-counter-zero handling, same as how nested interrupts are detected. This should also fix the interrupt counter issue (it needs manual verification).

Tasks:

- [x] Reimplement "lockout"
- [x] Test interrupt executor count match manually
- [x] Add explicit config option for debug/assert of executor count
- [x] Fix double executor halt issue in debugger Eval

### DIFF
--- a/config/config-options/DUK_USE_INTERRUPT_DEBUG_FIXUP.yaml
+++ b/config/config-options/DUK_USE_INTERRUPT_DEBUG_FIXUP.yaml
@@ -1,0 +1,10 @@
+define: DUK_USE_INTERRUPT_DEBUG_FIXUP
+introduced: 1.4.0
+default: false
+tags:
+  - development
+description: >
+  For Duktape development only: enable "interrupt fixup" in call handling
+  so that heap->inst_count_exec and heap->inst_count_interrupt can be
+  manually checked to match.  Only useful when debugging and/or asserts
+  are enabled.
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -1700,6 +1700,7 @@ DUK_INTERNAL void duk_debug_halt_execution(duk_hthread *thr, duk_bool_t use_prev
 	DUK_ASSERT(thr != NULL);
 	DUK_ASSERT(thr->heap != NULL);
 	DUK_ASSERT(DUK_HEAP_IS_DEBUGGER_ATTACHED(thr->heap));
+	DUK_ASSERT(thr->heap->dbg_processing == 0);
 
 	DUK_HEAP_SET_PAUSED(thr->heap);
 
--- a/src/duk_error_misc.c
+++ b/src/duk_error_misc.c
@@ -74,8 +74,9 @@ DUK_INTERNAL void duk_err_setup_heap_ljstate(duk_hthread *thr, duk_small_int_t l
 #if defined(DUK_USE_DEBUGGER_SUPPORT)
 	/* If something is thrown with the debugger attached and nobody will
 	 * catch it, execution is paused before the longjmp, turning over
-	 * control to the debug client. This allows local state to be examined
-	 * before the stack is unwound.
+	 * control to the debug client.  This allows local state to be examined
+	 * before the stack is unwound.  Errors are not intercepted when debug
+	 * message loop is active (e.g. for Eval).
 	 */
 
 	/* XXX: Allow customizing the pause and notify behavior at runtime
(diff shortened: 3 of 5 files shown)


svaarala:
Ok, this fixes the performance degradation from 1.3.0:

```
test-array-read.js            : duk.O2  4.08 duk.O2.130  4.13
test-array-write.js           : duk.O2  4.47 duk.O2.130  4.40
test-bitwise-ops.js           : duk.O2  0.86 duk.O2.130  0.86
test-break-fast.js            : duk.O2  3.73 duk.O2.130  3.59
test-break-slow.js            : duk.O2  9.52 duk.O2.130 13.64
...
```

There's just one run there but non-break/continue/return stuff is as before and slow break/continue handling is faster, as expected.

svaarala:
@fatcerberus Did you have a specific test where this debugger/interrupt handler lockout was needed? I can come up with a case, but if you have one in mind it'd be nice to check that this revised check still does what it's supposed to :)

fatcerberus:
Yes, the lockout is required for Eval commands to work properly when paused at an error.

fatcerberus:
More specifically, without the lockout, executing an Eval while halted will cause a corrupt debug message (Throw notify interleaved with Eval reply) to be sent if the Eval throws.

svaarala:
Right, so pausing on an uncaught error and executing a long enough Eval should ensure this works as before?

fatcerberus:
Right, also test if the Eval throws, a good test might be to try to JSON-serialize a cyclic object. ;)

svaarala:
Ok, will do that :)

svaarala:
It seems there was an existing bug in master for Eval throw when there's no active catcher in the call stack (i.e. error is considered fatal). The current "error is fatal" condition only considers Ecmascript catchers, so that Eval - while using a protected native call - is not considered a catching call.

Anyway, in this case an error notify was written (which is OK but perhaps odd), after which the debugger halt helper would also be called - leading to a "double halt".

I added a condition to disable the Throw notify / uncaught error pause handling when debugger message loop is active; it's not really sensible for Eval I guess.

With this change the revised check seems to work as expected, e.g. I can Eval freely when paused on a native uncaught error.

fatcerberus:
That's weird... It was my experience that an error during Eval was marked nonfatal (I got a green check mark in the minisphere error list, fatal = 1 shows a red X) and no double halt occurred.

fatcerberus:
Never mind, it turns out minisphere's Eval helper has a try/catch in it:
https://github.com/Radnen/spherestudio/blob/master/minisphere%20Remote/Duktape/DuktapeClient.cs#L232-L244

fatcerberus:
Note that the exclusion of native catchers is convenient--it allows halting to work as expected even when there is a top-level duk_peval() or duk_pcall(), which will usually be the case in any application which uses JS for main program logic (node.js style).

svaarala:
Yeah, I know - it's hard to know when the native catcher is a dump/rethrow and when it actually catches. So this is why a protected Eval is considered "uncatching". For Duktape internal call sites this would of course be fixable by indicating the intended "catching" style in the protected call.

svaarala:
That would break the debugger protocol if I'm (1) in the middle of writing a response, and (2) make a protected (!) call to do something. If (2) does any debug writing, in any situation, it will break the message being written. I think the cleanest policy would be to never do recursive debugger related handling.

This particular case would be fixable by making the condition for sending a notify "... and not writing a message", but I'm not sure it's worthwhile to make this too refined?

fatcerberus:
That's true, and now that I think of it the error is part of the Eval response anyway, so the notify is probably superfluous.

fatcerberus:
With this change, will a debugger Eval contribute to the instruction counters?  Because that might wreak havoc with, e.g. timeout handling or other sandboxing scenarios, especially if the user wants the sandbox to be relaxed while under debugger control.

svaarala:
That was already the case before the Throw notify branch: the interrupt executor cumulative count was expected to match this: https://github.com/svaarala/duktape/blob/master/src/duk_js_executor.c#L2325-L2327

The Throw notify merge broke this correspondence. This branch fixes it back to the way it was - which is not necessarily the best semantics possible, but it's consistent at least, and I don't want to leave it inconsistent in master.

The reliable cumulative interrupt counter is not used for anything right now but it might become useful for e.g. sampling-based profiling. I agree that the current semantics are not necessarily the best, as one might argue debugger executed instructions should be discounted. That can be changed later in several alternative ways.

For example, the interrupt counter can be as is, but we can get a cumulative count before and after the debugger message loop and keep track of "debugger executed instructions", so that we can then provide "cumulative count without debugger". This might be easier than changes to ignore interrupt counts etc: the debugger can kick in at any moment so doing that on the counter level might be much more difficult.

fatcerberus:
Ah, okay, that's fine then, keeping consistent behavior is always a good thing.  We can't go breaking things yet, it's not 2.0. :)

svaarala:
Somewhat related, many requests for debugger execution statistics, #375.

PR #380, opened 2015-09-24
Improve bytecode emitted for assignment expressions with identifier LHS

### LINKED ISSUE #372: Warnings in master when compiled with -Wall
From commit hash: 91d7830

gcc -fomit-frame-pointer -Wall -Wextra   -c -g -Werror -Ilib -std=c99 -fPIC  -MMD -MP -MF "build/Debug/GNU-Linux-x86/lib/duktape.o.d" -o build/Debug/GNU-Linux-x86/lib/duktape.o lib/duktape.c
duk_api_stack.c:2103:27: error: ‘duk_safe_to_hstring’ defined but not used [-Werror=unused-function]
duk_error_misc.c:12:22: error: ‘duk__have_active_catcher’ defined but not used [-Werror=unused-function]

svaarala:
In Duktape 1.3.0 a plain assignment with a number source and a register-bound target produced two bytecode instructions per assignment:

```js
x = 123.0;  // Duktape 1.3.0 produces: LDCONST + LDREG
x = 123;    // Duktape 1.3.0 produces: LDINT + LDREG
```

Improve the compiler to produce:

```js
x = 123.0;  // Produce: LDCONST directly to register of 'x'
x = 123;    // Produce: LDINT directly to register of 'x'
```

This reduces bytecode size and also improves performance for constant loading.

Tasks:

- [x] Careful review of coercions in compilation
- [x] Check if similar fix for LDTRUE, LDFALSE, etc would be easy
- [x] Try to replace the "target is register bound, operation is none" code path with a `ivalue_toforcedreg` so that no special handling is needed
- [x] Read through emitted bytecode
- [x] Share assignment handling code with "for (var i = ..." code path => not much overlap
- [x] Ensure testcases cover all paths in compiler, inspect bytecode
- [x] Add the necessary perf tests to see the effect, and allow regressions to be noted
- [x] Ensure "toplevel assign" check is correct
- [x] Assign perftest
- [x] Test with asserts (also test262)
- [x] Releases entry

Fixes #372.


fatcerberus:
Out of curiosity why is it currently implemented to (I assume) use a temporary if the bytecode indeed has the ability to load a constant directly into the target register?

svaarala:
If you just load the RHS into the LHS, and use *either* RHS or LHS as the assignment expression value, you'll potentially run into issues like #381.

The only easy and safe approach is to load through a temporary, and use that temporary as the value of the assignment expression itself. Avoiding this needs to be done carefully, and only when it's safe.

svaarala:
What I'm doing in this branch is figuring out how the respect the correct semantics but still generate the best bytecode possible for common "top level" assignments (whose expression value is not used).

fatcerberus:
That makes sense.  On the upside, I'd guess that the majority of assignments in real world code are top-level assignments (and JS doesn't have pass-by-reference thankfully), so even just optimizing for that case should be a meaningful improvement.

fatcerberus:
I'm guessing this is another thing that would be easier to handle if the compiler used an IR. :)

svaarala:
Yes, it'd be possible to figure out quite exactly if the assignment expression value was actually read by anyone. If so, it would only the be loaded into a temp if necessary.

With the current compiler I can detect that toplevel assignment situation specifically, but there are other safe cases which can't be detected (but they're quite rare in practical code).

svaarala:
Quick perf testing on assignments only:

```
test-assign-add.js            : duk-O2  5.24 duk-O2.master  9.07 duk-O2.130  8.77 duk-O2.123 11.00 duk-O2.113 10.56 duk-O2.102 11.40 rhino  3.72 mujs 38.75 lua  3.17 python 13.79 perl 25.03 ruby 10.12
test-assign-addto.js          : duk-O2  6.45 duk-O2.master 11.95 duk-O2.130  9.47 duk-O2.123 11.70 duk-O2.113 10.63 duk-O2.102 11.43 rhino  2.20 mujs 38.46 lua   n/a python   n/a perl 24.24 ruby 10.14
test-assign-const-int.js      : duk-O2  2.34 duk-O2.master  2.79 duk-O2.130  6.97 duk-O2.123  8.06 duk-O2.113  9.75 duk-O2.102 10.29 rhino  0.34 mujs  9.65 lua  2.11 python  5.47 perl 22.31 ruby  4.04
test-assign-const.js          : duk-O2  2.72 duk-O2.master  2.79 duk-O2.130  5.99 duk-O2.123  9.52 duk-O2.113 10.92 duk-O2.102 11.53 rhino  0.34 mujs  9.78 lua  2.11 python  5.43 perl 23.40 ruby  4.03
test-assign-literal.js        : duk-O2  4.55 duk-O2.master  7.01 duk-O2.130  7.01 duk-O2.123  8.59 duk-O2.113 10.75 duk-O2.102 11.48 rhino  0.33 mujs  9.98 lua  2.61 python 11.98 perl   n/a ruby  4.44
test-assign-reg.js            : duk-O2  2.82 duk-O2.master  5.32 duk-O2.130  2.76 duk-O2.123  4.30 duk-O2.113  5.31 duk-O2.102  5.52 rhino  0.32 mujs  8.83 lua  2.27 python  5.73 perl   n/a ruby  4.08
```

svaarala:
Basic const loading is now pretty close to Lua 5.1, which is nice. The remaining difference is caused partially by reference counting but dispatch may also be a little bit less efficient.

For `test-assign-literal.js` the difference is caused by the literal loads (like `LDTRUE`) being extra opcodes so that they go through double dispatch. Using a single dispatch level and tweaking the opcode bit allocations a bit (e.g. steal 1 bit from A) would help there.

PR #381, opened 2015-09-25
Fix assignment expression reg-bound RHS handling

svaarala:
When (1) assignment RHS is a register-bound variable, (2) it is also used elsewhere in the same expression, and (3) its value is changed also in the same expression, Duktape 1.3.0 and prior can emit incorrect bytecode. See attached bug testcase for an illustration.

Fix the bug, also outputs slightly better bytecode for some assignments.


fatcerberus:
What are the concrete ramifications of this bug?  You say "incorrect bytecode", but does that come with incorrect behavior too?

svaarala:
Yes, it wouldn't be a bug unless there was incorrect behavior.  The pull description refers to the bug testcase for an illustration.

fatcerberus:
> `print( (x = y) + (y = 30) );`

Interesting.  I would have expected this to be undefined behavior, it surprises me that the result is well-defined.  Maybe it's expectations from my C background coming through... :)

svaarala:
It's not very common in practical code, and wasn't covered by any existing test or test262. But it still needs fixing of course.

The required semantics are actually a bit awkward for producing the best bytecode possible, but I think I found a reasonable solution for that in #380, so that unnecessary intermediate results are avoided even for `x = y;` when the assignment is a "top level assignment" and not buried inside an expression.

svaarala:
Passes assertion tests (API + qecmatest), and test262test.

PR #383, opened 2015-09-27
Unroll perf test loops to reduce loop overhead

svaarala:
With less loop overhead the micro benchmarks provide clearer results on how the actual operation being tested behaves.


fatcerberus:
Loops have that much overhead that they can fuzz the perf tests like this?  How come?  Is it just because of checking the loop conditions on every iteration?

svaarala:
Yes, if the loop consists e.g. of 1 assignment only, more time will be spent in the loop condition checking than assignment.

PR #388, opened 2015-10-02
Add LuaJIT and Node.js (V8) performance comparisons

svaarala:
- Split perf output into more useful columns: Duktape current (with optimization variants), Duktape previous -O2, other interpreting engines, JIT engines
- Add sleeps to perftest targets to reduce effects on heat throttling when executed e.g. on a laptop


fatcerberus:
I'm actually curious now, with all the performance work, how Duktape stacks up against Lua under Emscripten.

PR #389, opened 2015-10-02
Change value stack init policy (garbage after current top)

svaarala:
- [x] Improve value stack negative/positive index resolution (use unsigned shared comparison)
- [x] Change value stack policy in API manipulation
- [x] Change value stack policy in garbage collection (refcount and mark-and-sweep)
- [x] Perf testcases to exercise value stack resizes
- [x] Test both possible policies (garbage/undefined above top)
- [x] If clear winner, remove other policy from code
- [x] Resolve FIXMEs
- [x] Assert runs
- [x] Releases


svaarala:
To elaborate a bit, the current value stack initialization policy is:

- Values between [top,end[ are initialized to "unused".

This is very old behavior and was useful when originally developing Duktape. Values above top can be dumped and can be seen to be properly initialized. Better policy candidates are:

- Values between [top,end[ are initialized to "undefined". The benefit of this policy is that when increasing value stack top, no actions are needed because the values are already "undefined". When decreasing value stack top, the previous value is DECREF'd and then overwritten as "undefined".
- Values between [top,end[ are considered garbage. The benefit of this policy is that when decreasing value stack top, we can just DECREF the previous value without rewriting it. But when value stack top increases we must overwrite garbage data with "undefined".

For reachability purposes only [0,top[ is considered reachable, [top,end[ is considered unreachable. (Previously [top,end[ was set to unused *and* was marked by mark-and-sweep.)

fatcerberus:
I don't understand how initializing to `undefined` is an improvement?  Duktape differentiates between "unused" and "undefined", but I thought as far as user code is concerned they were identical?

As for the second option, leaving garbage in there would make for faster unwinds, but at the expense of push performance.  Hard to say which is better.

svaarala:
When it's initialized to undefined and the value stack grows, nothing needs to be done.

When it's initialized to unused and the value stack grows, "unused" values need to be overwritten with undefined, which obviously takes longer than doing nothing :)

fatcerberus:
So hypothetically, if you didn't overwrite the "unused"s, they wouldn't simply read back as undefined?

svaarala:
As for performance I've measured the two options with refcounting enabled/disabled (4 options). Overall:

- Leaving values above top "undefined" performs better with refcounting enabled. This is somewhat predictable because increasing top is "free" and decreasing top does DECREF anyway, so combining DECREF with "set to undefined" is memory local.
- Leaving values above top "undefined" performs mostly better with refcounting disabled, but there are cases where "garbage" policy works better.

The differences for refcount disabled are small enough to probably not justify using a different value stack policy for refcounts and without refcounts. I have this behavior now behind a config option but I'll strip it out in a separate commit.

svaarala:
"Unused" is technically a different tagged type than "undefined", so that's not really an option.

I'm reworking that unused/undefined distinction after this pull anyway. It's only really needed for managing arrays, to indicate gaps in the array (which is different from having an actual undefined value). The API also uses "unused" for typing invalid indices and such but that has no internal impact.

svaarala:
Anyway, with the "leave as undefined" policy it works just as you describe: when increasing value stack top, nothing needs to be done except update a pointer.

svaarala:
Correction to above: the unused type is mapped to DUK_TYPE_UNDEFINED so it's not really visible in the API. But "unused" values don't get pushed to value stack in practice, they're only used for denoting missing entries in an array, and previously for unused parts of the value stack.

fatcerberus:
Not really related, but while I'm thinking about it: The debug protocol includes an "Unused" dvalue, so removing it entirely would be a breaking change.

svaarala:
Also, leaving garbage vs. undefined doesn't actually affect push performance at all: the push code will just overwrite the value and doesn't need to care about the old value's refcount/type anyway.

But the policy does affect what happens when value stack top is increased/decreased.

svaarala:
It doesn't need to be removed from the debugger protocol - you shouldn't encounter such values in practice even at the moment though.

fatcerberus:
So if you leave garbage on top, then when the valstack grows you need to initialize the new slots to undefined, same as now.  But *decreasing* top doesn't need to do anything.  The potential for faster unwinds is tempting here. :)

svaarala:
Here are some numbers for fun (x64 -O2):

- "und"=undefined above top , "grb"=garbage above top
- norc=no refcounting
- the testcases make function calls with an increasing target function nregs/nargs count (= calling a small vs. large target function)

```
test-call-basic-1.js          : duk.O2.vs_und_norc  9.18 duk.O2.vs_grb_norc  9.13 duk.O2.vs_und 10.02 duk.O2.vs_grb 10.05 duk.O2.master 10.06 duk.O2.130 12.11
test-call-basic-2.js          : duk.O2.vs_und_norc  9.04 duk.O2.vs_grb_norc  9.27 duk.O2.vs_und 10.13 duk.O2.vs_grb 10.11 duk.O2.master 10.16 duk.O2.130 12.24 
test-call-basic-3.js          : duk.O2.vs_und_norc 11.55 duk.O2.vs_grb_norc 11.82 duk.O2.vs_und 15.81 duk.O2.vs_grb 17.34 duk.O2.master 17.44 duk.O2.130 19.63 
test-call-basic-4.js          : duk.O2.vs_und_norc 19.17 duk.O2.vs_grb_norc 19.64 duk.O2.vs_und 35.69 duk.O2.vs_grb 43.58 duk.O2.master 43.31 duk.O2.130 46.00 
```

svaarala:
Decreasing the top *always* has to DECREF with refcounts enabled.

fatcerberus:
Hm, so the "undefined" policy seems to be the overall winner, based on those numbers.

svaarala:
Yes, that's what I said above :-)

Dropped one line though accidentally:

```
test-call-native.js           : duk.O2.vs_und_norc 14.69 duk.O2.vs_grb_norc 13.98 duk.O2.vs_und 15.61 duk.O2.vs_grb 15.86 duk.O2.master 16.53 duk.O2.130 19.66
```

Here, with refcounts disabled, garbage policy is slightly better. With refcounts enabled "undefined" policy is again (very slightly) better.

fatcerberus:
Sorry, I just woke up an hour ago and I'm still a bit hazy.  I'm not really a morning person. :)

svaarala:
Me neither :-) :+1:

PR #390, opened 2015-10-02
Change error accessor properties to make .fileName, .lineNumber, and .stack normally writable

### LINKED ISSUE #387: Unable to modify fileName and lineNumber in errCreate
I'm not sure if this is by design, but I've found that this doesn't work:

```javascript
Duktape.errCreate = function(e) { e.fileName = "the pig"; e.lineNumber = 812; return e; };
```

Similarly, using `duk_put_prop()` to modify those properties (my actual errCreate is implemented in C) doesn't work either.  However, this does work:
```c
	duk_push_string(ctx, "fileName"); duk_push_string(ctx, "the pig");
	duk_def_prop(ctx, 0, DUK_DEFPROP_HAVE_VALUE
		| DUK_DEFPROP_HAVE_WRITABLE | DUK_DEFPROP_WRITABLE
		| DUK_DEFPROP_HAVE_CONFIGURABLE | DUK_DEFPROP_CONFIGURABLE
		| DUK_DEFPROP_HAVE_ENUMERABLE | 0);
	duk_push_string(ctx, "lineNumber"); duk_push_int(ctx, 812);
	duk_def_prop(ctx, 0, DUK_DEFPROP_HAVE_VALUE
		| DUK_DEFPROP_HAVE_WRITABLE | DUK_DEFPROP_WRITABLE
		| DUK_DEFPROP_HAVE_CONFIGURABLE | DUK_DEFPROP_CONFIGURABLE
		| DUK_DEFPROP_HAVE_ENUMERABLE | 0);
```

The fact that the latter snippet works means the properties are configurable, but not writable, by default?  Seems odd, and makes error augmentation more difficult.

svaarala:
In Duktape 1.3.0 you can't simply overwrite an error instance's `.fileName`, `.lineNumber`, or `.stack`. The write will be captured by an inherited setter (whose getter counterpart virtualizes these property reads) which ignores the write:

```
$ ./duk.O2.130
duk> err = new Error('aiee'); err.fileName = 'dummy.js';
= dummy.js
duk> print(err.fileName);
input
= undefined
```

You could still set "own properties" overwriting these properties using Object.defineProperty:

```
$ ./duk.O2.130
duk> err = new Error('aiee'); Object.defineProperty(err, 'fileName', {value:'dummy.js',writable:true,configurable:true});
= Error: aiee
duk> err.fileName
= dummy.js
```

This pull changes the inherited setter so that it automatically invokes a `duk_def_prop()` call to allow ordinary property assignment to work as expected.

Note, however, that fileName/lineNumber values in a `.stack` dump are based entirely on the internal `_Tracedata` so that overwriting `.fileName` and `.lineNumber` won't change them in the stack trace.

Tasks:

- [x] Finalize change
- [x] Finalize testcase
- [x] Polyfill to restore 1.3.0 behavior
- [x] Polyfill to ensure 1.4.0 behavior
- [x] Website error description update
- [x] Releases entry

Fixes #387.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1185,6 +1185,10 @@ Planned
 * Add convenience API calls to detect specific error subtypes, e.g.
   duk_is_eval_error() (GH-340, GH-433)
 
+* Make Error instance .filename, .lineNumber, and .stack directly writable
+  to match V8 and Spidermonkey behavior; the previous behavior is provided
+  by polyfills/duktape-error-setter-nonwritable.js (GH-390)
+
 * Add a combined duktape.c without #line directives into the dist package,
   as it is a useful alternative in some environments (GH-363)
 
--- a/polyfills/duktape-error-setter-nonwritable.js
+++ b/polyfills/duktape-error-setter-nonwritable.js
@@ -1,0 +1,20 @@
+/*
+ *  Ensure Error .fileName, .lineNumber, and .stack are not directly writable,
+ *  but can be written using Object.defineProperty().  This matches Duktape
+ *  1.3.0 and prior.
+ *
+ *  See: https://github.com/svaarala/duktape/pull/390.
+ */
+
+(function () {
+    var err = new Error('test');
+    err.fileName = 999;
+    if (err.fileName !== 999) { return; }  // already non-writable
+
+    var fn = new Function('');  // nop
+    Object.defineProperties(Error.prototype, {
+        fileName: { set: fn },
+        lineNumber: { set: fn },
+        stack: { set: fn }
+    });
+})();
--- a/polyfills/duktape-error-setter-writable.js
+++ b/polyfills/duktape-error-setter-writable.js
@@ -1,0 +1,19 @@
+/*
+ *  Ensure Error .fileName, .lineNumber, and .stack are directly writable
+ *  without having to use Object.defineProperty().  This matches Duktape
+ *  1.4.0 behavior.
+ *
+ *  See: https://github.com/svaarala/duktape/pull/390.
+ */
+
+(function () {
+    var err = new Error('test');
+    err.fileName = 999;
+    if (err.fileName === 999) { return; }  // already writable
+
+    Object.defineProperties(Error.prototype, {
+        fileName: { set: new Function('v', 'Object.defineProperty(this, "fileName", { value: v, writable: true, enumerable: false, configurable: true });') },
+        lineNumber: { set: new Function('v', 'Object.defineProperty(this, "lineNumber", { value: v, writable: true, enumerable: false, configurable: true });') },
+        stack: { set: new Function('v', 'Object.defineProperty(this, "stack", { value: v, writable: true, enumerable: false, configurable: true });') },
+    });
+})();
--- a/src/duk_bi_error.c
+++ b/src/duk_bi_error.c
@@ -93,7 +93,7 @@ DUK_INTERNAL duk_ret_t duk_bi_error_prototype_to_string(duk_context *ctx) {
 	return 1;
 }
 
-#ifdef DUK_USE_TRACEBACKS
+#if defined(DUK_USE_TRACEBACKS)
 
 /*
  *  Traceback handling
@@ -114,7 +114,7 @@ DUK_INTERNAL duk_ret_t duk_bi_error_prototype_to_string(duk_context *ctx) {
 #define DUK__OUTPUT_TYPE_FILENAME    0
 #define DUK__OUTPUT_TYPE_LINENUMBER  1
 
-DUK_LOCAL duk_ret_t duk__traceback_getter_helper(duk_context *ctx, duk_small_int_t output_type) {
+DUK_LOCAL duk_ret_t duk__error_getter_helper(duk_context *ctx, duk_small_int_t output_type) {
 	duk_hthread *thr = (duk_hthread *) ctx;
 	duk_idx_t idx_td;
 	duk_small_int_t i;  /* traceback depth fits into 16 bits */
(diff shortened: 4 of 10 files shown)


fatcerberus:
Not modifying the trace data is good behavior, I think.  I often manipulate Error filename and line numbers so that the most relevant code is displayed in the debugger (e.g. invalid parameters are blamed on the caller), but having the REAL stack trace still available is useful.

svaarala:
Agreed. It'd actually be a bit hard to modify the tracedata because it lists a function and the *PC* in that function (not a line). That PC is then turned into a line number only later when looking up `.stack`. So to modify the stack trace one would need to either find a suitable replacement PC or change the tracedata format to begin with. The reason a pc2line lookup is not done when creating the tracedata is that most errors don't get their `.stack` accessed so that lookup would unnecessarily slow down error handling.

fatcerberus:
Hm, it actually might be preferable to *not* allow `.stack` to be changed directly.  One of the implied guarantees mentioned in the documentation is that `.stack` will always contain something useful and string-coercible (if not a stack trace, then at least the error message), and because this is documented, native code might actually rely on it.  Allowing it to be "modified" at will breaks the guarantee.

svaarala:
I don't see difference to `.fileName` and `.lineNumber` there though: user code expects them to contain useful values? One would only modify them if the new value would be more useful than the default one. Same goes for `.stack`?

This is actually one reason I like the current behavior: having to go through `Object.defineProperty()` ensures you don't overwrite the values by accident so that it's a conscious choice to change the values.

Overall I can think of three reasonable policies for `.fileName`, `.lineNumber`, and `.stack` protection:

1. Allow normal property write to work, either directly or using a transparent setter as in this branch.
2. Make the value configurable but require `Object.defineProperty()` or `duk_def_prop()` to change the value.
3. Make the value non-configurable and require `duk_def_prop()` with `DUK_DEFPROP_FORCE` to change the value. This would protect the values from change from all Ecmascript code.

Regardless of the default, user code can change the properties to conform to any of these simply by changing the property attributes and/or defining a setter.

fatcerberus:
For `.stack` specifically, I was referring to this bit in the guide:

> In builds where tracebacks are disabled, the `stack` accessor will return the same value as calling `toString()` on the error would. This means you can always print `e.stack` and get a useful output.

And later:

> Attempt to write to `stack` is silently ignored. You can still override the accessor by defining an own property of the same name explicitly with Object.defineProperty(). This behavior differs from V8 where `stack` is an own property of the Error instance, and if you assign a value to `stack`, the value reads back as assigned.

This is an explicit guarantee.  For `.fileName` and `.lineNumber` I see the "useful value" guarantee as less important, because they might not always contain useful data, and in fact never do in the case of:

* Duktape/C calls (fileName is undefined, lineNumber is 0)
* Eval code (fileName is "input")
* Lightfuncs (I think?)

svaarala:
The quoted sections are actually just trying to say that `.stack` is present and has some printable value (instead of being absent, an empty string, or something useless) even when stack traces are disabled. It's mentioned because one might equally expect `.stack` to be absent without stack trace support. It's not trying to say `.stack` is more "accurate" than `.fileName` or `.lineNumber`.

There's no guarantee made for what's actually in the stack trace, and this is intentional because it'd be very awkward to make guarantees there. Call stack structures change over time etc:

> The printable traceback format is intended to be human readable only. You shouldn't rely on an exact traceback format as it may change between versions. As an example of the current traceback format, the program: [...]

svaarala:
That bit about needing to use `Object.defineProperty()` - I didn't intend that to read as "this is good behavior and guaranteed", but rather to point out that the current behavior differs from V8 which might be unexpected.

But it's arguably part of versioning guarantees, so changing that would maybe be best to do in a major version bump.

I'd find it a bit odd if someone relied on this behavior though: it's not providing actual protection against changing, you just need to use `Object.defineProperty()` to do it.

fatcerberus:
Alright, that makes sense.

I have no further comments on this-- you can do whatever you feel is appropriate.  Maybe someone else could weigh in on this?  I'll continue using duk_def_prop() in the meantime.

svaarala:
By the way, when stack traces are disabled, the `.fileName` and `.lineNumber` properties are plain "own" properties of the error object. Their attributes are: writable and configurable. So when stack traces are disabled, you can write them directly without using `Object.defineProperty()`.

If this branch is not merged, it would be consistent to make them non-writable and configurable in master so they'd behave the same way as the current accessors do.

If this branch is merged, the behavior without stack traces is already consistent with the accessor change in this branch.

I'm quite happy with the behavior in master myself: I like the fact that accidental writes are prevented. The downside is that e.g. V8 behaves differently which may be an unnecessary surprise:

```
> err = new Error('aiee')
[Error: aiee]
> err.stack = 'foo'
'foo'
> err.stack
'foo'
```

fatcerberus:
How does Spidermonkey (Firefox) behave in this case?

svaarala:
Like V8: `.stack` is assignable:

```
err = new Error('aiee');
Error: aiee
Stack trace:
@debugger eval code:1:7

err.stack
"@debugger eval code:1:7
"
err.stack = 'foo'
"foo"
err.stack
"foo"
```

svaarala:
It's been quite a while since I implemented the current behavior - but I remember I compared to other engines and noticed that the common behavior is to allow property writes normally. That's why there's that `.stack` note in the guide for people coming from V8.

I would have implemented that behavior directly but it was preferable to use an inherited accessor to minimize property count. (I didn't think of changing the setter as in this branch at the time.)

fatcerberus:
Any decision on this yet?  It'll be nice to have these natively writable, but if you think it's better to keep them gated behind a defprop, that's fine too.

svaarala:
I'll merge the change, it's pretty close now - the changed behavior is more consistent with V8/Spidermonkey. Also Duktape has writable properties when tracebacks are disabled.

Issue #260, opened 2015-08-06
Suggestions what to deprecate in Duktape 2.0


svaarala:
With 2.0 release it's possible to deprecate some features as compatibility doesn't have to be guaranteed. This issue is for suggestions what features can/should be removed: rarely used features, features which aren't useful in practice, awkward APIs that should be replaced, etc.

I'll be doing a source code scan to come up with candidates for removal too (both compatible and incompatible).

Issues:
- [x] File I/O API calls (move to separate, optional module) - #734
- [x] Dependency of DUK_FPRINTF() - #734
- [x] duk_dump_context_stdout() and duk_dump_context_stderr(): they rely on DUK_FPRINTF() and I/O and are a portability issue (duk_dump_context() can be kept, but might be better as a separate extra) - #736
- [x] duk_safe_call() without userdata argument - #727
- [x] Remove `Duktape.Buffer`, replace with `ArrayBuffer` throughout, e.g. plain buffers should object coerce to `ArrayBuffer` - #739
- [x] Reduce number of custom error types (`DUK_ERR_UNCAUGHT_ERROR` etc) and make remaining custom errors actual Error subclasses: https://github.com/svaarala/duktape/issues/476#issuecomment-162260223 - #740
- [x] Remove option for refcounting only, as there are current features which depend on a back-stop mark-and-sweep (e.g. debugger paused state GC handling) #1168
- [x] Remove error code from fatal errors / panics
- [x] Remove panic vs. fatal distinction, and only use fatal errors?

fatcerberus:
Deprecate everything and rewrite it all from scratch!  It'll be more fun than a barrel of half-witted (spider)monkeys! ;-)

Okay, no, but seriously, I know you mentioned moving the file system functions into a separate module, would that count as deprecation?

svaarala:
I think so: some API calls currently depend on accessing files or stdout/stderr. In hindsight I don't think that's a good idea (at least by default) because an embedded engine should really consult the host environment for such things. It's not just a theoretical issue either, I've personally worked in weird environments where the usual assumptions don't hold (e.g. writes to stdout end up on a device screen, and file system is *not* accessed using ANSI file calls but a proprietary API).

So either Duktape shouldn't provide APIs that depend on I/O (providing them as an optional extra would be fine IMO) or such dependencies need to be very easy and clean to retarget. Current date/time is a similar host dependency which is not very easy or clean to replace at the moment.

JoshEngebretson:
[...]

PR #396, opened 2015-10-09
Add a separate marker tag for 'unused' values

svaarala:
In Duktape 1.3.0 the 'undefined' value had a single internal tag value and two sub-types:

- "undefined unused" denoted an unused value, and was used to e.g. indicate missing array elements
- "undefined actual" denoted the Ecmascript `undefined` value

The sub-type was indicated in a field separate from the tag type field (similar to how booleans are handled).

Add a separate 'unused' tag type so that 'undefined' tag is only used for Ecmascript `undefined`. This improves performance with both packed and unpacked `duk_tval`:

- With packed `duk_tval` a 16-bit write can be used instead of a 32-bit write (may not be faster on all platforms in practice)
- With unpacked `duk_tval` one field write instead of two field writes

Also ensure that `unused` is not actually used anywhere in the code base other than to mark missing array elements. This means that the new marker tag doesn't need to be added to every type switch-case. Instead, add assertions to switch-cases so that the default clause only catches numbers, not 'unused' values.

The debugger protocol has a dvalue for `unused` values. Such values shouldn't be accepted as inputs (erroring on them should be fine; they didn't do anything useful before the change). They are currently sent as outputs in certain cases, check if they should be replaced with "undefined" (arguably a breaking change).

Tasks:

- [x] Duk_tval change
- [x] Change call sites for renamed macros
- [x] Go through every tag switch case and fix
- [x] Remove `duk_push_unused()` to avoid pushing unused values on the value stack
- [x] Debugger protocol exposes the "unused" type, figure out what to do with it
- [x] Fix "isFastint" polyfill for new tag numbers if necessary, test for both packed and unpacked
- [x] Internal documentation
- [x] Test on x64 (unpacked) and x86 (packed)
- [x] Assertion tests
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1171,6 +1171,9 @@ Planned
   so that values above current value stack top are set to "undefined" instead
   of "unused" to improve call performance (GH-389)
 
+* Internal performance improvement: use a separate 'unused' tag instead of a
+  sub-type of 'undefined' to e.g. mark gaps in arrays (GH-396)
+
 2.0.0 (XXXX-XX-XX)
 ------------------
 
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -685,9 +685,10 @@ some cases:
 |                       |           | length in network order and buffer    |
 |                       |           | data follows initial byte             |
 +-----------------------+-----------+---------------------------------------+
-| 0x15                  | unused    | Represents the internal "undefined    |
-|                       |           | unused" type which used to e.g. mark  |
-|                       |           | unused (unmapped) array entries       |
+| 0x15                  | unused    | Represents an unused/none value, used |
+|                       |           | internally to mark unmapped array     |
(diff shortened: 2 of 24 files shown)


svaarala:
Regarding the debugger protocol, the "unused" dvalue will still be around:

- It's now rejected for inbound requests. This should not be an incompatible change because there is no request where it would have been a valid input. (Because debugger code was the only call site for the internal `duk_push_unused()` call, that call was removed.)
- It's still used for outbound messages to indicate "missing/none" values. For example, if GetVar doesn't find the binding request, the result message will have "unused" for the value. Changing this would be incompatible.

It'd probably make sense to remove the "unused" value from the debugger protocol too. For example, for GetVar there's already a boolean indicating if the binding was found or not, so that the value could just be a normal "undefined" when the binding is not found. Added a note to #260 to consider it for Duktape 2.0.

PR #398, opened 2015-10-12
Avoid rdflib in non-official dist

svaarala:
This removes one more dependency for making a development dist. Might matter to e.g. Windows users; rdflib is not a common dependency and is only used for SPDX license creation.


fatcerberus:
I think I remember installing rdflib in Cygwin at some point, I was unclear what its purpose was but it was required to build so...

This is good, it will make it simpler to get a Duktape build environment set up.  The only other dependency that would be nice to do without is PyYAML, but that's probably unrealistic.

Issue #262, opened 2015-08-08
Option to clear partial state when debug connection restarted


fatcerberus:
Duktape currently parses debug commands piece by piece, often executing the command even before EOM is reached.  As a result, it's possible to end up with the engine in a state where it's in the middle of reading a command, but the transport link is severed.

In minisphere I have a system where if a TCP reset is detected during debugging, it will wait 30 seconds for the client to reconnect.  However, if Duktape hasn't received the full command by this time, the client won't know which bytes are missing to resend them.  Attempting to send a new command while Duktape is in this state will likely result in issues similar to that described in #244 (or in the worst case, a segfault).

What I suggest to solve this issue is some method to reset the command-reading state such that the next byte received by Duktape is assumed to start a new command.

svaarala:
Just to be sure, in your case the logical debugger transport is not severed, but the concrete transport (TCP) is? In other words, you'd want the debugger transport to support transparent TCP reconnect, so that the debug session won't become detached if a TCP connection breaks.

There's no easy way to reset the command-reading state - there is really no such state, as Duktape is just pulling in data (and writing data) as it needs using read calls. Resetting the reading state would be equivalent to longjmp'ing out of the current debug command with all the necessary cleanup (if any), and all commands would need to support it.

But you should still be able to solve this by writing an appropriate transport framing yourself. For example, you could send data in numbered chunks and buffer them on both sending sides so that a chunk can be resent if it wasn't received completely; this is essentially a pretty standard queue protocol rather than a plain stream. Debug read/write callbacks would then be oblivious to the reconnection process. This is one reason why the transport is abstract so that you can implement something like this if you want.

svaarala:
Btw, coming back to the original problem (recovering from a dropped TCP connection), I don't it'll be possible to resynchronize the debug connection using plain TCP alone even if Duktape changes were made. Some issues one would have:

- The current command being executed would need aborting like you say. Similarly the remote peer must be able to abort parsing a current reply/notify or other messages at any point, including in the middle of a dvalue.
- Aborting a command means the reply for that command won't be sent - or it may mean that a *partial* reply has been written out by Duktape (and it may be in the middle of a dvalue even). Note that replies are just streamed out so it's very normal for them to be written out in pieces.
- Any number of requests may have been lost; any number of replies may have been lost by TCP.
[...]

PR #399, opened 2015-10-12
Reorder detach cleanup and detach callback

svaarala:
This should make it easier to allow a detached callback to reattach immediately without leaving the debugger state inconsistent.

Tasks:

- [x] Check that the detached -> attach sequence works in practice (multiple times)
- [x] Document that the detached callback may reattach immediately
- [x] Delay detached callback to message handling loop so that it's called between two full messages (this avoids accidentally e.g. skipping to EOM or writing to the new connection in handling code dealing with data from the previous connection)
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1145,12 +1145,15 @@ Planned
   intended mainly for minimal compatibility with existing code using "const"
   (GH-360)
 
-* Add a debugger Throw notify for errors about to be thrown, and an option
-  to automatically pause before an uncaught error is thrown (GH-286, GH-347)
-
 * Add a human readable summary of object/key for rejected property operations
   to make error messages more useful for operations like "null.foo = 123;"
   (GH-210, GH-405)
+
+* Add a debugger Throw notify for errors about to be thrown, and an option
+  to automatically pause before an uncaught error is thrown (GH-286, GH-347)
+
+* Allow debugger detached callback to call duk_debugger_attach(), previously
+  this clobbered some internal state (GH-399)
 
 * Fix "debugger" statement line number off-by-one so that the debugger now
   correctly pauses on the debugger statement rather than after it (GH-347)
--- a/examples/cmdline/duk_cmdline.c
+++ b/examples/cmdline/duk_cmdline.c
@@ -593,8 +593,26 @@ static int handle_interactive(duk_context *ctx) {
 
 #ifdef DUK_CMDLINE_DEBUGGER_SUPPORT
 static void debugger_detached(void *udata) {
+	duk_context *ctx = (duk_context *) udata;
+	(void) ctx;
 	fprintf(stderr, "Debugger detached, udata: %p\n", (void *) udata);
 	fflush(stderr);
+
+#if 0  /* For manual auto-reattach test */
+	duk_trans_socket_finish();
+	duk_trans_socket_init();
+	duk_trans_socket_waitconn();
+	fprintf(stderr, "Debugger connected, call duk_debugger_attach() and then execute requested file(s)/eval\n");
+	fflush(stderr);
+	duk_debugger_attach(ctx,
+	                    duk_trans_socket_read_cb,
+	                    duk_trans_socket_write_cb,
+	                    duk_trans_socket_peek_cb,
+	                    duk_trans_socket_read_flush_cb,
+	                    duk_trans_socket_write_flush_cb,
+	                    debugger_detached,
+	                    (void *) ctx);
+#endif
 }
 #endif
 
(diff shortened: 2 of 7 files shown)


fatcerberus:
Wait... I do this right now in minisphere.  In what way can the state be left inconsistent?  Could I segfault?

svaarala:
I think the best answer to that is in the diff. Based on the previous code I think at least `heap->dbg_udata` would be overwritten as `NULL` if a non-NULL value was given in the attach call.

fatcerberus:
It also sets dbg_detached_cb to NULL - ouch.  Glad you caught this, that would have been a nasty bug to diagnose, figuring out why my detach callback wouldn't fire the second time. :)

svaarala:
I'll verify that detached callback theory before I merge.

svaarala:
There was one more bug, a rather trivial one: the `heap->dbg_udata` was `NULL`ed before being used in the delayed detached callback. Fixed, with manual testing multiple reattaches now work cleanly.

svaarala:
One more small issue: when the Detach has been processed, the message processing helper skips to EOM before returning to the message loop. This is incorrect for Detach when the user code reattaches inside the detach callback: we'd skip to EOM in the new connection which causes requests and responses to be out of sync. This caused some funny issues in the web UI, like local variables appearing in the breakpoint list :)

svaarala:
Hmm, just occurred to me that if a transport is lost inside some message handling function due to an internal reason (like a parse error in the debug message stream) we'd call detach, there would be a reattach, and we'd then still be in the middle of message processing and potentially incorrectly scan to EOM on the new connection.

To avoid this `DUK__SET_CONN_BROKEN()` should only set a flag indicating the connection is broken and the detached callback would need to be called in the message loop, cleanly between handling two messages.

fatcerberus:
Yeah, that's the kind of thing I was worried about with #262, and why I suggested an API to "clear the table" as it were.

svaarala:
There's no problem in the debug client side I think - as long as debug transport connection are modelled as separate objects. However, inside Duktape the state keeping is rather primitive, and there's just the heap-wide state for the current transport and it'd probably be overkill to support multiple active connections so that one could be handling message from one debug connection while doing a handshake on another :)

svaarala:
I think I have a working solution for that problem now -- detach callback is postponed to the message loop so that it always happens cleanly between messages.

fatcerberus:
Yeah, I knew there was no problem on the client side - I was concerned about the Duktape end too. :)

svaarala:
Ah, I see :)

The problem with that approach is that even if we wipe the state through an API call, Duktape will still be in the middle of processing some random request - and if we reattach within that request things will go wrong, so one would still need to delay the reattach somehow (or use longjmp-based error handling inside the debugger code throughout).

Here's the delayed detach callback fix BTW: https://github.com/svaarala/duktape/commit/31d1296fcc1fea971ebaae0e07871963292cea52

Explicitly requested detach can happen from outside the message loop, concretely from `duk_debugger_detach()` and from `duk_heap_free()`. In these cases the detached callback is called directly because there's no message loop to worry about.

svaarala:
Ok, I think this is merge ready for my part: all scenarios seem to work cleanly. I simulated the "syntax error in debug connection" by injecting an error into Eval command handling, then using Eval from the web UI, and checking that reattach works properly. (Unfortunately these are not covered by any automated regression tests right now.)

@fatcerberus Any chance you'd have time to see if this works in your reattach scenario in the near future? If so, I can wait before merging.

fatcerberus:
I'll test it out either tonight or in the next couple of days.  My debugger connects over loopback (which is special-cased in Windows) though so it might be difficult to simulate a dropped connection.

svaarala:
Ok, sounds good - there's no particular hurry as this branch won't easily go into conflict.

fatcerberus:
One scenario that might be worth testing is if the debugger is detached during a call to `duk_debug_halt_execution()`

svaarala:
Agree - it's using `duk_debug_process_messages()` so there shouldn't be much difference.

svaarala:
Ok, I tried the detach/reattach sequence by enabling "pause on uncaught" and the auto-reattach feature in "duk". Seems to work as expected.

fatcerberus:
That's good, I'll test it out soon myself.  I was going to the other day, but it turns out minisphere is currently programmed to exit on detach (mimicking MSVC's behavior), so I will have to set up a harness first.

svaarala:
Thanks - no particular hurry, this won't conflict easily.

svaarala:
@fatcerberus This is now rebased and merge ready (unless you have any tests you might want to run).

fatcerberus:
Oops, i forgot about this, i tested out the reattach the other day and it worked as expected (after I actually implemented it in my client anyway :)

PR #401, opened 2015-10-12
Avoid slow path GETVAR/CSVAR for function self-reference

svaarala:
Currently a function self reference compiles to a slow path GETVAR. Avoid this slow path by detecting self reference at compile time and emit a LDCURRFN (load current function) when it's safe to do so.

I'm not sure if there are actually many safe situations, I'll need to check the semantics. For example, a global function declaration is writable so there's no guarantee that it won't be mutated during self recursion.

Tasks:

- [ ] Compiler changes
- [ ] Executor changes
- [ ] Debugger opcode definitions, check emitted output for call setup and normal read reference
- [ ] Check and document official semantics, compare to V8 and Rhino
- [ ] Test cases to cover emitted opcodes, with/catch capture cases, unsafe self references
- [ ] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1266,6 +1266,12 @@ Planned
   by allowing indent value or gap string and by supporting JX/JC in the
   fast path (GH-445)
 
+1.5.0 (XXXX-XX-XX)
+------------------
+
+* Internal performance improvement: handle function self-reference specially
+  to avoid a slow path identifier lookup for self-recursive functions (GH-401)
+
 2.0.0 (XXXX-XX-XX)
 ------------------
 
--- a/debugger/duk_opcodes.yaml
+++ b/debugger/duk_opcodes.yaml
@@ -434,7 +434,9 @@ extra:
   - name: ENDLABEL
     args:
       - BC_I
-  - name: EXTRA34
+  - name: LDCURRFN
+    args:
+      - BC_R
   - name: EXTRA35
   - name: EXTRA36
   - name: EXTRA37
--- a/src/duk_js_bytecode.h
+++ b/src/duk_js_bytecode.h
@@ -176,6 +176,7 @@ typedef duk_uint32_t duk_instr_t;
 #define DUK_EXTRAOP_IN              31
 #define DUK_EXTRAOP_LABEL           32
 #define DUK_EXTRAOP_ENDLABEL        33
+#define DUK_EXTRAOP_LDCURRFN        34
 
 /* DUK_OP_CALL flags in A */
 #define DUK_BC_CALL_FLAG_TAILCALL           (1 << 0)
--- a/src/duk_js_compiler.c
+++ b/src/duk_js_compiler.c
@@ -2323,8 +2323,24 @@ DUK_LOCAL void duk__ivalue_toplain_raw(duk_compiler_ctx *comp_ctx, duk_ivalue *x
 			x->x1.t = DUK_ISPEC_REGCONST;
 			x->x1.regconst = (duk_regconst_t) reg_varbind;
 		} else {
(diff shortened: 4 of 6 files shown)


svaarala:
This is a proof-of-concept right now. I'll need to check what the required semantics are (there seems to be a difference in behavior between V8 and Rhino too, for example).

`tests/perf/test-fib.js` uses self-reference and with this change execution time drops from about 8.5 seconds to 4.5 seconds. About 50% of the Fibonacci test is spent in the slow path GETVAR before the change.

svaarala:
This would be a nice improvement to call performance but a bit odds with Ecmascript semantics. Fixing the behavior to be compliant needs a bit of thinking, so I'm dropping this from 1.4.

Function declarations are handled differently from named function expressions; for function declarations the function name or binding exists in e.g. the global object, and is not part of the function. That global variable binding is non-configurable (see E5 Section 10.5) but it's still writable, so that:

```
> function test() {}
undefined
> delete test;  // cannot delete
false
> test
[Function: test]
> test = 123;  // can assign
123
> test
123
```

This means that a function declaration doing a self reference cannot be sure that the reference actually resolves to the function itself:

```
function func() {
    recursive = function nop() {};
}
function recursive(n) {
    print(n);
    func();
    if (n > 0) { recursive(n - 1); }
}
recursive(10);
```

The correct output here is:

```
10
```

With the change in this branch currently the test would output:

```
10
9
8
7
6
5
4
3
2
1
0
```

because the function won't react to the changed external binding.

fatcerberus:
It seems like this might actually be impossible to implement without breaking Ecmascript semantics.  The compiler has no practical way to guarantee the name won't be reassigned without analyzing every single possible call path, a potentially intractable problem.

svaarala:
There may be cases where the compiler would be able to prove that the binding won't be changed. Duktape's compiler doesn't have enough state to do that though, it'd need to know that every function potentially writing to the binding didn't in fact write to it, didn't have eval() calls, etc.

If the binding (= variable) was non-configurable and non-writable, it'd be relatively safe to assume it would never change. This would be the case for Ecmascript code; but with Duktape C API it's possible to force changes to such protected variables too. It's a bit of a grey area whether the compiler should be allowed to optimize in these cases, when standard code wouldn't be able to detect the difference. (If so, the behavior should be predictable and intuitive for the application.)

But it'd be quite annoying (and impractical) for user code to make such binding non-writable explicitly to make the idiom perform better. So for now it's most realistic to optimize scope handling performance piece by piece so that this would be faster without any tricks.

fatcerberus:
You'd also have to prove that none of the *other* bindings in the call path were ever written to as well.  It could get very complicated very fast, even if the compiler had sufficient state to do it.

svaarala:
If there are no eval() calls and no with statements, new variables can't be introduced at runtime - so it's possible to analyze the variable lookup starting from the current scope up to the declaration even if there are intervening scopes.

But to optimize a certain lookup, one would need to know that no function in the same program could introduce a capturing variable anywhere in the lookup "scope path", and in addition being sure that nothing writes to the final binding after compilation, so it needs a global state of all the relevant functions.

svaarala:
By the way, that "scope path analysis" is one optimization I've noted in a todo list a long time ago: a GETVAR opcode could have a number indicating how many scope levels can be safely bypassed (skipped over without checking if a matching variable is found) because such an identifier didn't exist at compile time and the scopes in question didn't have "with" statements or eval() calls.

This would speed up identifier lookup a bit because actual property checks would be skipped and the code would just lookup the internal prototype N times before making an actual lookup - and the variable would almost always then be present when the first actual property lookup was done.

svaarala:
Closing as unmerged, keeping the branch around for future reference.

PR #405, opened 2015-10-17
Human readable object/key summary for property access errors

### LINKED ISSUE #210: Make "invalid base value" more informative
Currently the `invalid base value` error is a bit too vague--it doesn't tell you the name of the variable that you're trying to access and is just in general very difficult to diagnose.  Would it be possible to reword this error message to make it clearer, something like what SpiderMonkey says in this case: `'varName' has no properties`

svaarala:
In Duktape 1.3.0 an invalid base value for a property access causes a rather cryptic error:

```
duk> null.foo=123
TypeError: invalid base value
```

Add a human readable summary of the object (base value) and the key of a rejected property access.

This pull also adds an internal helper API call to summarize a `duk_tval` safely:

- Guaranteed string result for any `duk_tval`
- Guaranteed limited length
- No side effects or error throwing (except for internal errors like out-of-memory)

This summary is useful for error messages and may be useful elsewhere too. The call is not added to the public API yet. A public API would need to operate on a stack index instead of a `duk_tval *` and the summary guarantees need some more thinking.

Tasks:

- [x] Add a config option for including object/key summary in error message, allow disabling for security sensitive environments
- [x] Add summarization internal API call, `duk_push_tval_readable()`
- [x] Come up with a better name
- [x] Finalize summarization details, see comments below
- [x] Test with and without object/key summary; add regression test to matrix tester
- [x] Document security issue (for some rare environments) in sandboxing document
- [x] Testcases for summary string and error messages
- [x] Fix expect strings of existing testcases
- [x] Add separate pull for handling function call base value in a human readable manner ("not callable" -> more verbose)
- [x] Releases entry
- [x] Rework readable string sanitization to avoid temporary buffer (use stack buffer instead)

Fixes #210.


svaarala:
Some examples of new default behavior:

```
((o) Duktape 1.3.99 (v1.3.0-141-g8180be6)
duk> undefined.foo = 123
TypeError: cannot write property 'foo' of 'undefined'
	duk_hobject_props.c:3292
	global input:1 preventsyield
duk> null.foo
TypeError: cannot read property 'foo' of 'null'
	duk_hobject_props.c:2252
	global input:1 preventsyield
duk> delete undefined[123]
TypeError: cannot delete property '123' of 'undefined'
	duk_hobject_props.c:4421
	global input:1 preventsyield
duk> undefined.foobarfoobarfoobarfoobarfoobarfoobar = 123
TypeError: cannot write property 'foobarfoobarfoobarfoobarfoobarfo...' of 'undefined'
	duk_hobject_props.c:3292
	global input:1 preventsyield
```

The summarization internal call requires some more work:

- Codepoints below 0x20 should be sanitized, especially literal NUL (0x00) is probably harmful
- Should non-ASCII codepoints be sanitized - this would be useful in many environments but not if you're working with a codebase using non-ASCII identifiers and keys
- Add object class or some clarification to object (`Object.prototype.toString()` is probably best summary for objects)

andoma:
Much appreciated!

fatcerberus:
Looks like someone other than me appreciates this change. :)

@svaarala How do you access the base value ("property 'x' of 'y')?  When we first discussed this I remember you saying that wasn't possible because base value information is no longer available to the parser by that point.

svaarala:
@fatcerberus It's just the base value, not its name. So it can say "null" or "undefined". So for example:

```js
var obj = null;
print(obj.foo);
```

Would say base value is "null" in the error message, and wouldn't know the name "obj".

fatcerberus:
Ah, that makes sense.

svaarala:
Ok, changed the string sanitization to use `duk_substring()` + `duk_map_string()` so that ASCII control characters are replaced with `?` but non-ASCII Unicode is kept as is. This should be better behavior than escaping all non-ASCII because non-ASCII characters may be the norm for some applications.

svaarala:
Had to rework the string sanitization once more because `duk_map_string()` throws an error on invalid UTF-8, and it's important that the sanitization never throws, even for invalid UTF-8 (e.g. internal strings).

svaarala:
Here's the current behavior:

```
((o) Duktape 1.3.99 (v1.3.0-150-g3ae13fa)
duk> undefined.foo = 123
TypeError: cannot write property 'foo' of undefined
	duk_hobject_props.c:3292
	global input:1 preventsyield
duk> null.foo
TypeError: cannot read property 'foo' of null
	duk_hobject_props.c:2252
	global input:1 preventsyield
duk> delete undefined[123]
TypeError: cannot delete property 123 of undefined
	duk_hobject_props.c:4421
	global input:1 preventsyield
duk> undefined.foobarfoobarfoobarfoobarfoobarfoobar = 123
TypeError: cannot write property 'foobarfoobarfoobarfoobarfoobarfo...' of undefined
	duk_hobject_props.c:3292
	global input:1 preventsyield
duk> null['foo\u1234bar'] = 123
TypeError: cannot write property 'fooሴbar' of null
	duk_hobject_props.c:3292
	global input:1 preventsyield
duk> undefined[Duktape.dec('hex', 'ff') + 'InternalProperty'] = null
TypeError: cannot write property '?InternalProperty' of undefined
	duk_hobject_props.c:3292
	global input:1 preventsyield
```

Changes to the first paste:

- Single quotes for strings only, makes value type identification clearer
- Unicode characters are not sanitized, only valid UTF-8 is produced
- Invalid UTF-8 is sanitized as '?' characters

svaarala:
ASCII control characters, single quote, and backslash are escaped:

```
duk> null['\x00\x7f\\\'"']
TypeError: cannot read property '\x00\x7f\x5c\x27"' of null
	duk_hobject_props.c:2252
	global input:1 preventsyield
```

fatcerberus:
Good to see this was merged--I'll try it out ASAP.  Thanks. :)

fatcerberus:
This works great, it will make bugs easier to find now.  Good work as always!

svaarala:
Good to hear, it took a few iterations for the string sanitization, hopefully it's intuitive enough :)

fatcerberus:
Just following up to say this was indeed a great improvement.  I've had a long-standing bug in my game's battle system for a while, enemy AIs would sometimes error out at the following line with "invalid base value":
```javascript
mini.Console.append("targ: " + (targets.length > 1 ? "[multi]" : targets[0].name));
```

I managed to reproduce it again, this time the error was "cannot read property 'name' of undefined".  Long story short, I now know the cause of the bug. :)

PR #407, opened 2015-10-18
Fix local time offset on RISC OS

### LINKED ISSUE #406: Fix DST handling for RISC OS (and possibly other platforms)
RISC OS fix described in:

- http://permalink.gmane.org/gmane.comp.web.netsurf.devel/10339

The current code also assumes DST is 3600 seconds which is not always correct, not sure if this has practical impact.

svaarala:
Fixes #406.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1151,6 +1151,9 @@ Planned
   value is a register-bound variable which is used and mutated in the rest
   of the expression (GH-381)
 
+* Fix Unix local time offset handling which caused issues at least on RISC
+  OS (GH-406, GH-407)
+
 * Remove octal autodetection in parseInt(), also fixes incorrect octal
   autodetection in e.g. "parseInt('00e1', 16)" (GH-413, GH-414)
 
--- a/src/duk_bi_date_unix.c
+++ b/src/duk_bi_date_unix.c
@@ -152,6 +152,13 @@ DUK_INTERNAL duk_int_t duk_bi_date_get_local_tzoffset_gmtime(duk_double_t d) {
 	                     (long) tms[1].tm_mday, (long) tms[1].tm_mon, (long) tms[1].tm_year,
 	                     (long) tms[1].tm_wday, (long) tms[1].tm_yday, (long) tms[1].tm_isdst));
 
+	/* tm_isdst is both an input and an output to mktime(), use 0 to
+	 * avoid DST handling in mktime():
+	 * - https://github.com/svaarala/duktape/issues/406
+	 * - http://stackoverflow.com/questions/8558919/mktime-and-tm-isdst
+	 */
+	tms[0].tm_isdst = 0;
+	tms[1].tm_isdst = 0;
 	t1 = mktime(&tms[0]);  /* UTC */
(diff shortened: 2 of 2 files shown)


fatcerberus:
So I'm confused... Does the OS handle adjusting for DST in this case?  Since you removed the adjustment code from Duktape itself...

svaarala:
There's stil DST handling in there after the change. What the code in question is trying to do is figure out the time offset between local time and UTC time on a given UTC instant. This is done in a perhaps non-obvious way as follows:

1. The UTC time value is broken down into a UTC date/time representation using `gmtime()`. Example: `2015-10-18 11:22:33`.
2. The UTC time value is broken down into a local date/time representation using `localtime()`. Example (Finland): `2015-10-18 14:22:33`.
3. The broken down date/time in step 1 is converted back into a UTC time value `t1` using `mktime()`.
4. The broken down date/time in step 2 is intentionally "interpreted as UTC" (!) and converted into a UTC time value `t2` using `mktime()`.
5. The local time difference is computed as `t2 - t1`.

The problem in the original code was in steps 2 and 4:

- In step 2 the result from `localtime()` had `tm_isdst > 0` indicating that the broken down time has DST active.
- In step 4 Linux `mktime()` interprets an incoming `tm_isdst` in `mktime()` which we don't want. The original code added `t2 += 3600` to counteract this effect and this works on Linux (at least when DST offset is 3600 seconds).
- In step 4 RISC OS `mktime()` probably ignores `tm_isdst` in `mktime()`, but the `t2 += 3600` fix is still applied which causes the time offset to be off by 1 hour. (This is my best guess at the moment.)

The fix in this branch forces `tm_isdst` to 0 before `mktime()` so that the behavior will be the same on both Linux and RISC OS and eliminates the `t2 += 3600` fixup entirely. Assuming this works as intended, this should be more portable and avoids assuming DST offset is 3600 seconds.

fatcerberus:
Ah, I see now.  Quite a clever algorithm there. :)

svaarala:
Don't know about it being clever really, but I couldn't figure out a better portable approach. I'm not really an expert on datetime handling, so I'd welcome suggestions on how to do it better :-)

Ecmascript actually places quite a lot of requirements on Date handling.

The valid Date range for example is roughly year 1970 +/- 273000 years. The platform often doesn't support that large a range so Duktape and other engines often need some workaround like choosing an "equivalent year" to replace an out-of-range year for the purposes of DST handling (for example). The equivalent year needs to have the same leap day, same week day, etc. Platforms also vary how well they handle DST (especially historical DST).

Another interesting issue is that local time to UTC conversion is not uniquely defined and it'd be nice to have a consistent result for that. This currently means iteration, but luckily it's constrained to the generic Date implementation; the platform Date primitives don't need to handle it. There are (not surprisingly) differences in native platform behavior for this conversion too.

fatcerberus:
Ecmascript is strict like this everywhere, which is annoying for implementors but at the same time as a programmer, I like that I can write JS code and know that as long as I stick to standard features it will work identically on any implementation (even later standards) without changes.

Very few mainstream programming languages can make that claim. Even C isn't that portable. :)  So it's kind of a double-edged sword.

Issue #408, opened 2015-10-19
Version of duk_json_encode() allowing custom indent


fatcerberus:
Current implementation of `duk_json_encode()` uses default encoding settings, putting the entire JSON literal on one long line.  Setting indent requires using the valstack API to call JSON.stringify() directly from C, which is not optimal since if a script reassigns the `JSON` object an unwanted error will be thrown.

What I suggest here is a version of `duk_json_encode()` allowing custom indent to be specified, useful for programs such as my Cell compiler, which generates source maps and manifests in JSON, where I need the output to be human-readable.

svaarala:
I agree it'd be nice. What would you suggest for an API -- considering that there'd certainly be further options other than indent down the line?

I didn't add options to the API for this reason originally: a plain flags field won't be enough and adding individual options one by one as they come along would be awkward for compatibility.

fatcerberus:
`JSON.stringify()` itself has only two options: indent and reviver.  I'm not certain a reviver function is useful from the native side, although it would be easy to have the new API accept a function pointer for it, or perhaps a boolean flag which, if true, uses a function pushed on the valstack.

For indent, the challenge is that in JS, it is either an arbitrary string or an integer specifying the number of spaces to indent per level.  This would require 2 functions at the least, which is a bit excessive for such a basic primitive.

One option that would be fairly elegant might be to get encoding options from the valstack, and accept a bitmask specifying which ones are present, similar to how `duk_def_prop()` is set up.

svaarala:
I've come to pretty similar conclusions, I guess the best options would be:

- A flags + variable value stack approach similar to `duk_def_prop()`. This is quite flexible but not easy to read because the value stack behavior depends on flags.
- Provide an optional value stack index pointing to an object encoding control arguments like: `duk_json_encode(ctx, -3 /*obj_idx*/, -7 /*opt_idx*/)`. The value at "opt_idx" (if != DUK_INVALID_INDEX) would be something like: `{ indent: 4, reviver: myReviver, customOptionX: true }`.

The latter approach would also work well for the Ecmascript API, so that you could provide a table of arguments to Duktape.enc(), like:

    Duktape.enc('jx', myVal, {
        indent: 4, asciiOnly: true, allowCycles: true
    });  // arg examples are bogus :)

The downside of this is the relative cost of creating and extracting the arguments. The arguments can often be shared/reused for a specific call site.

This would be a good place to add a "multi-get" primitive to extract multiple values using a format string, like:

[...]

PR #409, opened 2015-10-19
Add property multi get/put API calls

svaarala:
For now this is a quick prototype branch for adding property multi get/put API calls, so that there's something concrete to reflect on.

Issues:

- [ ] Jansson API provides keys as va_args too (e.g. "s:s" -> key string, value string) which is more flexible but less literal
- [ ] What should be the "get" operation: should it be a "duk_to_int" vs. "duk_get_int" for integers, for example? Lowercase for "get", uppercase for "to"?
- [ ] Characters for all relevant types
- [ ] Ability to specify property attributes (would be useful for internals where attributes are often important)
- [ ] Finalize get and put operations
- [ ] Find at least a few internal call sites for each
- [ ] API documentation, link to wiki examples
- [ ] API testcases
- [ ] Releases entry

### DIFF
--- a/src/duk_api_public.h.in
+++ b/src/duk_api_public.h.in
@@ -654,6 +654,9 @@ DUK_EXTERNAL_DECL duk_bool_t duk_has_prop_string(duk_context *ctx, duk_idx_t obj
 DUK_EXTERNAL_DECL duk_bool_t duk_has_prop_index(duk_context *ctx, duk_idx_t obj_index, duk_uarridx_t arr_index);
 DUK_EXTERNAL_DECL void duk_def_prop(duk_context *ctx, duk_idx_t obj_index, duk_uint_t flags);
 
+/* FIXME: naming; constants for characters */
+DUK_EXTERNAL_DECL void duk_get_prop_multi(duk_context *ctx, duk_idx_t index, const char *fmt, ...);
+
 DUK_EXTERNAL_DECL duk_bool_t duk_get_global_string(duk_context *ctx, const char *key);
 DUK_EXTERNAL_DECL duk_bool_t duk_put_global_string(duk_context *ctx, const char *key);
 
--- a/src/duk_api_stack.c
+++ b/src/duk_api_stack.c
@@ -4387,6 +4387,7 @@ DUK_EXTERNAL duk_bool_t duk_instanceof(duk_context *ctx, duk_idx_t index1, duk_i
 DUK_INTERNAL void duk_push_lightfunc_name(duk_context *ctx, duk_tval *tv) {
 	duk_c_function func;
 
+	DUK_ASSERT_CTX_VALID(ctx);
 	DUK_ASSERT(DUK_TVAL_IS_LIGHTFUNC(tv));
 
 	/* Lightfunc name, includes Duktape/C native function pointer, which
@@ -4409,6 +4410,7 @@ DUK_INTERNAL void duk_push_lightfunc_name(duk_context *ctx, duk_tval *tv) {
 }
 
 DUK_INTERNAL void duk_push_lightfunc_tostring(duk_context *ctx, duk_tval *tv) {
+	DUK_ASSERT_CTX_VALID(ctx);
 	DUK_ASSERT(DUK_TVAL_IS_LIGHTFUNC(tv));
 
 	duk_push_string(ctx, "function ");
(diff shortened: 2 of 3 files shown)


svaarala:
Idiom in the current prototype:

```
duk_int_t my_int = 123; /* default */

duk_get_prop_multi(ctx, -1, "integerValue:d", &my_int);
printf("my_int: %ld\n", (long) my_int);
```

fatcerberus:
So if the property doesn't exist (or is the wrong type?), it just leaves the value alone, that's convenient.

In #408 you said this could be footprint-neutral.  Would it be performance-neutral also?  That is to say, would the processing saved by combining multiple get/put calls into one operation offset the performance hit due to format string parsing overhead?

svaarala:
I'm not sure about the performance side, my guess it would be slightly slower. Parsing the format string is a string scan which has a cost, but on the other hand the object doesn't need to be looked up again for each property access.

Re: what to do for missing properties etc, I listed a bunch of questions in the description. Not sure what the best answers are :) I guess it comes down to what is most useful and predictable for calling code.

svaarala:
For some internals the exact behavior for missing/invalid types actually matters quite a bit, because the Ecmascript specification mandates certain checks and specific errors (usually `TypeError`) when some criteria are not met. Unfortunately these strict criteria may not be the most useful for applications though, so choices, choices .. :)

fatcerberus:
I would say the most intuitive behavior at the application level would be to type coerce.  Presumably a programs knows the nature of the objects it's reading from

fatcerberus:
...and would expect the value to be coerced if a different type is specified. (Sorry, the Comment button is very easy to hit by accident on mobile)

fatcerberus:
On the other hand, now that I thought on this some more, it might make more sense for it to behave the same as ‘duk_get_xxx‘ and write a falsy value if the types don't match.

svaarala:
Yeah, the choice is not obvious :) Doing a "get" would have the nice benefit of being side effect free.

It's also an option to have both behaviors, e.g. by "d" meaning an uncoerced integer and "D" meaning a coerced integer (or somethng like "*D"). Is that simpler and more useful to applications? Not sure.

fatcerberus:
An issue that occurs to me with the proposed `duk_get_prop_multi()`, if a failure occurs in a non-strict get (i.e. in the case that an individual duk_get_prop would return false), there's no way to tell which lookup failed.  So that might limit the function's utility somewhat.

The same issue applies for `duk_put_prop_multi()`, although it's less important there since that would normally be used with newly created objects pushed onto the stack where success can be guaranteed (outside of exotic circumstances anyway).

svaarala:
Hmm - when `duk_get_prop()` returns false it's not actually a failure as such, but a missing property. That would lead to defaulting, but like you said, the caller wouldn't know which value(s) defaulted.

An actual error, like a getter throwing, would lead to an error to be propagated outwards, and the "out" variables would be left in a state with no guarantees of which ones have been written to.

fatcerberus:
Yeah, I know it's not a failure per se, although it could be in strict mode (a missing property will throw).  But that was indeed what I'm getting at, there's no way to tell which property(s) are missing.

As for the inconsistent state of the variables in case of a throw, I think that's not really an issue--that's just expected behavior when dealing with exceptions. :)

svaarala:
Duktape C API is always in "strict mode" but `duk_get_prop()` doesn't throw on missing property. And a missing property is not an error in strict mode for Ecmascript either: `obj.notDefined` will just be `undefined`. You might be thinking about variable accesses, not property accesses :-)

fatcerberus:
Oh, for some reason I thought a missing property was an error in strict mode, huh.

As for the C API always being in strict mode, I thought there was a time or two where an API behaved differently depending on whether the calling ecma function was strict or not, but I might be imagining it.  If I notice it again I'll let you know because apparently that would be a bug. :)

svaarala:
Many versions ago (0.11.0?) the API would be strict when a Duktape/C function call was in progress and non-strict otherwise. That's been fixed so that the API is always strict, even when no call is in progress.

svaarala:
Revised this work-in-progress a bit. Added an initial set of "type characters":

- d: integer
- u: unsigned integer
- n: number (double)
- b: boolean
- s: string
- x: buffer
- p: pointer
- v: any value type; value is pushed on value stack (covers remaining types reasonably)

Testcase example:

- https://github.com/svaarala/duktape/blob/18987340c4417067303c4a2ee34473aa10f37cfe/tests/api/test-get-prop-multi.c#L40-L56

However, due to pointer safety the "string" and "buffer" type characters probably need to be omitted: if the string/buffer read from the object is provided by a getter/Proxy and the string/buffer is not on the caller value stack, the pointer read might then be stale and memory unsafe behavior will happen. Types coerced into a non-heap-based value have no such issues.

The "v" type character (= push on value stack) still allows strings, buffer, functions, arrays, etc to be read reasonably conveniently and also acts as a fallback to types not directly support:

    duk_int_t level;
    duk_get_prop_multi(ctx, -3, "name:s level:d", &level);
    printf("Name '%s', rank %d\n", duk_get_string(ctx, -1), level);

One could argue the string/buffer safety is up to the caller to ensure, but that would leave the API error prone. A crafted input could cause memory unsafe behavior which could be potentially exploitable.

The footprint difference to master is now about +1000 bytes (the function itself is ~730 bytes) on x64 and will decrease slightly if/when string/buffer types are removed.

Issue #334, opened 2015-09-11
Improve case insensitive RegExp unicode performance


svaarala:
Case insensitive RegExp compilation normalizes a case sensitive codepoint range into a case insensitive codepoint range by simulating all 16-bit codepoints and seeing what comes out of each conversion. Ranges are coalesced based on the output.

This is done because a single range may split into multiple ranges in case conversion, so it's not enough to simply case normalize the range endpoints. This approach is not ideal, and it won't scale to non-BMP codepoints (but since those are not part of the standard, this is not a big limitation).

There's a concrete performance issue with the current approach because every one of those 65536 codepoint normalization lookups will walk the bit-packed Unicode case conversion ranges.

This can be mitigated in many way:

- Making the case conversion ranges indexed which would improve performance.
- Making the normalization algorithm aware of the Unicode ranges so that a single pass through the normalization ranges (in order) would be enough.
- Reworking the normalization algorithm to be aware of Unicode ranges so that the RegExp compiler didn't need to walk through 65536 codepoints.
- Reworking normalization so that it happens at runtime instead of at compilation time - leading to a different performance tradeoff between compilation and execution.

svaarala:
The workaround in #411 will go into Duktape 1.4 release (it's used by NetSurf).

svaarala:
This issue also affects compilation of http://momentjs.com/ - compilation time is roughly 0,4 seconds my laptop with default options. With `DUK_OPT_REGEXP_CANON_WORKAROUND` enabled (which improves case insensitive regexp compilation performance) it is about 0,025 (~15x faster).

svaarala:
#1616 provides a ~50x speedup with 300 byte footprint impact.

fatcerberus:
So, even more of a speedup than the 128kB lookup table?  Nice.

svaarala:
No, the 128kB lookup table is much faster. But it's actually a separate mechanism:
- The mechanism in this pull speeds up the case insensitive regexp case specifically.
- The 128kB lookup table speeds up per codepoint canonicalization which improves regexps but also other uppercase conversions.

PR #411, opened 2015-10-22
Add an optional 128kB lookup to optimize regexp canonicalization

svaarala:
In Duktape 1.3.0 and prior case-insensitive RegExps with large unicode ranges have serious performance issues. Example regexp:

```js
var re = /[\u0000-\uffff]/i;
```

This is most likely the root cause behind this Netsurf issue (and at least another similar one):

- http://bugs.netsurf-browser.org/mantis/view.php?id=2366

The performance issue is caused by two algorithms interacting (neither alone would be a major issue):

- A continuous Unicode range may be split into multiple discontinuous ranges when it is canonicalized, so it's not enough to canonicalize the endpoints. The current simple approach is to run through the range codepoints, normalize each, and then detect the canonicalized ranges from the conversion output. This is a loop of ~64k canonicalizations for each `[\u0000-\uffff]` range in the case insensitive regexp source.
- Each canonicalization in turn runs through the size-optimized bit-packed Unicode data which doesn't have an index, so potentially hundreds of ranges are scanned. This exacerbates the problem quite a bit.

This branch has a very simple fix where a 65536-entry lookup table is used for the canonicalization operation, which should fix the immedate issue for environments where this is a practical issue:

```
$ time ./duk.O2.130 tests/perf/test-regexp-case-insensitive.js

real	0m23.962s
user	0m23.918s
sys	0m0.003s

$ time ./duk.O2 tests/perf/test-regexp-case-insensitive.js

real	0m0.019s
user	0m0.018s
sys	0m0.003s
```

This is a short term solution and the lookup table won't be enabled by default because the footprint impact is quite large.

A proper medium term solution is to use a better algorithm for handling case-insensitive RegExp ranges, and to make the Unicode data indexed and faster to work with for non-low-memory targets. It should be possible to get a similar improvement with much, much smaller footprint impact.

Partially addresses #334. The workaround is enabled using `DUK_OPT_REGEXP_CANON_WORKAROUND` (or `DUK_USE_REGEXP_CANON_WORKAROUND` if editing `duk_config.h` directly).

Tasks:

- [x] Add a config option, disabled by default
- [x] Ensure the change fixes the original Netsurf issue (also validates that root cause is understood)
- [x] Try to find a better short term solution without reworking the Unicode tables yet => future work
- [x] Resolve FIXMEs
- [x] Releases entry
- [x] Test manually without canonicalization fast path
- [x] Test manually with canonicalization fast path
- [x] Assert run with fast path
- [x] Assert run without fast path


fatcerberus:
Footprint impact is what, around 256kb? Not an issue for PC platforms but definitely impractical for embedded targets.

That speedup is crazy though, almost 24s down to less than a second.

svaarala:
The footprint impact is 64k entries of 16-bit codepoints -> 128kB.

fatcerberus:
Oh, I was thinking of the full Unicode range which is 32-bit, I forgot Duktape only supports the BMP(?).

svaarala:
Duktape strings support codepoints above BMP which is non-standard: individual codepoints in the string can be up to 31 bits (or so). Ecmascript strings only support 16-bit codepoints so surrogate pairs are normally used.

E5.1 treats strings as 16-bit codepoint sequences only, and doesn't support non-BMP codepoints through surrogate pairs in most operations. So for compliance, 16-bit support is enough for E5.1.

Non-BMP support is necessarily out-of-specification, so the support doesn't need to be 100%. It's fairly complete though, but for this RegExp case normalization process only the BMP is supported because the current small-but-slow algorithm would need to loop through over a million codepoints instead of just 65536 :-)

If the algorithm gets better, there should be no problem supporting codepoints above BMP for case-insensitive RegExps too -- but that's out-of-specification functionality anyway so it's not a compliance issue.

fatcerberus:
I just noticed `duktape.c` recently ballooned from ~3MB to around 3.5MB... is that due to this change?

svaarala:
Yes, it's the 64k entry lookup.

PR #412, opened 2015-10-24
Implement callstack level for debugger commands

fatcerberus:
Implement an optional callstack level for debugger commands dealing with the current function activation (such as GetLocals and Eval), allowing e.g. retrieval of variables from earlier activations and not just the topmost one.  This is especially useful when paused at an uncaught error, to ascertain the underlying cause of the crash by walking the stack.

This is implemented in a clean, backwards-compatible manner such that the debugger protocol version remains at v1.

Checklist:
- [x] Implement callstack level for GetLocals
- [x] Implement callstack level for GetVar/PutVar
- [x] Implement callstack level for Eval
- [x] Document the new parameter in `debugger.rst`
- [x] Manual testing to ensure correct and compatible functionality
- [x] Await results of branch review and clean up as necessary

### DIFF
--- a/src/duk_bi_global.c
+++ b/src/duk_bi_global.c
@@ -450,11 +450,10 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 	}
 
 #if defined(DUK_USE_DEBUGGER_SUPPORT)
-	/*  XXX: level is used only by the debugger and should never be present
-	 *  for eval() called from Ecmascript code. is there a way to assert for
-	 *  this?
+	/*  NOTE: level is used only by the debugger and should never be present
+	 *  for an Ecmascript eval().
 	 */
-	level = -2;
+	level = -2;  /* by default, use caller's environment */
 	if (duk_get_top(ctx) >= 2 && duk_is_number(ctx, 1)) {
 		level = duk_get_int(ctx, 1);
 	}
--- a/src/duk_bi_global.c
+++ b/src/duk_bi_global.c
@@ -429,6 +429,7 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 	duk_hobject *outer_var_env;
 	duk_bool_t this_to_global = 1;
 	duk_small_uint_t comp_flags;
+	duk_int_t level = -2;
 
 	DUK_ASSERT_TOP(ctx, 1);
 	DUK_ASSERT(thr->callstack_top >= 1);  /* at least this function exists */
@@ -448,15 +449,27 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 		return 1;  /* return arg as-is */
 	}
 
+#if defined(DUK_USE_DEBUGGER_SUPPORT)
+	/*  XXX: level is used only by the debugger and should never be present
+	 *  for eval() called from Ecmascript code. is there a way to assert for
+	 *  this?
+	 */
+	level = -2;
+	if (duk_get_top(ctx) >= 2 && duk_is_number(ctx, 1)) {
+		level = duk_get_int(ctx, 1);
+	}
+	DUK_ASSERT(level <= -2);
+#endif
+
 	/* [ source ] */
 
 	comp_flags = DUK_JS_COMPILE_FLAG_EVAL;
 	act_eval = thr->callstack + thr->callstack_top - 1;    /* this function */
-	if (thr->callstack_top >= 2) {
+	if (thr->callstack_top >= (duk_size_t) -level) {
 		/* Have a calling activation, check for direct eval (otherwise
 		 * assume indirect eval.
 		 */
-		act_caller = thr->callstack + thr->callstack_top - 2;  /* caller */
+		act_caller = thr->callstack + thr->callstack_top + level;  /* caller */
 		if ((act_caller->flags & DUK_ACT_FLAG_STRICT) &&
 		    (act_eval->flags & DUK_ACT_FLAG_DIRECT_EVAL)) {
 			/* Only direct eval inherits strictness from calling code
@@ -486,14 +499,14 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 	act = thr->callstack + thr->callstack_top - 1;  /* this function */
 	if (act->flags & DUK_ACT_FLAG_DIRECT_EVAL) {
 		DUK_ASSERT(thr->callstack_top >= 2);
-		act = thr->callstack + thr->callstack_top - 2;  /* caller */
+		act = thr->callstack + thr->callstack_top + level;  /* caller */
 		if (act->lex_env == NULL) {
 			DUK_ASSERT(act->var_env == NULL);
 			DUK_DDD(DUK_DDDPRINT("delayed environment initialization"));
 
 			/* this may have side effects, so re-lookup act */
 			duk_js_init_activation_environment_records_delayed(thr, act);
-			act = thr->callstack + thr->callstack_top - 2;
+			act = thr->callstack + thr->callstack_top + level;
 		}
 		DUK_ASSERT(act->lex_env != NULL);
 		DUK_ASSERT(act->var_env != NULL);
@@ -508,7 +521,7 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 			                     "var_env and lex_env to a fresh env, "
 			                     "this_binding to caller's this_binding"));
 
-			act = thr->callstack + thr->callstack_top - 2;  /* caller */
+			act = thr->callstack + thr->callstack_top + level;  /* caller */
 			act_lex_env = act->lex_env;
 			act = NULL;  /* invalidated */
 
@@ -559,7 +572,7 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 	} else {
 		duk_tval *tv;
 		DUK_ASSERT(thr->callstack_top >= 2);
-		act = thr->callstack + thr->callstack_top - 2;  /* caller */
+		act = thr->callstack + thr->callstack_top + level;  /* caller */
 		tv = thr->valstack + act->idx_bottom - 1;  /* this is just beneath bottom */
 		DUK_ASSERT(tv >= thr->valstack);
 		duk_push_tval(ctx, tv);
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -1146,7 +1146,7 @@ DUK_LOCAL void duk__debug_handle_get_var(duk_hthread *thr, duk_heap *heap) {
 
 	str = duk_debug_read_hstring(thr);  /* push to stack */
 	DUK_ASSERT(str != NULL);
-	if (duk_debug_peek_byte(thr) != 0x00) {
+	if (duk_debug_peek_byte(thr) != DUK_DBG_MARKER_EOM) {
 		level = duk_debug_read_int(thr);  /* optional callstack level */
 		if (level >= 0 || -level > (duk_int32_t) thr->callstack_top) {
 			DUK_D(DUK_DPRINT("invalid callstack level for GetVar"));
@@ -1197,7 +1197,7 @@ DUK_LOCAL void duk__debug_handle_put_var(duk_hthread *thr, duk_heap *heap) {
 	duk_debug_read_tval(thr);           /* push to stack */
 	tv = duk_get_tval(ctx, -1);
 	DUK_ASSERT(tv != NULL);
-	if (duk_debug_peek_byte(thr) != 0x00) {
+	if (duk_debug_peek_byte(thr) != DUK_DBG_MARKER_EOM) {
 		level = duk_debug_read_int(thr);  /* optional callstack level */
 		if (level >= 0 || -level > (duk_int32_t) thr->callstack_top) {
 			DUK_D(DUK_DPRINT("invalid callstack level for PutVar"));
@@ -1282,7 +1282,7 @@ DUK_LOCAL void duk__debug_handle_get_locals(duk_hthread *thr, duk_heap *heap) {
 
 	DUK_UNREF(heap);
 
-	if (duk_debug_peek_byte(thr) != 0x00) {
+	if (duk_debug_peek_byte(thr) != DUK_DBG_MARKER_EOM) {
 		level = duk_debug_read_int(thr);  /* optional callstack level */
 		if (level >= 0 || -level > (duk_int32_t) thr->callstack_top) {
 			DUK_D(DUK_DPRINT("invalid callstack level for GetLocals"));
@@ -1339,9 +1339,12 @@ DUK_LOCAL void duk__debug_handle_eval(duk_hthread *thr, duk_heap *heap) {
 
 	DUK_D(DUK_DPRINT("debug command Eval"));
 
-	/* The eval code must be executed within the specified activation.
-	 * For now, use global object eval() function, with the eval considered
-	 * a 'direct call to eval'.
+	/* The eval code is executed within the lexical environment of a specified
+	 * activation. For now, use global object eval() function, with the eval
+	 * considered a 'direct call to eval'.
+	 *
+	 * Callstack level for debug commands only affects scope--the callstack as
+	 * seen by, e.g. Duktape.act() will be the same regardless.
 	 */
 
 	/* nargs == 2 so we can pass a callstack level to eval(). */
@@ -1349,7 +1352,7 @@ DUK_LOCAL void duk__debug_handle_eval(duk_hthread *thr, duk_heap *heap) {
 	duk_push_undefined(ctx);  /* 'this' binding shouldn't matter here */
 
 	(void) duk_debug_read_hstring(thr);
-	if (duk_debug_peek_byte(thr) != 0x00) {
+	if (duk_debug_peek_byte(thr) != DUK_DBG_MARKER_EOM) {
 		level = duk_debug_read_int(thr);  /* optional callstack level */
 		if (level >= 0 || -level > (duk_int32_t)thr->callstack_top) {
 			DUK_D(DUK_DPRINT("invalid callstack level for Eval"));
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -49,6 +49,7 @@ DUK_INTERNAL void duk_debug_do_detach(duk_heap *heap) {
 	heap->dbg_step_thread = NULL;
 	heap->dbg_step_csindex = 0;
 	heap->dbg_step_startline = 0;
+	heap->dbg_have_next_byte = 0;
 
 	/* Ensure there are no stale active breakpoint pointers.
 	 * Breakpoint list is currently kept - we could empty it
@@ -149,6 +150,21 @@ DUK_INTERNAL void duk_debug_skip_byte(duk_hthread *thr) {
  *  Debug connection read primitives
  */
 
+/* Peek ahead in the stream one byte. */
+DUK_INTERNAL uint8_t duk_debug_peek_byte(duk_hthread *thr) {
+	/* It is important not to call this if the last byte read was an EOM.
+	 * Reading ahead in this scenario would cause unnecessary blocking if
+	 * another message is not available.
+	 */
+
+	duk_uint8_t x;
+
+	x = duk_debug_read_byte(thr);
+	thr->heap->dbg_have_next_byte = 1;
+	thr->heap->dbg_next_byte = x;
+	return x;
+}
+
 /* Read fully. */
 DUK_INTERNAL void duk_debug_read_bytes(duk_hthread *thr, duk_uint8_t *data, duk_size_t length) {
 	duk_heap *heap;
@@ -165,7 +181,12 @@ DUK_INTERNAL void duk_debug_read_bytes(duk_hthread *thr, duk_uint8_t *data, duk_
 		goto fail;
 	}
 
+	/* NOTE: length may be zero */
 	p = data;
+	if (length >= 1 && heap->dbg_have_next_byte) {
+		heap->dbg_have_next_byte = 0;
+		*p++ = heap->dbg_next_byte;
+	}
 	for (;;) {
 		left = (duk_size_t) ((data + length) - p);
 		if (left == 0) {
@@ -202,6 +223,11 @@ DUK_INTERNAL duk_uint8_t duk_debug_read_byte(duk_hthread *thr) {
 	if (heap->dbg_read_cb == NULL) {
 		DUK_D(DUK_DPRINT("attempt to read 1 bytes in detached state, return zero data"));
 		return 0;
+	}
+
+	if (heap->dbg_have_next_byte) {
+		heap->dbg_have_next_byte = 0;
+		return heap->dbg_next_byte;
 	}
 
 	x = 0;  /* just in case callback is broken and won't write 'x' */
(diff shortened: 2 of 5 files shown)


fatcerberus:
Regarding the FIXME: There doesn't seem to be any way to read ahead in the debug stream?

I'd like to implement this so that the callstack level is optional (defaulting to -1) to maintain protocol compatibility, but I can't risk reading ahead because if I happen to retire the EOM, then best I can tell the subsequent skip_to_eom() would skip the entire following message.

svaarala:
Yeah, there isn't a read window or an actual "peek for byte(s)" primitive now, the "peek" is for availability of bytes only. This would need some changes to make it easier to do conditional parsing.

For this change specifically it'd actually be enough to just consume the EOM and then somehow flag that it was already consumed so that it didn't need to be scanned for again. But this'd be quite awkward, and wouldn't work in other cases.

Adding the ability to push back a limited number of bytes should work and be relatively clean. We'd then just detect those pushed back bytes in `duk_debug_read_bytes()` (before the actual transport read loop) and `duk_debug_read_peek()` (returning > 0 if there were pushed back bytes).

A limited pushback window won't be enough to handle a generic "unconsume dvalue" case because dvalues have no upper size limit - but it would be enough to unconsume an EOM specifically. It would also allow a conditional dvalue read, e.g. "read dvalue unless EOM" or "read integer dvalue (or fail if not integer)" because one byte is enough to detect the EOM. For these two conditional reads a 1-byte pushback would be enough.

We could also use a read-ahead window so that we'd always try to keep more data available than was consumed. One would have to be careful with blocking here though, as we can't always guarantee there's e.g. at least one extra byte waiting if none are available. So a limited pushback might be easier to implement?

fatcerberus:
The EOM pushback would probably be sufficient in general for the foreseeable future.  This is really the only use case I imagine for it, to ensure `duk__debug_skip_to_eom()` doesn't eat the following message.  For any other dvalue type, if the bytes are consumed then it's a pretty safe assumption that the debug command needed them, otherwise it wouldn't have initiated the read.  It's just that if that read comes back with an EOM instead of a parameter, we need to put it back.

Another issue I can see is that there is apparently no generic helper for reading dvalues.  Concretely in this case, the `duk_debug_read_int()` call will break the connection if it encounters an EOM instead of the expected integer.  Not sure how to go about fixing that yet, I'll have to give the debugger code another lookover.

svaarala:
Right, if peeking for EOM is the only case for the foreseeable future, then just flagging that "oops, EOM is already eaten" and checking for that in the message loop would be sufficient. Hopefully there'd be some clean way to do that.

There's a generic `duk_debug_read_tval()` which will push the value read on the value stack - you can use that for a type independent dvalue read. I wanted to avoid creating yet another intermediate tag/value representation in the debugger code so that's why most read calls will read a specific type, and the few places where an arbitrary type is accepted `duk_debug_read_tval()` is used. It won't accept EOMs etc however, because it's not a *dvalue* reader but a `duk_tval` reader.

The current deubgger implementation is a reflection of the basic idea that Duktape wouldn't accumulate too much version dependent and optional backwards compatible features. Rather, the debug protocol version would be bumped (and announced in the protocol), and the debug client would then know what Duktape expects. That's why the only way to extend messages is by appending new values (which is not very well supported by the current reading code as you noted); there's no structured value support etc.

I won't go into it too much, but my hunch is that at the moment fork lift upgrading the debugger rather doing a lot of incremental development to incorporate features like heap walking would be a better option. For backwards compatibility it'd be relatively simple to support both protocol version 1 (current) and 2 (next) at the same time, with separate `duk_debugger.c` files providing the debugger implementation. Even going forward supporting "current" and "previous" fork lift version might be reasonable, because the debugger integrates quite narrowly to other internals -- at least for the time being.

svaarala:
All that said, I won't object to this pull if it can be implemented cleanly :-)

fatcerberus:
I'm generally good at maintaining compatibility in a clean way; I managed to maintain full Sphere 1.5 compatibility in minisphere 2.0 without any ugly hacks, so I'll see what I can do here.

My main thing is I don't want to bump the protocol version unnecessarily for such a tiny change, and it probably makes sense for the call index to be optional anyway (using the top activation by default), even for clients which are aware of the extra parameter.

svaarala:
It's not so much the difficulty but the effort related in updating documentation incrementally, regression testing, etc. Sometimes this is unavoidable, and it does affect other areas of Duktape.

Sometimes it's cleaner and more efficient to do "feature packs" rather than individual features. With the debugger this might be the case, because the current debugger version really is just a "first pass" at debugger support. It was quite tightly timeboxed too, so several compromises had to be made. To improve on the basic structure, backwards incompatible changes will be needed pretty soon, and I hope to avoid a too long backwards compatibility baggage going forwards.

Obviously this is balancing between user convenience and overall project progress. That's why I've been considering support a relatively static "previous" (or "stable") debugger protocol version and a "current" (or "in progress") debugger protocol version simultaneously. Incompatible changes could be made freely for the "in progress" version while the "stable" version wouldn't gain much features but we wouldn't be causing debug clients too much hassle until they're ready to upgrade. At any point we'd only support two protocol versions so old features would phase out naturally.

To be clear, I don't mean the breaking changes will be "rewrite 50% of the debug client" :) But rather, we wouldn't need to hesitate about reorganizing some command format if it needs to be done. Longer term this might be more desirable than shoehorning additional parameters in, or establishing new commands with slight parameter variations (for example), because all that stuff adds up over time.

The other approach would be for the protocol to be more easily extensible (support structured values with optional keys for example instead of relying on a list of dvalues only). But that has other costs, of course.

Anyway, this was more of a high level issue, not really an issue for this pull.

fatcerberus:
I figured out a way to do this cleanly and generically:

1. Implement a one-byte pushback, like C `ungetc()`.
2. Add a helper to read the next byte from the stream, and immediately push it back using the aforementioned mechanism.  Conceptually this is simply a one-byte lookahead.
3. Based on the byte returned to the caller, it can decide how to proceed.  Since dvalue type is always encoded by a single byte and Duktape assumes it can read an entire message at a time, unnecessary blocking isn't an issue.
4. In any case, the terminating EOM is never accidentally eaten (because.Step 2 puts it back).

svaarala:
That sounds good to me -- the pushback byte would then need special handling in the lowest level read primitive but nothing above that (including "skip bytes" which just reads and will just work).

fatcerberus:
This is one of the reasons I like NUL-terminated strings (and terminators in general) too: If you can assume the terminator is always present, then you can always safely look one element ahead.

fatcerberus:
Quick question that I can't ascertain from looking over the debugger code: Is `duk_debug_read_bytes()` guaranteed to be called with `length >= 1`?  Or can the length be zero in the case of, e.g. empty strings?

fatcerberus:
@svaarala Here's a much better implementation.  Let me know what you think.

svaarala:
A zero-read is possible yes, at least for strings. It might be good to note that in a comment because as you say it's not apparent.

svaarala:
Looks pretty good to me at a glance :) Adding initialization/assert for at least "have next byte" is needed to handle multiple attaches.

fatcerberus:
Calling attach while already attached is a no-op right?  `heap->dbg_have_next_byte` is set to 0 (false) on detach, so an assert should be sufficient I think.

svaarala:
Hmm, I guess attach should actually check for an attach without previous detach - but it doesn't now.

Is there a guarantee `dbg_have_next_byte != 0` if, say, an error is thrown during debugger message handling and a followup detach?

I guess the consistent handling would be to zero them out in `duk_debug_do_detach()` and initialize them in `duk_debugger_attach()` as is done for other variables. That pushback state is associated with a specific transport so it should never carry over a transport change.

fatcerberus:
`duk_debug_do_detach()` in the current branch sets `dbg_have_next_byte` to zero, so as long as something doesn't go horribly wrong causing the detach handler not to be called, that part should be fine.

I agree that it should be explicitly initialized to zero on attach, I'll add that.

fatcerberus:
Somewhat related: I tried to implement a callstack level for Eval as well, but ran into an issue: the Eval handler simply calls the built-in `eval()` which naturally always runs in the context of the current activation.  I first considered a hack like this:

```javascript
	old_top = thr->callstack_top;
	thr->callstack_top -= (level + 1);
	call_ret = duk_handle_call(thr, 1 /*num_stack_args*/, call_flags);
	thr->callstack_top = old_top;
```

But then realized that would clobber the activations above `level`.  So it doesn't seem like implementing this for Eval will be an easy change.  I'll look into it further after this is merged.

fatcerberus:
Alright, I cleaned up the changes, added some comments, and also implemented callstack level for GetVar while I was at it.

fatcerberus:
By the way, I managed to get callstack level working for Eval in my local copy, as well. :)  As I said, I'll open a separate pull for that once everything has been sorted out in this one.

svaarala:
Setting the top would also change GC reachability which would probably be unfixable.

The easiest change is probably to refactor eval() a little bit - basically you want to compile a function (eval code) and give it an explicit environment record. I can help with that change - maybe we could do it in a separate pull?

fatcerberus:
I managed to get it working in my local copy by passing a callstack index as a second argument to the built-in Duktape/C eval function.  Ecmascript evals never see this because the built-in function is pushed with nargs == 1. :)

I didnt add it to this pull though because I think it's probably best to do the Eval change in a separate pull, there are a lot more things that can go wrong. JS eval is a very strange beast.

By the way, how do I add the mergeready label?  It doesn't seem to let me edit the labels for the pull.

fatcerberus:
There may be a bug in this related to running GetLocals/GetVar against a native activation (Duktape doesn't send a reply), further testing will be needed to ensure the bug is in Duktape code and not my client implementation.

fatcerberus:
If you're curious, here's a prototype I came up with for Eval callstack level:
https://github.com/fatcerberus/duktape/commit/a62a146e5cdbf3882af08ea0719f2ea01870b051

fatcerberus:
Follow up to above: The implementation works flawlessly, it was my glitchy thrown-together client. :)

I consider this merge-ready from my end.  We'll do the Eval change in a separate pull, the current setup I have for that is kind of hacky.

svaarala:
@fatcerberus Sounds good :) I've been quite busy the past few days (and will be for a few more), I'll try to look into this ASAP. As for the labeling, I guess you can't enable public labeling which is too bad :/

src/duk_debugger.c
```diff
@@ -149,6 +150,22 @@ DUK_INTERNAL void duk_debug_skip_byte(duk_hthread *thr) {
  *  Debug connection read primitives
  */
 
+/* Peek ahead in the stream one byte. */
+DUK_INTERNAL uint8_t duk_debug_peek_byte(duk_hthread *thr)
+{
```
svaarala: Style trivia: opening brace on same line as function args.

svaarala:
After a first read-through of the diff, everything seemed logical. That pushback byte worked its way in there pretty cleanly after all :) I'll give it a bit more thorough read-through tomorrow and see if I could add some support into the web UI (useful for manual regression testing too).

fatcerberus:
Okay, I fixed the style issue you mentioned and also implemented callstack level for PutVar as well, for symmetry with GetVar.

fatcerberus:
After quite a bit of real-world testing I changed my mind: I cleaned up the Eval changes and merged them into this branch.  I still consider this merge-ready.

src/duk_debugger.c
```diff
@@ -1284,27 +1346,41 @@ DUK_LOCAL void duk__debug_handle_eval(duk_hthread *thr, duk_heap *heap) {
...
+	/* nargs == 2 so we can pass a callstack level to eval(). */
+	duk_push_c_function(ctx, duk_bi_global_object_eval, 2 /*nargs*/);
 	duk_push_undefined(ctx);  /* 'this' binding shouldn't matter here */
 	(void) duk_debug_read_hstring(thr);
+	if (duk_debug_peek_byte(thr) != 0x00) {
+		level = duk_debug_read_int(thr);  /* optional callstack level */
+		if (level >= 0 || -level > (duk_int32_t)thr->callstack_top) {
+			DUK_D(DUK_DPRINT("invalid callstack level for Eval"));
+			duk_debug_write_error_eom(thr, DUK_DBG_ERR_NOTFOUND, "invalid callstack level");
+			return;
```
svaarala: When this error happens the current code leaves temporaries on the value stack. These should cause an assert failure in the debugger "process messages" loop.

It might actually be less error prone to change the "process messages" loop to store the top-at-entry value, and forcibly reset the value stack top when completing a message handle: this would be more robust and would allow error paths to just return instead of carefully pop stuff first.

fatcerberus: So in other words, the same way Duktape handles unmatched pushes in a Duktape/C function.  That's probably a good idea, it will make error handling cleaner at least.  That said, I'll fix the bug here. :-)

svaarala:
Re: that value stack thing, the best place might be in "duk__debug_process_message".

I'll merge a cleanup step into master with this change - I'll also clean the other message handlers.

fatcerberus:
So do you want me to fix the bug here, or...?

svaarala:
It's probably best to make the structure more robust - I'll post a branch shortly with the message loop change. If you rebase on that there should be no need to fix the issue.

svaarala:
Ok, this is the pull: https://github.com/svaarala/duktape/pull/419

I'll test the debugger functionality before merging, should be in master soon.

svaarala:
Ok that value stack change is in master now.

fatcerberus:
Thanks, that was quick. :-)  Give me a minute to fix the conflict and make sure everything's still correct and then I'll push the rebased branch.

fatcerberus:
Okay, rebased on top of latest master, waiting on Travis checks.

fatcerberus:
With regard to the protocol version discussion earlier, is that versioned independently of Duktape itself?  e.g. Would be considered acceptable in general to bump the protocol version in a minor release?

svaarala:
Versioning is still a bit of a work in progress. The explicit version is there so that the wire protocol can be identified directly without having a Duktape version table or similar in the debug client which is good for forwards compatibility (how would the debug client know how future Duktape versions map to protocol versions).

Having a separate version also makes it possible for a certain Duktape version to support two debugger protocol versions (selectable through config options) if that turns out to be a good approach - which means that a Duktape version isn't necessarily 1:1 with the debug protocol version.

For now, I think it'd be best to make incompatible debugger protocol changes in Duktape major version bumps only.

fatcerberus:
One thing that probably needs cleanup is the documentation for the Eval command.  I wasn't sure how to properly reword it to reflect the new functionality.

fatcerberus:
Did you need me to do anything else for this?

svaarala:
Not off the top of my head, I'm hoping to add something to the web UI to be able to manually test this feature for regressions. Or maybe I'll just use the JSON proxy for that.

svaarala:
Did some testing with the JSON proxy, and mostly things seem to be working fine.

However, found the following issue with Eval. The callstack here is: global -> foo() -> bar():

```
Error: aiee
	eval input:1 preventsyield
	anon  native strict directeval preventsyield
	bar _test.js:5
	foo _test.js:10
	global _test.js:13 preventsyield
```

Doing an Eval request with an arbitrary call stack index:

```
{"request":"Eval","args":["print(new Error('aiee').stack)",-2]}
{"notify":"Print","command":2,"args":["Error: aiee\n\teval input:1 preventsyield\n\tanon  native strict directeval preventsyield\n\tbar _test.js:5\n\tfoo _test.js:10\n\tglobal _test.js:13 preventsyield\n"]}
```

When the eval level is given as e.g. -2 or -3, the error stack trace will still reflect the whole call stack so that it's independent of what level is given.

After all, the call stack is in place and hasn't been dismantled for the duration of Eval, so that any internal code dealing with the call stack will still see a full call stack. Also any new calls made by the code being Eval'd will conceptually be on top of the full call stack - only the lexical environment of the Eval has been changed. A similar issue would happen for `Duktape.act()` where the indices would map to the full call stack regardless of the given Eval level.

I think this needs at least a short note in the documentation, e.g. the Eval happens in the lexical scope indicated by the call stack level index, but the call stack itself is unchanged so that e.g. tracebacks and `Duktape.act()` will see the full call stack.

svaarala:
Here's a JSON proxy example for reading local variables off every level:

```
$ telnet localhost 9093
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
{"notify":"_Connected","args":["1 10399 v1.2.0-778-gd7838fe duk command built from Duktape repo"]}
{"notify":"Status","command":1,"args":[1,"_test.js","global",1,0]}
{"request":"AddBreak","args":["_test.js", 7]}
{"reply":true,"args":[0]}
{"request":"Resume"}
{"reply":true,"args":[]}
{"notify":"Status","command":1,"args":[0,"_test.js","global",1,0]}
{"notify":"Status","command":1,"args":[1,"_test.js","bar",7,1]}
{"request":"GetLocals","args":[]}
{"reply":true,"args":["b",{"type":"number","data":"401c000000000000"},"y",{"type":"number","data":"4000000000000000"}]}
{"request":"GetLocals","args":[-1]}
{"reply":true,"args":["b",{"type":"number","data":"401c000000000000"},"y",{"type":"number","data":"4000000000000000"}]}
{"request":"GetLocals","args":[-2]}
{"reply":true,"args":["a",{"type":"number","data":"4059000000000000"},"bar",{"type":"object","class":6,"pointer":"000000000260c9e0"},"x",{"type":"number","data":"3ff0000000000000"}]}
{"request":"GetLocals","args":[-3]}
{"reply":true,"args":[]}
{"request":"GetLocals","args":[-4]}
{"error":true,"args":[3,"invalid callstack level"]}
```

src/duk_bi_global.c
```diff
@@ -448,15 +449,27 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 		return 1;  /* return arg as-is */
 	}
 
+#if defined(DUK_USE_DEBUGGER_SUPPORT)
+	/*  XXX: level is used only by the debugger and should never be present
+	 *  for eval() called from Ecmascript code. is there a way to assert for
+	 *  this?
+	 */
```
svaarala: I think it'd be just fine to comment shortly here that level is present only for debugger-initiated calls. The actual eval() binding has nargs==1 but it's not easy to check here.

fatcerberus:
The callstack behavior mentioned above seems intuitive to me, at least.  Agreed that a note in the documentation should be sufficient.

src/duk_debugger.c
```diff
@@ -1113,16 +1139,27 @@ DUK_LOCAL void duk__debug_handle_get_var(duk_hthread *thr, duk_heap *heap) {
...
 	duk_bool_t rc;
+	duk_int32_t level;
 
 	DUK_UNREF(heap);
 	DUK_D(DUK_DPRINT("debug command GetVar"));
 
 	str = duk_debug_read_hstring(thr);  /* push to stack */
 	DUK_ASSERT(str != NULL);
+	if (duk_debug_peek_byte(thr) != 0x00) {
+		level = duk_debug_read_int(thr);  /* optional callstack level */
```
svaarala: Here 0x00 could be `DUK_DBG_MARKER_EOM`.

src/duk_bi_global.c
fatcerberus: I always thought nargs for Duktape/C functions was pointless (I just use `DUK_VARARGS` for everything), but it came in really handy here. :)

src/duk_debugger.c
```diff
@@ -1159,10 +1197,20 @@ DUK_LOCAL void duk__debug_handle_put_var(duk_hthread *thr, duk_heap *heap) {
 	duk_debug_read_tval(thr);           /* push to stack */
 	tv = duk_get_tval(ctx, -1);
 	DUK_ASSERT(tv != NULL);
+	if (duk_debug_peek_byte(thr) != 0x00) {
+		level = duk_debug_read_int(thr);  /* optional callstack level */
```
svaarala: 0x00 -> `DUK_DBG_MARKER_EOM`.

src/duk_debugger.c
```diff
@@ -1229,15 +1277,28 @@ DUK_LOCAL void duk__debug_handle_get_call_stack(duk_hthread *thr, duk_heap *heap
...
 	duk_activation *curr_act;
+	duk_int32_t level;
 	duk_hstring *varname;
 
 	DUK_UNREF(heap);
 
-	duk_debug_write_reply(thr);
-	if (thr->callstack_top == 0) {
-		goto callstack_empty;
+	if (duk_debug_peek_byte(thr) != 0x00) {
```
svaarala: 0x00 -> `DUK_DBG_MARKER_EOM`.

src/duk_debugger.c
```diff
@@ -1272,28 +1333,44 @@ DUK_LOCAL void duk__debug_handle_eval(duk_hthread *thr, duk_heap *heap) {
...
+	 * a 'direct call to eval'.
 	 */
 
-	duk_push_c_function(ctx, duk_bi_global_object_eval, 1 /*nargs*/);
+	/* nargs == 2 so we can pass a callstack level to eval(). */
+	duk_push_c_function(ctx, duk_bi_global_object_eval, 2 /*nargs*/);
 	duk_push_undefined(ctx);  /* 'this' binding shouldn't matter here */
+
 	(void) duk_debug_read_hstring(thr);
+	if (duk_debug_peek_byte(thr) != 0x00) {
```
svaarala: 0x00 -> `DUK_DBG_MARKER_EOM`.

src/duk_debugger.c
```diff
@@ -1272,28 +1333,44 @@ DUK_LOCAL void duk__debug_handle_eval(duk_hthread *thr, duk_heap *heap) {
...
 
 	DUK_D(DUK_DPRINT("debug command Eval"));
 
-	/* The eval code must be executed within the current (topmost)
-	 * activation.  For now, use global object eval() function, with
-	 * the eval considered a 'direct call to eval'.
+	/* The eval code must be executed within the specified activation.
+	 * For now, use global object eval() function, with the eval considered
+	 * a 'direct call to eval'.
 	 */
```
svaarala: Perhaps note here, too, that the callstack index affects the scope seen by Eval but not the call stack (i.e. full call stack will be visible in tracebacks etc).

svaarala:
All right, the trivia is commented in commits and the other stuff I had was:

- Comment on the Eval call stack behavior briefly in the debugger document.
- Remove call stack level references from debugger future work as appropriate (there are a few clear references to this feature)

Good stuff :)

src/duk_bi_global.c
svaarala: Yeah, it's intended for situations where the function really just wants a fixed configuration in its entry value stack and doesn't want to worry about missing/extra arguments. Both are useful I guess :)

svaarala:
Action points for separate pulls (for me) are:

- Rebase the reattach fix branch and ensure `dbg_have_next_byte` and `dbg_next_byte` are initialized properly.
- Work in minimal callstack level support into the web UI somehow. Perhaps so that if you click on a callstack level, it will be used in followup Eval/GetVar/PutVar calls.

src/duk_debugger.c
fatcerberus: See, this was my first instinct too, but then my OCD got the better of me (because the `peek`'d byte isn't necessary always a dvalue marker, although it is in this case).

svaarala: That'd be true in the general case, but isn't it necessarily an "initial byte" here (and markers are initial bytes)?

svaarala: I have a little bit of OCD myself when it comes to code trivia, so I can symphatize on the hard but small choices :-)

fatcerberus: Yes, in this case it would always be an IB.  I just tend to think in the general case "by default".  Anyway I'll fix these along with the other issues you brought up.

fatcerberus:
Regarding the debugger UI, how I implemented it in the minisphere debugger is if you double-click a callstack entry, it updates the local variable list by calling GetLocals at that level, and then subsequent Eval commands operate at that level.

svaarala:
Sounds reasonable - I'd need to maybe add highlighting for the current level, etc. The web UI was written pretty much as a throw-away prototype though so I'm not sure how far I can take it without improving the structure a bit first :)

fatcerberus:
Okay, rebased and trivia addressed.  How's this?

src/duk_debugger.c
```diff
@@ -149,6 +150,21 @@ DUK_INTERNAL void duk_debug_skip_byte(duk_hthread *thr) {
...
 
+/* Peek ahead in the stream one byte. */
+DUK_INTERNAL uint8_t duk_debug_peek_byte(duk_hthread *thr) {
+	/* It is important not to call this if the last byte read was an EOM.
+	 * Reading ahead in this scenario would cause unnecessary blocking if
+	 * another message is not available.
+	 */
+
+	duk_uint8_t x;
+
```
svaarala: By the way, should there be a `DUK_ASSERT(thr->heap->dbg_have_next_byte == 0);` here? In other words, it's not allowed to call `duk_debug_peek_byte()` twice without consuming the pushed back byte first?

svaarala:
One more commit note on an issue I just spotted, and a short note on Eval call stack behavior on `debugger.rst` would be nice. But otherwise good to go :)

src/duk_debugger.c
fatcerberus: Actually that's safe--it will just read back the pushback byte. :)

svaarala: Ah true, never mind :)

fatcerberus:
Ah, forgot to add a comment on that in debugger.rst, sorry about that.  I'll do that after my dinner. :)

fatcerberus:
Okay, that's done.

svaarala:
Ok, is this merge ready or any incoming changes?

fatcerberus:
It's merge ready as long as toy have no further issues.

svaarala:
Ok, I'll merge this and start a separate pull for the web UI stuff etc. Thanks for the nice work :)

fatcerberus:
"Toy" should be "you", my phone is weird. :)

PR #417, opened 2015-10-29
More verbose error message for "not callable"

svaarala:
Use `duk_push_string_tval_readable()` summary when a call target is not callable.

Tasks:

- [x] Separate config option or merge with property config option? => replace DUK_USE_VERBOSE_PROP_ERRORS with DUK_USE_PARANOID_ERRORS
- [x] Add to config examples
- [x] Testcase fixes and improvements
- [x] Check whether CSVAR, CSPROP could indicate TypeError cause directly for better error messages
- [x] Releases entry


fatcerberus:
For the config option, for my part I think it should be merged with the property option.  As you mentioned in another issue discussion, summarizing things is a sandboxing/security concern, so splitting up the options for this and property names doesn't really make sense.  If you disable property names in errors, you'd almost certainly want to disable this too.

svaarala:
I tend to agree - though the cases are slightly different because with property errors the *key* is included in the message (e.g. "'mySecretKey' of undefined"), but for these function calls only the target *value* (e.g. "undefined") is in the message.

fatcerberus:
That's true.  Maybe a better option then would be to have two generic config options--one to disable displaying keys in errors, and another to disable summarization.

svaarala:
Dunno, will have to think about that a bit.

Ideally even the call errors would indicate what property or variable name lead to the call failure. That's not trivial right now because "call setup" is a separate step from actually performing the call, and after call setup we only have the function (or whatever value) and the requested "this" binding.

A partial fix would be throw the `TypeError` already from the call setup, at least for the most common cases. In that case the error message could include e.g. the property key in `foo.bar()` (-> `'bar' of undefined not callable`). In this case the error would include a key so the cases would then be similar-

For security sensitive environments one might want to have otherwise verbose errors but never reveal any data or key names, so a basic "verbose" vs. "non-verbose" division is also not that great. Perhaps there should be two options:

- Minimal vs. verbose errors (= DUK_USE_VERBOSE_ERRORS)
- Keys and other potentially unsafe stuff in verbose errors (= DUK_USE_XXX_ERRORS)

fatcerberus:
What would a "verbose" error look like if it didn't have the sensitive data in it?  If that just means rewording the existing messages to be clearer, I don't see why `DUK_USE_VERBOSE_ERRORS` would even be necessary?

svaarala:
Right now if you turn `DUK_USE_VERBOSE_ERRORS` off, the error messages are dropped so that you get e.g.:

```
duk> aiee
ReferenceError: 103
	global input:1 preventsyield
```

This is for low memory environments where both ROM and RAM footprint matter a great deal.

So if you drop the keys etc off the property error messages, it'd be awkward to drop into that directly if the motivation for doing so is security instead of footprint. It'd be more useful to get e.g. `ReferenceError: identifier undefined`.

fatcerberus:
Oh, that make sense, I didn't realize `DUK_USE_VERBOSE_ERRORS` was an existing option.  I see what you mean now.

svaarala:
This would probably make most sense to me:

- Default: verbose errors with maximal readability, summarizing keys and unexpected values, etc.
- Paranoid errors: avoid exposing values etc in error messages, but otherwise verbose.
- Minimal errors: drop error messages for low memory footprint.

Mapping to options:

- DUK_USE_VERBOSE_ERRORS is already used and defaults to true. Setting to false turns on minimal errors for low memory environments.
- DUK_USE_PARANOID_ERRORS would replace DUK_USE_VERBOSE_PROP_ERRORS and default to false. Setting to true with verbose errors enabled would avoid keys, values, etc in error messages for security conscious environments.

In any case, the default would be maximal verbosity because that's the most common case.

fatcerberus:
That sounds good to me. :)

svaarala:
Right now this is what happens:

```
$ ./duk
((o) Duktape 1.3.99 (v1.3.0-238-g29436de)
duk> (undefined)()
TypeError: undefined not callable
	duk_js_call.c:722
	global input:1 preventsyield
duk> (123)()
TypeError: 123 not callable
	duk_js_call.c:722
	global input:1 preventsyield
```

But:

```
duk> obj={}; obj.foo();
TypeError: undefined not callable
	duk_js_call.c:722
	global input:1 preventsyield
```

Ideally the message here would be something like "property 'foo' of [object Object] not callable".

Ecmascript requires that the callability of the call target is only checked once the argument list has been evaluated. This doesn't work well with the current compiler opcode sequence which is: CSPROP, argument evaluation, CALL. When the CALL opcode is executed, there is no "memory" of the call setup anymore.

The error could also be thrown from CSPROP directly. Or perhaps the opcode sequence could be: arguments, CSPROP, CALL instead (with potentially the two opcodes merged to CALLPROP).

These changes are best made in a separate pull.

fatcerberus:
Hm, looks like a few testcases are failing judging by the Travis logs.  Also, Travis hasn't figured out that one of the builds has finished, weird.

svaarala:
Yeah, I haven't fixed the testcases yet (expect strings change for at least one case). I also want to add a testcase for these new error messages - both to document current behavior and to notice if something inadvertently changes the behavior.

rbloch:
Hello,

I'm commenting here because it's directly linked to this merge request. I've a lot of 'undefined is not callable' functions when calling C functions from ECMAScript (the functions are indeed not yet implemented) and i would like to have the function names. After reading this merge request it seemed to me that i just have to change

`#undef DUK_USE_PARANOID_ERRORS`

to this

`#define DUK_USE_PARANOID_ERRORS`

to get better message. However, when i change it in the latest release, it keeps showing me the same error. I tried to tweak the code to force it to display the advanced errors, no chance.

Is there something else that i've to do? I didn't find any doc on how to really setup this new verbose mode. 

thanks a lot.

svaarala:
There's no option which will give a function name e.g. when you do:

```js
myGlobalFunction()
```

The only change was that if you call with the property style:

```js
foo.bar()
```

the key name ('bar') is included in the error (this may be different than the function's `.name` property).

This behavior is enabled in master by default: DUK_USE_VERBOSE_ERRORS defines, and DUK_USE_PARANOID_ERRORS undefined.

rbloch:
well that's exactly my point: i call

`var annualValue = task.getDatafieldValueAsBigDecimal('CGEBFI.BUDGET13');`

and i display the result in a message box : 



shouldn't i get `getDatafieldValueAsBigDecimal` in the message? or did i totally miss something there

svaarala:
Hmm, you should - this is from master:

```
((o) Duktape [linenoise] 2.1.99 (v2.1.0-361-gbf6c1ec7-dirty)
duk> task={}
= {}
duk> var annualValue = task.getDatafieldValueAsBigDecimal('CGEBFI.BUDGET13');
TypeError: undefined not callable (property 'getDatafieldValueAsBigDecimal' of [object Object])
    at [anon] (prep/nondebug/duktape.c:73980) internal
duk> 
```

Are you running from master?

svaarala:
> However, when i change it in the latest release, [...]

This seems to imply you're not running from master; if so, this change in error messages is not yet in any release. It will be part of 2.2.0 which is not yet released.

rbloch:
i'm using it through C api with a custom fatal handler but yeah i'll check again

svaarala:
I meant which Duktape release are you using - e.g. 2.1.0 or snapshot from master?

svaarala:
If that's the case, it was tracked as future work in #1633. I've opened pull #1644 to extend the error message improvement to duk_(p)call_prop() too. I should be able to get that into master quite soon.

rbloch:
i create a heap

```c++
duk_context* newHeap = duk_create_heap(NULL, NULL, NULL, NULL, this->execution_error);`
```
with the following error function

```c++
void JavaScriptKernel::execution_error(void* udata, const char* msg)
{
    throw(std::logic_error(msg));
}
```

I then display the message msg in a windows dialog as i showed.

To call the script i showed i used a simple 

```c++
duk_call(context, args.size());
```
after evaluating the content of a file that contains the function

I wasn't on master, i was on master from 1week ago, i need to try tomorrow with the latest version.

svaarala:
Hmm, assuming the call itself is in Ecmascript code (and not a duk_(p)call_prop()) then current master should work. But #1630 was merged 5 days ago so maybe you didn't have it yet.

rbloch:
i'll tell you tomorrow.. i probably got master from 6 days ago -_-'

rbloch:
Ok it's working. I had trouble seeing it as the error message in the fatal is always truncated for some reasons (as we can see in the following screenshot.. really truncated, not just a preview of the value)

truncated-error

svaarala:
There's a sanity limit on the fatal error message (it's formatted to a stack based buffer). It should be increased in size now with more verbose errors.

svaarala:
Hmm, the message in your case is actually truncated to 47 bytes which is smaller than the fatal error buffer (64 bytes). The truncation actually seems to happen when the fatal error handler is using the internal "summarize value to string safely" primitive to come up with a description of the error. That summarization helper has limits for string lengths so that error messages don't become too long (for example if a property name is 1kB long it's not desirable in a summary string).

But those limits are perhaps now too small to handle uncaught errors + more verbose errors properly.

rbloch:
On master the 

`#define DUK__READABLE_STRING_MAXCHARS 32`

is set to 32. By increasing the constant to 128 i managed to get a bit more chars but not everything which is weird as i have way less than 128 chars. Thus i guess that there is a limit set elsewhere. As you said, i also think that you should increase the limit. 

buffer-size

svaarala:
There are two limits that affect the result:
- There's a 64-byte buffer for fatal error message formatting in general (for uncaught errors). I think you're hitting that limit now, after your change.
- The generic internal primitive to summarize values safely without side effects had a 32-char limit. With the change you made you're no longer hitting that.

If you have extra time on your hands, #1653 should improve the situation a bit for both fatal error message formatting and the error summarization which fatal error formatting relies on. It still won't capture the longest possible keys/object names but it'd be nice to hear if it worked for your case.

I tried simulating the previous case by making 'duk' fail with an uncaught error instead of cleanly catching it:
```
((o) Duktape [linenoise] 2.1.99 (v2.1.0-383-g983db958-dirty)
duk> task={};
= {}
duk> var annualValue = task.getDatafieldValueAsBigDecimal('CGEBFI.BUDGET13');
*** FATAL ERROR: uncaught: 'undefined not callable (property \x27getDatafieldValueAsBigDecimal\x27 of [object Object])'
```

It's not very pretty because there's a double escaping of single quotes but that's a separate matter.

On a general note, I'd recommend using a duk_safe_call() wrapper or other protected calls instead of relying on the fatal error handler for diagnostics. It's just much cleaner, and you can get tracebacks, etc.

PR #419, opened 2015-10-31
More robust value stack handling for debugger message loop

svaarala:
- Change duk__debug_process_message() to store/restore value stack top for message handlers.  This is more robust than requiring that every debug message handler code path always manages the value stack 100% correctly.

- Change existing message handlers so that trivial value stack cleanups are not done in the message handler.  Value stack popping is still necessary for loops where the value stack might grow without bound.

### DIFF
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -982,6 +982,10 @@ DUK_LOCAL void duk__debug_skip_to_eom(duk_hthread *thr) {
 
 /*
  *  Process incoming debug requests
+ *
+ *  Individual request handlers can push temporaries on the value stack and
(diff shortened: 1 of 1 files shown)


svaarala:
Ok, everything I could think of passes with assertions enabled - @fatcerberus: look good to you?

svaarala:
Added an assert just before re-setting the value stack top: the top before resetting must be >= entry top.

In other words, message handlers can push stuff on the stack but must not consume existing entries.

fatcerberus:
Wow, that's much cleaner, got rid of a lot of code noise (all the extra pops).  I like it, it appeals to my love of generalization. :-)

PR #421, opened 2015-10-31
Add a few convenience flags to duk_def_prop()

svaarala:
Add convenience flags to set/clear writable, enumerable, and configurable. These defines make call sites a bit more readable:

```
duk_def_prop(ctx, -3, DUK_DEFPROP_SET_WRITABLE);
```

instead of:

```
duk_def_prop(ctx, -3, DUK_DEFPROP_HAVE_WRITABLE | DUK_DEFPROP_WRITABLE);
```

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1155,6 +1155,9 @@ Planned
 * Allow debugger detached callback to call duk_debugger_attach(), previously
   this clobbered some internal state (GH-399)
 
+* Add defines DUK_DEFPROP_{SET,CLEAR}_{WRITABLE,ENUMERABLE,CONFIGURABLE} for
+  duk_def_prop() to improve call site readability (GH-421)
+
 * Add a combined duktape.c without #line directives into the dist package,
   as it is a useful alternative in some environments (GH-363)
 
--- a/src/duk_api_public.h.in
+++ b/src/duk_api_public.h.in
@@ -176,6 +176,12 @@ struct duk_number_list_entry {
 #define DUK_DEFPROP_HAVE_GETTER           (1 << 7)    /* set getter (given on value stack) */
 #define DUK_DEFPROP_HAVE_SETTER           (1 << 8)    /* set setter (given on value stack) */
 #define DUK_DEFPROP_FORCE                 (1 << 9)    /* force change if possible, may still fail for e.g. virtual properties */
+#define DUK_DEFPROP_SET_WRITABLE          (DUK_DEFPROP_HAVE_WRITABLE | DUK_DEFPROP_WRITABLE)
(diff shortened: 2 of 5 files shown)


fatcerberus:
This is nice.  I always wondered why duk_def_prop() wasn't set up like this to start, honestly.

fatcerberus:
Actually now that I think about it, are the CLEAR_ variants needed?  I thought the default if a field isn't provided is false?

svaarala:
The CLEAR variants are the same as just saying e.g. `DUK_DEFPROP_HAVE_WRITABLE` but the call site intent will be clearer:

```
duk_def_prop(ctx, -3, DUK_DEFPROP_CLEAR_WRITABLE);
```

vs.

```
duk_def_prop(ctx, -3, DUK_DEFPROP_HAVE_WRITABLE);
```

svaarala:
Note that the attributes need to be three-state: "set", "clear", or "don't touch", and that's why there are separate flags for indicating that you want to change an attribute (e.g. DUK_DEFPROP_HAVE_WRITABLE) and a separate one providing the value (e.g. DUK_DEFPROP_WRITABLE).

fatcerberus:
So is this: 

‘‘‘javascript
duk_def_prop(ctx, -3, DUK_DEFPROP_CLEAR_WRITABLE);
‘‘‘

Different than:
‘‘‘javascript
duk_def_prop(ctx, -3, 0);
‘‘‘

?

svaarala:
Yes, the latter doesn't change any attributes of the target object - previous values will be kept.

svaarala:
In Javascript terms, the first example call would be equivalent to:

```js
Object.defineProperty(obj, "prop", { writable: false });
```

while the second would be:

```js
Object.defineProperty(obj, "prop", { });
```

which would be a no-op for most purposes.

svaarala:
Btw, note that both `Object.defineProperty()` and `duk_def_prop()` are used to (a) create new properties, and (b) edit existing properties. That's why a three-state flag approach is needed: for creating new properties defaulting works fine, but for editing you need to be able to signal what is changed and what is not touched.

fatcerberus:
Ah, I was unclear on the semantics, for some reason I thought not providing a flag for defineProperties was the same as setting it to false.

Thanks for clearing that up. :)

svaarala:
That is effectively the case when you're creating a new property because the default attributes are `false`. But it's not the case for editing existing properties.

fatcerberus:
Not directly relevant to this pull but related: Will the old flags be deprecated in Duktape 2.0?

svaarala:
The new flags are just shorthands for the existing ones, so I don't think they need to be deprecated.

PR #426, opened 2015-11-02
Fix gcc warnings for -Wcast-qual

svaarala:
Current goal for warning free compilation with GCC is `-Wall -Wextra`. Add `-Wcast-qual` to this set and fix the compile warnings issued.

Tasks:

- [x] Fix `const` related issues, mostly theoretical / cosmetic issues
- [x] Fix `volatile` related issues, potential porting issue when volatile-to-nonvolatile casts are used
- [x] Update code issues documentation for the best approaches
- [x] Releases entry
- [x] Test on Windows / MSVC
- [x] Test run using assertions


svaarala:
There's a Windows `snprintf()` regression in master, I'll fix that in a separate pull.

fatcerberus:
Regarding `duk_push_pointer()` in particular, is there a reason it doesn't just accept a `const void*` directly?  The pointer is meant to be completely opaque to Duktape so that seems like it would be cleaner than changing the call sites to cast away `const`-ness.

svaarala:
I could change that in the API, but then the caller would need to deal with `duk_get_pointer()` returning a `const void *` and potentially casting away the `const` there.

A similar issue exists for a volatile pointer: `duk_push_pointer()` doesn't accept a volatile pointer and casting from volatile to non-volatile has potential portability issues. It should be safe if you then cast back into a volatile on the pointer-reading end before dereferencing the pointer at all, but that cast will still be there (causing warnings). Changing `duk_push_pointer()` to accept a `const volatile *` pointer, and `duk_get_pointer()` to return one would be odd IMO.

The two functions could be asymmetric though, e.g. `duk_push_pointer()` accepting a `const volatile *` (which has the least issues casting to) and `duk_get_pointer()` returning a `void *` (which has the least issues casting from). But I'd find this asymmetry confusing.

fatcerberus:
That's true, I guess there's no perfect solution.

svaarala:
It's also more or less a standard idiom to use `void *userdata` (instead of having a const there) so I figured that'd be the most expected typing.

I actually ended up changing an internal helper's userdata pointer into a `const void *` to avoid some const-to-non-const casts. Some user code will need to deal with similar issues too. Luckily, as far as I know, const-to-non-const casting is pretty harmless as long as you don't actually use the pointer incorrectly. Volatile-to-non-volatile casting can actually make a difference in the generated code (assuming the pointer is dereferenced that is).

Issue #427, opened 2015-11-03
Nested property assignment evaluates incorrectly


svaarala:
(Originally issue filed in: https://github.com/svaarala/duktape-wiki/issues/59; see also http://stackoverflow.com/q/33489083/449161)

```
var NodeType;
(function (NodeType) {
    NodeType[NodeType["none"] = 0] = "none";
    NodeType[NodeType["text"] = 1] = "text";
})(NodeType || (NodeType = {}));
```

Duktape will resolve NodeType["text"] => 1 but not NodeType[1] => "text". The same code works fine in a browser such as Chrome.

svaarala:
This bug seems to exist in Duktape 1.3.0 but seems to work correctly in master.

svaarala:
Duktape 1.3.0 bytecode for the inner function:

```
; c0 "none"
; c1 {"type":"number","data":"0000000000000000","_value":0}
; c2 "text"
; c3 {"type":"number","data":"3ff0000000000000","_value":1}
00000 00000040   LDREG      r1, r0
00001 80c00056   PUTPROP    r1, c0, c1
00002 00000080   LDREG      r2, r0
00003 000040c0   LDREG      r3, r1
00004 8000c096   PUTPROP    r2, r3, c0
00005 00000040   LDREG      r1, r0
00006 81c08056   PUTPROP    r1, c2, c3
00007 00000080   LDREG      r2, r0
00008 0000c0c0   LDREG      r3, r3
00009 8100c096   PUTPROP    r2, r3, c2
00010 0000006f   RETURN     0x1, r0          ; have_retval
```

Bytecode in master:

```
; c0 "none"
; c1 {"type":"number","data":"0000000000000000","_value":0}
; c2 "text"
; c3 {"type":"number","data":"3ff0000000000000","_value":1}
00000 00000040   LDREG      r1, r0
00001 80c00056   PUTPROP    r1, c0, c1
00002 00000080   LDREG      r2, r0
00003 80404096   PUTPROP    r2, c1, c0
00004 00000040   LDREG      r1, r0
00005 81c08056   PUTPROP    r1, c2, c3
00006 00000080   LDREG      r2, r0
00007 8140c096   PUTPROP    r2, c3, c2
00008 0000002f   RETURN     0x0, r0
```

svaarala:
Just to ensure this is not working in master because the assignment value is mapped to a constant, also checked that there's no issue with dynamic results:

```js
// Slightly modified
var obj2;
[...]

PR #428, opened 2015-11-03
Add a testcase for already fixed nested property assignment bug

svaarala:
Nested property assignment bug #427 happens in Duktape 1.3.0 but is already fixed in master. This pull provides a test case for the bug, which was probably fixed in #381.


fatcerberus:
Now that you have a testcase, it might be interesting to do a reverse bisection to figure out exactly which commit fixed the bug.

svaarala:
It is 99% certainly the one for #381, because there was a bug there in assignment RHS handling.

PR #429, opened 2015-11-03
Fix property access handling of inner expression mutating a register-bound variable holding the base value

svaarala:
A property assignment LHS base value must be evaluated before the rest of the expression, and the expression (e.g. the key of the LHS) may mutate a variable containing the base value. The test case illustrates this best.

This is relatively uncommon in actual code but worth fixing. The fix is to coerce the base value to "temp or const" instead of just "plain" which also allows a register-bound variable (which can then get mutated).

Tasks:

- [x] Add testcase
- [x] Add initial fix
- [x] Assert testcase run
- [x] Perf test run
- [x] Releases entry (bug present in 1.3.0 too)

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1218,6 +1218,10 @@ Planned
 
 * Fix nested property assignment handling (GH-427, GH-428)
 
+* Fix property access expression handling when a variable holding the base
+  value is mutated by other parts of the expression, in both LHS and RHS
+  positions (GH-429)
+
 * Fix Unix local time offset handling which caused issues at least on RISC
   OS (GH-406, GH-407)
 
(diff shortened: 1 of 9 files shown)


fatcerberus:
Why is the LHS evaluated first? Assignment is right associative so I would think it would be evaluated right-to-left as well? (I know the two concepts are technically unrelated, I'm speaking in terms of practical implementation)

svaarala:
Ultimately the explicit order is specified in: http://www.ecma-international.org/ecma-262/5.1/#sec-11.2.1.

svaarala:
Detecting the `foo.bar = 123;` should be relatively easy. The seemingly same case `foo['bar']` is a bit more difficult because instead of just finding a string token the key is a bracket expression, and the compiler would need to know that whatever was inside there was safe. If ivalues carried a "side effect free" flag this would be easy: e.g. plain strings and compile-time concatenation of literal strings would carry the "side effect free" flag.

svaarala:
Actually this doesn't actually have a performance change over master for typical cases. While master uses an incorrect ivalue for `foo['bar']`, with the ivalue pointing directly to the register holding "foo", an explicit load is generated by assignment expression handling before proceeding to the RHS so the end result is the same opcode. There's only a difference for weird keys with potential side effects, so there doesn't seem to be any relevant performance impact here.

I'll run a perftest before merging.

Issue #423, opened 2015-11-01
Add a Detach notify to debugger protocol


fatcerberus:
It is currently impossible to differentiate between a deliberate target-initiated detach, as by closing the application window, versus an unexpected disconnection.  In both cases the client side will only see that the remote socket was closed, with no indication of why.

To solve this, I propose an additional Detach notify, which is sent just prior to detachment, either because the target called `duk_debugger_detach()` or the heap is destroyed.  If the socket is closed without receiving the notify, the client can then safely assume the connection was lost and try to reconnect.

svaarala:
Sounds good to me, it could include a string message to explain the detach too.

fatcerberus:
Unless there will be a hundred different possible detach scenarios, I think it might be simpler to just include an integer constant specifying the reason, like in a Status notify.

svaarala:
That's true. We can add a string message later if that becomes relevant, e.g. if Duktape tries to send a Detach notify on a parse error and could then try to explain the error briefly in the message ("invalid initial byte 0xNN").

fatcerberus:
Hm, you're right that we might want to describe error conditions, so they can be displayed by the client.  Here might be a good format:

`NFY <int: 6> <int: 0/1 iserror> [<str: msg>]`

If the first argument `iserror` is 0, it's a normal, expected detach so no message is needed or present.  Otherwise `iserror` is 1 and `msg` describes the error condition.

fatcerberus:
Just posted a pull for this, #430.

PR #430, opened 2015-11-04
Implement debugger Detach notify

fatcerberus:
This implements #423.

When debugger detach is initiated, a notify (`DUK_DBG_CMD_DETACHING = 0x06`) is sent out to let the client know the target is shutting down its end of the transport.  The notify includes an integer flag indicating the reason for the detach, 0 for an intentional detach (Detach command, heap destruction) or 1 for detach due to error (syntax error in debug message, read/write error, etc.).

### DIFF
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -23,12 +23,12 @@ typedef union {
  *  Detach handling
  */
 
-#define DUK__SET_CONN_BROKEN(thr) do { \
+#define DUK__SET_CONN_BROKEN(thr, is_err) do { \
 		/* For now shared handler is fine. */ \
-		duk__debug_do_detach1((thr)->heap); \
+		duk__debug_do_detach1((thr)->heap, (is_err)); \
 	} while (0)
 
-DUK_LOCAL void duk__debug_do_detach1(duk_heap *heap) {
+DUK_LOCAL void duk__debug_do_detach1(duk_heap *heap, duk_bool_t is_err) {
 	/* Can be called multiple times with no harm.  Mark the transport
 	 * bad (dbg_read_cb == NULL) and clear state except for the detached
 	 * callback and the udata field.  The detached callback is delayed
@@ -37,7 +37,25 @@ DUK_LOCAL void duk__debug_do_detach1(duk_heap *heap) {
 	 * inside the detached callback.
 	 */
 
+	if (heap->dbg_detaching) {
+		return;
+	}
+
 	DUK_D(DUK_DPRINT("debugger transport detaching, marking transport broken"));
+
+	heap->dbg_detaching = 1;
+
+	/* avoid sending multiple Detaching notifys */
+	if (heap->dbg_write_cb != NULL) {
+		duk_hthread *thr;
+
+		thr = heap->heap_thread;
+		DUK_ASSERT(thr != NULL);
+
+		duk_debug_write_notify(thr, DUK_DBG_CMD_DETACHING);
+		duk_debug_write_int(thr, is_err);
+		duk_debug_write_eom(thr);
+	}
 
 	heap->dbg_read_cb = NULL;
 	heap->dbg_write_cb = NULL;
@@ -85,10 +103,12 @@ DUK_LOCAL void duk__debug_do_detach2(duk_heap *heap) {
 		DUK_D(DUK_DPRINT("detached during message loop, delayed call to detached_cb"));
 		detached_cb(detached_udata);
 	}
+
+	heap->dbg_detaching = 0;
 }
 
 DUK_INTERNAL void duk_debug_do_detach(duk_heap *heap) {
-	duk__debug_do_detach1(heap);
+	duk__debug_do_detach1(heap, 0);
 	duk__debug_do_detach2(heap);
 }
 
@@ -230,8 +250,9 @@ DUK_INTERNAL void duk_debug_read_bytes(duk_hthread *thr, duk_uint8_t *data, duk_
 #endif
 		got = heap->dbg_read_cb(heap->dbg_udata, (char *) p, left);
 		if (got == 0 || got > left) {
+			heap->dbg_write_cb = NULL;  /* squelch further writes */
 			DUK_D(DUK_DPRINT("connection error during read, return zero data"));
-			DUK__SET_CONN_BROKEN(thr);
+			DUK__SET_CONN_BROKEN(thr, 1);
 			goto fail;
 		}
 		p += got;
@@ -243,33 +264,10 @@ DUK_INTERNAL void duk_debug_read_bytes(duk_hthread *thr, duk_uint8_t *data, duk_
 }
 
 DUK_INTERNAL duk_uint8_t duk_debug_read_byte(duk_hthread *thr) {
-	duk_heap *heap;
-	duk_size_t got;
 	duk_uint8_t x;
 
-	DUK_ASSERT(thr != NULL);
-	heap = thr->heap;
-	DUK_ASSERT(heap != NULL);
-
-	if (heap->dbg_read_cb == NULL) {
-		DUK_D(DUK_DPRINT("attempt to read 1 bytes in detached state, return zero data"));
-		return 0;
-	}
-
-	if (heap->dbg_have_next_byte) {
-		heap->dbg_have_next_byte = 0;
-		return heap->dbg_next_byte;
-	}
-
 	x = 0;  /* just in case callback is broken and won't write 'x' */
-	DUK_ASSERT(heap->dbg_read_cb != NULL);
-	got = heap->dbg_read_cb(heap->dbg_udata, (char *) (&x), 1);
-	if (got != 1) {
-		DUK_D(DUK_DPRINT("connection error during read, return zero data"));
-		DUK__SET_CONN_BROKEN(thr);
-		return 0;
-	}
-
+	duk_debug_read_bytes(thr, &x, 1);
 	return x;
 }
 
@@ -316,7 +314,7 @@ DUK_INTERNAL duk_int32_t duk_debug_read_int(duk_hthread *thr) {
 	}
 
 	DUK_D(DUK_DPRINT("debug connection error: failed to decode int"));
-	DUK__SET_CONN_BROKEN(thr);
+	DUK__SET_CONN_BROKEN(thr, 1);
 	return 0;
 }
 
@@ -361,7 +359,7 @@ DUK_INTERNAL duk_hstring *duk_debug_read_hstring(duk_hthread *thr) {
 
  fail:
 	DUK_D(DUK_DPRINT("debug connection error: failed to decode int"));
-	DUK__SET_CONN_BROKEN(thr);
+	DUK__SET_CONN_BROKEN(thr, 1);
 	duk_push_hstring_stridx(thr, DUK_STRIDX_EMPTY_STRING);  /* always push some string */
 	return duk_require_hstring(ctx, -1);
 }
@@ -395,7 +393,7 @@ DUK_LOCAL void *duk__debug_read_pointer_raw(duk_hthread *thr) {
 
  fail:
 	DUK_D(DUK_DPRINT("debug connection error: failed to decode pointer"));
-	DUK__SET_CONN_BROKEN(thr);
+	DUK__SET_CONN_BROKEN(thr, 1);
 	return (void *) NULL;
 }
 
@@ -503,7 +501,7 @@ DUK_INTERNAL void duk_debug_read_tval(duk_hthread *thr) {
 
  fail:
 	DUK_D(DUK_DPRINT("debug connection error: failed to decode tval"));
-	DUK__SET_CONN_BROKEN(thr);
+	DUK__SET_CONN_BROKEN(thr, 1);
 }
 
 /*
@@ -547,8 +545,9 @@ DUK_INTERNAL void duk_debug_write_bytes(duk_hthread *thr, const duk_uint8_t *dat
 #endif
 		got = heap->dbg_write_cb(heap->dbg_udata, (const char *) p, left);
 		if (got == 0 || got > left) {
+			heap->dbg_write_cb = NULL;  /* squelch further writes */
 			DUK_D(DUK_DPRINT("connection error during write"));
-			DUK__SET_CONN_BROKEN(thr);
+			DUK__SET_CONN_BROKEN(thr, 1);
 			return;
 		}
 		p += got;
@@ -572,7 +571,7 @@ DUK_INTERNAL void duk_debug_write_byte(duk_hthread *thr, duk_uint8_t x) {
 	got = heap->dbg_write_cb(heap->dbg_udata, (const char *) (&x), 1);
 	if (got != 1) {
 		DUK_D(DUK_DPRINT("connection error during write"));
-		DUK__SET_CONN_BROKEN(thr);
+		DUK__SET_CONN_BROKEN(thr, 1);
 	}
 }
 
@@ -1024,7 +1023,7 @@ DUK_LOCAL duk_bool_t duk__debug_skip_dvalue(duk_hthread *thr) {
 	return 0;
 
  fail:
-	DUK__SET_CONN_BROKEN(thr);
+	DUK__SET_CONN_BROKEN(thr, 1);
 	return 1;  /* Pretend like we got EOM */
 }
 
@@ -1447,7 +1446,7 @@ DUK_LOCAL void duk__debug_handle_detach(duk_hthread *thr, duk_heap *heap) {
 	duk_debug_write_eom(thr);
 
 	DUK_D(DUK_DPRINT("debug connection detached, mark broken"));
-	DUK__SET_CONN_BROKEN(thr);
+	DUK__SET_CONN_BROKEN(thr, 0);  /* not an error */
 }
 
 #if defined(DUK_USE_DEBUGGER_DUMPHEAP)
@@ -1795,7 +1794,7 @@ DUK_LOCAL void duk__debug_process_message(duk_hthread *thr) {
  fail:
 	DUK_ASSERT(duk_get_top(ctx) >= entry_top);
 	duk_set_top(ctx, entry_top);
-	DUK__SET_CONN_BROKEN(thr);
+	DUK__SET_CONN_BROKEN(thr, 1);
 	return;
 }
 
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -1325,6 +1325,31 @@ taken directly from the thrown object if it is an Error instance (after
 augmentation), otherwise these values are calculated from the bytecode
 executor state.
 
+Detaching notification (0x06)
+-----------------------------
+
+Format::
+
+    NFY <int: 6> <int: reason> [<str: msg>] EOM
+
+Example::
+
+    NFY 6 1 "error parsing dvalue" EOM
+
+Reason is one of:
+
+* 0x00: normal detach
+
+* 0x01: detaching due to stream error
+
+Duktape sends a Detaching notification when the debugger is detaching. If the
+target drops the transport without the client seeing this notification, it can
+assume the connection was lost and react accordingly (for example by trying to
+reestablish the link).
+
+``msg`` is an optional string elaborating on the reason for the detach. It may
+or may not be present depending on the nature of detachment.
+
 Commands sent by debug client
 =============================
 
--- a/src/duk_debugger.h
+++ b/src/duk_debugger.h
@@ -20,6 +20,7 @@
 #define DUK_DBG_CMD_ALERT         0x03
 #define DUK_DBG_CMD_LOG           0x04
 #define DUK_DBG_CMD_THROW         0x05
+#define DUK_DBG_CMD_DETACHING     0x06
 
 /* Initiated by debug client */
 #define DUK_DBG_CMD_BASICINFO     0x10
--- a/src/duk_heap.h
+++ b/src/duk_heap.h
@@ -434,6 +434,7 @@ struct duk_heap {
 	duk_bool_t dbg_paused;                  /* currently paused: talk with debug client until step/resume */
 	duk_bool_t dbg_state_dirty;             /* resend state next time executor is about to run */
 	duk_bool_t dbg_force_restart;           /* force executor restart to recheck breakpoints; used to handle function returns (see GH-303) */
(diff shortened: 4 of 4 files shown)


fatcerberus:
REALLY not merge-ready. :)

I overlooked that `duk__debug_do_detach1()` takes a heap pointer instead of a thread pointer, meaning this currently won't even compile.  Oops!

svaarala:
It might make sense for the integer to be an enumerated "cause" and the message being an optional explanation? For example:

- 0: Transport error
- 1: Detach requested by API call
- 2: Detach from heap destruction

I guess what you'd like to see is the ability to detect a recoverable error, which would prompt an automatic reconnect? If so, is an error/not-error the correct thing for that, i.e. are there errors which don't warrant a reconnect?

svaarala:
With that debug client action in mind, what are the possible actions for a detach? I can think of:

- Indicate successful detach in a debug client UI
- Indicate an error in the UI without a prompt
- Indicate an error with a prompt, perhaps option to reconnect
- Indicate transport lost, automatic reconnect

fatcerberus:
This is mostly for the benefit of the client; i.e., if the remote socket shuts down without a Detach notify being received, the client can assume the connection was lost and try to reconnect.  If any Detach notify is received at all (whether error or not), then the detach was intentional and reconnecting might not work.

Of course that's still not pretty vague and not necessarily actionable--hence why I added the "error/not error" distinction.  If the notify indicates "no error", then the detach was definitely intentional, and you should not reconnect.  If it indicates an error, that's usually a syntax error or something else (e.g. due to the client being buggy) and reconnecting might be in order.

fatcerberus:
Concretely, here are the scenarios I can think of:

* Target's heap is destroyed (intentional, is_err = 0, don't reconnect)
* Client sends Detach command (intentional, is_err = 0, don't reconnect)
* Syntax error in debug stream (unintentional, is_err = 1, maybe reconnect?)
* No Detach notify received at all (connection lost, attempt reconnection)

svaarala:
Based on these it would seem the only really relevant information is whether the detach was intentional or not (= error). For errors the client would indicate an error and/or automatically reconnect. For intentional detaches it wouldn't.

I wonder if there's a need for the client to know the reason of an intentional detach? There's some difference in being detached because: (1) the client requested it using the Detach command, (2) user code requested it using the API, and (3) the program finished and the heap was destroyed.

By the way, because there's a command named `Detach` maybe this notify should be `Detaching`?

fatcerberus:
My solution to the name clash was to rename all notifys to DUK_DBG_NFY_XXX.  I was originally going to call it Detached but this seemed more consistent with the naming of the other notifys (i.e. present tense)

I'm not averse to an additional description by the way, I didn't implement it in the prototype because I wasnt sure of the cleanest way to do it.  Should we pass the string on the valstack, as a direct parameter, etc.

svaarala:
Well we could just leave a space for the message in the protocol document. It doesn't have to be present in all cases anyway, but maybe it could be added to e.g. the "syntax error" detach case because there's minimal plumbing in that case and it'd be quite useful when implementing a debug client.

As for the naming, I know what you mean and I had the same thoughts on that. However, "Print" notify is a bit different from "Detaching" because the Print notify is describing the actual event while Detaching is informing the client of a future event - so maybe the naming would be more logical that way. But it's not a big deal, I mainly just thought it'd be confusing to have the exact same name for a command and a notify.

fatcerberus:
I realized this might cause infinite recursion in the case that the error is "write_cb returned zero", so there should probably be a special case for that.

svaarala:
I think the easiest approach would be to flag "detach notify sent" and never send it twice per connection? In any case the *first* detaching reason is probably the most valid one, additional ones could be caused by unclean tear down etc.

svaarala:
This will also make it easier to implement because let's say the detach is initiated by user code calling the API - if we write the detached notify first, there's no need to track it down the line in other parts. Any other reasons for sending a detached notify will automatically be ignored and won't confuse the client.

fatcerberus:
Yeah, I just meant that if Duktape detaches due to the write callback not writing anything, it will attempt to send the notify, which will again write zero bytes, causing a another notify, etc. and smash the stack.

svaarala:
I see, but my point was that when you start writing the first notify, you flag it as "detach sent". When the recursion point happens, one would notice that a notify was already sent and one wouldn't call write for the second time. But maybe I don't understand the issue.

fatcerberus:
Thats true, although the write callback would still be called a second time after returning zero, which the target may not expect.

svaarala:
Ah, that's true. That'd have to be handled specially so that when a zero is received from the write callback, that'd be flagged somehow so that we'd never end up calling the write callback again. It'd be OK to then try to send a notify - as long as the write callback was not actually called.

svaarala:
I don't know how important that "don't call write callback after zero return" guarantee really is - but it seems like the cleanest guarantee Duktape can give.

fatcerberus:
I'm at work now but I'll revisit this tonight and make an actually working implementation. :)

svaarala:
Yeah, I'm sure this is solvable using pretty small tracking state to guarantee (1) at most one Detach notify is sent, and (2) write callback is never called once transport is severed. But it's nice to find a robust solution and not pepper checks here and there :)

fatcerberus:
Okay, just pushed a working branch.  No extra state variables are necessary, it simply sets `heap->dbg_write_cb` to `NULL` to prevent anything else from being written to the errored transport.

I left the 0x00 reason code as the normative case (no error) because this matches the semantics of other notifys where nonzero is indication to the client that "the ball is in your court" so to speak (e.g. Status 1 means paused, Throw 1 means fatal, etc.) and 0 means "no further action needed".

Unfortunately, I'm unable to test this properly myself because my sockets implementation (https://github.com/rxi/dyad) forcibly closes all sockets on shutdown, preventing the Detach notify from going through.  My debugger never sees it.  :/  I *was* able to verify with a breakpoint that the notify is sent out by Duktape, at least.

svaarala:
Sounds good, I'll give it a test later today.

fatcerberus:
By the way do you still want to rename the notify to Detaching?  I'm okay with that if so.

svaarala:
I think Detaching would be better - I used a virtual `_Disconnecting` in the JSON proxy for when the target connection lost (virtual because the notify was generated by the proxy) for the same reason.

svaarala:
Actually, to be really pedantic, this notify is actually about the transport being disconnected. For example: syntax error in the incoming debug stream causes the transport to be dropped. Right now that also results in a debugger detach, but that didn't need to be the case (Duktape could pause waiting for a reconnect, for example).

Right now Detaching/Disconnecting are pretty much equivalent, and detaching is the best default if it's not configurable. With an immediate re-attach supported in the detached callback, user code can already avoid the detach causing execution to auto-resume.

fatcerberus:
Alright, I'll change that during pre-merge cleanup, have you had a chance to test this yet?

svaarala:
Not yet, I was away for the weekend. I'll try it out later tonight or tomorrow.

svaarala:
After a quick test, seems to work but the following triggered during Eval:

```
PANIC 54: assertion failed: (duk_idx_t) duk_get_top((ctx)) == (duk_idx_t) (1) (duk_bi_global.c:434) (segfaulting on purpose)
Segmentation fault
```

Caused by:

```
DUK_ASSERT_TOP(ctx, 1);
```

That assert is no longer true with debugger support.

svaarala:
The best assert would probably be (since we can't easily check the caller):

```c
DUK_ASSERT(duk_get_top(ctx) == 1 || (duk_get_top(ctx) == 2 && duk_is_number(ctx, 1)));
```

fatcerberus:
Huh, somehow I missed that assert when I was reviewing the eval() implementation, and I must have disabled them because it never triggered during my tests...

Another check that might work is to assert that `top == 1` but only if the callee is the built-in `eval`.  Since we push a new function object for a debug Eval (pointing at the same C function), such a check should be possible, no?

svaarala:
You'd need to find the built-in eval() Ecmascript function reference by reading it from the global object (and it might no longer be there if it was reassigned to some other variable and removed).

fatcerberus:
You couldn't read it from `thr->builtins` like I did for the doublefault check in throw notify?

svaarala:
Individual functions are not there, it's just the top level built-in objects.

fatcerberus:
Ah, I guess then, yeah, your assert above is the closest we can get.

svaarala:
Hmm, playing around I got this loop:

```
[D] <42583889302069> duk_debugger.c:40 (duk__debug_do_detach1): debugger transport detaching, marking transport broken
[D] <42583889314896> duk_debugger.c:585 (duk_debug_write_byte): connection error during write
[D] <42583889328941> duk_debugger.c:40 (duk__debug_do_detach1): debugger transport detaching, marking transport broken
[D] <42583889341351> duk_debugger.c:585 (duk_debug_write_byte): connection error during write
...
```

svaarala:
JSON proxy explicit Detach request looks like this:

```
$ telnet localhost 9093
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
{"notify":"_Connected","args":["1 10399 v1.2.0-806-g2662557 duk command built from Duktape repo"]}
{"notify":"Status","command":1,"args":[1,"test-dev-mandel2-func.js","global",58,0]}
{"request":"Detach"}
{"reply":true,"args":[]}
{"notify":"Detaching","command":6,"args":[0]}
{"notify":"_Error","args":["Error: read ECONNRESET"]}
{"notify":"_Disconnecting"}
```

svaarala:
#434 adds minimal web UI support for Detaching.

fatcerberus:
Yeah, that loop is what I was afraid of earlier.  How did you trigger it?

svaarala:
I think it was through an intentional protocol syntax error - I can reproduce it if needed.

svaarala:
Some log of the situation:

```
[D] <42546314562399> duk_heap_alloc.c:899 (duk_heap_alloc): allocated heap: 0x2453010
Debugger enabled, create socket and wait for connection
Waiting for debug connection on port 9091
Debug connection established
Debugger connected, call duk_debugger_attach() and then execute requested file(s)/eval
[D] <42547856437950> duk_debugger.c:1915 (duk_debug_process_messages): paused, process debug message, blocking if necessary
dist/examples/debug-trans-socket/duk_trans_socket.c: debug read failed, ret == 0 (EOF), closing connection
[D] <42564654761233> duk_debugger.c:278 (duk_debug_read_byte): connection error during read, return zero data
[D] <42564654786045> duk_debugger.c:40 (duk__debug_do_detach1): debugger transport detaching, marking transport broken
[D] <42564654801381> duk_debugger.c:585 (duk_debug_write_byte): connection error during write
[D] <42564654814214> duk_debugger.c:40 (duk__debug_do_detach1): debugger transport detaching, marking transport broken
```

To reproduce, I did the following:

- Start "duk --debugger"
- Attach with web UI debugger
- Once attached, CTRL-C the duk_debug.js debug client, which severes the transport from the debug client end

I think you should be able to reproduce this by abruptly dropping the debug connection from your debug client, so that a Duktape attempt to read from the debug transport returns an error.

svaarala:
The sequence of events is probably something like this:

- A read helper notices a read error, and calls DUK__SET_CONN_BROKEN which maps to duk__debug_do_detach1().
- duk__debug_do_detach1() notices write_cb != NULL and tries to write out an error. Both read_cb and write_cb are still intact because we haven't marked them bad yet.
- Each write call in the write_cb != NULL branch will encounter a transport error and call DUK__SET_CONN_BROKEN.
- On entry to duk__debug_do_detach1() the read_cb and write_cb callbacks are still intact so we'll keep on looping.

svaarala:
What would suffice is the flag "detaching notify sent" I described earlier:

- If write_cb != NULL && !detaching_nfy_sent -> set detaching_nfy_sent=1; try to send detaching notify

This would cause the write helpers still to call DUK__SET_CONN_BROKEN() for the second time though which is not ideal.

So perhaps the best flag would be `heap->dbg_detach_in_progress`. In duk__debug_do_detach1():

- If `dbg_detach_in_progress` -> don't do recursive detach, return silently
- Set `dbg_detach_in_progress = 1` and proceed as currently

This would cause the write helpers to call into duk__debug_do_detach1() but return directly without any state changes.

svaarala:
That `dbg_detach_in_progress` flag would prevent multiple Detaching messages in the current implementation because there's only one site sending a Detaching notify.

But if we wanted other parts of the code to be able to send a Detaching notify (*before* calling DUK__SET_CONN_BROKEN()) another flag would be needed for the detaching notify. Having the ability to send the notify before shutting down the transport would be nice because it'd avoid the need to plumb e.g. the detaching message to the duk__debug_do_detach1() helper.  One could simply:

```
thr->heap->dbg_detaching_sent = 1;
duk_debug_write_notify(thr, DUK_DBG_NFY_DETACH);
duk_debug_write_int(thr, 0 /*is_err*/);
duk_debug_write_cstring(thr, "detach requested from API"); 
duk_debug_write_eom(thr);
...
```

fatcerberus:
Regarding the eval() issue: That's a regression related to #412.  Should it be fixed here or in a separate branch/pull?

fatcerberus:
Okay, this is weird.  The loop described above should NOT be possible in the current branch:
https://github.com/fatcerberus/duktape/commit/63fee270681c92ba7c0973d685bffb2481debb5b#diff-df880d4c6b6e3b1ee1453743eba813ebR242

For a read or write error, `dbg_write_cb` is set to `NULL` *before* the connection is marked broken.  This should cause `duk_debug_do_detach1()` to skip writing the Detaching notify, but it's clearly trying to do it anyway.

I'm not averse to that extra flag (it would make the intent of the code clearer, in fact), but before going ahead with that I want to ascertain how this is happening right now because it may indicate a sleeper bug elsewhere.

svaarala:
That write_cb NULLing is missing from `duk_debug_read_byte()` so it would have to be duplicated there.

Then again, maybe the better fix would to make `duk_debug_read_byte()` just call `duk_debug_read_bytes()`. It's not performance critical and having it call read_cb on its own duplicates the pushback and detach logic for minimal gain.

svaarala:
Btw, just to be clear, I was suggesting two flags :-)

- Detach in progress: detach has started, don't recurse into `duk__debug_do_detach1()` at all.
- Detaching notify written (or write in progress): don't try to send a second detaching notify.

svaarala:
Also note that the current code will enter that infinite loop whenever we're trying to write a detaching notify and a transport error occurs during the write operations - which is entirely possible and up to the transport. This is the case because write_cb is only NULLed once the detaching notify has been sent, so that any recursion into `duk__debug_do_detach1()` will repeat the same sequence of trying to send a detaching notify etc.

fatcerberus:
Barring the bug, the first transport error will immediately NULL the write callback (before calling a detach helper), so fixing read_byte() should suffice here.

I'll go the extra mile with the flag(s) anyway though because, as I said, it will make the code clearer.  Mostly I was just trying to avoid introducing footprint creep in the debugger.  We did just add that pushback byte after all. :)

fatcerberus:
Hm, maybe we *should* rename the notify "Disconnect"?  Since notify messages appear to actually be just be commands sent in the opposite direction (a Print notify, for example, is a request by Duktape to print text in the UI), this would be a request to disconnect the transport (since it can no longer be reused).

fatcerberus:
Okay, this is a better implementation that avoids calling into the detach handler recursively.

svaarala:
Re: the write callback nulling, note that a transport error may occur on any transport write callback too. This will now cause an error loop, regardless of that write_cb nulling in read helper calls.

svaarala:
Re: naming, I think Disconnecting would still be more appropriate because it's not a *request* to disconnect, but more letting the other end know the connection is going down no matter what.

fatcerberus:
The write helper nulls it on the first error too, or at least it should unless I screwed up

svaarala:
Yeah, I was referring to the snapshot I was testing earlier. In the detach handler that snapshot would first try to send a notify if write_cb != NULL, and once all the notify-related writes were complete it would set write_cb to NULL. If a transport error occurred during those notify writes, they'd recurse into the disconnect handler and go through the same path over and over again. But this was before - I'll check the current branch.

src/duk_debugger.c
```diff
@@ -37,8 +37,23 @@ DUK_LOCAL void duk__debug_do_detach1(duk_heap *heap) {
...
+	}
+
 	DUK_D(DUK_DPRINT("debugger transport detaching, marking transport broken"));
 
+	heap->dbg_detaching = 1;
+
+	/* avoid sending multiple Detach notifys */
+	if (heap->dbg_write_cb != NULL) {
+		duk_hthread *thr = heap->heap_thread;
+
```
svaarala: Humm, this was one thing blocking this earlier. I guess using heap_thread would be OK here, the thread doesn't have much bearing on the writing process. But maybe assert thr != NULL here as we're certainly requiring that.

src/duk_debugger.c
```diff
@@ -64,6 +79,8 @@ DUK_LOCAL void duk__debug_do_detach1(duk_heap *heap) {
 	 * XXX: clear breakpoint on either attach or detach?
 	 */
 	heap->dbg_breakpoints_active[0] = (duk_breakpoint *) NULL;
+
+	heap->dbg_detaching = 0;
```
svaarala: One perhaps more robust alternative for this flag would be "detach in progress or completed" (not sure how to name it shortly). The difference would be that we'd leave the flag to 1 until the next attach, so that any code path ending up trying to detach again (perhaps because of a bug) wouldn't unnecessarily go through the motions.

svaarala:
Right, the current branch seems to be cleanly protected against recursion. Though I'm wondering if we should only set `dbg->detaching = 0` in an attach and leave it 1 to prevent recursion even after the detach handler exits (and we remain detached).

src/duk_debugger.c
fatcerberus: Oh, okay.  I assumed `heap_thread` would always be non-NULL by definition so didn't add the assert.

svaarala: It might be zero during heap destruction - of course we shouldn't be attached at that point but since we do detach in heap destruction too, we're basically assuming that the destruction sequence is well ordered in this respect. The assert would communicate that intent :)

fatcerberus:
Alternately, we could rename the flag - the behavior described above would require setting to 1 during initialization, whereas `attached = 0` is cleanly handled by the memzero on heap init.

svaarala:
Hmm? I meant that the flag would be "detach in progress or completed" and zero at attach. It would then be set to 1 when the *first* detach call starts, and remain 1 until the next attach. But perhaps we're talking cross purposes :)

fatcerberus:
Well the idea is that you don't do detach processing while already detached.  Since we start up detached, the flag should logically be set == 1 on startup.  That's what I meant.

svaarala:
I see what you mean - I didn't mean that to be done, didn't think about that case being really possible because we should never venture into the debugger code in that initial state. We also shouldn't venture into the debugger code when we're detached, but the difference might be that we're already there, recovering from a detach and thus more likely to call repeatedly into the detach handling.

But it's probably not worth the change: the detach handling should already be a no-op if we call into it again after a *completed* previous detach.

fatcerberus:
Actually though, should detaching = 0 instead be done in `duk__debug_do_detach2()` since it's not technically complete until the second stage?

fatcerberus:
I'm away from my computer right now but I'll make any necessary changes when I get home, so if you see anything else feel free to point it out. :)

svaarala:
Setting detaching=0 in `duk__debug_do_detach2()` would actually reduce the chances of a bug re-entering the detach handling a bit, so that might be a good idea. But on the whole we're talking about a very minor thing because the behavior would be only triggered by a bug, and even then it would most likely be a no-op :)

fatcerberus:
Alright, I fixed the issues you mentioned, check it out now.  I kept the name of the notify as "Detaching" because that matches what's implemented in the web UI right now (#434) and I don't want to leave master inconsistent, so I guess you can change it later if you want. :)

I did what you suggested and refactored `duk_debug_read_byte()` to just call read_bytes() with a buffer size of 1.  As you say it's not really a hot path, more time will likely be spent blocking on reads than the read calls themselves. :)

I was going to add a helper to write Detaching notifys from outside like we already have for Status and Throw, but changed my mind.  It's bad form to allow a Detach notify to be easily sent from outside the debugger I think, you could easily end up with a bug where some code sends out a Detach notify without actually detaching, confusing the client.  This way seems cleaner to me.

fatcerberus:
Oh, I meant to add: This should be merge-ready.

svaarala:
Ok, good to know - I'll try to get into this soon. This will probably have to be the last debugger change for 1.4.0, there's still a lot of backlog for the next ~3 weeks and I'll need to do some pruning anyway :-)

But there's of course room for debugger stuff in 1.5.0 :) I'm still hoping to prototype the heap walk stuff in a branch and try to get that into master pretty soon after 1.4.0 release. That'd be a nice debugger feature in 1.5.0 I think.

svaarala:
Oh, and maybe I wasn't very explicit, thanks for all the effort :)

fatcerberus:
No problem, I have a lot of fun implementing debugger stuff, it's probably my favorite Duktape feature!

svaarala:
The diff looks good - I'll open a separate pull to improve the web UI handling and perhaps renaming Detaching -> Disconnecting. I'll merge this today once I've tested it manually a bit more. Thanks!

svaarala:
One thing we might handle better in the future is that if there's a syntax error in parsing the stream, it'd be nice to send a Detaching notify with an explanation. There's no guarantee that the debug client will be able to parse it (a syntax error is a pretty bad situation and some causes may imply the debug client or stream state is broken in whatever way), but it'd probably go through in most cases.

However, the current transport guarantees are that if the transport indicates an error (e.g. read returns zero) no further calls are made into the transport. This guarantee prevents trying to send this notify:

- https://github.com/svaarala/duktape/blob/master/doc/debugger.rst#read-callback-semantics

We can remove the guarantee but it'd need at least a release note so that users know to check their transport for that change. Anyway, this is best handled as future work.

fatcerberus:
Ah, so the "don't call transport again in case of error" guarantee was already in place then, you had me under the impression there was no special treatment specified for that earlier in the discussion. :)

Anyway, good to see this is merged, now I just have to figure out how to get my client to see the notify.  Thanks!

PR #431, opened 2015-11-04
Fix "duk" cmdline Windows snprintf() issue for older MSVC versions

svaarala:
Older MSVC versions don't have `snprintf()` but do have `_snprintf()` which has different NUL termination semantics. Because the NUL terminator differences don't affect `duk_cmdline.c` use `_snprintf()` with MSVC.

See: http://stackoverflow.com/questions/2915672/snprintf-and-visual-studio-2010

### DIFF
--- a/examples/cmdline/duk_cmdline.c
+++ b/examples/cmdline/duk_cmdline.c
@@ -6,7 +6,7 @@
  *  support for example allocators, grep for DUK_CMDLINE_*.
  */
 
-#ifndef DUK_CMDLINE_FANCY
+#if !defined(DUK_CMDLINE_FANCY)
 #define NO_READLINE
 #define NO_RLIMIT
 #define NO_SIGNAL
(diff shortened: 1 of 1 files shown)


fatcerberus:
Yeah, MSVC 2015 is the first version with an actual `snprintf ()`.  No idea what took them so long... :)

svaarala:
Which reminds me, Duktape should probably not use the `_snprintf()` replacement when the real thing is available.

fatcerberus:
By the way if you ever do need NUL termination I believe `sprintf_s()` has the required semantics.

svaarala:
Currently Duktape relies on the `snprintf()` not necessarily doing NUL termination and uses the idiom:

```
DUK_SNPRINTF(buf, len, ...);
buf[len - 1] = 0;
```

It would be good to change all of these to use some other macro which guaranteed NUL termination and then implement the macro using `sprintf_s()` on Windows.

PR #433, opened 2015-11-08
Add duk_is_xxx_error() convenience calls

### LINKED ISSUE #340: Add duk_is_syntax_error() etc convenience macros
There's a convenience macro `duk_is_error()` which maps to `duk_get_error_code() != 0`. But to check for a specific error code one has to:

```c
if (duk_get_error_code(ctx, -3) == DUK_ERR_SYNTAX_ERROR) {
    /* ... */
}
```

Would be nicer to:

```c
if (duk_is_syntax_error(ctx, -3)) {
    /* ... */
}
```

svaarala:
Currently to detect e.g. a URI error:

```c
if (duk_get_error_code(ctx, -3) == DUK_ERR_URI_ERROR) {
    /* ... */
}
```

Using the convenience calls:

```c
if (duk_is_uri_error(ctx, -3)) {
    /* ... */
}
```

Fixes #340.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1158,6 +1158,9 @@ Planned
 * Add defines DUK_DEFPROP_{SET,CLEAR}_{WRITABLE,ENUMERABLE,CONFIGURABLE} for
   duk_def_prop() to improve call site readability (GH-421)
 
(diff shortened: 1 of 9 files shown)


fatcerberus:
I like that they are implemented as macros so as not to affect footprint. :)

PR #437, opened 2015-11-13
Minor debugger (and other) cleanups

svaarala:
Summary of changes:
- Debugger cleanups for callstack level changes, add ability to test callstack level using JSON proxy
- Debugger cleanups for Detaching notify
- String hash declaration warning fix (declared but not defined when dense string hash not in used)
- -Wcast-qual fix for duk_debug_fixedbuffer.c
- Releases entry for Detaching notify

Tasks:
- [x] Review changes
- [x] Decide whether to implement callstack level support into debugger web UI in this pull => separate pull
- [x] Rename is_err to reason in detach macro, change typing

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1152,6 +1152,9 @@ Planned
 * Add a debugger Throw notify for errors about to be thrown, and an option
   to automatically pause before an uncaught error is thrown (GH-286, GH-347)
 
+* Add a debugger Detaching notify which informs a debug client of an orderly
+  (or non-orderly) pending detach (GH-423, GH-430, GH-434)
+
 * Allow debugger detached callback to call duk_debugger_attach(), previously
   this clobbered some internal state (GH-399)
 
--- a/debugger/duk_debug.js
+++ b/debugger/duk_debug.js
@@ -1206,16 +1206,16 @@ Debugger.prototype.sendBasicInfoRequest = function () {
     });
 };
 
-Debugger.prototype.sendGetVarRequest = function (varname) {
+Debugger.prototype.sendGetVarRequest = function (varname, level) {
     var _this = this;
-    return this.sendRequest([ DVAL_REQ, CMD_GETVAR, varname, DVAL_EOM ]).then(function (msg) {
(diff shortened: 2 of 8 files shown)


fatcerberus:
For the Detaching message, would it be better to have a string description as specified in the doc right now, or just have multiple reason codes?  A reason code would be easier to parse for the client, and could easily be converted to a localized error string by the client.

svaarala:
Do you mean adding a second reason code in addition to the existing one (and maybe renaming the existing "reason" field to "isError"?). Or adding more reason codes to the existing field?

With respect to the message, my hunch would be to use a string message for details, similarly to how thrown errors work in Ecmascript. There are quite many error sources so that updating an enumeration might be tedious (and often in such situations people, me included, tend to map slightly different errors to existing enumeration values to avoid adding a new code ;-).

Further, I can see the debugger code going inside a setjmp catch point later on, and at that point the easiest approach would be to handle an error by detaching and converting the error thrown (an internal error) into a string for the detach reason. Converting a thrown error to an enumeration isn't currently possible because errors don't carry a numeric reason or other programmatic information to distinguish errors reliably.

svaarala:
All that said, what level of detail were you thinking of for a detach reason?

fatcerberus:
I just meant to expand the existing "is_error" field to an enumeration - 0 for orderly detach, nonzero for error cases.  Code that just needs to know whether it's an error doesn't have to change this way - just compare against zero.

I guess we could leave the string message in with this setup as well.

svaarala:
Hmm, but it's already specified as a "reason" field right now?

```
Format::

    NFY <int: 6> <int: reason> [<str: msg>] EOM

Example::

    NFY 6 1 "error parsing dvalue" EOM

Reason is one of:

* 0x00: normal detach

* 0x01: detaching due to stream error
```

I thought it was intended to be open ended already :)

svaarala:
The code uses "is_err" though, so maybe rename that internally to "reason" too to make it clearer it maps directly to this detach reason field.

fatcerberus:
Haha, I guess I already had that in mind when I wrote the docs then.  So it's just some code trivia then, that's easy to fix.

svaarala:
Yeah I think we changed the field at least once. I'll change the macro argument naming to match.

fatcerberus:
While you're at it you also might want to change the reason code type to `duk_int_t` (it's currently a bool).

PR #438, opened 2015-11-16
Further improvements for error messages of non-callable targets

svaarala:
First improvement was in #417, but it'd be nice to emit a better error message for e.g. `obj.noSuchProperty()` so that the error message would include the property name. Some of the difficulties are explained here: https://github.com/svaarala/duktape/pull/417#issuecomment-157075027

This is a work in progress pull with no merge ready solution yet.

Tasks:
- [x] Decide on basic approach, e.g. CALLPROP
- [x] Move opcodes for CALL<N> and CALLPROP<N> next to each other, simplifies compiler
- [x] Debugger opcode metadata update
- [x] Debug print opcode update
- [x] Low memory variant, avoid entire opcode?
- [ ] Manual perf test
- [ ] Test debugger opcode dump manually
- [ ] Test executor %!I debug dump manually
- [ ] Testcase fixes
- [ ] Releases entry
- [ ] Future work: include base value identifier if available
- [ ] Future work: property key for `new foo.noSuch()`, not shown now
- [ ] Future work: duk_(p)call_prop() error message, not handled like Ecmascript `obj.noSuch()` yet


fatcerberus:
You mentioned in #417 that one potential solution is to throw the error directly from the CSPROP.  But you also said ES5.1 requires arguments to be evaluated first (which strikes me as backwards but what can you do), throwing a noncallable error in CSPROP would prevent argument evaluation and cause non-compliant behavior.

svaarala:
What was the question? :)

fatcerberus:
The question was how is that a feasible solution if it breaks spec?  You mentioned the "throw in CSPROP" solution *after* explaining the standard behavior, so it seemed like you viewed it as workable.

svaarala:
I guess it'd be possible to split the CSPROP handling into two: the compiler would first extract the property value but keep the key intact, and after evaluating arguments there'd be an explicit "call check" with the known key which could then throw a nicer error. Or instead of a CALL, there'd be a CALLPROP variant which would convey the key to the call handling.

Nothing is really clean though, and I'm not sure if these changes are worth the improved error message.

fatcerberus:
The CALLPROP method seems pretty clean to me.  Since spec mandates arguments be evaluated first, merging CSPROP and CALL into one opcode actually seems to make more sense.  Although... Does ES5 say anything about evaluating the CallExpression before the arguments?  If so that complicates matters.

Ultimately I think I agree that this isn't a terribly pressing concern, but it's worth discussing at least. :)

svaarala:
I haven't thought about it too much but one'll probably need both CSPROP and CALLPROP, or CSPROP and some sort of CHECKCALL+CALL sequence.

Yeah, it's definitely worth figuring out. It'd be nice to get that key into the error message, because property based calls are very common.

svaarala:
Still intend to look into this for 2.0 but I want to rework the call opcodes a bit first, #896.

svaarala:
Current draft is using a separate 16 opcode block for CALLPROP<N> where:
- Base register + 0: key
- Base register + 1: base object

The executor does the base[key] lookup and an initial "is callable" check. If the value is not callable, it can throw an error message which includes the key. This is not a 100% check because it might be that call handling throws a TypeError for some reason but it catches the common causes. Further, bound functions and Proxies now have a CALLABLE flag which is copied from their target so ordinarily the direct CALLABLE flag should match whatever is resolved down the line.

The error message is a placeholder and looks like:

```
((o) Duktape [linenoise] 2.1.99 (v2.1.0-324-g1367413a)
duk> Math.noSuch(123)
TypeError: undefined not callable (property 'noSuch')
    at [anon] (duk_js_executor.c:4746) internal
```

The error message illustrates another limitation: the CALLPROP opcode no longer sees the name of the base object (assuming it's a slow path variable lookup) so only the property name is included in the error message.

fatcerberus:
Having the key in the error message is still very useful even without the object name, and is a big improvement over "undefined not callable" :)

svaarala:
Yeah, the common case is that some property is undefined and the property name provides at least a small hint of what might be wrong.

The base value name ('Math') isn't currently available, but the value itself can be summarized which should (partially) cover the other common case where the base value is accidentally wrong:

```
((o) Duktape [linenoise] 2.1.99 (v2.1.0-324-g1367413a-dirty)
duk> Math.noSuch(123)
TypeError: undefined not callable (property 'noSuch' of [object Math])
    at [anon] (duk_js_executor.c:4748) internal
duk> x = 'oopsString'
= "oopsString"
duk> x.noSuch()
TypeError: undefined not callable (property 'noSuch' of 'oopsString')
    at [anon] (duk_js_executor.c:4748) internal
```

But this won't really help if one [object Object] is mistakenly used instead of another [object Object].

fatcerberus:
Interestingly, Firefox/SpiderMonkey seems to have enough information at runtime to show the offending *identifier* in the error message:
```js
>> var foo = {};
>> var bar = {};
>> (foo, bar).getEaten()
```
> TypeError: bar.getEaten is not a function

It can even distinguish the variables when they refer to the same object:
```js
>> foo = bar
>> foo.getEaten()
>> bar.getEaten()
```
> TypeError: foo.getEaten is not a function
> TypeError: bar.getEaten is not a function

I don't think it's easy to implement that in the current Duktape compiler, just throwing it out there. :)

svaarala:
Sure, there's no fundamental reason why the base identifier couldn't be carried along the way into call handling. It's just not easily available right now.

svaarala:
Basically the compiler would just need to remember which identifier was used to look up the base value. That's syntactically always available, even for comma expressions, and could be carried through ivalues in a relatively straightforward manner.

The bytecode executor would then need to know the base value identifier for error throwing. There are two basic approaches for that.

One can load the identifier name into a register for error throwing or point to the constant table for it (which narrows the available bits a bit). The downside is that one is doing extra work which is only needed when an error is actually thrown; often the base value is register bound and there's otherwise no need to deal with its identifier at runtime.

Another approach would be to associate metadata related to an opcode which wouldn't be used unless the executor looked it up explicitly. For example, the base identifier could be associated with the opcode doing the property-based call. This wouldn't have a runtime cost which would be nice. However, there's no current data structure to do that (though I've prototyped an approach like this to make all error catching implicit without any opcodes like TRYCATCH).

So.. definitely a feature that might be nice at some point, but probably way too much effort to be added to the current compiler/executor.

fatcerberus:
Maybe the error message doesn't need to have the value in it if the key is known?  It seems too verbose, I would be happy with just:
```
property 'noSuch' of 'oopsString' not callable
```
Which is pretty close in spirit to the Firefox error.

svaarala:
To me it makes sense to include the value because it's also included in other call error cases, e.g.:

```
duk> var x = 123;
= undefined
duk> x()
TypeError: 123 not callable
    at [anon] (duk_js_call.c:1316) internal
```

With this pull that just gets extended with some extra information of the base value and the key, e.g.:

```
duk> Math.x = 123;
= 123
duk> Math.x()
TypeError: 123 not callable (property 'x' of [object Math])
    at [anon] (duk_js_executor.c:2648) internal
```

To me it'd be weird if the value wasn't summarized in this case. A matter of taste to be sure :)

svaarala:
Hmm, but maybe the error would be more readable in a different order. Current is:

```
TypeError: 123 not callable (property 'x' of [object Math])
```

Maybe what you propose with a tweak to include the value would be more readable:

```
TypeError: property 'x' of [object Math] (= 123) not callable
```

or:

```
TypeError: property 'x' of [object Math] not callable (123)
```

Then again the current form has the benefit of being similar to the non-property-based call TypeError which is also useful.

svaarala:
Again a matter of taste, but IMO it's also good if 'not callable' is relatively early in the error message rather than at the very end.

svaarala:
This basic approach has a measurable impact on call performance (microbenchmarks at least):

```
duktape@keppihevonen:~/duktape$ time ./duk.O2.master tests/perf/test-call-prop.js 
10274.852298998201 calls per millisecond

real	0m3.899s
user	0m3.896s
sys	0m0.000s

duktape@keppihevonen:~/duktape$ time ./duk.O2 tests/perf/test-call-prop.js 
7525.870178739417 calls per millisecond

real	0m5.322s
user	0m5.316s
sys	0m0.000s
```

The extra "is callable" check and duk_replace() seem like a small overhead, but call handling is now relatively fast so the difference turns up in the microbenchmark. The difference seems maybe too large to be merged into master :-/

Another implementation approach would be to use normal call handling as before, but when call setup fails (which is a slow path where additional overhead no longer matters) we'd somehow be able to access the related key and possibly the base value. There's no convenient way to do that at present, unfortunately.

One would at least need to make a copy of the key. A reference to the constants table is not enough in the general case because the key may be computed (though supporting that case is probably not critical):

```
duk> Math['foo'+'bar']()
TypeError: undefined not callable (property 'foobar' of [object Math])
    at [anon] (duk_js_executor.c:2648) internal
```

svaarala:
Optimizing the CALLPROP<N> opcode handling a bit the difference shrinks a bit but is still measurable:

```
duktape@keppihevonen:~/duktape$ time ./duk.O2.master tests/perf/test-call-prop.js 
10149.708195889369 calls per millisecond

real	0m3.947s
user	0m3.944s
sys	0m0.000s

duktape@keppihevonen:~/duktape$ time ./duk.O2 tests/perf/test-call-prop.js 
8658.008658008657 calls per millisecond

real	0m4.623s
user	0m4.620s
sys	0m0.000s
```

svaarala:
Some more optimization:

```
duktape@keppihevonen:~/duktape$ time ./duk.O2 tests/perf/test-call-prop.js 
9409.550693954363 calls per millisecond

real	0m4.256s
user	0m4.252s
sys	0m0.000s
```

With this the microbenchmark is < 10% slower.

fatcerberus:
FWIW Firefox doesn't seem to show the value in the error message if the identifier is known.  But it's probably useful for logging if it's included anyway. :)

Maybe this is a good format:
`property 'x' (= 123) not callable on [object Math]`

🚲🏠 :)

svaarala:
Looks like a good compromise if one wants to include all three items and try to put 'not callable' as early as possible :)

svaarala:
Hmm... I may have an idea of how to reduce the up front overhead and handle the whole thing almost entirely in normal call setup processing. Let's see if it pans out.

fatcerberus:
Sorry for all the bikeshedding in these error message issues, I realize that error messages are ultimately a compromise.  You want your errors to be as informative as possible because sometimes an error message is all you have to debug an issue, but you also don't want to introduce a bunch of extra checks and state *just* for the sake of specificity, either.

svaarala:
Not at all, I appreciate argumentation and suggestions for these. It's always a compromise and also somewhat arbitrary what looks good; consistency also matters so it's not trivial to find the shortest, most consistent, and most informative error messages.

I find error messages quite important to make an engine (or any library generally) nice to use. One often picks up scripting in order to make things easier, and good error messages, tracebacks, etc are some of the things that make scripting nice. Without useful tracebacks, for example, I find scripting much less useful.

svaarala:
Nope, not too much improvement with that approach. Came up with yet another idea though ;)

svaarala:
Ok, the latest idea seems to work. The basic idea is to avoid touching call opcodes or call handling at all but instead introduce a GETPROPC opcode which is the same as GETPROP but with a callability check for the property value read. This happens before any call processing and is relatively cheap because the all the values (base, key, property value) are at hand automatically. It's also nicer in that much fewer new opcodes are introduced.

There's barely a measurable difference in execution time; for test-call-prop.c master is 3.92 and the prototype is 3.94 seconds (average of 5).

I'll play with the approach a bit more, and if it seems like a better idea, open a fresh pull for it.

svaarala:
Hmm.. that alternative approach (which is fast) has the issue discussed previously that callability check will happen before argument evaluation. Have to think about this some more :-)

svaarala:
The GETPROPC approach could be salvaged with some trickery:
- Do a GETPROPC first. If the result is not callable, construct the error and place it e.g. into the call target slot (idx_func). Don't throw the error yet.
- Evaluate arguments normally.
- Do a CALL operation normally. Call setup will detect that the target is non-callable (= error object). In that code path, somehow figure out that we're supposed to throw the object as is without affecting the case where user code attempts to call an error object.

One simple way would be an internal property on the error object which is written and checked only when an error is going to be thrown.

svaarala:
#1630 takes this pull's place, so closing without merge.

PR #439, opened 2015-11-16
Flip download list order to match releases list

svaarala:
Most users seem to prefer newest-first ordering.


fatcerberus:
Yeah, this is preferable to me, too.  With each release you have to scroll further down, whereas this ensures the latest is always up top.

svaarala:
For some reason people prefer this, I've never understood it myself :) I also append to log files instead of prepending etc ;)

PR #440, opened 2015-11-16
Detect and greet NetSurf users on index page

svaarala:
Add a little easter egg.


fatcerberus:
Hehe, just a little "Hey, bet you didn't know you were using Duktape right now did ya?" :)

svaarala:
Yeah, I was also wondering what'd be the best way to detect Duktape when the `Duktape` object had been removed.  Tracebacks are still a give-away :)

PR #441, opened 2015-11-16
Improve error messages for invalid index and require errors in API calls

### LINKED ISSUE #28: duk_require_xxx little more verbose error string
Hi! I have a simple idea:

Why not making the error strings of duk_require_xxx functions a little more verbose by including the stack index.

For example, when doing duk_require_int(ctx, 0) in place of "not a number" we would have "stack id 0: not a number"

(In duktape.c, "not a number" will become "stack id %d: not a number")

Thanks :)

svaarala:
Fixes #28, also improves other API call errors.

Laundry list of tasks:

- [x] Add more verbose errors for index API calls
- [x] Add more verbose errors for duk_require_xxx() API calls
- [x] Add `duk_require_function()` and `duk_require_callable()` to API
- [x] Improve `Array.prototype.forEach()` error message by using `duk_require_callable()`
- [x] Add surrounding parentheses for tval summary of a pointer to ensure NULL pointer is distinguishable from a null value (`(null)` vs `null`)
- [x] Remove unused strings
- [x] Fix duk_hobject_props.c uninitialized warning (which is incorrect, but annoying)
- [x] Fix warnings
- [x] Check verbose errors
- [x] Check paranoid errors
- [x] Check minimal errors
- [x] Check if shared helper/macros could be used to reduce code duplication
- [x] Paranoid messages should not summarize values, use type names instead
- [x] Code footprint difference, minimize (started from close to +1,5kB)
- [x] Testcase fixes and additions
- [x] Releases

Future work notes:

- Non-verbose error messages are currently defined in code but don't make it into error objects when DUK_USE_VERBOSE_ERRORS is not set (not any longer anyway). For low memory targets it'd actually be useful to distinguish between "no error messages" and "very spartan messages". For the latter one could use e.g. "unexpected type" for all type-related errors etc. Non-verbose errors could use a bit of cleanup.
- Invalid index error message for DUK_INVALID_INDEX refers to index -2147483648 which is not ideal, and the constant is also potentially platform specific.


svaarala:
Right now errors look like:

```
TypeError: null not pointer at value stack index -3
```

I think it'd be better to start the error message with the expected type directly ("pointer" here):

```
TypeError: pointer required at value stack index -3, found null
TypeError: pointer required, found null at value stack index -3
```

Suggestions for better wording?

Is "value stack index -3" too verbose? Would "index -3" be OK? Example:

```
TypeError: pointer required at index -3, found null
TypeError: pointer required, found null at index -3
```

Or perhaps even:

```
TypeError: pointer required @-3, found null
TypeError: pointer required, found null @-3
```

Suggestions welcome :)

svaarala:
The error message could also either indicate the offending type or the offending value:

```
TypeError: pointer required at index -3, found "foobar"  // summarize offending value
TypeError: pointer required, found string at index -3    // show offending type only
```

svaarala:
I think I prefer these:

```
TypeError: pointer required, found null (at index -3)
TypeError: pointer required, found null (at value stack index -3)
```

The value stack index is the least informative to an end user so it should probably be last in the message.

fatcerberus:
Generally speaking this error won't be triggered by Ecmascript code alone, it indicates a bug in either the application or Duktape itself.  So the specific message is less important than for, say, a noncallable error.

I agree "value stack index" is too verbose though.  Honestly I would probably drop the "at" also, leaving:

```
pointer required, found null (index -3)
```

svaarala:
Hmm, I'm not sure what you mean, but these error messages are not necessarily related to bugs? It's quite OK to say:

```
const char *my_msg = duk_require_string(ctx, 0);
```

and if the argument is not a string, get one of these error messages.

For me "index" would also be enough, but it might maybe be confused with an array index, at least if operating on an array.

svaarala:
For example:

```
((o) Duktape 1.3.99 (v1.3.0-270-g7ac6bd1-dirty)
duk> require(new ArrayBuffer())
TypeError: string required, found [object ArrayBuffer] (index 0)
	duk_api_stack.c:1252
	require  native strict preventsyield
	global input:1 preventsyield
```

svaarala:
The error message above would actually make more sense if it indicated "arg 1", but the value stack index doesn't necessarily refer to arguments and can also be negative, etc.

fatcerberus:
Oh, sorry for the confusion, I was getting these mixed up with "invalid index". :)

Hm, maybe split the difference and say "stack index"?  I just think if you make it too verbose the index starts to dominate the error message, regardless of its positioning.

svaarala:
"stack index" works for me; I was also considering "vs index" which is somewhat technical. But making sense of the index requires understanding of what a value stack is anyway, so perhaps that'd be OK. 

The index related errors are distinct, but they can also be thrown out in argument error situations even without an application/Duktape bug. For example, a called function may want to ensure that a certain index exists (= is provided by caller) for a required argument.

fatcerberus:
For me at least, "vs index" would be confusing, since I normally just say "the stack" in reference to the Duktape valstack--I only tend to qualify it if I need to differentiate between it and, say, the callstack.

Maybe that's just me though.

svaarala:
Yeah, "stack" is understandable for a developer using the API. But these errors may also be shown to end users etc who have never heard of the concept. So they will either ignore the stack index or incorrectly assume it refers to some array index or something ;)

svaarala:
Concrete example of the best suggestion so far:

```
((o) Duktape 1.3.99 (v1.3.0-271-g596cac5-dirty)
duk> Duktape.enc(Duktape.Pointer('dummy'), {foo:'bar'})
TypeError: string required, found 0x1841060 (stack index 0)
	duk_api_stack.c:1460
	enc  native strict preventsyield
	global input:1 preventsyield
```

svaarala:
In 1.3.0 that would produce:

```
duk> Duktape.enc(Duktape.Pointer('dummy'), {foo:'bar'})
TypeError: unexpected type
	duk_api_stack.c:1432
	enc  native strict preventsyield
	global input:1 preventsyield
```

fatcerberus:
I like that, with the value shown.  That'll help when debugging, to find when something weird is passed to a function.

Definitely a big improvement over "unexpected type" :-)

svaarala:
Price in code footprint is >1kB so far :neckbeard:

fatcerberus:
On the other hand, maybe include the type with the value?  (e.g. "found number 812"). Since the value is string coerced, the type could potentially be ambiguous.

fatcerberus:
Wow, why so much overhead?

svaarala:
The "summarization" results should not be ambiguous: strings are quoted etc. For example:

- If the value is `undefined`, the summary is `undefined`
- If the value is the string `"undefined"` the summary is `'undefined'`.

svaarala:
There are a lot of error strings that get added, e.g.:

```
buffer required, found %s (stack index %ld)
string required, found %s (stack index %ld)
compiledfunction required, found %s (stack index %ld)
regexp required, found %s (stack index %ld)
thread required, found %s (stack index %ld)
nativefunction required, found %s (stack index %ld)
undefined required, found %s (stack index %ld)
null required, found %s (stack index %ld)
boolean required, found %s (stack index %ld)
number required, found %s (stack index %ld)
pointer required, found %s (stack index %ld)
heapobject required, found %s (stack index %ld)
```

These alone are ~0.5kB. Then there's additional plumbing for making helper calls to convey the expected types, etc.

svaarala:
I could refactor the code to avoid full format strings and build the message from smaller parts, not sure if that leads to a better result though.

fatcerberus:
I tried to test this branch, but it didn't compile:
```
duk_api_stack.c(481): error C2065: 'index': undeclared identifier (compiling source file ..\src\duktape.c)
duk_api_stack.c(878): error C2065: 'index': undeclared identifier (compiling source file ..\src\duktape.c)
```

svaarala:
Yeah the branch is still in progress and I haven't pushed all local changes yet. I'll push an update in a few minutes.

svaarala:
Ok, hopefully compiles now - but just to be sure, the branch is not yet complete :)

fatcerberus:
Thanks, I knew it wasn't complete but wanted to test it out anyway.

fatcerberus:
Oh, crud, this breaks my custom error augmentation - minisphere automatically checks for error messages of the form "not XXX" (with exclusions for, e.g. "not callable") and calls `Duktape.act()` to get the correct filename and line number for a failed `duk_require_xxx()`.  As C has no native regex support, the check will be more complicated once this makes it to master.

svaarala:
That's one price of the Ecmascript error model: the sub-typing is very simple which is nice in many ways, but error messages play a more prominent role. And since one cannot rely on any specific message contents, there's no reliable way to programmatically distinguish errors unless the throw site is using a custom Error subclass specifically.

svaarala:
Some notes on "blame" handling in #454.

svaarala:
Testcase expect strings are now fixed for API calls and tests now pass. I'll try to get this merged tomorrow.

fatcerberus:
Why the distinction between "is function" and "is callable" anyway?  That makes me think it's possible to have a non-callable function, which would be stupid and dumb.

svaarala:
IsCallable is defined in E5 Section 9.11, returning true for any object with a `[[Call]]` internal method. For standard objects it's equivalent to being a Function. There might be some object types in the future which can be called but are not (e.g. from inheritance perspective) Function instances. That's the reasoning anyway.

In any case, `duk_is_callable()` and `duk_is_function()` were already present before this branch.

svaarala:
Full Proxy support will actually probably make a distinction between the two methods, for a callable Proxy:

- `typeof` will be `function`, as it is required to be `function` whenever the target implements `[[Call]]`
- The Proxy is callable, so `duk_is_callable()` will be true
- However, from a strict C API typing perspective `duk_is_function()` would, at least currently, be false because the value is not a Function object (Ecmascript function, C function, or a bound function)

It may be necessary to change the API semantics slightly when callable Proxies are implemented though.

svaarala:
Anoither difference is also that it would be possible to have a function which had `[[Construct]]`, i.e. you could `new Foo()`, but didn't have `[[Call]]`, i.e. you couldn't `Foo()`.

This may seem a bit pointless, but it'd sometimes be useful to define a function which could only be used as a constructor and attempt to call it as a normal function would automatically `TypeError`. For example, some of the typed array bindings require this behavior but since there's no way to prevent those calls automatically, the bindings need to check the call style explicitly (using `duk_is_constructor_call()`).

If this separation was made (so that you could create such functions), it'd also nice to have `duk_is_constructable()` which would then be different from `duk_is_callable()`.

fatcerberus:
Okay, that all makes a lot more sense now.

I knew the distinction was already there before, I just wanted some clarification on it and this seemed like a good place to ask about it.

svaarala:
Well don't know how much sense it makes, it'll probably need revisiting later on. The reason `duk_is_callable()` exists is partially because I originally implemented all the specification algorithms regardless of whether they were needed or not, and things like Proxy were not in the picture then.

By the way, if there was a Function object which couldn't be called normally - in specification terms it didn't have `[[Call]]` internal method - the specification would require that `typeof` of such a function be `object` instead of `function`, even if the function was callable as a constructor... :)

fatcerberus:
Interesting, so this means according to ES spec, testing whether `typeof f === "function"` is exactly equivalent to asking if `f` is callable.  That's good to know for writing polyfills.

Getting back on topic, did you come up with a solution to mitigate the 1kb+ footprint overhead?

svaarala:
The size difference is roughly the same still. It's probably caused by the error call sites; with verbose errors they need to push a format string, potentially format a `duk_tval` using a function call etc. For example, the error call path for `duk_require_lstring()` is:

```
  4174a7:       89 ee                   mov    %ebp,%esi
  4174a9:       48 89 df                mov    %rbx,%rdi
  4174ac:       e8 5f a7 fe ff          callq  401c10 <duk_push_string_readable>
  4174b1:       48 89 6c 24 08          mov    %rbp,0x8(%rsp)
  4174b6:       48 89 04 24             mov    %rax,(%rsp)
  4174ba:       41 b9 56 f0 42 00       mov    $0x42f056,%r9d
  4174c0:       41 b8 02 f6 42 00       mov    $0x42f602,%r8d
  4174c6:       b9 69 00 00 00          mov    $0x69,%ecx
  4174cb:       48 89 da                mov    %rbx,%rdx
  4174ce:       be fa 04 00 00          mov    $0x4fa,%esi
  4174d3:       bf 62 ee 42 00          mov    $0x42ee62,%edi
  4174d8:       31 c0                   xor    %eax,%eax
  4174da:       e8 ba 42 ff ff          callq  40b799 <duk_err_handle_error>
```

which is 51 bytes. There are about 20 call sites like this, which amounts to 1kB of code; add strings to that and it pretty much explains the difference.

I'll try to see if I could put the majority of that call site into a helper. Even so, each call site will need to pass at the minimum the following to the helper for the call site to be properly identified:

- `__FILE__`
- `__LINE__`
- Offending index, possibly offending `tv` pointer
- Expected type name

Maybe call sites could be 50% of their current size, but not much less. Ignoring `__FILE__` and `__LINE__` would be possible. What would be lost is the reference to Duktape internal file/line in the traceback (which is often useful at least when figuring out ultimate error causes).

svaarala:
Okay, with that helper change applied the difference is now +656 bytes over master. This is pretty reasonable I think, and I think it'd be possible to shave several kilobytes of footprint by applying this approach elsewhere too. That's a matter for a separate pull though :)

svaarala:
I think I'll move the "require type at index" error macro stuff into the generic error headers and helpers, and then add more helpers there in a follow up pull to reduce error call site size. Those macros are a bit messy so it'd be nice if they didn't "pollute" the API call code, and there's already a bunch of config dependent handling in the generic error headers and helpers.

svaarala:
Couldn't resist the temptation to test what the impact for the helpers would be - by converting all DUK_ERR_API_ERROR throws into helper calls (which saves one argument per call site, the error type), around 150 bytes are saved.

svaarala:
Ok, so the final difference to master (with the API error change included), is +352 bytes.

svaarala:
I won't do any actual code changes in this pull anymore, but just remove obvious warnings and go through all the config combinations to see errors come out reasonably in each. This would then be ready for merging.

svaarala:
Verbose errors:

```
$ ./duk tests/ecmascript/test-dev-api-verbose-error-messages-gh441.js 
TypeError: string required, found none (stack index 0)
TypeError: string required, found undefined (stack index 0)
TypeError: string required, found null (stack index 0)
TypeError: string required, found true (stack index 0)
TypeError: string required, found false (stack index 0)
TypeError: string required, found 123 (stack index 0)
TypeError: string required, found [object Object] (stack index 0)
TypeError: string required, found [object Array] (stack index 0)
TypeError: string required, found [object Function] (stack index 0)
TypeError: string required, found (PTR) (stack index 0)
TypeError: string required, found [buffer:4] (stack index 0)
TypeError: string required, found [object Pointer] (stack index 0)
TypeError: string required, found [object Buffer] (stack index 0)
TypeError: object required, found '' (stack index 0)
TypeError: object required, found 'foo' (stack index 0)
TypeError: object required, found 'foo<4660>' (stack index 0)
TypeError: object required, found 'foobarfoobar\x00foobarfoobar<51966>foobar...' (stack index 0)
TypeError: RegExp required, found 123 (stack index -1)
```

Paranoid errors:

```
$ ./duk tests/ecmascript/test-dev-api-verbose-error-messages-gh441.js 
TypeError: string required, found none (stack index 0)
TypeError: string required, found undefined (stack index 0)
TypeError: string required, found null (stack index 0)
TypeError: string required, found boolean (stack index 0)
TypeError: string required, found boolean (stack index 0)
TypeError: string required, found number (stack index 0)
TypeError: string required, found object (stack index 0)
TypeError: string required, found object (stack index 0)
TypeError: string required, found object (stack index 0)
TypeError: string required, found pointer (stack index 0)
TypeError: string required, found buffer (stack index 0)
TypeError: string required, found object (stack index 0)
TypeError: string required, found object (stack index 0)
TypeError: object required, found string (stack index 0)
TypeError: object required, found string (stack index 0)
TypeError: object required, found string (stack index 0)
TypeError: object required, found string (stack index 0)
TypeError: RegExp required, found number (stack index -1)
```

Minimal errors:

```
$ ./duk tests/ecmascript/test-dev-api-verbose-error-messages-gh441.js 
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
TypeError: 105
```

fatcerberus:
Only 352 bytes, nice.  Believe me, I know how much of a hit 1kb is for limited memory environments--I still remember my DOS days with only ~400kb on average of usable RAM. :)

PR #445, opened 2015-11-17
Add indent and JX/JC support into the JSON.stringify() fast path

svaarala:
With these limitations removed the fast path is pretty versatile; the only things not supported are actual mutation risks such as:

- Replacer callback
- `.toJSON()` methods
- Accessor properties
- Gappy arrays inheriting ancestor values

Tasks:

- [x] Implement indent/gap support to fast path
- [x] Use an internal "always inline" helper and a parameter to ensure best performance without gap/indent and for plain JSON (the most important case) => does not seem worth it
- [x] Add JX/JC support to fast path
- [x] Improve fast path test cases to cover all new code paths
- [x] Perf test for fast path with indent - existing test cases cover that
- [x] Perf test for fast path JX and JC
- [x] Ensure manually (using debug prints) that fast path is correctly taken for all new paths
- [x] Update fast path limitation documentation on Duktape Wiki - https://github.com/svaarala/duktape-wiki/pull/63
- [x] Update fast path limitation documentation on website and config options
- [x] Compilation without JX/JC support
- [x] Check binary size (-Os) for with and without fast path
- [x] Check performance change against master (especially for plain JSON.stringify() fast path which is the most common case)
- [x] Assertion test run
- [x] Releases

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1227,6 +1227,10 @@ Planned
 * Internal performance improvement: improve JSON.stringify() default slow
   path indent handling (GH-444)
 
+* Internal performance improvement: improve JSON.stringify() fast path
+  by allowing indent value or gap string and by supporting JX/JC in the
+  fast path (GH-445)
+
 2.0.0 (XXXX-XX-XX)
 ------------------
 
--- a/config/config-options/DUK_USE_JSON_STRINGIFY_FASTPATH.yaml
+++ b/config/config-options/DUK_USE_JSON_STRINGIFY_FASTPATH.yaml
@@ -8,8 +8,8 @@ tags:
   - lowmemory
 description: >
   Enable fast path for JSON.stringify() serialization.  The fast path is used
-  when there is no "replacer" or indent; in practice, when JSON.stringify() is
-  called with only a value argument.  The fast path increases code footprint by
+  when there is no "replacer" argument.  Indent argument and JX/JC format is
+  supported since Duktape 1.4.0.  The fast path increases code footprint by
   roughly 1.5 kB but is up to 4-5x faster than the slow path.
 
   Current limitation: assumes "long long" type exists (and covers duk_int64_t
--- a/src/duk_bi_json.c
+++ b/src/duk_bi_json.c
@@ -2230,8 +2230,14 @@ DUK_LOCAL duk_bool_t duk__json_stringify_fast_value(duk_json_enc_ctx *js_ctx, du
 				tv_val = DUK_HOBJECT_E_GET_VALUE_TVAL_PTR(js_ctx->thr->heap, obj, i);
 
 				prev_size = DUK_BW_GET_SIZE(js_ctx->thr, &js_ctx->bw);
-				duk__enc_quote_string(js_ctx, k);
-				DUK__EMIT_1(js_ctx, DUK_ASC_COLON);
+				if (DUK_UNLIKELY(js_ctx->h_gap != NULL)) {
+					duk__enc_emit_newline_indent(js_ctx, js_ctx->recursion_depth);
+					duk__enc_quote_string(js_ctx, k);
+					DUK__EMIT_2(js_ctx, DUK_ASC_COLON, DUK_ASC_SPACE);
+				} else {
+					duk__enc_quote_string(js_ctx, k);
+					DUK__EMIT_1(js_ctx, DUK_ASC_COLON);
+				}
 				if (duk__json_stringify_fast_value(js_ctx, tv_val) == 0) {
 					DUK_DD(DUK_DDPRINT("prop value not supported, rewind key and colon"));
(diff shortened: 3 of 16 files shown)


svaarala:
These perf tests now execute using the fast path:

<table>
<tr><th></th><th>fastpath</th><th>master</th><th>130</th></tr>
<tr><td>test-json-serialize-indented.js</td><td>8.49</td><td>61.84</td><td>67.43</td></tr>
<tr><td>test-json-serialize-indented-deep25.js</td><td>3.42</td><td>40.60</td><td>46.93</td></tr>
<tr><td>test-jx-serialize-indented.js</td><td>3.67</td><td>30.62</td><td>33.09</td></tr>
<tr><td>test-jx-serialize.js</td><td>2.49</td><td>28.98</td><td>29.24</td></tr>
<tr><td>test-jc-serialize-indented.js</td><td>3.81</td><td>33.51</td><td>36.04</td></tr>
<tr><td>test-jc-serialize.js</td><td>2.55</td><td>29.55</td><td>30.95</td></tr>
</table>

while these execute as before:

<table>
<tr><th></th><th>fastpath</th><th>master</th><th>130</th></tr>
<tr><td>test-json-serialize-indented-deep100.js</td><td>3.00</td><td>2.96</td><td>3.78</td></tr>
<tr><td>test-json-serialize-indented-deep500.js</td><td>1.91</td><td>1.99</td><td>3.36</td></tr>
<tr><td>test-json-serialize.js</td><td>6.03</td><td>5.96</td><td>6.27</td></tr>
</table>

Deep objects (> 32) cause fast path to be abandoned as before, which is why there's no improvement for deeply intended objects. Basic serialize test was already executing using the fast path; it now executes slightly slower because the fast path has additional checks for indent and JX/JC features.

(Note that the numbers are not meaningfully comparable between test cases, only within a test, because loop counts differ between tests.)

svaarala:
Effect on Josh's AST test; without indent (as before):

```
duktape@keppihevonen:~/duktape$ time ./duk.O2.fastpath stuff/_josh.js 
real	0m1.672s
user	0m1.662s
sys	0m0.004s

duktape@keppihevonen:~/duktape$ time ./duk.O2.master stuff/_josh.js 
real	0m1.646s
user	0m1.637s
sys	0m0.004s

duktape@keppihevonen:~/duktape$ time node stuff/_josh.js 
real	0m1.310s
user	0m1.279s
sys	0m0.028s
```

With indent-by-4:

```
duktape@keppihevonen:~/duktape$ time ./duk.O2.fastpath stuff/_josh_indent.js 
real	0m2.967s
user	0m2.506s
sys	0m0.444s

duktape@keppihevonen:~/duktape$ time ./duk.O2.master stuff/_josh_indent.js 
real	0m17.775s
user	0m17.171s
sys	0m0.519s

duktape@keppihevonen:~/duktape$ time node stuff/_josh_indent.js 
real	0m14.561s
user	0m13.679s
sys	0m0.959s
```

It appears V8 also has a fast path for non-indented JSON but indented JSON goes through a different path?

fatcerberus:
I guess that makes sense, if you're just using JSON for general serialization you probably don't need indentation, so it's the most obvious candidate for a fast path.

svaarala:
Yes, that was the reasoning for the original fast path limitation too.

svaarala:
Notes for future improvement:

- Fast path aborts when encounting a lightfunc because a lightfunc might inherit a `.toJSON()` method. It would be better to check for `.toJSON()` explicitly and continue in the fast path in the (very likely) case where no such method exists.
- Fast path aborts on presence of `.toJSON()  even when it's not callable. This is probably not worth fixing.
- JX automatic unquoted key emitter first checks the key for being compatible and then emits the key as is or emits an ordinary string. This could be done in one pass: start emitting the unquoted key, and if an incompatible character is encountered, rewind the bufwriter and write the normally quoted key.
- Slow path now uses a "enc1 + enc2" helper approach which could now be reworked to just emit in one step and backtrack if the value should actually have been dropped. Refactor in a separate pull.

svaarala:
Hmm, there's a minor bug in the 1.3.0 fast path which causes a boxed Duktape.Pointer to be emitted when doing standard JSON serialization:

```
duk> JSON.stringify([ Duktape.Pointer('dummy'), new Duktape.Pointer('dummy') ])
= [null,{}]
duk> JSON.stringify([ Duktape.Buffer('dummy'), new Duktape.Buffer('dummy') ])
= [null,null]
```

Disabling the fast path with a replacer function:

```
duk> JSON.stringify([ Duktape.Buffer('dummy'), new Duktape.Buffer('dummy') ], function id(k,v) { return v; })
= [null,null]
duk> JSON.stringify([ Duktape.Pointer('dummy'), new Duktape.Pointer('dummy') ], function id(k,v) { return v; })
= [null,null]
```

Added a bug testcase, the issue is fixed in the pull.

svaarala:
Binary size impact on x64:

- Fast path enabled: +704 bytes over master
- Fast path disabled: +16 bytes over master

Issue #454, opened 2015-11-23
Improve selection of .fileName and .lineNumber for errors


svaarala:
Choosing the best fileName/lineNumber to "blame" is a bit difficult and there's room for much improvement.

For the `duk_require_xxx()` calls Duktape now blames the caller:

```
duk> Duktape.enc.fileName = 'foo.c'
= foo.c
duk> try{Duktape.enc(123);}catch(e){err=e}
= TypeError: string required, found 123 (stack index 0)
duk> err.fileName
= foo.c
duk> err.lineNumber
= 0
```

By default built-in methods don't have a `.fileName` so without assigning one the error's `.fileName` will actually be undefined now.

So here we blame `Duktape.enc()` which is a reasonable choice.

On the other hand, a user working only with Ecmascript code would maybe want to blame the nearest Ecmascript caller (and something else if there's no Ecmascript caller). Even further, if you're writing code on top of an Ecmascript framework, a user might want to blame the nearest user code line outside the Ecmascript framework.

It'd be good to improve on this somehow. One relatively simple approach would be to allow a function to be tagged "don't blame me" and when an Error was created, Duktape would find the most recent function without this tag. Instead of the current hardcoded blame approach, this would allow even Ecmascript to be tagged like this.

fatcerberus:
For the function tagging, maybe it could work like the way strict mode is requested, where you add a bare string literal to the top of the function and the compiler treats it as a tag: I think you already do this for tailcall flagging actually?

svaarala:
That'd be quite doable yes - it would ultimately then either set an object flag or an (internal?) property to indicate the status.

fatcerberus:
For reference, what is the default `.fileName` for Duktape/C functions?  Is it also `undefined`?  If so, the "closest function with filename" behavior mentioned above would be adequate for my use case at least.

svaarala:
They don't have a `.fileName` by default so that might work.

fatcerberus:
For now I could probably implement this myself by calling `Duktape.act()` successively (in an errCreate handler) until it returns a valid filename.

svaarala:
@fatcerberus There's a draft change for this behavior in #455. It seems like a pretty easy change - though the logic needs to be duplicated for the case where there are no tracebacks.

svaarala:
So, #455 was merged and what's left for improvement is providing a specific way to control blaming regardless of presence of `.fileName`: these two issues (blaming vs. presence of `.fileName`) usually, but not always, correlate.

Issue #149, opened 2015-03-24
Add the equivalent of duk_error_raw() and duk_error_va_raw() to the public API


svaarala:
See #147 for some discussion,

fatcerberus:
Now that I'm thinking about this, could I get the same behavior as `duk_error_raw()` by pushing an error onto the stack using `duk_push_error_object()` and then modifying the filename and line number manually?

svaarala:
The end result would be the same for the most part, with respect to tracebacks, etc. However, internally there would be a difference because normally call site `__FILE__` and `__LINE__` are stored in `_Tracedata` and not as own properties.

PR #455, opened 2015-11-23
Better error .fileName and .lineNumber default

svaarala:
Use the first traceback entry which actually has a `.fileName` to provide virtual `.fileName` and `.lineNumber`. See #454.

This works better for expressions such as `Duktape.enc(123)` which currently get `undefined` as `.fileName`: the value comes from the `Duktape.enc()` function, which has no `.fileName` property.

Tasks:

- [x] New behavior for traceback-based `.fileName` and `.lineNumber` getters
- [x] New behavior for error augmentation when tracebacks are disabled
- [x] Fix stack trace summarization: previously Duktape would print `[...]` before traceback depth was reached (code didn't account for compilation error or C call site)
- [x] Testcases and test case fixes
- [x] Testcase for empty callstack case
- [x] Testcase for the case where a callstack has entries but none has `.fileName`
- [x] Testcase for the case where a callstack has entries both with and without `.fileName`
- [x] Assert testcase run, with and without tracebacks
- [x] Final code review
- [x] Document the "error file/line attribution" issue in internal error document
- [x] Review and cleanup the internal error document
- [x] Code issues notes on preprocessor defines (somewhat unrelated to this issue)
- [x] Documentation changes (internal docs, website), if necessary
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1194,6 +1194,11 @@ Planned
 * Improve error message verbosity for API index check calls, duk_require_xxx()
   calls, and Array iterator calls (GH-441)
 
+* Improve error object .fileName and .lineNumber attribution: if a callstack
+  function is missing a .fileName property, scan the callstack until a
+  function with a .fileName is found which improves error reporting for e.g.
+  "[1,2,3].forEach(null)" (GH-455)
+
 * Add a combined duktape.c without #line directives into the dist package,
   as it is a useful alternative in some environments (GH-363)
 
@@ -1215,6 +1220,11 @@ Planned
 
 * Fix Unix local time offset handling which caused issues at least on RISC
   OS (GH-406, GH-407)
+
+* Fix a bug in stack trace ellipsis ("[...]") handling: previously the
+  ellipsis might be emitted up to 2 callstack levels too early because
+  the presence of a compilation error and/or a C call site was not taken
+  into account in stack trace creation (GH-455)
 
 * Remove octal autodetection in parseInt(), also fixes incorrect octal
   autodetection in e.g. "parseInt('00e1', 16)" (GH-413, GH-414)
--- a/config/config-options/DUK_USE_TRACEBACK_DEPTH.yaml
+++ b/config/config-options/DUK_USE_TRACEBACK_DEPTH.yaml
@@ -13,3 +13,7 @@ description: >
   Define traceback collection depth.  A large number causes tracedata to be
   larger, taking more time to create and consuming more memory.  A small
   number makes tracebacks less useful.
+
+  When tracebacks are disabled this option affects .fileName and .lineNumber
+  blaming.  Error augmentation code won't look deeper than this value to find
+  a function to blame for error .fileName / .lineNumber.
--- a/config/header-snippets/unsorted_flags.h.in
+++ b/config/header-snippets/unsorted_flags.h.in
@@ -130,12 +130,10 @@
 #undef DUK_USE_VERBOSE_ERRORS
 #endif
 
-#if defined(DUK_USE_TRACEBACKS)
 #if defined(DUK_OPT_TRACEBACK_DEPTH)
 #define DUK_USE_TRACEBACK_DEPTH  DUK_OPT_TRACEBACK_DEPTH
 #else
 #define DUK_USE_TRACEBACK_DEPTH  10
-#endif
 #endif
 
 /* Include messages in executor internal errors. */
--- a/doc/code-issues.rst
+++ b/doc/code-issues.rst
@@ -60,10 +60,10 @@ interest only can have a local name or have a double underscore after "DUK"::
   /* select DUK_FOO provider */
   #define DUK_FOO  DUK_FOO_ALT2
 
-There is only one space after a ``#define``, ``#ifdef``, etc, but there
-may be multiple spaces between the a macro name and its definition.  There
-is no strict rule on the alignment of a macro value; successive definitions
-usually keep values in the same column.
+There is only one space after a ``#define``, ``#if``, etc, but there may be
+multiple spaces between the a macro name and its definition.  There is no
+strict rule on the alignment of a macro value; successive definitions usually
+keep values in the same column.
 
 Comments are always traditional C comments, never ``//``, as they are not
 portable to older compilers::
@@ -126,7 +126,7 @@ This is more macro compatible.  Example::
 
 Multi-statement macros should use a ``do-while(0)`` construct::
 
-  #define  FROBNICATE(x,y)  do { \
+  #define FROBNICATE(x,y)  do { \
                   x = x * x; \
                   y = y * y; \
           } while (0)
@@ -135,18 +135,18 @@ When the body of a macro is sometimes empty, use an empty do-while so that
 the macro still yields a statement::
 
   #if defined(DUK_USE_FROB)
-  #define  FROBNICATE(x,y)  do { \
+  #define FROBNICATE(x,y)  do { \
                   x = x * x; \
                   y = y * y; \
           } while (0)
   #else
-  #define  FROBNICATE(x,y)  do { } while (0)
+  #define FROBNICATE(x,y)  do { } while (0)
   #endif
 
 Use parentheses when referring to macro arguments and the final macro
 result to minimize error proneness::
 
-  #define  MULTIPLY(a,b)  ((a) * (b))
+  #define MULTIPLY(a,b)  ((a) * (b))
 
   /* Now MULTIPLY(1 + 2, 3) expands to ((1 + 2) * (3)) == 9, not
    * 1 + 2 * 3 == 7.  Parentheses are used around macro result for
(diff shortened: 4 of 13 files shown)


svaarala:
Before the change:

```
duk> try { Duktape.enc(123); } catch (e) { print(e.fileName, e.lineNumber); }
undefined 0
```

After the change:

```
duk> try { Duktape.enc(123); } catch (e) { print(e.fileName, e.lineNumber); }
input 1
```

If one adds a `.fileName` to `Duktape.enc`, that will get blamed:

```
duk> Duktape.enc.fileName = 'dummy.js';
= dummy.js
duk> try { Duktape.enc(123); } catch (e) { print(e.fileName, e.lineNumber); }
dummy.js 0
```

fatcerberus:
Off-topic but I didn't realize `print()` accepted multiple parameters, learn something new every day!

svaarala:
Yeah it simply joins the values with spaces and appends a newline. The sole exception to this is if print() is called with a single buffer value; in that case the buffer is written as-is to stdout without adding a newline, which allows exact byte output. (Not sure if this is such a good idea, but that's the current behavior.)

svaarala:
@fatcerberus If you have a chance, it'd be nice to know if this branch would work for your issue in practice.

fatcerberus:
Alright, I'll try this out a bit later.

And thanks for that (presumably undocumented) tip on using a buffer with print() - I was looking for a way a while back to avoid printing the newline but it seemed impossible.

svaarala:
Ok, let me know what happens :)

That print() behavior is documented :-) => http://duktape.org/guide.html#builtin-print-alert

fatcerberus:
Looks like this works as expected.  I disabled my `errCreate()` handler, then called minisphere's `Abort()` function which has the following signature:

```
Abort(message, callstack_index);
```

`callstack_index` is checked using `duk_require_int()`.  I passed a string instead.  Using Duktape master, the Throw notify indicates "undefined" for the filename, while in this branch I get "script/main.js" line 30, which indeed matches the location of the offending `Abort()`.  So everything seems to be in order. :-)

Now all I need is a version of `duk_error()` that doesn't use `__FILE__` so I can stop calling `duk_error_va_raw()` directly.  I sometimes need to blame errors thrown from C on Ecmascript code... when range-checking function arguments, for example.

svaarala:
Do you mean just making the raw variant a part of the public API (#149) or to add some separate variant which doesn't "blame" the call site?

fatcerberus:
Implementing #149 would work, too.  Mainly I'm just uncomfortable relying on internal Duktape functions for the long-term, but it's not a terribly pressing concern. :)

svaarala:
The current behavior for `duk_error()` is to "blame" the error call site (`__FILE__` / `__LINE__` of the call site). Whether this is a good default or not really depends on the kind of program one is writing. If C code is doing a lot of actual logic, it's natural to pin errors on the C code. But if C code is mostly thin wrappers and such, it's more useful to blame the Ecmascript call site.

Perhaps the `duk_error()` default should be not to blame the C call site?

svaarala:
A short overview of how the "blaming" is handled internally:

Functions like `duk_error()` and `duk_push_error_object()` will automatically add the `__FILE__` and `__LINE__` into `_Tracedata`. A flag added to the line number indicates whether this C call should be "blamed" or not when reading back `.fileName` and `.lineNumber`.

When an error is created the "blame or not" flag is carried as part of the error code, e.g.:

```
DUK_ERR_API_ERROR | DUK_ERRCODE_FLAG_NOBLAME_FILELINE
```

This flag is now internal. It'd be trivial to expose it as a part of the API, though I'm not sure that's the best approach for the public API (maybe it is, there are other ORed flags too). I guess API options range from:

- Adding flags to the error code field, which needs no new API calls or arguments in calls
- Adding new API calls for blame / no-blame variants
- Adding (or rather exposing) a raw error throw / create call which has enough flags and parameters to provide flexibility for this and future situations

fatcerberus:
Regarding that "noblame" flag, won't that be made redundant by #455?

svaarala:
Not as #455 is now: if a C function is "blamed" for the error, `.fileName` and `.lineNumber` will match `__FILE__` and `__LINE__` regardless of whether the Duktape/C function has a `.fileName`or not. All user code `duk_error()` and `duk_push_error_object()` call sites are "blamed" at the moment in this respect.

That blaming behavior could of course be removed entirely - and Duktape/C functions would then only be blamed if the user explicitly added a `.fileName` to the Duktape/C function in question. The `.lineNumber` would still need to be magic'ed from the error call site `__LINE__` somehow.

Maybe that'd be a good change? It would affect the behavior of existing code quite a bit though. While there are no hard promises for `.stack`, `.fileName`, or `.lineNumber`, it's in the grey area whether the change is "surprising" for a minor version.

fatcerberus:
Ah, I hadn't realized the tracedata included a separate file and line, I thought it was always taken from the call stack.  But now that I think of it that doesn't make sense, since then `duk_error_raw()` wouldn't work.  Ignore my stupidity. :)

svaarala:
I think this issue is worth documenting in the internal errors document, with typical call scenarios and current/preferred behavior. It's kind of hard to figure out the best approach without that.

Luckily we're really only talking about a relatively minor issue here - the traceback always contains the relevant information (including the C call site's `__FILE__` and `__LINE__`, even if it's not blamed). But it'd still be great to have file/line properties be as useful as possible :)

By the way, the `.fileName` of a Duktape/C function (set by the user) may not have anything to do with `__FILE__`: it's entirely possible for the Duktape/C function to call a helper in another C file and the error to be thrown from there. So carrying the `__FILE__`/`__LINE__` combination seems mandatory in any case, but whether to "blame" it is another matter.

Anyway, I'll write up a section on file/line blaming because it's probably worth explaining in a bit more detail.

fatcerberus:
Agreed on the above points.  In any case, this pull as it's implemented right now is adequate for my use case, so whatever you decide to do beyond that is fine with me. :)

fatcerberus:
@svaarala
Thanks for making all these error message improvements by the way, Duktape's vague error messages were always one of my pet peeves with it compared to other engines.

svaarala:
You're welcome, it's a pretty painful but hopefully useful process :)

svaarala:
First draft of describing the issue, the Duktape 1.3.0 behavior, and the proposed Duktape 1.4.0 behavior: https://github.com/svaarala/duktape/blob/e86fa305226a42eaff03c69da94b680ee29cae51/doc/error-objects.rst#choosing-filename-and-linenumber-for-an-error

The main issue in the end is that Duktape would need to be able to distinguish between "infrastructure functions" (which are not "blamed" for error file/line) and "user functions" for file/line attribution. Both C and Ecmascript functions may be user and infrastructure functions, so no fixed rule can work for all cases. The same applies to the C call site: some call sites are infrastructure (like all Duktape internal `DUK_ERROR()` sites which are now correctly ignored for file/line purposes) while others are not.

fatcerberus:
Found a typo:
`These rules have a first shortcomings.`

I'm assuming that should be `a few shortcomings`.

svaarala:
Yeah that was a typo, fixed.

By the way, there are actually *three* potential sources for file/line information; I just remembered the third and added that to the document too:

- Call stack entries
- C call site `__FILE__`/`__LINE__`
- SyntaxError source text file/line

SyntaxErrors source text file/line is included in the stack trace, and is also attributed as the error `.fileName`/`.lineNumber`.

The stack trace also reflects this:

```
duk> try { eval('\n\nfoo='); } catch (e) { print(e.stack); print(e.fileName, e.lineNumber); }
SyntaxError: parse error (line 3)
        input:3                                        <-- file/line of source text (SyntaxError)
        duk_js_compiler.c:3612                         <-- __FILE__ / __LINE__ of DUK_ERROR() call site
        eval  native strict directeval preventsyield   <-- innermost activation, eval() function
        global input:1 preventsyield                   <-- second innermost activation, caller of eval()
input 3
```

svaarala:
Internal documents updated, and there's now an implementation for the "no tracebacks" case too. Testcases next.

svaarala:
Alright, test case should now cover the new behavior. It should cover the relevant code paths and stack depth corner cases.

fatcerberus:
I assume the stack trace for a SyntaxError is simulated?  Because I remember a while back they didn't include file/line info at all (I requested that actually, now that I think of it).

fatcerberus:
Regarding the traceback depth, what happens if in the case tracebacks are disabled and the depth limit is exceeded without finding a viable function to blame?  Does Duktape then fall back on the 1.3 behavior?

svaarala:
Hmm - not sure what you mean about the SyntaxError?

```
((o) Duktape 1.3.99 (v1.3.0-299-g7c04241)
duk> try { eval('\n\n\nfoo='); } catch (e) { print(e.stack); }
SyntaxError: parse error (line 4)
	input:4
	duk_js_compiler.c:3655
	eval  native strict directeval preventsyield
	global input:1 preventsyield
```

The stack trace includes the source file/line (input:4), the C call site (duk_js_compiler.c:3655), and the actual call stack.

svaarala:
When tracebacks are disabled and Duktape can't find a function to blame within the depth limit, the `.fileName` and `.lineNumber` will both be `undefined`. This matches how Duktape would behave with tracebacks enabled. In that case, the runtime lookup for `.fileName` and `.lineNumber` would be handled by the inherited accessors which would consult `_Tracedata` and run out of candidates in the (truncated) tracedata.

fatcerberus:
*Is* there a call stack in the case of a syntax error though?  Since it's a compilation error no actual code is being executed (yet).  I assumed that was the cause of SyntaxErrors not including that info in... I think it was v1.1?

svaarala:
I just pasted one above :)

svaarala:
The call stack in a SyntaxError will reflect the code path which requests compilation. In case of `eval()` it's the callstack leading to the eval.

When you compile functions from C code, it's whatever call stack is present when you call `duk_compile()` or its variants.

svaarala:
Syntax errors didn't use to have file/line information (because that's not actually a part of the callstack) but that was added a while ago - but they should've had callstack information pretty much always.

fatcerberus:
That's what I meant by simulated - does Duktape generate a dummy callstack entry in the tracedata, or is there a separate code path to handle file/line lookup for syntax errors?  It doesn't matter for user code either way, I just like to reverse engineer stuff. :)

svaarala:
Ah ok, you just meant how is that handled internally. There's no substitute for the actual source of course, but it's described in the error document:

- https://github.com/svaarala/duktape/blob/cb16e8bc000c2d81e1846e4aacd3727ca253cf61/doc/error-objects.rst#traceback-format-_tracedata

I added an example to that in this pull.

fatcerberus:
I tested this again a bit more thoroughly.   I removed my custom error handler from minisphere code entirely and I still get correct file/line attribution for all `duk_require_xxx()` errors, so this checks out!  I also tested setting a `.fileName` to various C API functions, which had the expected results (this was to ensure minisphere wasn't tampering with the error in some other unforeseen way).

svaarala:
Assert runs caught a case with tracebacks disabled and callstack empty, fixed now :) This is pretty ready to go, I'll give it a read-through once more before merging.

fatcerberus:
What was the failed assert, if you don't mind me asking?

svaarala:
The "no tracebacks" path pushes a file/line to be added to the error object and then writes the properties. The file/line pushing part had separate cases for compilation error, C call site, and scanning the call stack. But if the call stack was empty, nothing got pushed which fouled up the value stack discipline.

PR #462, opened 2015-11-26
Move compiler error (usually SyntaxError) error message handling to error augmentation

svaarala:
Before the change compiler errors (usually SyntaxError) had their message modified in the compiler, with the line number added to the error message. The compiler would do this by catching the error, modifying it, and rethrowing it. The downside of this is that a debugger Throw notify won't see the line number in the error message because the message hasn't been modified on the first throw.

This pull request change the message modification to happen in ordinary error augmentation. This is a slight change in behavior if error augmentation is disabled, because compiler error messages won't then have a line number.

### DIFF
--- a/src/duk_error_augment.c
+++ b/src/duk_error_augment.c
@@ -403,6 +403,43 @@ DUK_LOCAL void duk__add_fileline(duk_hthread *thr, duk_hthread *thr_callstack, c
 #endif  /* !DUK_USE_TRACEBACKS */
 
 /*
+ *  Add line number to a compiler error.
+ */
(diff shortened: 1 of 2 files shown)


fatcerberus:
Is there a reason the line number isn't just added at the original throw site in the compiler?

svaarala:
While it'd be possible to change all the call sites in the compiler, note that the line number is added to *any* error thrown during compilation. That means the call site can be e.g. in a `duk_require_hstring()` called by the compiler, or anywhere else inside Duktape.

PR #472, opened 2015-11-30
Add Makefile.sharedlibrary example

### LINKED ISSUE #94: Make Duktape more Unix distribution friendly
Some possible improvements:

- [x] More source/binary compatibility guarantees. Perhaps guarantee binary compatibility for patch releases.
- [x] Avoid the need to duplicate feature options when compiling an application (#64). Not a blocker if defaults are good enough.
- [x] Document how Duktape is intended to be used in a distribution.

Avoiding macros in API calls would be good for distributions and writing language bindings, but it also has downsides, see #87.

Any other suggestions?

### LINKED ISSUE #300: Add example Makefile for system-wide install in a Unix system
See https://github.com/svaarala/duktape/issues/94#issuecomment-82882138.

svaarala:
Add a minimal example of how to compile Duktape as a shared library and compile "duk" command line executable against the shared library. The example is limited to Unix environments. I'm not sure what else is needed for #94 - perhaps someone more interested in this could comment.

Fixes #300. Fixes #94.

### DIFF
--- a/dist-files/Makefile.sharedlibrary
+++ b/dist-files/Makefile.sharedlibrary
@@ -1,0 +1,71 @@
+#
+#  Example of how to build and install locally as a shared library
+#
+#  Usage:
+#
+#    $ make -f Makefile.sharedlibrary
+#    $ sudo make -f Makefile.sharedlibrary install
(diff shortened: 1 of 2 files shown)


svaarala:
Well, no comments so far so I'll merge this and if issues come up we'll fix them then :)

Issue #108, opened 2015-01-19
Finalization issue, telling when an object has been GC'd


JoshEngebretson:
I have objects with native class instances bound and need to know when the object has been garbage collected and will no longer be referenced in script.

The initial approach has been to use a finalizer.  However, this is problematic as the objects are being rescued outside of the finalizer.  I understand that this can happen, as the guide says "this may happen in special cases even when the object is not rescued by the finalizer".  I am not sure what this means.

I am wondering if there is any reliable way to tell if an object has been GC'd.  (I am using heapptr native side to track the script objects)

Thanks!

sva-p:
The quote relates to multiple execution of a finalizer:

>A finalizer may be called multiple times (this may happen in special cases even when the object is not rescued by the finalizer).

There are some special cases where a finalizer may run twice even when the object was not rescued. This used to happen when an object had been finalized by mark-and-sweep (but not yet wiped) and `duk_destroy_heap()` was called, which then ensured all objects are finalized and ended up calling the finalizer for a second time (IIRC this has been fixed but there may be similar cases).

The fix right now is to make the finalizer capable of detecting it's being executed for the second time (e.g. write a NULL over a native resource pointer or such).

The finalizer is currently the only signal of an object being GC'd - with the catch the Duktape may tell that you twice :)

Fixing the finalizer guarantees to be exact would be nice, I'll add it to the 1.2 release.

svaarala:
The quote relates to multiple execution of a finalizer:

>A finalizer may be called multiple times (this may happen in special cases even when the object is not rescued by the finalizer).

There are some special cases where a finalizer may run twice even when the object was not rescued. This used to happen when an object had been finalized by mark-and-sweep (but not yet wiped) and duk_destroy_heap() was called, which then ensured all objects are finalized and ended up calling the finalizer for a second time (IIRC this has been fixed but there may be similar cases).

The fix right now is to make the finalizer capable of detecting it's being executed for the second time (e.g. write a NULL over a native resource pointer or such).

The finalizer is currently the only signal of an object being GC'd - with the catch the Duktape may tell that you twice :)

Fixing the finalizer guarantees to be exact would be nice, I'll add it to the 1.2 release.

JoshEngebretson:
Hi Sami,

The problem isn't that the finalizer may be called twice, which is pretty easy to catch.  It is that the finalizer is being called while the object is still valid and in use.  So, the native object is gone when it actually shouldn't have been deleted.

[...]

Issue #310, opened 2015-08-29
Finalizer which creates new garbage can cause infinite loop


fatcerberus:
So, remembering how the finalizer torture test had to take pains to avoid generating new refzero entries, I realized this can be an issue in Ecmascript code as well, and came up with this testcase to illustrate it:

```javascript
function Majin(name, power) {
	this.blowUpTheEarth = function() {
		if (power < 1000000000) { print(name + " isn't powerful enough..."); }
		else { print(name + " blew up the Earth!"); }
	}
}
Duktape.fin(Majin.prototype, function() {
	var buu = new Majin("Majin Buu", 1000000000);
	buu.blowUpTheEarth();
});
new Majin("Vegeta", 9000).blowUpTheEarth();
Duktape.gc();
```

The end result: Buu proceeds to blow up the Earth from now until the end of time. :)

This testcase is admittedly a bit contrived (it's half a step away from being deliberate infinite recursion), but it would easily be possible to cause this by accident if the finalizer needed to create other kindsof objects with finalizers, which themselves have finalizers, etc.

fatcerberus:
I can also generate a segfault from this: minisphere performs a longjmp if you click the game window's Close button and cleans up (this was the easiest way I could find to unconditionally back out of JS execution).  Duktape then crashes because the refzero queue wasn't empty when I attempted to shut down (an assert is triggered if those are enabled).

fatcerberus:
I assume this kind of thing is why objects with finalizers in .NET take two GC passes to free... strange, this never occurred to me before now.

svaarala:
That's more or less a "feature": what else could Duktape do other than finalize the new garbage (with a finalizer) user code created?

svaarala:
I found that crash using torture testing yesterday and there's a fix for it in the #287 work branch.

svaarala:
It would be possible to have a sanity limit for the refzero list processing so that it'd only process, say, 10000 objects and then exit, picking up finalizing objects when it next started.

But that wouldn't really solve the problem. For example, if my finalizer creates *two* finalizable objects, you'll run out of memory regardless of how the finalization works. That is the case at least if you *guarantee* finalizers will be called. If you don't, then it's easy to fix, but will be very awkward for users ("so what are the guarantees for finalization again?" :-)

svaarala:
The current refzero/finalizer behavior is "immediate" so that:

- Refzero may cause other objects to have refcount zero, and that whole cascade is processed immediately
[...]

PR #473, opened 2015-11-30
Avoid finalizer rerun unless object is rescued

### LINKED ISSUE #109: Make finalizer guarantees exact
Currently a finalizer may run twice in special cases even when the object is not rescued. Make the guarantees exact:

- If an object is not rescued, its finalizer is executed once.
- If an object is rescued, its finalizer is executed once per rescue cycle.

svaarala:
Change `duk_hobject_run_finalizer()` so that:

- If `DUK_HEAPHDR_FLAG_FINALIZED` is already set, do nothing.
- Set `DUK_HEAPHDR_FLAG_FINALIZED` just before running finalizer.

This should guarantee the semantics described in #109 (= at most one finalizer call per rescue cycle) because:

- The only place where a finalizer is called is `duk_hobject_run_finalizer()`.
- The `FINALIZED` flag is only cleared by an explicit rescue, detected by either mark-and-sweep or refcounting. 

Other changes:

- Heap destruction finalizer behavior changes: allow finalizers to create new finalizable objects whose finalizers are also executed, sanity limit.
- Explicit Proxy check, don't call finalizer on a Proxy object.

Tasks:

- [x] Review changes
- [x] Try to come up with a repro case for finalizer re-run, most obvious cases are already fixed
- [x] Test case for runaway finalizer which spawns 1 or 2 new objects per finalization
- [x] Test case for new finalizer argument
- [x] Ensure "mark-and-sweep running" is checked even when mark-and-sweep not enabled
- [x] Website changes for new guarantee
- [x] Internal doc changes for new guarantee
- [x] Add second parameter to finalizer indicating if rescue is possible or not (true: heap being destroyed, can't rescue; false: normal rescue possible)
- [x] Finalizer documentation changes for new 2nd argument
- [x] Run heap destruction finalizers in a loop to allow any new finalizable objects created by previous finalizers to also be finalized
- [x] Add a sanity limit to the finalization loop
- [x] Releases
- [x] Assert run with mark-and-sweep + refcounting
- [x] Assert run with mark-and-sweep only
- [x] Assert run with refcounting only

Fixes #109. See #108.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1220,6 +1220,20 @@ Planned
   function with a .fileName is found which improves error reporting for e.g.
   "[1,2,3].forEach(null)" (GH-455)
 
+* Provide a stronger finalizer re-entry guarantee than before: a finalizer
+  is called exactly once (at the latest in heap destruction) unless the
+  target object is rescued, in which case the finalizer is called once per
+  "rescue cycle" (GH-473)
+
+* Better finalizer behavior for heap destruction: finalized objects may
+  create new finalizable objects whose finalizers will also be called
+  (GH-473)
+
+* Add a second argument to finalizer calls, a boolean value which is true
+  when the finalizer is called during heap destruction as part of forced
+  finalization; the finalized object cannot be rescued by the finalizer
+  in such cases (GH-473)
+
 * Add a combined duktape.c without #line directives into the dist package,
   as it is a useful alternative in some environments (GH-363)
 
--- a/doc/memory-management.rst
+++ b/doc/memory-management.rst
@@ -1122,21 +1122,23 @@ The following flags in the heap element header are used for controlling
 mark-and-sweep:
 
 * ``DUK_HEAPHDR_FLAG_REACHABLE``:
-  element is reachable through the reachability graph
+  element is reachable through the reachability graph.
 
 * ``DUK_HEAPHDR_FLAG_TEMPROOT``:
   element's reachability has been marked, but its children have not been
-  processed; this is required to limit the C recursion level
+  processed; this is required to limit the C recursion level.
 
 * ``DUK_HEAPHDR_FLAG_FINALIZABLE``:
   element is not reachable after the first marking pass (see algorithm),
   has a finalizer, and the finalizer has not been called in the previous
   mark-and-sweep round; object will be moved to the finalization work
-  list and will be considered (temporarily) a reachability root
+  list and will be considered (temporarily) a reachability root.
 
 * ``DUK_HEAPHDR_FLAG_FINALIZED``:
   element's finalizer has been executed, and if still unreachable, object
-  can be collected
+  can be collected.  The finalizer will not be called again until this
+  flag is cleared; this prevents accidental re-entry of the finalizer
+  until the object is explicitly rescued and this flag cleared.
 
 These are referred to as ``REACHABLE``, ``TEMPROOT``, ``FINALIZABLE``,
 and ``FINALIZED`` below for better readability.  All the flags are clear
--- a/doc/sandboxing.rst
+++ b/doc/sandboxing.rst
@@ -216,7 +216,9 @@ Suggestions for sandboxing:
   reference directly.
 
 * Write finalizers very carefully.  Make minimal assumptions on which
-  thread they run, i.e. which global object they see.
+  thread they run, i.e. which global object they see.  It's also best
+  practice to tolerate re-entry (although Duktape 1.4.0 and above has
+  a guarantee of no re-entry unless object is rescued).
 
 * For sandboxed environments it may be sensible to make all finalizers
   native code so that they can access the necessary thread contexts
--- a/src/duk_heap.h
+++ b/src/duk_heap.h
@@ -19,6 +19,7 @@
 #define DUK_HEAP_FLAG_REFZERO_FREE_RUNNING                     (1 << 2)  /* refcount code is processing refzero list */
 #define DUK_HEAP_FLAG_ERRHANDLER_RUNNING                       (1 << 3)  /* an error handler (user callback to augment/replace error) is running */
 #define DUK_HEAP_FLAG_INTERRUPT_RUNNING                        (1 << 4)  /* executor interrupt running (used to avoid nested interrupts) */
+#define DUK_HEAP_FLAG_FINALIZER_NORESCUE                       (1 << 5)  /* heap destruction ongoing, finalizer rescue no longer possible */
 
 #define DUK__HEAP_HAS_FLAGS(heap,bits)               ((heap)->flags & (bits))
 #define DUK__HEAP_SET_FLAGS(heap,bits)  do { \
@@ -33,18 +34,21 @@
 #define DUK_HEAP_HAS_REFZERO_FREE_RUNNING(heap)            DUK__HEAP_HAS_FLAGS((heap), DUK_HEAP_FLAG_REFZERO_FREE_RUNNING)
 #define DUK_HEAP_HAS_ERRHANDLER_RUNNING(heap)              DUK__HEAP_HAS_FLAGS((heap), DUK_HEAP_FLAG_ERRHANDLER_RUNNING)
 #define DUK_HEAP_HAS_INTERRUPT_RUNNING(heap)               DUK__HEAP_HAS_FLAGS((heap), DUK_HEAP_FLAG_INTERRUPT_RUNNING)
+#define DUK_HEAP_HAS_FINALIZER_NORESCUE(heap)              DUK__HEAP_HAS_FLAGS((heap), DUK_HEAP_FLAG_FINALIZER_NORESCUE)
 
 #define DUK_HEAP_SET_MARKANDSWEEP_RUNNING(heap)            DUK__HEAP_SET_FLAGS((heap), DUK_HEAP_FLAG_MARKANDSWEEP_RUNNING)
 #define DUK_HEAP_SET_MARKANDSWEEP_RECLIMIT_REACHED(heap)   DUK__HEAP_SET_FLAGS((heap), DUK_HEAP_FLAG_MARKANDSWEEP_RECLIMIT_REACHED)
 #define DUK_HEAP_SET_REFZERO_FREE_RUNNING(heap)            DUK__HEAP_SET_FLAGS((heap), DUK_HEAP_FLAG_REFZERO_FREE_RUNNING)
 #define DUK_HEAP_SET_ERRHANDLER_RUNNING(heap)              DUK__HEAP_SET_FLAGS((heap), DUK_HEAP_FLAG_ERRHANDLER_RUNNING)
 #define DUK_HEAP_SET_INTERRUPT_RUNNING(heap)               DUK__HEAP_SET_FLAGS((heap), DUK_HEAP_FLAG_INTERRUPT_RUNNING)
+#define DUK_HEAP_SET_FINALIZER_NORESCUE(heap)              DUK__HEAP_SET_FLAGS((heap), DUK_HEAP_FLAG_FINALIZER_NORESCUE)
 
 #define DUK_HEAP_CLEAR_MARKANDSWEEP_RUNNING(heap)          DUK__HEAP_CLEAR_FLAGS((heap), DUK_HEAP_FLAG_MARKANDSWEEP_RUNNING)
 #define DUK_HEAP_CLEAR_MARKANDSWEEP_RECLIMIT_REACHED(heap) DUK__HEAP_CLEAR_FLAGS((heap), DUK_HEAP_FLAG_MARKANDSWEEP_RECLIMIT_REACHED)
 #define DUK_HEAP_CLEAR_REFZERO_FREE_RUNNING(heap)          DUK__HEAP_CLEAR_FLAGS((heap), DUK_HEAP_FLAG_REFZERO_FREE_RUNNING)
 #define DUK_HEAP_CLEAR_ERRHANDLER_RUNNING(heap)            DUK__HEAP_CLEAR_FLAGS((heap), DUK_HEAP_FLAG_ERRHANDLER_RUNNING)
 #define DUK_HEAP_CLEAR_INTERRUPT_RUNNING(heap)             DUK__HEAP_CLEAR_FLAGS((heap), DUK_HEAP_FLAG_INTERRUPT_RUNNING)
+#define DUK_HEAP_CLEAR_FINALIZER_NORESCUE(heap)            DUK__HEAP_CLEAR_FLAGS((heap), DUK_HEAP_FLAG_FINALIZER_NORESCUE)
 
 /*
  *  Longjmp types, also double as identifying continuation type for a rethrow (in 'finally')
@@ -69,8 +73,9 @@
 
 #define DUK_MS_FLAG_EMERGENCY                (1 << 0)   /* emergency mode: try extra hard */
 #define DUK_MS_FLAG_NO_STRINGTABLE_RESIZE    (1 << 1)   /* don't resize stringtable (but may sweep it); needed during stringtable resize */
-#define DUK_MS_FLAG_NO_FINALIZERS            (1 << 2)   /* don't run finalizers (which may have arbitrary side effects) */
-#define DUK_MS_FLAG_NO_OBJECT_COMPACTION     (1 << 3)   /* don't compact objects; needed during object property allocation resize */
+#define DUK_MS_FLAG_NO_OBJECT_COMPACTION     (1 << 2)   /* don't compact objects; needed during object property allocation resize */
+#define DUK_MS_FLAG_NO_FINALIZERS            (1 << 3)   /* don't run finalizers; leave finalizable objects in finalize_list for next round */
+#define DUK_MS_FLAG_SKIP_FINALIZERS          (1 << 4)   /* don't run finalizers; queue finalizable objects back to heap_allocated */
 
 /*
  *  Thread switching
--- a/src/duk_heap_alloc.c
+++ b/src/duk_heap_alloc.c
@@ -133,7 +133,7 @@ DUK_LOCAL void duk__free_allocated(duk_heap *heap) {
 	}
 }
 
-#ifdef DUK_USE_REFERENCE_COUNTING
+#if defined(DUK_USE_REFERENCE_COUNTING)
 DUK_LOCAL void duk__free_refzero_list(duk_heap *heap) {
 	duk_heaphdr *curr;
 	duk_heaphdr *next;
@@ -149,7 +149,7 @@ DUK_LOCAL void duk__free_refzero_list(duk_heap *heap) {
 }
 #endif
 
-#ifdef DUK_USE_MARK_AND_SWEEP
+#if defined(DUK_USE_MARK_AND_SWEEP)
 DUK_LOCAL void duk__free_markandsweep_finalize_list(duk_heap *heap) {
 	duk_heaphdr *curr;
 	duk_heaphdr *next;
@@ -173,16 +173,17 @@ DUK_LOCAL void duk__free_stringtable(duk_heap *heap) {
 DUK_LOCAL void duk__free_run_finalizers(duk_heap *heap) {
 	duk_hthread *thr;
 	duk_heaphdr *curr;
-#ifdef DUK_USE_DEBUG
-	duk_size_t count_obj = 0;
-#endif
+	duk_uint_t round_no;
+	duk_size_t count_all;
+	duk_size_t count_finalized;
+	duk_size_t curr_limit;
 
 	DUK_ASSERT(heap != NULL);
 	DUK_ASSERT(heap->heap_thread != NULL);
-#ifdef DUK_USE_REFERENCE_COUNTING
+#if defined(DUK_USE_REFERENCE_COUNTING)
 	DUK_ASSERT(heap->refzero_list == NULL);  /* refzero not running -> must be empty */
 #endif
-#ifdef DUK_USE_MARK_AND_SWEEP
+#if defined(DUK_USE_MARK_AND_SWEEP)
 	DUK_ASSERT(heap->finalize_list == NULL);  /* mark-and-sweep not running -> must be empty */
 #endif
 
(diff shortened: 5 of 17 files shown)


fatcerberus:
A question just occurred to me: What happens if a finalizer rescues an object during heap destruction, say by putting it into the global stash?  Will this create an infinite loop or perhaps a memory leak?  Or will it be handled gracefully?

svaarala:
It won't cause any trouble - the finalizer will be executed once, but not twice.

svaarala:
There was a potential issue in the heap destruction finalizer handling: if a refzero situation occurred an object might be queued into the refzero list which was not processed.

Fixed this so that the heap destruction code fakes a "mark-and-sweep running" flag which prevents a mark-and-sweep and also prevents refzero from moving objects around in the heap linked lists (in particular, it prevents an object from being moved from heap_allocated to refzero_list).

This change also means "mark-and-sweep running" flag needs to be checked for in refzero handling even when mark-and-sweep itself is disabled.

fatcerberus:
So that's a potential memory/resource leak for the application then, if the finalizer attempts to rescue the object (instead of freeing the resource) but then doesn't get called again because we're in heap destruction... Might be worth noting that in the documentation.

svaarala:
Agree, I'll describe the limitations for finalization during heap destruction on the website.

svaarala:
omewhat related - IIRC right now any new garbage/objects created during heap destruction won't get their finalizers executed.

That's useful in the sense that heap destruction won't get stuck if a finalizer always creates one or more new finalizable objects. On the other hand it violates the base finalization policy for these new objects. And, there are legitimate reasons to create finalizable garbage in a finalizer, and if the application is correctly written that process will eventually terminate.

I could change the heap destruction so that it'll walk through the `heap_allocated` list running finalizers until there are no more finalizers to run (or perhaps some sanity limit is reached, say 10 passes over the heap?). This would have the upside that new garbage would get their finalizer executed (which may be useful if only a *limited* amount of garbage is created); the downside would be the potential to get forever stuck destroying a heap. There's potential for that anyway though: a finalizer may simply block, for example.

svaarala:
Heh, I think I'm starting to remember why I didn't clean this up before... There's a lot of corner cases to consider, and the resulting behavior should be both reasonable (users probably expect heap destruction to succeed most of the time) and intuitive (finalizers are executed).

I'll try out a few alternatives to see what they look like; I'll update the website text and testcases to match.

svaarala:
The current limitations in heap destruction would be:

- A finalizer is only executed once if the object is rescued.
- If a finalizer creates new objects with finalizers, they will be prepended to `heap_allocated` and their finalizers won't run.

These can be tweaked in various ways, for example:

- Rescuing can be supported by simply looping finalization and forced mark-and-sweep until no finalizable objects remain. Add a sanity limit if it's useful to guarantee the process terminates forcibly; this will necessarily mean some finalizers won't run, and have an associated potential leak.
- New objects created by a finalizer can also be finalized simply by looping the `heap_allocated` list multiple times until no finalizable objects remain. A sanity limit can be applied to this too, with the same caveat that some finalizers won't run.

The main question is what guarantee is the most intuitive, given that there's a tradeoff in ensuring heap destruction eventually finishes vs. skipping some finalizers after a certain sanity round limit.

There's also a limitation when script timeout is active and the timeout `RangeError` is propagating:

- When the timeout error is propagating, any Ecmascript finalizers will start executing but will `RangeError` immediately. If this is a concern, finalizers should be Duktape/C functions.

This limitation could be mitigated by flagging the finalizer call so that the script timeout error wouldn't be thrown while the finalizer runs. This would be nice, but I'm not sure if this can be done cleanly.

Regardless, this needs to be described as a finalizer guarantee limitation (I added it to the website already in this pull, but I'll rewrite the section to be a bit easier to read with subsections).

fatcerberus:
I'm thinking the "infinite loop if a finalizer creates new finalizable garbage" behavior might be that "surprising" - see #310.

In the longer term, for sandboxing reasons we may still want to have some kind of timeout mechanism in place, however.

fatcerberus:
It would be interesting I think to find out how other environments, e.g. .NET, handle runaway finalization cases.

svaarala:
Regarding #310, I really think that behavior is really just like any other resource consumption bug, similar to an application just creating reachable garbage in a busy loop or something. There's not really much to do about those cases. I can write an application which exhausts memory quite easily and there's no general fix for that.

However, things are perhaps a bit different when you're trying to destroy the heap. It might be preferable to be able to destroy the heap (even at the cost of not running some finalizers) rather than hang in such a case. It'd be nice, for example, to be able to use a script timeout to ensure finalizers finish in a reasonable time, even when no script timeout is used in the normal execution mode.

But I guess I'm leaning on not making an exception in the heap destruction: just run finalizers until they're all done. If the application has a bug which creates garbage in an infinite loop, it's a bug that would most likely affect the program in normal operation too.

fatcerberus:
Okay, found this (.NET finalization behavior):
https://msdn.microsoft.com/en-us/library/system.object.finalize(v=vs.110).aspx#How

In short: .NET guarantees the finalizer will only be called once, and calls ALL finalizers on shutdown, regardless of accessibility.  This bit is interesting though:

```
The runtime continues to finalize objects during shutdown only while the number of finalizable objects continues to decrease.
```

Which implies that if the "number of finalizable objects" *increases* instead, it will bail out of finalization?  Hm.

Also, this:
```
Finalize can take almost any action, including resurrecting an object (that is, making the object accessible again) after it has been cleaned up during garbage collection. However, the object can only be resurrected once; Finalize cannot be called on resurrected objects during garbage collection. There is one action that your implementation of Finalize should never take: it should never throw an exception.
```

Rescued objects are not re-finalized (essentially, they are zombies).

svaarala:
Yeah, I was thinking about something like that earlier today. For example:

- When heap destruction starts, run a single round of finalization. Count the number of finalizers called as `n_initial`.
- Set `n_limit` as some multiple of that, e.g.  `4 * n_initial`.
- Run finalization rounds; keep on running as long as finalizers executed is below `n_limit`. Decrease `n_limit` on every round to ensure termination.

But I thought it was maybe a bit too complicated. Perhaps not? :)

fatcerberus:
Despite what I said above, it definitely makes sense to guarantee termination on heap destruction--I think we all know how annoying it is to hit the close box on a window and have it take an age and a day to actually terminate (assuming it ever does).  In most cases--on desktop platforms at least--any leaks will be cleaned up by the OS anyway.

svaarala:
One approach would be to run the finalizer rounds but call user a callback between rounds and let it decide whether or not to keep running or abort and ditch the remaining finalizers. This would require either an API change, a new API call, or it might be implemented as a config macro like script timeout.

svaarala:
Considering the kinds of applications where this matters: if there's a runaway finalizer, then there isn't a clean solution: either the heap destruction hangs or some finalizers won't run. But it doesn't maybe really matter if the source is an application bug.

Intended behavior where a finalizer might spawn new finalizable objects definitely exist. For example, you might be finalizing an object which requires a HTTP request to inform a server of a state change. That server request may have native state which needs finalization. Perhaps that HTTP request finalizer will use another resource which needs finalization. But in the end this chain of finalizers would normally terminate pretty quickly.

So to allow this to happen, the barebones solution where at most N finalization rounds are executed (say N=10) should mostly work. It will work badly if the number of finalizable resources doubles on every round; the number of finalizable garbage would go up 1000x before the sanity kicked in.

The .Net approach is a bit more refined, it doesn't impose a maximum round count, but keeps an eye on the application behaving badly. It'd also be quite easy to implemented roughly along the lines I pointed to above (get an initial limit based on first round and then start reducing that, e.g. by 20% on every round).

fatcerberus:
Duktape also has it a bit harder compared to .NET - rescued objects become eligible for finalization again.  So it's possible to have an accidental runaway just from normal rescue logic.  .NET avoids this issue by treating rescued objects as zombies.

I personally prefer the Duktape rescue behavior, just pointing out that it does make this problem more complicated. :)

svaarala:
Based on the discussion above I think the best approach so far is:

- Support finalizer rescue during heap destruction: provide the same guarantee of "one call per rescue cycle" for rescued objects.
- Allow finalizers to create more finalizable objects during heap destruction. Run their finalizers too.
- Impose a sanity limit to detect runaway finalizers using a shrinking limit for number of finalizers executed per finalization round. This should be lenient enough to not affect ordinary programs.

This would provide "normal" finalizer guarantees for heap destruction, except in the case of runaway finalizers. That'd need to be described in the documentation too, with the specific algorithm maybe left a bit vague so that it can be finetuned without breaking an explicit guarantee.

svaarala:
Regarding rescue behavior, at one point the behavior was to finalize only once. But for native resources that makes rescuing pointless. Native resources are of course not the only reason to rescue an object so that policy is still useful in some other situations.

fatcerberus:
Agree on keeping the specific algorithm vague.  The .NET doc linked above is the same - "finalize only while number of objects decrease" doesn't necessarily specify how soon it bails out, just describes the termination condition.

svaarala:
Hmm, rescuing doesn't actually make much sense in heap destruction: when the heap is destroyed finalizers are executed for what are typically *reachable* objects, so what would "rescuing" mean in that case? Concretely: once the finalizer has been executed for a reachable object, how to determine that it did something which warrants a second finalization?

fatcerberus:
Well, I just meant that a typical rescue (assuming a Duktape/C finalizer) might involve putting the object into the global stash, which could complicate matters during shutdown (is this even memory safe?)

svaarala:
I understand what the finalizer code does in that situation -- but how could the heap destruction code detect that it did something like that if the object is reachable in any case after the finalizer?

svaarala:
Just to clarify, the normal rescue situation is detected by (1) object becoming unreachable, (2) finalizer being executed, (3) object is then detected as being reachable again, i.e. rescued. But if the object is reachable even at step 1, which is the case during heap destruction, what would the criterion be at step 3?

fatcerberus:
Hm, I guess the confusion for me comes from the fact that I consider *everything* to be unreachable by definition during heap destruction.  But from an implementation standpoint, I guess that's not really the case.

This is kind of what I'm getting at though - it's not totally clear what should be done in this case.  A finalizer performing such a rescue assumes the rescue will be honored - i.e. the object will remain live and so doesn't have to clean up yet.  Heap destruction violates this assumption because the object is ultimately going to be collected anyway.

svaarala:
I guess conceptually the criterion could be that "new references" have been created. But there's no easy way to detect that reliably, especially without reference counts (even with reference counts some old references may be deleted and new ones added which would be ambiguous).

svaarala:
Yeah, there's actually a semantic difference between "you're unreachable and about to be collected" and "heap is going down, you're about to be collected no matter what". These could be separate callbacks - or perhaps the finalizer should be given a second argument indicating if the heap is going down.

svaarala:
Regarding "everything is unreachable": if that were the case, there'd be no way to rescue an object from a finalizer. Wherever a reference was written, it'd be by definition unreachable and not count as a rescue :-)

fatcerberus:
That's true.  The main thing is what you said, the finalizer code has no way of knowing whether it can successfully perform a rescue operation.  Adding an additional argument to the finalizer would probably be the cleanest solution without breaking backward compatibility with existing finalizers.

By the way, why do finalizers not have `this` bound to the object being finalized?

svaarala:
I guess we could add that second argument. To be fair, I haven't encountered a need to rescue a native object in practical code, so I'm somewhat hopeful this will be quite rarely needed in practical code.

The finalizer argument convention is more or less an arbitrary choice. I mostly prefer plain callbacks. Looking back it'd be more natural as a method call, i.e. with a `this` binding.

fatcerberus:
The main thing with it was that I have a standard prologue in all my native methods to get the object pointer using the `this` binding, which doesn't work when pasted into a finalizer.  It's not a huge issue, just something that caught me off guard at first.

svaarala:
The current argument can't be removed (at least until 2.0) but it shouldn't harm adding the `this` binding also.

fatcerberus:
So do we have a solution for the finalizer behavior yet? ;)

svaarala:
If there's no reasonable semantics for "rescuing", the best so far would be:

- Provide a second argument to finalizers, `true` if heap is being destroyed and rescuing is not possible, `false` otherwise.
- Don't support rescuing during heap destruction: the semantics are difficult to define and awkward to implement.
- Allow finalizers to create new finalizable objects and run their finalizers too.
- Impose some sanity limit for runaway finalizers.

Does that sound good? I think I'll need to sleep on it before answering myself ;)

fatcerberus:
That sounds good to me.  Sorry for bugging you with corner cases -- even in my own code I probably expend more effort than necessary on accounting for them, largely due to OCD. :)

svaarala:
Quite the contrary, I prefer going through a lot of alternatives and discussion when making changes. Often the issues and best choices are not obvious as in this case. I also reserve the right to change my mind when I figure out something I didn't realize berfore ;)

Anyway, thanks for bringing up issues on this, let's see if anything new pops up or if we arrived at a good approach :)

svaarala:
Prototyping a bit on the runaway finalizer limit: the initial limit is now 4*finalizers_on_first_round + 1000, and decreases 25% on every round. This is pretty arbitrary but allows for the finalizer count to grow initially while still tapering down quite quickly.

If the finalizer spawns one new finalizable object the behavior is like this:

```
((o) Duktape 1.3.99 (v1.3.0-357-g825c9c5-dirty)
duk> function mkObj() { var obj={}; Duktape.fin(obj, function (o,b) { var dummy = mkObj(); }); return obj; }
= undefined
duk> mkObj();
= [object Object]
duk> Cleaning up...
[D] <48940871453402> duk_heap_alloc.c:263 (duk_heap_free): free heap: 0x25e9010
[...]
[D] <48940873208264> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 0 complete, tried to execute 1 finalizers, current limit is 1004
[D] <48940873557571> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 1 complete, tried to execute 1 finalizers, current limit is 753
[D] <48940873898429> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 2 complete, tried to execute 1 finalizers, current limit is 564
[D] <48940874215226> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 3 complete, tried to execute 1 finalizers, current limit is 423
[D] <48940874507072> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 4 complete, tried to execute 1 finalizers, current limit is 317
[D] <48940874791627> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 5 complete, tried to execute 1 finalizers, current limit is 237
[D] <48940875076664> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 6 complete, tried to execute 1 finalizers, current limit is 177
[D] <48940875363507> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 7 complete, tried to execute 1 finalizers, current limit is 132
[D] <48940875648687> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 8 complete, tried to execute 1 finalizers, current limit is 99
[D] <48940876019058> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 9 complete, tried to execute 1 finalizers, current limit is 74
[D] <48940876216740> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 10 complete, tried to execute 1 finalizers, current limit is 55
[D] <48940876401576> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 11 complete, tried to execute 1 finalizers, current limit is 41
[D] <48940876586673> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 12 complete, tried to execute 1 finalizers, current limit is 30
[D] <48940876780209> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 13 complete, tried to execute 1 finalizers, current limit is 22
[D] <48940876968486> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 14 complete, tried to execute 1 finalizers, current limit is 16
[D] <48940877156826> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 15 complete, tried to execute 1 finalizers, current limit is 12
[D] <48940877351445> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 16 complete, tried to execute 1 finalizers, current limit is 9
[D] <48940877540673> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 17 complete, tried to execute 1 finalizers, current limit is 6
[D] <48940877732313> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 18 complete, tried to execute 1 finalizers, current limit is 4
[D] <48940877928315> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 19 complete, tried to execute 1 finalizers, current limit is 3
[D] <48940878121641> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 20 complete, tried to execute 1 finalizers, current limit is 2
[D] <48940878315486> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 21 complete, tried to execute 1 finalizers, current limit is 1
[D] <48940878323502> duk_heap_alloc.c:249 (duk__free_run_finalizers): finalizer count above limit, potentially runaway finalizer; skip remaining finalizers
[...]
```

If the finalizer spawns two new finalizable objects:

```
((o) Duktape 1.3.99 (v1.3.0-357-g825c9c5-dirty)
duk> function mkObj() { var obj={}; Duktape.fin(obj, function (o,b) { var dummy = mkObj(); dummy = mkObj(); }); return obj; }
= undefined
duk> mkObj()
= [object Object]
duk> Cleaning up...
[D] <48318547627776> duk_heap_alloc.c:263 (duk_heap_free): free heap: 0x1ffd010
[...]
[D] <48318551091522> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 0 complete, tried to execute 4 finalizers, current limit is 1016
[D] <48318553365703> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 1 complete, tried to execute 8 finalizers, current limit is 762
[D] <48318557766630> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 2 complete, tried to execute 16 finalizers, current limit is 571
[D] <48318566595678> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 3 complete, tried to execute 32 finalizers, current limit is 428
[D] <48318583355019> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 4 complete, tried to execute 64 finalizers, current limit is 321
[D] <48318603839325> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 5 complete, tried to execute 128 finalizers, current limit is 240
[D] <48318646559103> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 6 complete, tried to execute 256 finalizers, current limit is 180
[D] <48318646607061> duk_heap_alloc.c:249 (duk__free_run_finalizers): finalizer count above limit, potentially runaway finalizer; skip remaining finalizers
[...]
```

If the finalizer is well-behaved the behavior is simple:

```
((o) Duktape 1.3.99 (v1.3.0-357-g825c9c5-dirty)
duk> function mkObj() { var obj={}; Duktape.fin(obj, function (o,b) { print('finalized'); }); return obj; }
= undefined
duk> ref = mkObj();
= [object Object]
duk> Cleaning up...
[D] <49358976247311> duk_heap_alloc.c:263 (duk_heap_free): free heap: 0x21fa010
[...]
finalized
[D] <49358977506081> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 0 complete, tried to execute 1 finalizers, current limit is 1004
[D] <49358977651090> duk_heap_alloc.c:243 (duk__free_run_finalizers): finalizer round 1 complete, tried to execute 0 finalizers, current limit is 753
[...]
```

So I think this approach should work well - the parameters are up to debate of course. And I need to come up with a (fragile, I guess) testcase for ensuring this behavior.

fatcerberus:
Following up on the rescue behavior, I guess it makes the most sense to just say "rescue not allowed during shutdown" and codify that in documentation.  Heap destruction is unconditional and irreversible, so any attempt to rescue an object during that process is really just treading water--eventually the object will have to be collected.  So it's probably most intuitive at that point to just ignore the rescue attempt.

svaarala:
Agree, but to notpick "not allowed" is a bit misleading because there's no problem in trying to create new references to the object and e.g. no error is thrown. It's allowed to do so but there won't be a new call into the finalizer.

Here's what I drafted so far for the second finalizer argument: https://github.com/svaarala/duktape/blob/afee9cfda050452174ad6df279cbdce132a60101/website/guide/finalization.html#L38-L42

And for the heap destruction limitations: https://github.com/svaarala/duktape/blob/afee9cfda050452174ad6df279cbdce132a60101/website/guide/finalization.html#L76-L90

This is just draft text, I'll try to make it clearer and more compact. Wording suggestions welcome :)

fatcerberus:
"Such objects" in the third bullet point could probably be "these objects".  The second bullet seems awkwardly worded, but I'm not sure how to fix it.

svaarala:
Moved the algorithm details to the Wiki: https://github.com/svaarala/duktape-wiki/pull/68

svaarala:
The sanity limit is necessarily a bit arbitrary, but maybe the initial limit should not be based on the number of finalizable objects on the first round, but on the *total* heap object count instead?

This may be a practical issue for some applications. For example, an application might have a data structure holding a number of objects hanging from a single finalizable object. When that object is finalized, it may perform some operation on all the referred objects, and if this operation creates finalizers it's possible to go from having 1 finalizer to e.g. 100 finalizers in a valid application.

If the limit was set as e.g. `2 * heap_object_count` and then decreased 25% per round, it'd take this scenario into account.

fatcerberus:
I think that makes sense, using the total number of objects instead of the finalizer count to set the limit.  Judging by the diff, it seems you already made that change. :)

svaarala:
Yeah, I changed the initializer, this seems like a better approach than the original placeholder.

svaarala:
Here's a testcase which demonstrates the difference between 1.3.0 and 1.4.0:

- https://github.com/svaarala/duktape/blob/1f3c687eb424d8b378a556b395aea40a0ac26e35/tests/ecmascript/test-dev-finalizer-heapdestruct-runonce.js

For 1.3.0:

```
created object 1
object 1 finalized before heap destruction, create new object
created object 2
object 2 finalized before heap destruction, create new object
created object 3
object 2 finalized before heap destruction, create new object
created object 4
object 3 finalized before heap destruction, create new object
created object 5
```

The finalizer for object number 2 gets run twice. Object 5 is created but never finalized. (The testcase's finalizer doesn't know when to stop creating objects in 1.3.0 because there's no 2nd argument to the finalizer.)

For 1.4.0:

```
created object 1
object 1 finalized before heap destruction, create new object
created object 2
object 2 finalized before heap destruction, create new object
created object 3
object 3 finalized during heap destruction, don't create another object
```

Three objects are involved, all are finalized exactly once.

fatcerberus:
So you did manage to get 1.3 to double-finalize after all.  What are the triggering conditions for that?

svaarala:
It's explained in the test case, but basically what you need is an object which has been:

- Detected as unreachable by mark-and-sweep
- Mark-and-sweep has executed the object's finalizer once but the object is not yet collected
- Heap destruction forced finalizer run begins while the object is in the heap, finalized, but not yet collected

This is actually a bit difficult to arrange because heap destruction begins with two rounds of forced normal garbage collection (which doesn't forcibly finalize reachable objects) so the test case works around that by creating new finalizable objects whenever a previous finalizer runs.

svaarala:
This might not be the only double-finalize case though, but I couldn't come up with another one. In any case the explicit FINALIZED flag check in `duk_hobject_run_finalizer()` should cover all such situations. Only an explicit rescue which clears FINALIZED would allow a finalizer to run again.

fatcerberus:
Hm, that could explain the rare crash I see with minisphere on shutdown.  Oddly, the segfault usually happens if I close it during a debugging session.

I'll have to see if that stops after this change.

svaarala:
Do you have a hunch that it's finalizer related? It could also be debugger related I think. The heap destruction code does a forced detach before winding down, but it doesn't currently prevent a new attach!

That's actually something to fix: add a heap flag to indicate heap destruction is ongoing and reject a new debugger attach.

fatcerberus:
Hm, I'll have to review my shutdown code, you could be right about that.

fatcerberus:
Hm, could you look over this code and see if you notice anything wrong?  The code path during shutdown is actually quite convoluted and I'm not actually sure how this works at all without causing an infinite loop:

https://github.com/fatcerberus/minisphere/blob/master/src/debugger.c

Note: `exit_game()` doesn't return; it performs a longjmp() directly into cleanup code.

svaarala:
Hmm, if `exit_game()` longjmps out of Duktape a lot of stuff could go wrong, assuming that longjmp->setjmp call crosses Duktape setjmp catchpoints. Does that happen?

fatcerberus:
The setjmp point for `exit_game()` is established on startup (as the final initialization step before loading the game manifest), so I'm going to say... yes?  This was the cleanest way I could find to bail out of JS execution when the user clicks the window close button as otherwise game loops (which may be arbitrarily deep) will prevent termination.

svaarala:
Ok, then all bets are off :-)

I guess the "normal" solution to an exit like that is to have an exit mechanism on the script side. For example, if the game has a frameloop, that frameloop could detect a requested exit and finish.

If one wanted to do a "hard" exit, it'd be useful to be able to trigger something like script timeout manually. In other words, an error which propagates outwards, unwinding the call stack without being catchable (however, native protected calls will have a chance to clean up).

fatcerberus:
Yeah, for compatibility with older Sphere games, I unfortunately need a "hard exit" solution, as most frameloops look like this:
```javascript
while (condition) {
    UpdateMapEngine();
    RenderMap();
    FlipScreen();
}
```

Anyway, sorry for derailing the thread, as the issue seems to be unrelated to this change.

svaarala:
There are several future work items for the current script timeout - one of them is exposing the abiility to throw an uncatchable error from user code. I'll open an issue for that - there are several implementation options so it's not yet implementation ready.

svaarala:
Found and fixed one more corner case: objects finalized in the last normal GC must be rescued before doing forced finalization to ensure their finalizer gets called once more in forced finalization. This must be done because when the finalizers were called, the second finalizer argument was `false` (not forced finalization) so the finalizer is expecting rescuing to work.

What this needs in practice is one extra mark-and-sweep round which rescues such objects, but doesn't run any new finalizers. Instead, any finalizable objects are queued back `heap_allocated` for forced finalization.

Managed to write a test case for this too so it's regression covered.

(Did I ever mention providing strict finalizer guarantees is tricky? ;)

svaarala:
While there are tricky changes in this pull request, the concrete impact on code footprint is +200 bytes.

fatcerberus:
Could you avoid the two finalizer calls on shutdown if you added that "destroying heap" flag mentioned above?

svaarala:
Hmm, I don't think so. The core of the problem is that any finalized objects with FINALIZED flag set must be processed before forced finalization, but we don't want to finalize further objects when doing so. "Processing" concretely means clearing the FINALIZED flag, and detecting if the object was rescued; if so, move it back to `heap_allocated` so that it gets finalized again (as required by the finalizer guarantees).

svaarala:
Anyway, the pushed solution should work (and does work for the test case) - it's pretty simple in the end: mark-and-sweep will just handle rescue normally, but any new finalizable objects will be just ignored and moved back to `heap_allocated` for forced finalization.

fatcerberus:
So just to clarify: if in the situation described above, the object *isn't* rescued during mark-and-sweep, the finalizer won't be called again, correct?

svaarala:
Yes, the finalizer guarantees will hold: the second argument will be `false` on the last normal GC run to indicate rescue is possible; if the object is not rescued, no more finalizer calls will happen. If the object is rescued, one more finalizer call will happen, this time with the finalizer second argument being `true`.

fatcerberus:
Ah, so it's just an extra rescue cycle, that's no problem then.

svaarala:
I'd leave out the "extra" because the rescue is up to the finalizer :)

svaarala:
I think this is ready to be merged now - I'll merge in a bit.

fatcerberus:
Regarding the dangling heapptr pushed onto the stack in this test:
https://github.com/svaarala/duktape/commit/1e8499012adf43634ccc011aafb5d2b8bfce5bbc

How does this not segfault the minute `duk_push_heapptr()` is called?

svaarala:
The object referenced by `ptr` is intentionally in a reference loop so it can't be collected by refcounting (even when it's not reachable). When `duk_push_heapptr()` happens, the object becomes reachable again and Duktape is none the wiser.

This is not a "supported" process: anything triggering a mark-and-sweep before `duk_push_heapptr()` lead to memory unsafe behavior. It works in the test case well enough to cover a case which is otherwise difficult (if not impossible) to reproduce.

svaarala:
Sorry, there's an explicit GC in there before the `duk_push_heapptr()`, so it really is quite an unsafe test. It's the only way I found to exercise this particular finalizer re-entry case, but it's maybe too fragile to be in the test set; I'll disable it.

fatcerberus:
In theory that actually should be safe now that I look at it again - it takes two GC cycles to collect an object with a finalizer, you're just tricking Duktape into making the already-finalized but unfreed object reachable again.

svaarala:
Yeah, that's what intended to happen. But there are quite a few assumptions in there so I disabled the test (but left the test code inside an #if 0 block for manual testing in case this ever turns up).

svaarala:
Aiee!

fatcerberus:
What happened, forgot to merge the changes? :)

svaarala:
Took a leap of faith that the merge improves finalizer behavior and doesn't introduce new corner cases :)

fatcerberus:
If nothing else it'd be interesting to see if this fixes #108.

PR #479, opened 2015-12-05
Fix debugger Eval command 'this' binding handling

### LINKED ISSUE #478: Debugger Eval command does not see correct `this` binding
Evaluating `this` with a Debug Eval seems to return the global object instead of the current function's correct `this` binding.  For example, after setting a breakpoint on this line:

```javascript
this.aic.definePhases([ 3000, 2000, 1000, 500 ], 50);
```

Performing an Eval of `this.aic` produces `undefined`, which is obviously incorrect (the code runs fine after resume).  After looking a little deeper into it, this bug is present in 1.3.0 (and prior) as well.

svaarala:
Fixes #478.

Just a placeholder for an eventual fix now.

### DIFF
--- a/src/duk_bi_global.c
+++ b/src/duk_bi_global.c
@@ -430,8 +430,8 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 	duk_bool_t this_to_global = 1;
 	duk_small_uint_t comp_flags;
 	duk_int_t level = -2;
+	duk_idx_t entry_top;
 
-	DUK_ASSERT_TOP(ctx, 1);
 	DUK_ASSERT(thr->callstack_top >= 1);  /* at least this function exists */
 	DUK_ASSERT(((thr->callstack + thr->callstack_top - 1)->flags & DUK_ACT_FLAG_DIRECT_EVAL) == 0 || /* indirect eval */
 	           (thr->callstack_top >= 2));  /* if direct eval, calling activation must exist */
@@ -444,6 +444,7 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 	 *  activation doesn't exist, call must be indirect.
 	 */
 
+	entry_top = duk_get_top(ctx);
 	h = duk_get_hstring(ctx, 0);
 	if (!h) {
 		return 1;  /* return arg as-is */
@@ -496,7 +497,7 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 	/* E5 Section 10.4.2 */
 	DUK_ASSERT(thr->callstack_top >= 1);
 	act = thr->callstack + thr->callstack_top - 1;  /* this function */
-	if (act->flags & DUK_ACT_FLAG_DIRECT_EVAL) {
+	if (act->flags & DUK_ACT_FLAG_DIRECT_EVAL || entry_top >= 2) {  /* FIXME: hack */
 		DUK_ASSERT(thr->callstack_top >= 2);
 		act = thr->callstack + thr->callstack_top + level;  /* caller */
 		if (act->lex_env == NULL) {
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -1376,6 +1376,12 @@ DUK_LOCAL void duk__debug_handle_eval(duk_hthread *thr, duk_heap *heap) {
 	 * as seen by, e.g. Duktape.act() will be the same regardless.
 	 */
 
+	/* FIXME: the global eval() code will interpret this eval call as an
+	 * indirect eval so 'this' will be mapped to the global object.  Rework
+	 * the eval handling to avoid this.  It'd probably be best if we didn't
+	 * actually call eval() here.
+	 */
+
 	/* nargs == 2 so we can pass a callstack level to eval(). */
 	duk_push_c_function(ctx, duk_bi_global_object_eval, 2 /*nargs*/);
 	duk_push_undefined(ctx);  /* 'this' binding shouldn't matter here */


svaarala:
That hack test I just committed seems to fix the issue for me, so the root cause seems correctly understood (debugger Eval gets interpreted as an indirect eval() and not a direct eval).

Now for an actual fix :)

fatcerberus:
If it's actually interpreted as indirect, how come I can still Eval local variables (declared with `var`)?

svaarala:
Good question, I was assuming master didn't work properly - but in fact, when I compile from master my debugger Eval sees the "this" binding in this example (for example Eval'ing `this.name` yields "my object"):

```js
function test() {
    var local = 123;
    print(this.name);
    print(local);
}

try {
    var obj = { name: 'my object', test: test };
    obj.test();
} catch (e) {
    print(e.stack || e);
}
```

There's an explicit direct eval flag in the debugger call site: https://github.com/svaarala/duktape/blob/master/src/duk_debugger.c#L1407-L1414

So this is not obviously broken after all.

svaarala:
Tracing that sequence a bit:

- The "direct call to eval" is set on the topmost activation to the global eval() handler. This call is made by the debugger code and the flag is set here: https://github.com/svaarala/duktape/blob/master/src/duk_debugger.c#L1407-L1414. The flag is only set if the function at "level" is an Ecmascript function, as Duktape/C functions don't have a visible scope.
- The eval() code eventually checks "direct call to eval" from the callstack top: https://github.com/svaarala/duktape/blob/master/src/duk_bi_global.c#L498-L513. When set, `this_to_global = 0`.
- Finally, the `this` binding is initialized as: https://github.com/svaarala/duktape/blob/master/src/duk_bi_global.c#L571-L578. Assuming `this_to_global == 0` still here, I don't see why this wouldn't work.

@fatcerberus Could you add a few debug prints in this path with the code that fails for you? Could it be that the index you're giving to Eval actually points to an activation with a native function (such as `.forEach()` or something similar)?

fatcerberus:
The index my debugger gives to Eval is the exact same index given to, e.g. GetLocals (i.e. they are both driven by the same variable), and the latter indicates the correct scope, so that seems unlikely.  That said, I'll see if I can come up with a minimal repro case outside the context of my game (which is quite a large codebase).

fatcerberus:
Let me know what happens with this (minisphere debugger at least produces `undefined` for the Eval):

```javascript
function pig() {
	this.weight = 812;
	this.eat = function() {  // set breakpoint here, then Eval "this.weight"
		print("a pig just came into existence");
	}
}
function cow() {
	this.weight = 1000;
	this.eat = function() {
		new pig().eat();
		print("the cow ate the pig");
	}
}
function ape() {
	this.weight = 1208;
	this.eat = function() {
		new cow().eat();
		print("the ape ate the cow");
	}
}

new ape().eat();
```

fatcerberus:
Also try setting breakpoints at various other locations (for example inside the `.eat()` functions)

svaarala:
Eval'ing `this.weight` in each of the eat() functions seems to work correctly when I test it. Hmm.

fatcerberus:
What happens if you pass an explicit callstack index of -1?

svaarala:
Still seems to work:

```
{"notify":"_Connected","args":["1 10399 v1.3.0-376-g3588381-dirty duk command built from Duktape repo"]}
{"notify":"Status","command":1,"args":[1,"_test.js","global",1,0]}
{"request":"AddBreak","args":["_test.js", 4]}
{"reply":true,"args":[0]}
{"request":"Resume"}
{"reply":true,"args":[]}
{"notify":"Status","command":1,"args":[0,"_test.js","global",1,0]}
{"notify":"Status","command":1,"args":[1,"_test.js","",4,0]}
{"request":"Eval","args":["typeof this"]}
{"reply":true,"args":[0,"object"]}
{"request":"Eval","args":["this.weight"]}
{"reply":true,"args":[0,{"type":"number","data":"4089600000000000"}]}
{"request":"Eval","args":["this.weight", -1]}
{"reply":true,"args":[0,{"type":"number","data":"4089600000000000"}]}
```

fatcerberus:
Hm, this is baffling.  I'll have to investigate some more, looks like.

fatcerberus:
Wait, why is the weight coming back as `4089600000000000`?  Did you modify the test?

fatcerberus:
Hm, it seems the minisphere debugger implements an Eval as, in actuality, a nested `eval()`:
https://github.com/fatcerberus/minisphere/blob/master/plugin/Utility/DuktapeClient.cs#L233-L246

Could this be the cause of the issue?

svaarala:
That would cause the behavior if I understood correctly: the Eval'd code calls an anonymous function and the `this` binding is lost in the process:

```
((o) Duktape 1.3.99 (v1.3.0-376-g3588381-dirty)
duk> function test() { (function () { print(this.name); })(); }
= undefined
duk> test.call({ name: 'fake this' })
undefined
= undefined
```

If you want to keep that structure, you can pass the `this` onwards to the anonymous function:

```
duk> function test() { (function () { print(this.name); }).call(this); }
= undefined
duk> test.call({ name: 'fake this' })
fake this
= undefined
```

svaarala:
`4089600000000000` is the IEEE double encoding for the value:

```
duk> var buf = new ArrayBuffer(Duktape.dec('hex', '4089600000000000'))
= undefined
duk> var dv = new DataView(buf)
= undefined
duk> dv.getFloat64(0)
= 812
```

fatcerberus:
Okay, I see it now, the whole thing is wrapped in a self-executing function expression, how did I miss that?  Tacking on `.call(this)` to the expression indeed fixes the issue and I can now access the proper `this` binding at any callstack level.

So... yeah, it seems I reported a bug that doesn't exist and wasted both of our time. :(

svaarala:
No worries :)

PR #483, opened 2015-12-07
Accept buffer objects as duk_push_buffer_object() base values

svaarala:
- [ ] Accept any buffer object as a duk_push_buffer_object() base value; new buffer object will a previous offset into account so that it effectively creates a slice against the existing buffer object.
- [ ] When creating typed arrays and the base value is `ArrayBuffer`, should the result `.buffer` be the argument ArrayBuffer, or a fresh ArrayBuffer with the same backing buffer? In other words, should ArrayBuffer be the only special argument which doesn't trigger creation of a fresh ArrayBuffer?
- [ ] API testcase updates: basic cases
- [ ] API testcase updates: range/wrap checks
- [ ] Releases


fatcerberus:
Regarding the second bullet point, the most intuitive and least "surprising" solution to me would be to look at how the TypedArray constructors behave for `ArrayBuffer` argument and mimic that.

svaarala:
Yeah, I agree on that principle. A little more detail though:

TypedArray constructors only accept ArrayBuffer buffer objects. There's a separate question of whether they should accept *plain* buffers too and treat them like ArrayBuffers which seems intuitive and useful. There's naturally no standard what to do when the argument is a Node.js Buffer or Duktape.Buffer; perhaps treat roughly the same, sharing the same underlying buffer, but a new ArrayBuffer would need to be created for `.buffer` property of the result.

TypedArray constructors treat typed array views like arrays and just use them as value initializers. In effect the buffer is copied instead of sharing the same underlying buffer - and when the element types differ (e.g. Uint16Array created from Uint8Array input) there's an actual conversion. A new backing buffer is always created. This is not necessarily very intuitive, at least for C code, if the input is a Uint8Array (instead of ArrayBuffer).

To fully mimic that, `duk_push_buffer_object()` would need to:

- Accept a plain buffer as before. If the result is a typed array view, create a fresh ArrayBuffer and set it as `.buffer` of the result. The underlying plain buffer is shared.
- Accept an `ArrayBuffer` (respecting the slice offset; this is zero for standard ArrayBuffers). If the result is a typed array view, use the same ArrayBuffer instance as `.buffer`. The underlying plain buffer is shared.
- Perhaps accept Node.js Buffers and Duktape.Buffers similarly to `ArrayBuffer`: respect the slice offset, and shared the underlying plain buffer. I guess in this case a fresh ArrayBuffer needs to be created for `.buffer` to respect typing. Another option would be to treat these too like array arguments and make a buffer copy.
- Typed arrays should be treated as *initializers*: instead of using the same underlying buffer, create a new buffer and populate it through the typed array view. Create a fresh ArrayBuffer and use it for `.buffer`.

As a C programmer I'd like the ability to e.g. create a Uint16Array from an Uint8Array so that the underlying buffer is shared. The above behavior wouldn't do that, so I'd need some API to either get the plain buffer of the Uint8Array out first, or read the `.buffer` property to get the ArrayBuffer out.

In other words, the other API design viewpoint would be that `duk_push_buffer_object()` would be a low level operation which would *always* share the same underlying buffer and just create a different wrapper for it. This would be much more intuitive to me as a C programmer: it'd be odd if sometimes the underlying buffer is copied and sometimes shared.

svaarala:
Ah, and also, if `duk_push_buffer_object()` treated a typed array argument like an array initializer, it should probably also just accept plain array-like values like the actual typed array constructors?

This is how that works in Ecmascript:

```
duk> u16 = new Uint16Array([1, 2, 3])
= [object Uint16Array]
duk> Duktape.enc('jx', u16)
= |010002000300|
```

fatcerberus:
Hm, you're right, this gets complicated.  Maybe the current solution is best?  As in, keep it low-level and only accept a plain buffer as base.  Then you can just provide an API to get the underlying buffer for a given buffer object (if such a thing doesn't already exist).

svaarala:
I guess my views are biased towards C code working with byte buffers in a low level manner. That's why `duk_push_buffer_object()` arguments are byte based (byte offset and byte length) instead of element offset and element length.

svaarala:
The C API could also mimic `.subarray()` for typed views (instead of typed array constructor) which shares the same underlying buffer. The `subarray()` call doesn't convert between view types of course so it's not a perfect analogy.

svaarala:
Yeah, one option would be to just accept plain buffers, and provide a separate API to "take apart" a buffer object:

- Get the plain underlying buffer
- Get the element size/type
- Get the byte offset and byte length (and/or element offset and element length) so user code can make slice calculations on its own

svaarala:
I'll leave this pull open for discussion, and won't merge anything until there's been time to figure out the best approach.

svaarala:
I guess accepting ArrayBuffer, Node.js Buffer, and Duktape.Buffer in addition to plain buffers would be relatively easy to understand. But even with them one has to deal with existing view/slice offsets and lengths in the argument bufferobjects; a plain object never has a view/slice offset/length. It's not difficult to do that, but it may not be obvious what should happen in the API.

One downside of accepting just plain buffers is that if the user has an ArrayBuffer and wants to create a view into it, he'd first need to get the plain buffer and then create the view. As a result a *new* ArrayBuffer would always be created (instead of reusing an existing one). So accepting at least ArrayBuffers might be nice in this respect. (This is actually one reason I was on the fence whether `duk_push_buffer_object()` should set the `.buffer` property at all, because that's not a "low level" operation.)

fatcerberus:
Is the `.buffer` property not required by spec?

svaarala:
It is required, but the caller could set that explicitly if it wanted to. Not all code will need that property if the buffer is just used internally. This is what Duktape internals do too: a helper creates that buffer object and the `.buffer` property is then set using normal property mechanics.

Just to clarify that ArrayBuffer issue a bit, re: a second ArrayBuffer being created:

- You want to create a new `Uint8Array` using  `duk_push_buffer_object()` from an existing `ArrayBuffer` (call it A).
- Because `duk_push_buffer_object()` does not accept an ArrayBuffer, you need to get the plain buffer out first.
- When calling `duk_push_buffer_object()` to create a Uint8Array, use the plain buffer as a base.
- The resulting Uint8Array will have a `.buffer` property pointing to a new ArrayBuffer (call it B) with the same underlying plain buffer.

This is quite weird because `var u8 = new Uint8Array(myArrayBuffer)` doesn't create a new ArrayBuffer, but sets `u8.buffer` to `myArrayBuffer`.

fatcerberus:
Well what I meant was, if you `duk_push_buffer_object()` a typed array onto the stack (without setting `.buffer`) and that object then escapes into client Ecmascript code, you now have a non-compliant object floating around, which may break JS code that assumes it will be present because the standard guarantees it.

Although maybe it's the same situation as Duktape/C functions, which don't have a `.prototype` by default? (that caught me off-guard a while ago, hehe)  In other words, you could say because the typed array originated from native code as opposed to one of the ES constructors, the same guarantees don't apply.

svaarala:
Sure, that'd then be possible. But that's what I mean by having `duk_push_buffer_object()` be a "low level" primitive: you'd use it as part of building an object.

Or to look at it another way, I find it a bit annoying there's *no* way to create a Uint8Array without also creating an ArrayBuffer object and a `.buffer` property.

It's a compromise either way, of course.

fatcerberus:
On the other hand, having taken a look at the specified behavior, it's not mandated that `.buffer` be a valid ArrayBuffer, just that it's "the value of the [[ViewedArrayBuffer]] internal slot".  So it seems entirely acceptable that a typed array originating in C code doesn't have a meaningful value for this.

svaarala:
But doesn't that mean `.buffer` must be the same value that the view operations back to? So it can't be some arbitrary value?

For Duktape that happens through the view and the ArrayBuffer sharing the same underlying buffer though - view accesses go to the plain buffer directly without going through the ArrayBuffer, but the end result is mostly the same (at least as far as standard behavior is concerned). So Duktape doesn't need that ArrayBuffer because it's not the lowest level buffer type :)

fatcerberus:
Yeah, that's what I meant - Duktape is still technically in compliance not setting that value since the view in this case isn't backed by an ArrayBuffer but rather by a lower-level type.

svaarala:
I guess that depends on how you look at it :) As far as the Ecmascript code is concerned the ArrayBuffer *is* backing the view, because they ultimately share the same buffer, and that mapping cannot change dynamically.

fatcerberus:
I guess the main point of contention here is that, as far as the Ecmascript spec is concerned ArrayBuffer is the lowest-level type while in Duktape that role is taken by plain buffers.  So it gets confusing when trying to reconcile the two...

Oh well, I'm sure you'll come up with a workable solution. :)

svaarala:
It's maybe more accurate to say that Duktape implements ArrayBuffer using an underlying plain buffer, which is of course neither required nor prohibited by the spec (likewise for other implementation options). But, the plain buffers are then visible in the C API which makes it less obvious what's the best behavior - should the API look like typedarrays or like Duktape buffers?

PR #484, opened 2015-12-08
Accept a plain buffer value in typedarray constructor with same behavior as Duktape.Buffer

svaarala:
A `Duktape.Buffer` is accepted as a value initializer, so that a new ArrayBuffer is created and the view is created on top of that ArrayBuffer. The ArrayBuffer is initialized through the view treating the constructor argument as an array-like value initializer:

```
duk> var buf = new Duktape.Buffer('foobar');
= undefined
duk> Duktape.enc('jx', new Uint16Array(buf));
= |66006f006f00620061007200|
```

Extend this behavior to plain buffers too; in Duktape 1.3.0 plain buffer arguments are treated the same as string arguments: coerced to an integer to give an element length for the typed array.

Tasks:

- [x] Finalize source change for plain buffer in typedarray constructor
- [x] Also fix buffer zeroing for ArrayBuffer and typedarray constructors when DUK_USE_ZERO_BUFFER_DATA is not set
- [x] Test case update
- [x] Internal docs
- [x] Wiki buffer howto for typedarray argument and zeroing guarantees - https://github.com/svaarala/duktape-wiki/pull/71


svaarala:
This pull makes `Duktape.Buffer` and plain buffer values behave the same for typed array constructors, i.e. both are treated like array-like initializers.

I don't think that behavior is necessarily the most intuitive. I think it'd be more intuitive if all actual buffer values (plain buffers, Duktape.Buffer, Node.js Buffer, and ArrayBuffer) would be treated the same -- i.e., share the underlying buffer and create a view into it. The only weird thing is that since `.buffer` of the result must be an ArrayBuffer, a fresh ArrayBuffer needs to be created for the other types (but with that ArrayBuffer backing to the same underlying buffer).

There wasn't much time to figure out the best buffer-type mixing behaviors for 1.3.0 so I intentionally left these behaviors marked as subject to change. So this can still be changed if the above behavior makes more sense.

PR #485, opened 2015-12-08
Update footprints on website index page for 1.4.0

svaarala:


fatcerberus:
Looks like footprint actually *decreased* for 1.4.0.

svaarala:
It might be that I compiled the "duk" command line instead of a barebones "hello" program last time. I've been meaning to automate generating those numbers but haven't gotten around to it yet. In any case they are just rough estimates, ARM binaries are a bit smaller, etc.

PR #490, opened 2015-12-10
Fix incorrect debugger eval-related assert

svaarala:
Debugger Eval caused an assertion failure with asserts enabled. The failing assert was `DUK_ASSERT_TOP(ctx, 1)` in the `duk_bi_global.c` `eval()` function. That assert is incorrect when the debugger calls `eval()`; in that case top will be 2.


fatcerberus:
I thought we addressed this during the implementation of #412, weird.

PR #494, opened 2015-12-19
Don't "require" debugger support for debugger opts

svaarala:
DUK_USE_DEBUGGER_THROW_NOTIFY default is "true".  If the option "requires" DUK_USE_DEBUGGER_SUPPORT (which defaults to "false") the option configuraiton would be inconsistent and rejected by default.

Remove the "required" metadata for debugger options: the options don't really require debugger support.  Instead, if debugger support is disabled, the debugger config options are just ignored without causing a build error.


fatcerberus:
I'm getting throw notifys again, so looks like this fixes the issue.

PR #497, opened 2015-12-20
Add volatile to 'idx_func' to fix NetSurf issue

svaarala:
Technically the volatile shouldn't be needed: idx_func is written before setjmp() and only read afterwards so its value should be well defined.


fatcerberus:
How does this impact performance, if at all?

svaarala:
Don't know until it's measured - I don't think it could be reliably measured in microbenchmarks though.

svaarala:
Impact on code footprint is +60 bytes so it does have a measurable effect on the emitted code.

Quick performance test on call handling shows no measurable difference.

PR #499, opened 2015-12-23
Support C++ exceptions instead of setjmp/longjmp to allow automatic destructors etc

svaarala:
- [x] Add DUK_USE_CPP_EXCEPTIONS which causes Duktape to use C++ exceptions for internal long control transfers (instead of setjmp/longjmp or their variants)
- [x] Use `throw 1` (dummy value) and `catch (int dummy)` initially for the internal exception
- [x] Check what's the most portable value to throw; ideally it'd be a class user code cannot throw => Use a custom `duk_internal_exception` class which doesn't inherit from any base class, so that it's unlikely user code would accidentally catch it (except with `catch (...)`)
- [x] Should `duk_internal_exception` inherit from `std::exception`? => Initially no; user code catching `std::exception` won't accidentally catch Duktape internal exception
- [x] Finalize handling for user thrown exceptions which propagate to Duktape (this is initially not allowed because such exceptions would bypass protected calls); they should be converted to e.g. API error
- [x] `const char *what() const throw() { return "something"; }` => Not needed because Duktape internal exception does not inherit from `std::exception`
- [x] Sanity check: check for C++ compiler when DUK_USE_CPP_EXCEPTIONS is enabled
- [x] Rebase to master once native-call-outer-setjmp has been merged
- [x] Example with C++ exceptions, add to dist
- [ ] Website updates
- [ ] Wiki update - https://github.com/svaarala/duktape-wiki/issues/72
- [x] Releases entry


ddevienne:
Thanks! Waiting for 1.4 release to test the amalgamation.

svaarala:
Here's a prebuilt distributable in case you want to try it sooner :)

http://duktape.org/snapshots/duktape-1.3.99-20160106184924-v1.3.0-431-g5c863ed.tar.xz

Issue #504, opened 2015-12-31
Whether to support Chinese encoding


zero-ui:
My js code:
我() //Call success here
function 我(){
	print("ok")
}

My C code:
		duk_push_global_object(Global_JsContext);
		duk_push_string(Global_JsContext, "我");
		rc=duk_get_prop(Global_JsContext, -2);//Call failed here Return FALSE
		if (rc)
		{
			duk_pcall(Global_JsContext, 0);
		}

Encoding is the source file ANSI

svaarala:
Strings given through the Duktape C API must be in UTF-8 encoding - it is used throughout the API and also internally.

HouQiming:
I think an issue is that in such a situation, currently the error message looks like:
"SyntaxError: char decode failed (line 0)"

Then one has to consult the documentation to find out that the correct encoding is UTF-8. Changing it to "UTF-8 decode failed" could make it more clear to us Windows programmers working with CJK locale. When we see "char decode failed", the first thought tends to be "Did I feed a GB2312 file into a Shift-JIS system?", and the second thought tends to be "Was that western programmer expecting ISO-8859-1?". That kind of issues are much more common than UTF-8 stuff around here.

fatcerberus:
To be fair, JavaScript is UTF-8 by spec, but I agree the error message could probably be clearer.

HouQiming:
I don't think JavaScript specs cover the *source code*, or rather, C API encoding. In any case, there are a lot of Shift-JIS / GB2312 scripts on the web and browsers have no problem loading them. That usually works even if you don't specify the encoding (browsers have automatic encoding detection). I'd bet quite a few people would expect duktape to behave the same way unless otherwise notified in the error message.

zero-ui:
Thank you very much,Convert the parameters to UTF8,problem solving.

monkey2000:
Then please close this issue. @zero-ui

svaarala:
Supporting a lot of different encodings would have quite a high impact on code footprint unfortunately: there are many code pages and a bunch of multibyte encodings, endianness variants, etc. So that's why Duktape supports only UTF-8 both internally and in its C API. While it's not ideal, it's not very difficult to use e.g. [iconv](https://www.gnu.org/software/libiconv/) to convert source code into UTF-8 before compiling it.

If you look at the encodings supported by iconv (all presumably used by *someone*), I hope it's clear why I think sticking to UTF-8 only is the best approach :)

svaarala:
[...]

PR #506, opened 2016-01-01
Use clearer error message for source decode error

svaarala:
See #504.


svaarala:
Old error:

```
SyntaxError: char decode failed (line 0)
	/tmp/test.js:0
	duk_lexer.c:314
```

New error:

```
SyntaxError: utf8 decode failed (line 0)
	/tmp/test.js:0
	duk_lexer.c:314
```

fatcerberus:
I like this, it would have saved me a lot of confusion back when I first encountered this error due to extended characters (Win-1252).

PR #508, opened 2016-01-02
Improvements to C++ exception long control transfers

svaarala:
- [ ] Finalize handling for error throw when no catcher is active: for setjmp/longjmp the situation is detected based on jmpbuf pointer being NULL, but since there's no jmpbuf for C++ exceptions a manual state flag may be needed? As things are now, C++ uncaught exception handling is invoked instead of the Duktape fatal error handler.
- [ ] Try to avoid duplication in user exception catch blocks
- [ ] Add some public API exposed define (`DUK_CPP_EXCEPTIONS`) for checking that C++ exceptions are supported, so that user code can #ifdef check and avoid silent errors if Duktape options are incorrect for code depending on C++ automatic cleanups
- [ ] Testcase coverage: either custom test target, or add C++ support for runtests (and run them with DUK_USE_CPP_EXCEPTIONS)


fatcerberus:
For what it's worth, I think invoking the C++ uncaught error handling is fine and probably what a C++ programmer would expect.

Duktape's fatal handler is needed for C because there is no native exception support, so that gives the program one last opportunity to clean up.  C++ programs on the other hand would hopefully be using RAII so a custom handler isn't needed.

svaarala:
Hmm, if we left uncaught exception handling that way, there'd still be other cases where the fatal error handler is called (e.g. assertion errors and possibly some other conditions in the future). Those can't be easily converted into C++ uncaught error because there may be catch points along the way (both Duktape and user).

fatcerberus:
Ah, I see what you mean - I was thinking of actual uncaught JS errors, but there are indeed other cases where the fatal error handler may be called even with active catch points.  And many C++ developers, especially game developers, prefer to turn exception handling off, so we'd still need the handler anyway.

In any case I do think uncaught "userland" errors (i.e. `throw new Error()` with no catch point) would probably be most intuitive to allow to fall all the way through the stack and let the C++ runtime pick up the pieces.  Otherwise we run the risk of accidentally circumventing RAII (I think?)  Sorry if that was confusing, I know what I'm trying to say, but bad at articulating it. :(

svaarala:
I see what you mean, even though Duktape didn't have a catch point there might be one in a Duktape/C function in the (otherwise "unprotected") call stack.

One possibility would be to convert all fatal errors into a specific exception type and propagate that outwards. Duktape's catch points would just ignore it, and user code would be required to do the same. It'd OK to do a catch-all-and-rethrow for cleanups and such.

The problem with this is that it's very easy to violate, resume calling into Duktape, and get into a lot of trouble :)

svaarala:
Feedback on the current C++ exception mechanism would be welcome now that it's been available for a while. If there are no functional issue I'll probably postpone these cleanups to reduce 1.5.0 release scope.

mike-lischke:
Is this still relevant now that we have PR #1915?

svaarala:
There are still some improvements left but one is fixed (updated checkboxes).

Issue #74, opened 2014-11-13
Rework regular expression parsing/compiling to be more lenient


svaarala:
Duktape's regular expression parser/compiler conforms strictly to the E5/E5.1 regular expression syntax, and rejects many regular expressions that are leniently allowed by other engines. It's confusing to users that other engines parse several regexp forms that Duktape (technically correctly) rejects as a SyntaxError.

Examples of technically invalid regexps that work with most engines:
- `/^\_/`: underscore escape is not allowed, underscore is an IdentifierPart and identity escapes are not allowed for identifier part characters.
- `/\$/`: regexp escape is not allowed, dollar sign is an IdentifierPart. Duktape already has a fix to allow a dollar escape because it's so common.
- `/\\{/`: unescaped brace is not allowed (unless it is part of a quantifier like `/x{5}/`).
- `/[\2]/`: non-zero decimal escape is not allowed (E5 Section 15.10.2.19) but accepted by other engines.

Because these regexps are quite widely used, Duktape should probably parse regexps a bit more leniently. This change is non-trivial because lenient regexp parsing needs backtracking which doesn't fit easily into the current regexp parser/compiler code. For instance, when parsing `/{foo}/`, Duktape will see the left curly brace and think it's parsing a quantifier. When it fails to parse, it currently throws a SyntaxError. Instead, it should rewind and assume `{` is meant literally, i.e. parse the regexp as `/\{foo\}/`.

Related issues with notes on various failing regexps:
- #19
- #39
- #69
- #86
- #142
- #425
- #643

Tasks:
- [ ] Figure out what leniency is necessary for "real world" compatibility with other engines. The main goal is to avoid unnecessarily surprising users.
- [ ] Make the necessary changes to Duktape for more lenient parsing. The changes should be made with code footprint in mind (not growing the footprint if possible). There should be an option to restore compliance, if possible. There's probably no need to have an option to enable/disable individual leniency cases, just a lenient/compliant option.
- [ ] Update testcases.
- [ ] Update internal documentation (`doc/regexp.rst`) to reflect the changes in the regexp algorithm.
- [ ] Update external documentation. Document the non-compliant regexp forms allowed.

svaarala:
By the way, ES6 has optional support for literal braces in Section B.1.4:

- http://www.ecma-international.org/ecma-262/6.0/#sec-regular-expressions-patterns

This doesn't extend to unquoted square brackets etc though.

fatcerberus:
[...]

PR #513, opened 2016-01-05
Loosen rules for RegExp { } literals

### LINKED ISSUE #142: Duktape chokes on literal braces in RegExp
The following code works fine in all other JS engines, but Duktape chokes on it ("invalid quantifer"):

```
function FormatString(s) {
    var a = arguments, i = 1;
    return s.replace(/{\?}/g, function() { return a[i++]; });
}

Abort(FormatString("Hello {?} how are you?", "John"));
```

crazyjul:
Check the regex for valid quantifier, if none found, suppose it's a character literal.

If } is encountered outside of quantifier parsing, it is treated as a literal

This MR fixes #142 and partial for #74

### DIFF
--- a/src/duk_lexer.c
+++ b/src/duk_lexer.c
@@ -1609,26 +1609,34 @@ DUK_INTERNAL void duk_lexer_parse_re_token(duk_lexer_ctx *lex_ctx, duk_re_token 
 		duk_uint_fast32_t val1 = 0;
 		duk_uint_fast32_t val2 = DUK_RE_QUANTIFIER_INFINITE;
 		duk_small_int_t digits = 0;
+
+		/*
+		 *  Store lexer position, restoring if quantifier is invalid
+		 */
+
+#ifdef DUK_USE_NONSTD_REGEXP_BRACES
+		duk_lexer_point lex_pt;
+		DUK_LEXER_GETPOINT(lex_ctx, &lex_pt);
+#endif
+
 		for (;;) {
-			DUK__ADVANCECHARS(lex_ctx, 1);  /* eat '{' on entry */
+			DUK__ADVANCECHARS(lex_ctx, 1); /* eat '{' on entry */
 			x = DUK__L0();
 			if (DUK__ISDIGIT(x)) {
+				digits++;
+				val1 = val1 * 10 + (duk_uint_fast32_t) duk__hexval(lex_ctx, x);
+			} else if (x == ',') {
 				if (digits >= DUK__MAX_RE_QUANT_DIGITS) {
 					DUK_ERROR(lex_ctx->thr, DUK_ERR_SYNTAX_ERROR,
 					          "invalid regexp quantifier (too many digits)");
 				}
-				digits++;
-				val1 = val1 * 10 + (duk_uint_fast32_t) duk__hexval(lex_ctx, x);
-			} else if (x == ',') {
 				if (val2 != DUK_RE_QUANTIFIER_INFINITE) {
-					DUK_ERROR(lex_ctx->thr, DUK_ERR_SYNTAX_ERROR,
-					          "invalid regexp quantifier (double comma)");
+					goto invalid_quantifier;
 				}
-				if (DUK__L1() == '}') {
+				if ( DUK__L1() == '}') {
 					/* form: { DecimalDigits , }, val1 = min count */
 					if (digits == 0) {
-						DUK_ERROR(lex_ctx->thr, DUK_ERR_SYNTAX_ERROR,
-						          "invalid regexp quantifier (missing digits)");
+						goto invalid_quantifier;
 					}
 					out_token->qmin = val1;
 					out_token->qmax = DUK_RE_QUANTIFIER_INFINITE;
@@ -1639,9 +1647,12 @@ DUK_INTERNAL void duk_lexer_parse_re_token(duk_lexer_ctx *lex_ctx, duk_re_token 
 				val1 = 0;
 				digits = 0;  /* not strictly necessary because of lookahead '}' above */
 			} else if (x == '}') {
-				if (digits == 0) {
+				if (digits >= DUK__MAX_RE_QUANT_DIGITS) {
 					DUK_ERROR(lex_ctx->thr, DUK_ERR_SYNTAX_ERROR,
-					          "invalid regexp quantifier (missing digits)");
+						"invalid regexp quantifier (too many digits)");
+				}
+				if (digits == 0) {
+					goto invalid_quantifier;
 				}
 				if (val2 != DUK_RE_QUANTIFIER_INFINITE) {
 					/* val2 = min count, val1 = max count */
@@ -1655,8 +1666,7 @@ DUK_INTERNAL void duk_lexer_parse_re_token(duk_lexer_ctx *lex_ctx, duk_re_token 
 				DUK__ADVANCECHARS(lex_ctx, 1);
 				break;
 			} else {
-				DUK_ERROR(lex_ctx->thr, DUK_ERR_SYNTAX_ERROR,
-				          "invalid regexp quantifier (unknown char)");
+				goto invalid_quantifier;
 			}
 		}
 		if (DUK__L0() == '?') {
@@ -1666,6 +1676,18 @@ DUK_INTERNAL void duk_lexer_parse_re_token(duk_lexer_ctx *lex_ctx, duk_re_token 
 			out_token->greedy = 1;
 		}
 		advtok = DUK__ADVTOK(0, DUK_RETOK_QUANTIFIER);
+		break;
+invalid_quantifier:
+#ifdef DUK_USE_NONSTD_REGEXP_BRACES
+
+		/* Failed to match the quantifier, restore lexer */
+		DUK_LEXER_SETPOINT(lex_ctx, &lex_pt);
+		advtok = DUK__ADVTOK(1, DUK_RETOK_ATOM_CHAR);
+		out_token->num = '{';
+#else
+		DUK_ERROR(lex_ctx->thr, DUK_ERR_SYNTAX_ERROR,
+				"invalid regexp quantifier");
+#endif
 		break;
 	}
 	case '.': {
@@ -1806,8 +1828,10 @@ DUK_INTERNAL void duk_lexer_parse_re_token(duk_lexer_ctx *lex_ctx, duk_re_token 
 		}
 		break;
 	}
-	case ']':
-	case '}': {
+#ifndef DUK_USE_NONSTD_REGEXP_BRACES
+	case '}':
+#endif
+	case ']': {
 		/* Although these could be parsed as PatternCharacters unambiguously (here),
 		 * E5 Section 15.10.1 grammar explicitly forbids these as PatternCharacters.
 		 */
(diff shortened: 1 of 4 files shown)


svaarala:
Hmm, I might be reading the diff wrong but it seems that `{x}` would emit a matcher for `x\}` (not `/\{x\}/`, i.e. no "atom char" matcher is emitted for the left curly brace?

crazyjul:
I don't think so. The error at line 1658 was replaced by the same code as in default: case, that is : 
```
advtok = DUK__ADVTOK(1, DUK_RETOK_ATOM_CHAR);
out_token->num = x;
```

So the { is still added as a literal token, treated as a PatternCharacter.

I can add test for these if needed

svaarala:
```
duk> x = /{foo}/
= /{foo}/
duk> x.test('{foo}')
= false
duk> x.test('foo}')
= false
duk> x.test('ffoo}')
= true
```

Ok, so the index is handled correctly but the "bailout" clause shouldn't match the offending character but rather the left curly - and then resume parsing the rest.

src/duk_lexer.c
```diff
@@ -1652,12 +1653,15 @@ DUK_INTERNAL void duk_lexer_parse_re_token(duk_lexer_ctx *lex_ctx, duk_re_token
...
 					out_token->qmax = val1;
 				}
-				DUK__ADVANCECHARS(lex_ctx, 1);
+				DUK__ADVANCECHARS(lex_ctx, index + 1);
 				break;
 			} else {
-				DUK_ERROR(lex_ctx->thr, DUK_ERR_SYNTAX_ERROR,
-				          "invalid regexp quantifier (unknown char)");
+				advtok = DUK__ADVTOK(1, DUK_RETOK_ATOM_CHAR);
+				out_token->num = x;
```
svaarala: This should emit a matcher for the left curly instead?

svaarala:
A test case illustrating the new behavior (and its limitations if any) will definitely be needed - you can add one in this pull, or I can add one in master if this gets merged.

Another thing needed in master will be a compliance option to turn the behavior off. Again, you can add it in the pull or I can add it in master.

svaarala:
One limitation in the current approach is that if the quantifier looks like a number but turns out not to be, the regexp parse will fail with an "invalid quantifier" still so that the fix for #142 is partial, e.g.:

```
duk> x = /{1111111111111111111111111/
SyntaxError: invalid regexp quantifier (too many digits) (line 1)
	input:1
	duk_lexer.c:1619
```

This may be fixable in the pull. I don't think this will be a practical limitation though, so documenting it in a testcase would be good enough for now in my view.

svaarala:
Hmm - if I understand the new approach correctly, it postpones the `DUK__ADVANCECHARS` in the regexp quantifier parsing loop until we know whether it's a quantifier or something else, and `DUK_LOOKUP(lex_ctx, index + N)` lookups are used in the loop instead. However, the guaranteed window (DUK_LEXER_WINDOW_SIZE) is only 6 codepoints long so unless I'm mistaken the algorithm may overrun the window in some cases?

One simple fix for this would be to increase the window so it's always enough to scan valid quantifier expressions (~20 codepoints or so) but this has some performance impact.

Another alternative (which wouldn't need to change the window) would be to change the algorithm so that it snapshots the lexer position before the loop and then advances in the quantifier parsing loop as in master. When the quantifier turns out to be invalid, the lexer position can be rewound and we can resume parsing. Rewinding the lexer should have no side effects - it'd be more difficult if we needed to "unemit" bytecode but luckily that's not necessary here.

crazyjul:
Yes, that was the idea. I agree the proposed change will make this code
better. I wasn't aware the lexer could be rewound.

For the too many digits, does a test suffice for you?

The bug is a copy/paste issue :-(. I should have paid more attention.

Finally, to disable the option, should I create a preprocessor option?

I'll setup a linux machine and add the tests and the fixes

Thanks for the comments

svaarala:
As a bit of background, I haven't added support for loose parsing of RegExps because there are a lot of individual issues, and most of the cases require actual backtracking (including bytecode backtracking). I'm also not very enthusiastic about "dead end partial fixes", i.e. changes which fix a subset of cases but can't be developed into fully working solutions -- one can easily get lost in improving the fixes step by step but never quite reaching a fully working end result. That's why the main plan for RegExps is to add support for (optionally) hooking up an external regexp engine which will perform better and will be able to support a lot of non-standard RegExps; these are very difficult to achieve in the built-in engine which is size and portability optimized.

Even so, this RegExp curly brace issue comes up pretty often, and because the change is potentially relatively simple, it'd be nice to merge this into the built-in minimal regexp engine.

svaarala:
As for the concrete questions / issues:

IMO it'd be OK to fail for now for e.g. `/{1111111111/`; the test case should illustrate this current limitation. I think this limitation should actually be fixable, and I can give it a go once the basic change is in master. The memory safety fix will be necessary before merging, of course.

As for the config option, I can handle that bit in master (if you prefer), but the change is basically:

- Add a config metadata file, e.g. `config/config-options/DUK_USE_NONSTD_REGEXP_BRACES.yaml`, the option should be enabled by default. The YAML metadata is documented in `config/config-options/DUK_USE_EXAMPLE.yaml` but comparing to existing files is also useful.
- Modify the changes in the lexer so that the new/old behavior are behind `#if defined(DUK_USE_NONSTD_REGEXP_BRACES)`. Test both cases by modifying `duk_config.h`.

Once the remaining issues are resolved, I'll be happy to merge this pull, and I think a lot of users will appreciate it :-)

crazyjul:
I addressed all the remarks you made. Looks cleaner. From here, it would be easy to convert any quantifier parsing failure to literal {, but I believe it will be hard to debug and understand what's going on in the regex.

Code compiles without warnings and pass the tests

fatcerberus:
This is what I use as a reference point:
https://regex101.com/

I tested this pull, works pretty well.  There are two issues I noticed:
* If the regex starts with what would otherwise be a valid quantifier, e.g. `/{2}/`, Duktape throws an error "quantifier without preceding atom".  Regex101 also rejects this, so it almost certainly isn't an issue in practice.
* `/{2,,}/` is rejected with a "double comma in quantifier" error.  Regex101 accepts this as a literal match in all modes.

crazyjul:
As all those issues are already managed as error, it would be pretty easy to make them work as well. That was the subject of my comment after the last commit.

Again, I believe we should chose wisely which one to (dis)allow, as it may be harder to debug strange behaviors for people not used to this discrepancy.

fatcerberus:
That's why I like regex101, it serves as a useful point of reference for real-world regex engine behavior.

svaarala:
Like @crazyjul said I think the remaining issues should be relatively easy to fix. I'm OK with either fixing them in separate pulls or working them out in this pull.

crazyjul:
Ok, small vote here. Here is the list of error easily bypassed : 

- invalid regexp quantifier (too many digits)
- invalid regexp quantifier (double comma)
- invalid regexp quantifier (missing digits)
- invalid regexp quantifier (missing digits)

Which one do you think should trigger literal brace?

fatcerberus:
Double comma:
This is a tough call.  Regex101 treats it as literal, but it might be more useful to keep it as a syntax error, like it is now.  It'd be easy to make a typo and add an extra comma, and it would be confusing to have your quantifier suddenly turn into a literal.  I vote to keep the error.

Too many digits:
Probably best to leave this alone too.  Ideally something like `/a{11111111111111111111111111111111111}/` would trigger the error, but `/a{11111111111111111111111111111111111maggie}/` would not (i.e. literal curlies).  But since it's probably not easy to differentiate between the two cases, I vote again to leave the error.

Missing digits:
I'm not sure what triggers "missing digits", this would have to be clarified.

crazyjul:
In fact, the 'too many digits' is easy to change to get that behavior. Right now, it checks the digit count for each digit. Simply checking it at closure would be enough, any alien character would cancel the parsing.

For Missing digits, it triggers on ```{}``` and ```{,}```.

fatcerberus:
I vote literal semantics for missing digits then.

svaarala:
Because this behavior is out-of-specification, it'd probably be quite useful to compare against e.g. V8. Even if the behavior was different, it's good to know where the differences lie.

crazyjul:
node 4.2.2 has this behavior : 
Missing digits leads to literal
No limits for digits

fatcerberus:
Some quick experiments in Node.js v5.3.0:

```javascript
> /{}/.exec("a pig ate everyone at 8:12 {}")
[ '{}', index: 27, input: 'a pig ate everyone at 8:12 {}' ]
> /{1}/.exec("a pig ate everyone at 8:12 {1}")
SyntaxError: Invalid regular expression: /{1}/: Nothing to repeat
    at new RegExp (native)
    at repl:1:1
    at REPLServer.defaultEval (repl.js:252:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:417:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:210:10)
    at REPLServer.Interface._line (readline.js:549:8)
> /a{}/.exec("a{}")
[ 'a{}', index: 0, input: 'a{}' ]
> /a{111111a}/.exec("a{111111a}")
[ 'a{111111a}', index: 0, input: 'a{111111a}' ]
> /{1,,}/.exec("{1,,}")
[ '{1,,}', index: 0, input: '{1,,}' ]
```

tl;dr: Curlies are literal unless they form a valid quantifier atom.

crazyjul:
Here it is, I've implemented failure support as discussed here. A test is also attached to it.

To avoid code bloater with ```#ifdef #else #end```I've created a macro that either report error or jump to literal brace code. I was not sure about the coding standard here, so comment if it needs to change

svaarala:
@crazyjul Great, thanks! :) I'll look into this pull once I've got 1.4.0 finished, probably weekend or early next week.

Just as a quick comment on the error macro: I think the error handling could be simplified so that the quantifier syntax errors would just share a single error message string ("invalid regexp quantifier") when the non-standard curly support is disabled. That would make it unnecessary to use an error macro and the goto target site could just throw the shared error if non-standard curly support is disabled. The original error message precision is somewhat of an overkill compared to error message verbosity in other places.

But in case the macro remains: there are no `#define`s inside functions now as a convention, but it's probably OK here as it's quite local. I'd like an `#undef` after the macro is no longer used. Finally, internal macros are always prefixed `DUK__` (double underscore).

svaarala:
Stuff to clean up in master after this is merged:

- RegExp limitations documentation on website and internal regexp documentation
- Check if `make emscriptentest` and friends now work without a fixup step (there were other issues apart from RegExp syntax errors as well so might not work)

fatcerberus:
@crazyjul You might want to squash some of the commits.  The exception being the testcase(s), for which the convention from what I've seen is usually to commit them separately.

svaarala:
@fatcerberus Yeah, I've tried to keep code commits separate from other commits because it makes e.g. cherry picking easier. But it's not that important because one usually knows when committing something that will be cherry picked to a fix release later so that only that commit can go in separately. What I don't like too much is a bunch of fixup commits in the master :-)

crazyjul:
Ok, I'll do it this afternoon

crazyjul:
I've squashed all the commits to keep a single one.

svaarala:
@crazyjul I just noticed that you added your name to the wrong author list in AUTHORS.rst, could you please open a pull to add yourself to the first list in AUTHORS.rst (which is for code contributions)? Thanks! :)

Issue #306, opened 2015-08-28
Cleanups for 1.4.0 release


svaarala:
- [x] Move into merging through pull requests
- [x] Check fastint downgrade handling of yield values
- [x] Update index page LOC, lowmemory startup size, code footprint (with buffer support) - #485
- [x] Accept plain buffer as a typed array constructor initializer? - #484
- [x] Buffer zeroing behavior with DUK_USE_ZERO_BUFFER_DATA unset? - #484
- [x] Typing for `duk_push_buffer_object()`, accept a buffer object as a base value, see discussion https://github.com/svaarala/duktape/issues/190 - #483, not sure if will be merged but tracked in that pull
- [x] -Wcast-qual warnings for DDDEBUG build
- [x] Buffer behavior changes? => No more for 1.4.0
- [x] Release notes: buffer behavior changes
- [x] MSVC /W4 fixes - #514

svaarala:
Overall I'm planning the 1.4.0 release to be smaller than 1.3.0 which had a lot of changes. A good release interval would be around 3 months I think. Duktape 1.1 -> 1.2 took 3 months, 1.2 -> 1.3 took 5 months.

I'll be doing some release grooming and planning next week. I'll add a Wiki page for a roadmap overview so that this is a bit more transparent.

svaarala:
Had to prune a few more things from the 1.4.0 release - there's a nice chunk of performance work in the release already, and I'd like to finish the call handling improvements. Moved other performance stuff like reimplementing the string table, further dispatch optimization etc, to 1.5.0.

Rough goal would be to finish actual feature development by end of November so that the release could be early December.

fatcerberus:
Hm, I think I'm done with debugger features for the moment.  The only other one would be the heap walking for 1.5.0, but that's a little out of my league right now. :)

fatcerberus:
1.4.0 is coming together pretty quickly it seems. :)

svaarala:
I've been sidetracked this week. I'll probably need to postpone some performance / call handling related work to ensure the release gets done before the holidays. There's also a few large pending merges I'd like to start working into master, but only after the 1.4.0 release is done :)

svaarala:
Code commits for 1.4.0 are now done, I'll be re-running the performance tests and doing compilation tests etc. If something is broken in master, now is a good time to bring it up :)

fatcerberus:
Code freeze, nice.

[...]

PR #514, opened 2016-01-05
Fix some MSVC /W4 cast warnings

svaarala:
https://github.com/svaarala/duktape/issues/306#issuecomment-168883454

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1258,6 +1258,8 @@ Planned
 
 * Add a fastint downgrade check for yield/resume values (GH-482)
 
+* Improve error message for source code UTF-8 decode error (GH-504, GH-506)
+
 * Fix a segfault (and assertion error) caused by compiler intermediate value
   handling bug; the bug was triggered when a temporary register was required
   by the compiler, but an existing "intermediate value" referred to a const
@@ -1295,8 +1297,6 @@ Planned
 
 * Fix base64 decode reject for whitespace between padding characters
   (e.g. "Zm= =") (GH-465)
-
-* Improve error message for source code UTF-8 decode error (GH-504, GH-506)
 
 * Internal performance improvement: rework RETURN opcode handling to avoid
   longjmp() calls, improving performance slightly on ordinary platforms and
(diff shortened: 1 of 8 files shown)


fatcerberus:
Going to test this now, I'll let you know what happens. :)

svaarala:
Thanks :)

fatcerberus:
Yep, that only warning left is:

```
1>duk_api_stack.c(3948): warning C4127: conditional expression is constant
```

I'm impressed.  Most people ignore `/W4` completely because it generates a lot of spurious warnings, but you went the extra mile anyway.  Good work as always!

fatcerberus:
I forgot I have LTCG (i.e. full-program optimization) on, it turns out a bunch more warnings are generated at link time.  Almost every single one is for "unreachable code", but there's one that sticks out:

```
1>d:\src\minisphere\msvc\duk_js_compiler.c(2188): warning C4701: potentially uninitialized local variable 'd3' used
```

svaarala:
Hmm, that uninitialized case should never occur: `d3` is only referenced if `accept != 0` and the switch-case ensures `d3` is set or `accept` is set to 0. But I'll see if I can restructure it.

fatcerberus:
I've noticed similar issues with MSVC's static analyzer where it claims a variable is uninitialized when it really isn't.  Just figured I'd point it out anyway. :)

fatcerberus:
Regarding the lightfunc varargs fix: Could that have potentially caused issues in practice?  You seem to have given it special attention.

svaarala:
At first I thought it might have caused a bug but it didn't - it got cast from `duk_idx_t` to `duk_uint16_t`, and then into `duk_int16_t`, so that the varargs marker (-1) ended up correct.

fatcerberus:
Did you figure out what to do (if anything) about the `d3` uninitialized var warning?

svaarala:
I got lazy and decided to change it later. It'd be trivial for example to just initialize d3 to `0.0` but that generates dead code which would be nice to avoid. I'll look into it in master after 1.4.0 release.

PR #519, opened 2016-01-08
Fix src-noline DUK_SINGLE_FILE define

svaarala:


fatcerberus:
What was the bug here?  I glanced at the diff, but it's not clear to me what was fixed.

svaarala:
`DUK_SINGLE_FILE` was not defined for the combined `duktape.c` with no `#line` directives. It mainly affects symbol visibility macros.

PR #520, opened 2016-01-09
Miscellaneous v1.4.0 release preparations

svaarala:


fatcerberus:
Regarding the compiler warnings: Should the MSVC `/W4` warning(s) be documented as well?

svaarala:
Good point, I'll add them too.

PR #522, opened 2016-01-10
Post-release cleanups after 1.4.0 release

svaarala:


fatcerberus:
Nice, the performance improvements in 1.4 are worth the upgrade alone

fatcerberus:
Not to mention my debugger enhancements. :)

JoshEngebretson:
I haven't had any time to check our Emscripten performance on Duktape 1.4 in depth, though eyeballing it is *definitely* smoother, need some better performance metrics from here! :)

svaarala:
@JoshEngebretson I think the next step for Emscripten performance would be to figure out how well the executor switch-case structure compiles and performs, as that seemed to be slower (relatively speaking) than other native code such as array algorithms etc.

svaarala:
@fatcerberus Yeah, thanks for all that debugger work :)

PR #531, opened 2016-01-12
Fix heap_allocated reinsertion for refzero finalizer rescued objects

svaarala:
Fix heap_allocated reinsertion bug which affected refzero finalizer rescued objects. See test case in: https://github.com/oleavr/dukrecycle/ (a bug test case will be adapted in this branch based on that).

- [x] Fix the re-queueing of refzero finalizer rescued objects
- [x] Adapt test case from dukrecycle, test that bug appears before and gets fixes by the diff
- [x] Add linked list consistency assertion helper which checks that prev/next and curr are linked correctly
- [x] Add more heap_allocated consistency asserts to linked list manipulation points, ensure assertion tests pass with them in place and that asserts would have caught the original bug
- [x] Releases entry
- [x] Schedule backports to 1.4.1, 1.3.3, and 1.2.6

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -64,6 +64,7 @@ bugs, provided ideas, etc; roughly in order of appearance):
 * Michael Drake (https://github.com/tlsa)
 * https://github.com/chris-y
 * Laurent Zubiaur (https://github.com/lzubiaur)
+* Ole André Vadla Ravnås (https://github.com/oleavr)
 
 If you are accidentally missing from this list, send me an e-mail
 (``sami.vaarala@iki.fi``) and I'll fix the omission.
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1367,6 +1367,11 @@ Planned
 1.5.0 (XXXX-XX-XX)
 ------------------
 
+* Fix potentially memory unsafe behavior when a refcount-triggered finalizer
+  function rescues an object; the memory unsafe behavior doesn't happen
+  immediately which makes the cause of the unsafe behavior difficult to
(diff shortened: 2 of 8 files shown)


svaarala:
@joshengebretson If the diagnosis/fix here is correct, this may be related to the finalizer issues you had a way back :) Here's hoping!

oleavr:
This fixes the stability issues we were seeing in the [frida-gum](https://github.com/frida/frida-gum) test-suite! :heart:

svaarala:
@oleavr Big thanks for coming up with the repro case!

fatcerberus:
@svaarala What is the bug here, exactly?  I'm curious because I haven't yet seen the finalizer issues Josh had with Atomic in minisphere.

svaarala:
The diff really says it all, but in short, object rescued during refcount-triggered finalization need to be queued into the heap_allocated list. That queueing operation failed to update the "prev" pointer of the previous list head which resulted in an inconsistent heap_allocated list. With a suitable mix of later operations (e.g. finalizing again) this could trigger a read/write-after-free situation. It's a pretty significant bug and could have various weird side effects -- but only if an object is rescued during refcount-triggered finalization.

fatcerberus:
Could that potentially lead to a segfault?

svaarala:
I think so, though the problems were captured using ASAN/valgrind in practice.

svaarala:
Added the test case and checked that the test case catches the original issue.

Also added linked list consistency asserts where `duk_heaphdr` linked list is manipulated. Verified that test cases pass with the asserts in place, and that the asserts would have caught the original bug.

PR #536, opened 2016-01-13
Convert make_dist.sh to Python make_dist.py

svaarala:
Convert make_dist.sh straightforwardly into Python make_dist.py and change dist to use that instead. This reduces dependency on bash for the dist process.


fatcerberus:
Would this allow Duktape to be built natively in Windows, i.e. without Cygwin?

svaarala:
The goal is that the dist process would run on Python 2 (and maybe 2 + 3 later) without using bash commands, etc. This pull should make that easier but I haven't yet tested if the whole dist now works without Cygwin.

svaarala:
Test using python.org Python 2.7.11 (https://www.python.org/downloads/release/python-2711/). Add Python 2.7 to PATH manually. Results:

- Creating `genconfig_metadata.tar.gz` fails because of missing "tar" command; use tarfile: https://docs.python.org/2/library/tarfile.html (comes with Windows Python)
- `import yaml` fails, need to `pip install PyYAML`

fatcerberus:
Nice, I'll have to test this.  Anyone would call me crazy for saying this, but I much prefer the simplicity of cmd.exe over bash.

svaarala:
Added two improvements:

- If the YAML import fails, make_dist.py will suggest the command to install it ("pip install PyYAML" for Windows)
- Use the tarfile module to avoid depending on "tar" being in PATH

With these fixes, what works for me:

- Install Python 2.7.x (python.org), add it to PATH
- `pip install PyYAML`
- `python util/make_dist.py`

With these I get a working `dist/` directory as a result. I'll move more of the Makefile functionality into the Python script later.

fatcerberus:
Will `make_dist.py` detect changes to source files?  Because currently with the makefile I have to do `make clean dist`, otherwise it assumes everything is still up to date.

svaarala:
One more needed change is to update the README to describe how to "make dist" on Linux/UNIX, OSX, and Windows without using the Makefile.

Other stuff in the Makefile will still be Linux oriented.

svaarala:
`make_dist.py` will always make a fresh dist directory, it doesn't track changed files etc, and will delete an existing `dist/` if present.

svaarala:
Alright, the README is now revised a little to describe how to dist in Linux, OSX, and Windows.

svaarala:
@fatcerberus Let me know if the latest push works for you. The README is rather short on the dist steps but it should be quite straightforward now.

I intentionally left out a few details from the README:

- If you use `--create-spdx` to create an SPDX license `pip install rdflib` is needed. I don't want to add that to the README because it's only really needed for official dists (and very few users will care about it).
- The "init JS" code used internally by Duktape won't be minified by default. The impact of this is a few hundred bytes of code footprint so I don't want to clutter the README with this. Official dist packages will use a minifier and that option is available in `make_dist.py` too if you really want to use it.

But with the defaults the only dependencies should be Python + PyYAML.

fatcerberus:
Yeah, I don't really care about a couple hundred bytes when my primary target platform is x64. :)

I vaguely remember having to install rdflib in Cygwin to get the dist working.  Was it not optional before?

svaarala:
It was required earlier yes, but now only if you specify `--create-spdx`.

fatcerberus:
Alright, I'm going to try this now.

fatcerberus:
Everything seems to be working!

One thing: Is it normal that I get a `duktape.c` with Unix line endings (LF) instead of CRLF?  I'm using Windows versions of both Git and Python, so the LF line endings seem odd.

svaarala:
The line endings (LF) are intentionally the same as for an official dist package (and the repo itself). This could of course be changed but then a Windows dist package would be different from a Linux one.

fatcerberus:
Ah, okay.  MSVC doesn't complain either way and any line endings in minisphere are normalized to LF anyway when I commit (satisfying my OCD :) ), so it's not an issue.  I was just wondering if that was normal.

fatcerberus:
I will say having `pip` to install Python packages is nice, I didn't get that luxury in Cygwin for whatever reason.  I don't even feel the PyYAML dependency now since it's just one command and about 20 seconds to install it.

fatcerberus:
Thanks, this will make it easier for me to work on Duktape. :)

Issue #530, opened 2016-01-12
Duktape doesn't appear to respect pointer alignment


davidchisnall:
I'm trying to use DukTape on a very weird architecture[1] which, among other things, has strict pointer alignment requirements.  The `DUK_HOBJECT_A_GET_VALUE_PTR` macro apparently expands to:

```
(&((duk_tval *) ( ((((obj)))->props) + ((((obj)))->e_size) * (sizeof(duk_hstring *) + sizeof(duk_propvalue) + sizeof(duk_uint8_t)) + ((4 - (((((obj)))->e_size))) & 0x03) ))[(arr_idx)])
```

For us (with `sizeof(void*) == 32`), this evaluates to 196, which is not a multiple of 32.  Is there an easy change possible to fix this?

[1] Pointers are 256 bits, containing a 64-bit virtual address, base, bounds and permissions.  `sizeof(void*)` is 32.  Similarly, `sizeof(intptr_t)` is 32 and permits arithmetic on the pointer value.  Pointers may only be materialised

svaarala:
That's indeed an interesting platform :) Duktape assumes currently that align by 8 would be sufficient for pointers too. Because memory areas contain a mix of tagged values and pointers, some alignment assumptions are unfortunately necessary (or at least quite difficult to avoid).

I can look into supporting stronger alignment guarantees, I'll have to give it some thought :)

Is there any way (say an emulator) where I could reproduce this?

davidchisnall:
We have a QEMU port that is almost ready: https://github.com/CTSRD-CHERI/qemu.  Currently it requires a lot of assembly, but we should have pre-packaged images available soon.  I found the `DUK_USE_ALIGN_8` macro, so I can take a look at the places where it might need to be a bit stronger.

fatcerberus:
> Pointers are 256 bits, containing a 64-bit virtual address, base, bounds and permissions.

Wow, I hope this platform has huge amounts of RAM if programs are passing 32-byte pointers (and integers) around everywhere!

Also, encoding permissions in the pointer sounds exploitable...

svaarala:
@davidchisnall I guess `duk_tval` will use the unpacked alternative and be 512 bytes to host a tag and a pointer so it would always be properly aligned.

[...]

PR #538, opened 2016-01-14
Avoid (u)intptr_t for const-to-nonconst casts

svaarala:
- [ ] Add a test file with various const-to-nonconst cast alternatives
- [ ] Check if `DUK_LOSE_CONST` fill-in could cast through a union to avoid dependency on `(u)intptr_t` type which is not guaranteed to exist (see https://github.com/svaarala/duktape/issues/530#issuecomment-171674359)
- [ ] Update code issues: best practice for losing const, note that intptr_t are not guaranteed to exist so avoid them
- [ ] Open a pull for the remaining intptr_t usages to remove remaining dependencies on the type

### DIFF
--- a/misc/lose_const_warnings.c
+++ b/misc/lose_const_warnings.c
@@ -1,0 +1,42 @@
+/*
+ *  Doing a cost-to-non-const cast without warnings, and without portability
+ *  or compiler specific issues is a bit challenging.
+ *
+ *  $ gcc -o/tmp/test -std=c99 -Wall -Wextra -Wcast-qual misc/lose_const_warnings.c -lm
+ *  misc/lose_const_warnings.c: In function ‘main’:
+ *  misc/lose_const_warnings.c:24:7: warning: cast discards ‘__attribute__((const))’ qualifier from pointer target type [-Wcast-qual]
+  p1 = (void *) p2;
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+
+typedef union {
+	void *x;
+	const void *y;
+} lose_const;
+
+int main(int argc, char *argv[]) {
+	void *p1;
+	const void *p2;
+	lose_const lc;
+
+	(void) argc; (void) argv;
+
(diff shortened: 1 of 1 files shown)


svaarala:
@fatcerberus: https://github.com/svaarala/duktape/issues/530#issuecomment-171677586

Why is casting through a union hacky? Converting between e.g. 8-byte array and IEEE double is also normally done using a union, so it's a standard approach in my book :-) I'd imagine representing pointers as integers would be challenging at least for segmented pointer models where a memory address doesn't necessarily have a unique pointer value (different segment bases can be used), etc. Perhaps this is what David is referring to.

The C standard apparently just guarantees that you can cast from pointer to `(u)intpr_t` and back to pointer without losing information. There are apparently no guarantees what e.g. the lowest bits represent, so perhaps it'd be entirely legal for the lowest 16 bits of a `(u)intptr_t` to represent a segment selector instead of offset within segment, etc.

But even so I don't understand why the argument order for bitwise AND should matter, as it seems to.

fatcerberus:
I don't know what the C spec says about this, but it was my understanding that it's undefined behavior to access a different union member than the last one you assigned to.  That's why I said it was hacky.  In practice there will probably be no issues, but... eh.  Undefined behavior scares me.

svaarala:
@fatcerberus Right, I'd have to recheck what the actual guarantees are.

However, Duktape relies casting through union in a number of places to e.g. perform bitwise operations on IEEE double representations and then read them back as `double` values. So this dependency currently exists anyway, and adding one for pointer conversions shouldn't depend any more on compiler behavior re: unions. (At least if the union is volatile.)

fatcerberus:
I found this Stack Overflow question which gives some insight into the issue of type punning with unions:
http://stackoverflow.com/questions/25664848/unions-and-type-punning

Notably, C89 says it's implementation-defined, but later versions of the spec including C99 allow it explicitly.  In C++, however, it seems to be UB--so that's a potential pitfall to keep in mind.

PR #539, opened 2016-01-14
Fix utf8 clen fast path pointer arithmetic issue

svaarala:
See discussion: https://github.com/svaarala/duktape/issues/530#issuecomment-171654860

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1372,6 +1372,10 @@ Planned
   immediately which makes the cause of the unsafe behavior difficult to
   diagnose (GH-531)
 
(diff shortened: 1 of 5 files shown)


fatcerberus:
This is very odd.  Bitwise AND is by definition commutative.

svaarala:
Here's a description of the underlying issue: https://github.com/svaarala/duktape/issues/530#issuecomment-171697759

Instead of switching the bitwise AND order, the better fix is to cast to `size_t` instead of `(u)intptr_t` which doesn't rely on the AND ordering. To be fixed in the commits.

PR #543, opened 2016-01-15
Add -Wunreachable-code to gcc default compile

svaarala:


fatcerberus:
I got a ton of "unreachable code" warnings last time I ran cl.exe with `/W4`.  Have those been fixed?

svaarala:
I'm not sure what you're referring to so not sure :-)

There were warnings related to linked list asserts in master when assertions were disabled (these were unused variables rather than unreachable code I believe); I fixed that in master already.

If master has warnings now with `/W4` that would need to be fixed.

fatcerberus:
I was referring to the `/W4` warnings I found just prior to the 1.4.0 release.  Besides the warnings that were already fixed, there were about 50+ warnings for unreachable code.  There were so many that I figured it was known and only mentioned it in passing.

Anyway, I'll get a full dump of all `/W4` warnings and post an issue.

PR #547, opened 2016-01-17
Cleanups for literal regexp curly brace handling

svaarala:
- [x] Support two more cases of invalid regexp quantifier, so that the lexer falls back to literal brace
- [x] Fix quantifier digit check (rejected 9 digits), caused by check being moved into a different place
- [x] Update testcase to test for the two additional cases
- [x] Fix config option metadata version (1.5.0)
- [x] Check whether Emscripten "fixups" can now be removed from Makefile => no, at least this fixup is still needed (not a regexp leniency issue, but a function .toString() parsing issue): https://github.com/svaarala/duktape/blob/master/util/fix_emscripten.py#L24-L32
- [x] Update Emscripten compatibility documentation re: regexp fixes
- [x] Internal documentation
- [x] Website
- [x] Releases entry
- [x] Rebase and squash
- [x] Assertion runs


svaarala:
All the Ecmascript RegExp syntax fixups can now be removed, but one unrelated fixup is still needed: Emscripten has some non-standard expectations of what a function's `.toString()` result contains: https://github.com/svaarala/duktape/blob/master/util/fix_emscripten.py#L24-L32

fatcerberus:
I don't understand the Emscripten `.toString()` issue, even after reading the fixup code.  What exactly is the problem?

svaarala:
`sourceRegex` assumes more of `.toString()` output than is guaranteed by specification.

svaarala:
This is probably the offending regexp:

```
var sourceRegex=/^function\s*\(([^)]*)\)\s*{\s*([^*]*?)[\s;]*(?:return\s*(.*?)[;\s]*)?}$/
```

As an example of what it matches and doesn't match:

```
> var sourceRegex=/^function\s*\(([^)]*)\)\s*{\s*([^*]*?)[\s;]*(?:return\s*(.*?)[;\s]*)?}$/
undefined
> sourceRegex.exec('function () { /* dummy */ }')
null
> sourceRegex.exec('function () { return 1; }')
[ 'function () { return 1; }',
  '',
  '',
  '1',
  index: 0,
  input: 'function () { return 1; }' ]
```

Duktape's function output won't match this regexp.

svaarala:
This is what's guaranteed by specification (http://www.ecma-international.org/ecma-262/5.1/#sec-15.3.4.2):

> An implementation-dependent representation of the function is returned. This representation has the syntax of a FunctionDeclaration. Note in particular that the use and placement of white space, line terminators, and semicolons within the representation String is implementation-dependent.

fatcerberus:
It seems to be assuming there's a return statement in the function, whereas Duktape only has a comment.  Seems pretty misguided, if an implementation happens to provide actual source for a function it might not always have an explicit return. :/

svaarala:
That's not the issue, the return part is a conditional match. Rather it's assuming there's no comment (star characters in particular) inside the function body (which Duktape uses to indicate the type of the function, e.g. `/* native code */`):

```
> var sourceRegex=/^function\s*\(([^)]*)\)\s*{\s*([^*]*?)[\s;]*(?:return\s*(.*?)[;\s]*)?}$/
undefined
> sourceRegex.exec('function () { /* dummy */ }')
null
> sourceRegex.exec('function () { }')
[ 'function () { }',
  '',
  '',
  undefined,
  index: 0,
  input: 'function () { }' ]
```

JoshEngebretson:
@svaarala I am pulling to update for the regex changes, would you recommend I pull master or this PR?

svaarala:
@JoshEngebretson Either will probably work fine - the changes in this branch are pretty much corner cases.

svaarala:
@JoshEngebretson I'll be merging this branch probably later today though, so it will be in master in a short while :)

JoshEngebretson:
@svaarala Great, thanks will pull this hash :+1:

PR #550, opened 2016-01-19
Fix feature option defaults when already provided

svaarala:
This caused an issue with a few config options when feature option support (DUK_OPT_xxx) was enabled.  For example, DUK_USE_PACKED_TVAL is provided by platform detection; if neither DUK_OPT_PACKED_TVAL nor DUK_OPT_NO_PACKED_TVAL was defined, the default clause would #undef DUK_USE_PACKED_TVAL instead of keeping the platform detected value. Workaround is to use `-DDUK_OPT_PACKED_TVAL`.

Affected config options:

- DUK_USE_PACKED_TVAL
- DUK_USE_SETJMP
- DUK_USE_UNDERSCORE_SETJMP

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1381,6 +1381,10 @@ Planned
   exotic pointer models by avoiding arithmetic and binary operations on
   (u)intptr_t values (GH-530, GH-539)
 
+* Fix a genconfig legacy feature option bug in Duktape 1.4.0 which caused
+  DUK_USE_PACKED_TVAL to default to false unless forced using
+  DUK_OPT_PACKED_TVAL (GH-550)
+
 2.0.0 (XXXX-XX-XX)
 ------------------
 
--- a/config/config-options/DUK_USE_ALIGN_BY.yaml
+++ b/config/config-options/DUK_USE_ALIGN_BY.yaml
@@ -1,5 +1,4 @@
 define: DUK_USE_ALIGN_BY
-# FIXME: resolve wrapper
 feature_snippet: |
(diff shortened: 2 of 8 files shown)


svaarala:
Example issue caused by not using packed `duk_tval` when pointer compression enabled:

- http://stackoverflow.com/questions/34884143/porting-duktape-getting-duk-create-heap-error-during-js-compilation-of-builtin/34887806?noredirect=1#comment57519918_34887806

I'll add an `#error` in a later pull to check for pointer compression + unpacked `duk_tval`.

fatcerberus:
I don't understand why losing packed tval causes a bug--isn't the packing just an optimization?

svaarala:
It is "just" an optimization, but critical for low memory targets.

svaarala:
Also pointer compression without packed `duk_tval` is not currently supported, but there's no `#error` sanity check so it leads to difficult to diagnose issues.

fatcerberus:
Yeah, I knew it was critical for low-memory environments, by "just an optimization" I meant I didn't expect it to cause a change in behavior.  I don't doubt the existence of the bug of course, just curious is all.

svaarala:
Yeah, in this case unpacked `duk_tval` and compressed pointers don't currently work together (it's not well documented and there's no sanity check which I think is a bug).

But there are also some small differences in behavior between packed and unpacked `duk_tval`s, for example the handling of NaN normalization. Internal type tags, visible through `Duktape.info()`, will also differ. For the most part it is intended to be fully transparent, that's true.

PR #554, opened 2016-01-24
Change Function .toString() output to be Emscripten compatible

svaarala:
Change Function `.toString()` output in two minor ways to make it Emscripten compatible without the need for fixups:

- If the function is anonymous, don't use a dummy `"anon"` name for it. For example, instead of `function anon() {...}` output `function () {}`.
- Use a string literal (similar to a directive) instead of a comment to indicate function type. For example, instead of `function foo() {/*ecmascript*/}` output `function foo() {"ecmascript"}`

While both the old and new formats are equally valid from a compliance perspective, it'd be nice if no Emscripten fixups were needed. The anonymous function behavior is, in my opinion, an improvement; the function type indication is more a matter of style.

Tasks:
- [x] Finalize change
- [x] Fix breaking testcases
- [x] Add a dev testcase for new format with explanation
- [x] Internal documentation, esp. Emscripten status
- [x] Website/wiki update for Emscripten status
- [x] Update `fix_emscripten.py` to indicate it's no longer required (at least with current Duktape and Emscripten versions)
- [x] Update Makefile to avoid the hopefully unnecessary fixups
- [x] Releases entry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -678,8 +678,8 @@ emscriptentest: emscripten duk
 	emscripten/emcc $(EMCCOPTS) tests/emscripten/helloworld.c -o /tmp/duk-emcc-test.js
 	cat /tmp/duk-emcc-test.js | $(PYTHON) util/fix_emscripten.py > /tmp/duk-emcc-test-fixed.js
 	@ls -l /tmp/duk-emcc-test*
-	./duk /tmp/duk-emcc-test-fixed.js
-	#./duk /tmp/duk-emcc-test.js
+	#./duk /tmp/duk-emcc-test-fixed.js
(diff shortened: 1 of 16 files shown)


fatcerberus:
I agree with the anonymous function change.  The old way was misleading: an anonymous function has no name by definition, while Duktape actually returned a named function. (the name happened to be "anon" but it was named nonetheless).

fatcerberus:
What interests me about the Emscripten behavior is that it can potentially break under any engine which provides the real function source code in `.toString()`.  SpiderMonkey does this, for example.  It amazes me that this hasn't yet caused real-world issues outside of Duktape.

PR #559, opened 2016-01-26
Add support for placing builtin strings and objects into ROM (read only data section)

svaarala:
Add preliminary and experimental support for moving built-in strings and objects into ROM (read-only data section) which drops low memory startup RAM footprint from ~27kB to ~3kB.  For example with "ajduk":

```
[...]
Pool dump after heap creation
======= dump of 25 heap pools ======
heap[0] pool[8] unused
heap[0] pool[12] used=0 free=10 high-water=1 min-alloc=12 max-alloc=12
heap[0] pool[16] unused
heap[0] pool[20] used=3 free=397 high-water=3 min-alloc=20 max-alloc=20
heap[0] pool[24] unused
heap[0] pool[28] used=1 free=199 high-water=1 min-alloc=28 max-alloc=28
heap[0] pool[32] unused
heap[0] pool[40] unused
heap[0] pool[48] unused
heap[0] pool[52] unused
heap[0] pool[56] unused
heap[0] pool[60] unused
heap[0] pool[64] unused
heap[0] pool[128] used=1 free=79 high-water=1 min-alloc=80 max-alloc=80
heap[0] pool[256] unused
heap[0] pool[392] used=1 free=0 high-water=1 min-alloc=392 max-alloc=392
heap[0] pool[512] used=1 free=15 high-water=1 min-alloc=320 max-alloc=320
heap[0] pool[964] used=1 free=0 high-water=1 min-alloc=964 max-alloc=964
heap[0] pool[1024] used=1 free=5 high-water=1 min-alloc=1024 max-alloc=1024
heap[0] pool[1780] unused
heap[0] pool[2048] unused
heap[0] pool[4096] unused
heap[0] pool[8192] unused
heap[0] pool[16384] unused
heap[0] pool[32768] unused
======= heap hwm = 3120 use = 3108 waste = 240 ======
```

The initial 3kB breaks down roughly as follows:

- Duktape heap (964 bytes) and initial thread (392 bytes), total about 1.3kB.
- Initial value stack, 1kB.  The initial value stack will resize to 2kB (256 tagged values) quite quickly when you compile/run any code.
- Writable global object (no own properties but inherits from a ROM global object), global environment record and its property table, other trivia, total about 0.6kB.

In addition to low memory targets the ROM object support may be useful for environments where a large number of Duktape heaps are used simultaneously: ROM objects can be used with and without pointer compression.  For example, x64 build without pointer compression or any other low memory options reports about 8kB startup memory usage (compared to ~110kB without ROM objects) using valgrind massif.

The pull also provides groundwork to later move user strings and objects into ROM to allow custom bindings to be free in terms of RAM usage.  This is already possible by editing the strings/objects YAML files directly, but later pulls will make it possible to provide custom bindings through separate YAML files.

Impact on ordinary (no ROM objects) build should be minimal, but there are very minor changes in the RAM built-in initialization format to support more flexible initialization in the future for RAM objects too.

More detailed changes:

- Refactor genstrings.py and genbuiltins.py to read string and object data from YAML files.
- Add support for ROM objects/strings into genbuiltins.py, which generates a set of static initializers for builtin objects/strings when --rom-support is enabled (this expands the size of duktape.c considerably so it's not enabled by default).  Also add --rom-support to make_dist.py which forwards the ROM support request to genbuiltins.py.
- Also make RAM builtin initializer format more flexible to allow arbitrary strings to be encoded into the built-ins init data without needing explicit DUK_STRIDX_xxx values.
- Add runtime support for ROM objects by adding a new duk_heaphdr flag (DUK_HEAPHDR_FLAG_READONLY) and avoid writing to objects which have this flag set.  Rework heap and thread initialization to support ROM and RAM builtins, depending on what defines are active.
- Several small fixes and improvements.

The ROM support is still quite rough and may not work in all configurations. Limitations and future work: #560.

Tasks:

- [x] Assertion runs for non-ROM build
- [x] Valgrind runs for non-ROM build
- [x] Performance test for non-ROM build
- [x] Releases


svaarala:
Code footprint impact on x64 default build (with ROM builtins disabled) is about +500 bytes which is mostly explained by an increase in the RAM init data size: strings not needing a runtime string index (stridx) will appear as literals and may appear multiple times. This is more than offset by the reduction in `duk_heap` struct size though.

fatcerberus:
Does this have any performance benefit?

svaarala:
It shouldn't, unless ROM memory accesses are faster than RAM. This is just about reducing RAM usage so that it's possible to run in smaller devices.

fatcerberus:
For what it's worth duplicate literals should have negligible effect on x86/x64 code size--and possibly other platforms--since many compilers can intern/pool strings.

svaarala:
These duplicates will be inside the opaque init data blob so the compiler can't really do anything about it. The duplication only affects code footprint: the strings in question will be interned into the Duktape heap stringtable and won't be duplicated there. So it's just about the init data size.

PR #562, opened 2016-01-26
Fix debugger eval locking

### LINKED ISSUE #558: Uncaught error during debugger Eval generates fatal Throw notify
...it's actually even worse than that: Duktape apparently halts execution in the Eval code since the filename in the Status notify is "input".  This confuses my debug client to no end (it doesn't expect execution to pause there ;) ), somehow ultimately leading to Duktape sending a Detaching notify with the error flag set.  I haven't pinned down the precise chain of events that causes the errant detach yet, my message handling is currently a bit rickety. :)

svaarala:
`heap->dbg_processing` was not set to 1 when debugger message loop was entered from the bytecode executor. When the debugger Eval command was then used it was possible for error handling code to send Throw notifys while running the eval code; more seriously, if "pause on uncaught" was enabled the debugger might try to halt while already halted.

Fixing `dbg_processing` flag fixes the uncaught error bug. As a side effect there won't be Throw notifys during debugger Eval.

Fixes #558.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1394,6 +1394,10 @@ Planned
   DUK_USE_PACKED_TVAL to default to false unless forced using
   DUK_OPT_PACKED_TVAL (GH-550)
 
+* Fix debugger Eval handling issue where an uncaught error during Eval
+  command (with "pause on uncaught" option enabled) would cause a recursive
+  attempt to halt execution (GH-558, GH-562)
+
 * Portability improvement for Atari Mint: avoid fmin/fmax (GH-556)
 
 * Change OS string (visible in Duktape.env) from "ios" to "osx" for non-phone
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -1877,6 +1877,7 @@ DUK_INTERNAL duk_bool_t duk_debug_process_messages(duk_hthread *thr, duk_bool_t 
 	DUK_UNREF(ctx);
 	DUK_ASSERT(thr->heap != NULL);
 	DUK_ASSERT(DUK_HEAP_IS_DEBUGGER_ATTACHED(thr->heap));
+	DUK_ASSERT(thr->heap->dbg_processing == 1);  /* caller ensures */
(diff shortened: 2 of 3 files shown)


fatcerberus:
Thanks, I'll test this out in a bit.

fatcerberus:
The bug seems to be fixed:

```
D:\src\spectacles-i>ssj dist
SSJ 2.99.812 Sphere Game Debugger x64
A powerful JavaScript debugger for minisphere
(c) 2016 Fat Cerberus

Starting minisphere... OK.
Connecting to 127.0.0.1:1208... OK.
Handshaking... OK.
  Attached to minisphere 2.99.812
  Duktape v1.4.0-63-g4dd5b55

scripts/main.js:6 ssj$ eval pig
Evaluating JS expression... OK.
ReferenceError: identifier 'pig' undefined

scripts/main.js:6 ssj$ eval [].forEach()
Evaluating JS expression... OK.
TypeError: function required, found undefined (stack index 0)

scripts/main.js:6 ssj$ _
```

fatcerberus:
Looks like it's back to using Duktape masters again for me. :)

svaarala:
I'll need to think about the couple of FIXMEs in this pull and I'll then merge this.

I'll also merge this to 1.4.1 maintenance release so you can use that too if you want soonish :)

fatcerberus:
I've generally been okay with using Duktape master for minisphere releases in the past as it's very well-tested and stable, but for my 3.0 release I want everything to look as polished as possible, thus an official versioned Duktape release is preferred. :)

svaarala:
When is that release - I can make the 1.4.1 release earlier if necessary.

fatcerberus:
It's a ways off still - I have to finish the console debugger first.  No rush.

svaarala:
Ok, when it's closer we can make sure there's a maintenance release with any necessary fixes :)

svaarala:
@fatcerberus I've been a bit busy but finally had time to finish this branch - I'll merge this once the tests complete.

PR #564, opened 2016-01-27
Add support for HTML comment syntax

svaarala:
https://www.mail-archive.com/netsurf-users@netsurf-browser.org/msg06825.html

HTML comment syntax is not part of ES5 but is an optional part of ES6:
- http://www.ecma-international.org/ecma-262/6.0/#sec-html-like-comments

### DIFF
--- a/src/duk_lexer.c
+++ b/src/duk_lexer.c
@@ -975,6 +975,7 @@ void duk_lexer_parse_js_input_element(duk_lexer_ctx *lex_ctx,
 		advtok = DUK__ADVTOK(1, DUK_TOK_COMMA);
 		break;
 	case DUK_ASC_LANGLE:  /* '<' */
+		/* FIXME: add HTML comment <!-- ... --> support here. */
 		if (DUK__L1() == '<' && DUK__L2() == '=') {
 			advtok = DUK__ADVTOK(3, DUK_TOK_ALSHIFT_EQ);
 		} else if (DUK__L1() == '=') {


fatcerberus:
Huh, I didn't even know this was a thing.  You have this tagged realworld - Is there really JS code in the wild that uses HTML comments?

svaarala:
Yes, this came up in NetSurf testing.

chris-y:
Old advice was to encapsulate everything in <script> tags in an HTML comment, eg:
<script>
</script>

If you want a real world example, there are a couple of scripts on http://www.terrestrialtv.uk/dtt.cgi which do exactly this.

fatcerberus:
Hm.  Although if the JS parser treats  as a normal comment (instead of ignoring it), wouldn't that then turn the entire block of code into a comment?

svaarala:
@fatcerberus The semantics are a bit interesting:

`<!--` starts a *line comment* so it doesn't extend to the `-->` end marker (!)

fatcerberus:
Ah.  That makes more sense then.

andwhyisit:
Is this bug ever going to be fixed? It has been almost a year and it seems silly to suggest that 1/4 of inline javascript can't even be parsed because of some practice from 20 years ago.

svaarala:
@andwhyisit I'm not sure what you mean by a bug - but HTML syntax is not a feature of ES5 or core ES6, but rather part of ES6 Annex B of optional, legacy features. Duktape is not a full ES6 engine (yet) so several ES6 features are not yet implemented.

In any case, HTML comment syntax is now supported in master. Closing this FIXME pull as unmerged.

PR #569, opened 2016-01-31
Add debugger Break notify

fatcerberus:
This is an initial prototype for a Break notify which is sent whenever a breakpoint is hit during execution.  The notify message indicates the breakpoint number that was hit, information which is not available in a Status notify.

### DIFF
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -1350,6 +1350,21 @@ reestablish the link).
 ``msg`` is an optional string elaborating on the reason for the detach.  It may
 or may not be present depending on the nature of detachment.
 
+Break notification (0x07)
+-------------------------
+
+Format::
+
+    NFY <int: 7> <int: index> EOM
+
+Example::
+
+    NFY 7 3 EOM
+
+Duktape sends a Break notification whenever a breakpoint is hit during checked
+execution. ``index`` is the number of the breakpoint which was hit, as returned
+in the response for AddBreak (see below).
+
 Commands sent by debug client
 =============================
 
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -906,6 +906,16 @@ DUK_INTERNAL void duk_debug_send_status(duk_hthread *thr) {
 	duk_debug_write_eom(thr);
 }
 
+DUK_INTERNAL void duk_debug_send_break(duk_hthread *thr, duk_uint32_t index) {
+	/*
+	 *  NFY <int: 7> <int: index> EOM
+	 */
+
+	duk_debug_write_notify(thr, DUK_DBG_CMD_BREAK);
+	duk_debug_write_int(thr, index);
+	duk_debug_write_eom(thr);
+}
+
 #if defined(DUK_USE_DEBUGGER_THROW_NOTIFY)
 DUK_INTERNAL void duk_debug_send_throw(duk_hthread *thr, duk_bool_t fatal) {
 	/*
--- a/src/duk_debugger.h
+++ b/src/duk_debugger.h
@@ -21,6 +21,7 @@
 #define DUK_DBG_CMD_LOG           0x04
 #define DUK_DBG_CMD_THROW         0x05
 #define DUK_DBG_CMD_DETACHING     0x06
+#define DUK_DBG_CMD_BREAK         0x07
 
 /* Initiated by debug client */
 #define DUK_DBG_CMD_BASICINFO     0x10
@@ -87,7 +88,9 @@ DUK_INTERNAL_DECL void duk_debug_write_notify(duk_hthread *thr, duk_small_uint_t
 DUK_INTERNAL_DECL void duk_debug_write_eom(duk_hthread *thr);
 
 DUK_INTERNAL_DECL duk_uint_fast32_t duk_debug_curr_line(duk_hthread *thr);
+
 DUK_INTERNAL_DECL void duk_debug_send_status(duk_hthread *thr);
+DUK_INTERNAL_DECL void duk_debug_send_break(duk_hthread *thr, duk_uint32_t index);
 #if defined(DUK_USE_DEBUGGER_THROW_NOTIFY)
 DUK_INTERNAL_DECL void duk_debug_send_throw(duk_hthread *thr, duk_bool_t fatal);
 #endif
--- a/src/duk_js_executor.c
+++ b/src/duk_js_executor.c
@@ -1484,6 +1484,7 @@ DUK_LOCAL void duk__interrupt_handle_debugger(duk_hthread *thr, duk_bool_t *out_
 	duk_activation *act;
 	duk_breakpoint *bp;
 	duk_breakpoint **bp_active;
+	duk_uint32_t bp_idx;
 	duk_uint_fast32_t line = 0;
 	duk_bool_t send_status;
 	duk_bool_t process_messages;
@@ -1532,6 +1533,7 @@ DUK_LOCAL void duk__interrupt_handle_debugger(duk_hthread *thr, duk_bool_t *out_
 			 * described in: https://github.com/svaarala/duktape/issues/263.
 			 */
 			bp_active = thr->heap->dbg_breakpoints_active;
+			bp_idx = 0;
 			for (;;) {
 				bp = *bp_active++;
 				if (bp == NULL) {
@@ -1543,8 +1545,11 @@ DUK_LOCAL void duk__interrupt_handle_debugger(duk_hthread *thr, duk_bool_t *out_
 					DUK_D(DUK_DPRINT("BREAKPOINT TRIGGERED at %!O:%ld",
 					                 (duk_heaphdr *) bp->filename, (long) bp->line));
 
+					duk_debug_send_break(thr, bp_idx);
 					DUK_HEAP_SET_PAUSED(thr->heap);
 				}
+
+				bp_idx++;
 			}
 		} else {
 			;


fatcerberus:
My initial thought was to fold this functionality into the Status notify (with a status code of say, 0x2 to mean a breakpoint), but as mentioned above, there is no place to put the breakpoint number which was hit, and tacking it onto the end of the Status notify would be awkward.  This seemed cleaner.

svaarala:
Sounds reasonable. One alternative would be to have some sort of "Paused" notify which would more generally indicate why we're pausing: breakpoint, debugger statement, uncaught error, etc.

fatcerberus:
That's probably the best way to go about it, rather than incrementally adding notify types as new scenarios arise.  I'll review this branch and see if I can generalize it.

svaarala:
Hmm, I guess adding pause reason fields to Status instead of a separate Paused notify would be more or less equivalent? In both cases the pause reason fields will need to indicate the breakpoint index, and are potentially a versioning issue because we don't know beforehand what additional fields are needed. This doesn't seem easy to avoid expect by using different notify types which has its own downsides.

fatcerberus:
The difference would be that the new Paused notify would be a clean slate and we could specify, e.g. that the message contents will vary based on the reason code.  It's a bit more awkward with the Status notify because there are existing fields that need to be honored. (i.e. the "variable" part would be appended to the end instead of next to the reason code which seems more logical)

svaarala:
Compatibility is good but it's also useful to think beyond that to figure out what the cleanest protocol would look like. There's going to be a (small) incompatible break for Duktape 2.x for both the API and the debugger protocol, so it'd be good to unify Paused and Status at least there to keep the mechanisms used minimal. (I don't think that break is going to be big, but it's necessary from time to time to remove overlapping clutter.)

So if the only motivation for a separate notify is a clean slate, perhaps it'd be more efficient (compared to adding a new notify type and then deprecating it) to add things into Status and then reshuffle the Status fields in that compatibility break?

fatcerberus:
Hmm.  If I were designing the protocol from scratch, with no compatibility baggage to worry about, I probably wouldn't have implemented the Status notify at all.  Conceptually, it's too broad and I usually end up building a wrapper around it that splits it into separate "paused" and "resumed" events.  I also tend to ignore all Status notifys beyond the first with a 0 status ("running"), considering them to be noise.

With that in mind, I think the cleanest solution for Duktape 2.0 would be to have separate pause and resume notifys and rework Status to only be for tracing live execution, rather than serving three different purposes as it does now.

In fact a dedicated Resume notify may not even be needed, since a resume is always initiated by the client anyway.

svaarala:
The way I see Status is that it describes the current execution state. Logically that includes at least:

- Are we paused or running?
- If we are paused, why did we pause?
- Where are are we paused or currently running

I don't see these as being separare things but rather describing the execution state in an atomic and stateless fashion so that the client doesn't need to piece together the current state from previous notifys.

svaarala:
As an example where having an atomic state update is useful:

Consider the situation where the client becomes paused because a debugger statement was hit. If Paused and Status are separated and don't provide overlapping information, the Paused notify will just say "paused because debugger statement was hit" without providing the file/line which will be provided by the Status notify later on.

But when processing Paused, the client needs to update its UI state to indicate we're paused but doesn't yet know where. It will then take some time (how much depends on latency, packet drops, etc) before it will know where we're paused.

On the other hand, if Paused includes file/line information, it will then be duplicating the Status fields which is a bit awkward.  So one either ends up duplicating fields or forcing the debug client to deal with partially out-of-sync information arriving in independent notifications.

fatcerberus:
Hm, that's true.  I definitely understand the need to avoid redundancy, that was a big bugbear for me while I was implementing the Throw notify, at a few points I was tempted to leave the file/line info out of it. (I ultimately decided against that because whenever possible, that info is taken from the thrown Error object and is thus not actually redundant)

Anyway, when I have a bit of free time later I'll look at this branch again and see if I can come up with a more robust implementation.

svaarala:
By the way, I'm not sure what the best approach is for Status really is - which is why I raised the question. I'm hoping we'll agonize a little bit over design decisions so that we'll avoid a lot of boring maintenance later :-)

Stepping back, it may be that the "list of dvalues" model for the debugger protocol is not the best choice overall. Extensibility would be easier if messages were streams of key/value pairs (which was one design option early on). The upside would be that adding new keys would be easy and there'd be no concern for ordering etc. The downside would be that the client might need to allocate memory if it didn't have a guaranteed key order - which unfortunately conflicts with one of the design goals (minimal memory churn).

fatcerberus:
The flat list of dvalues is indeed a problem w.r.t. maintaining compatibility when making enhancements.  You either tack on atoms to the end of a message, which can lead to a chaotic ordering of values, or add completely new commands, which has its own problems (adding new API/protocol guarantees).

String-keyed dvalues would be a good solution, I think.  The memory churn issue is something to consider, though if we specify that the keys are guaranteed never to exceed a certain length, say 16 characters, allowing the client to pre-allocate buffers for them, the problem can be worked around.  Admittedly that's more complex than indexed dvalues though.

svaarala:
I don't the the debug client is a big issue, but Duktape's debug message parsing would either need multiple passes (awkward to add to the transport abstraction), buffering (awkward because means memory churn and unbounded memory usage), or an intermediate representation to iron out key ordering issues (unbounded memory usage).

Using string keyed key/value lists with ordering guarantees would also be possible but I think that'd be quite odd and difficult to understand, and maintaining ordering guarantees would also be a versioning problem.

svaarala:
And also, a single key/value space would not really be enough because some commands have lists or structured values so it must be possible to repeat "units" which same the same keys.

Letting go of the minimal/bounded memory usage goal would make the best choice an actual structured and nested format which supported rich typing.

svaarala:
One practical option (and part of the current assumptions) would also be to consider the debug protocol less version stable as the API and other parts, so that incompatible changes are made freely and the version is simply bumped. This means debug clients will initially break but the fixes are relatively small, e.g. shuffling a few fields or changing some command semantics slightly.

fatcerberus:
Well that *was* the idea of having a separate version number sent with the handshake, wasn't it?  Even so, if we have to bump the protocol version number I'd like for it to be a meaningful restructuring and not just "we needed to move some dvalues around" ;)

svaarala:
That was the reason for having a protocol version yes. I agree it shouldn't be bumped for moving a dvalue around - but on the other hand, I also don't want to create a large amount of baggage because each incremental change is not a big enough reason to bump the version.

fatcerberus:
Anyway, since you opened the door, the problem with the Status notify right now is best illustrated by my code to handle it:
* https://github.com/fatcerberus/minisphere/blob/master/src/ssj/session.c#L387-L403
* https://github.com/fatcerberus/minisphere/blob/master/src/plugin/Debugger/DebugSession.cs#L185-L229

The second one, from minisphere's GUI debugger, is especially bad.  As you say, Status is supposed to be stateless, however in practice handling it often requires stateful logic (Were we already paused before? Are we just now resuming? Was the last notify a Throw? and so on).  Having a separate "Pausing" notify would ease this a lot.

To make it easier on the client, we'd just have to guarantee that an up-to-date status is sent _before_ the pause notify goes out.

svaarala:
It's the classic question of whether to use a level-based or edge-based design.

With an edge-based design the question "did we become paused?" is easy and the client doesn't need to remember previous state to do that. But if it wants to know "are we paused now?" it needs to keep track of all edge transitions and no transitions can ever be lost or we're out of sync possibly for a long time. Duktape also needs to be 100% correct in generating the edges because sending "we became paused" twice without resuming in between can break client state (depending on how it's implemented).

With a level-based design every Status update ensures the client is in sync so "are we paused now?" is easy; "did we become paused?" means tracking previous state and comparing. For the sender level-based design is easier because you can always re-send the state if you're unsure. Another advantage is that you don't need to anticipate and provide explicit messages for all possible edge triggers but can just provide current state and let the client decide what transitions are worth an event.

What bothers me most with edge-triggered events is that a single Status message is replaced by potentially a lot of separate event notification messages. And unless those messages essentially duplicate all fields in Status (which potentially change on every event) the client won't get an up-to-date status atomically but will e.g. first get an event that we became paused, and only later get to know where. What's worse, there's a tendency to replicate only a subset of state fields into the notifications making them inconsistent; the client then knows we became paused and a few other fields of state information (but not all).

Another practical complication is that sometimes sending an edge notify from a certain point in Duktape code is awkward: if we're in the middle of processing a previous debugger message we can't send the notify right away. Instead we either need to drop it (bad if the debug client is expecting it) or queue it to be sent later (awkward and means memory churn).

These points are more or less why I personally favor a level-based design. It's certainly a trade-off: a level-based design makes the protocol less verbose (fewer messages etc) and the Duktape side quite simple to implement robustly and without cold code paths (also no need for 100% accurate events), but any edge-related information needs to be detected by the debug client as illustrated by your links.

One interesting design question is to figure out what things are best modelled as state vs. events. For example, "Duktape just did a mark-and-sweep" seems intuitively best modelled as an event (edge notify). But it can also be implemented by keeping a cumulative "mark_and_sweep_count" counter and providing that in Status which also resolves the problem of sending the notification potentially while already processing another debug command.

All that said I understand a level-based approach makes the client code potentially more complicated if it is especially interested in edges. I'm just worried fixing that complexity by providing edge notifications may not be worth it at least for all state transitions. Either way the protocol should have a consistent and minimal set of mechanisms if at all possible.

svaarala:
Re: sending a Status just before a Pause, that'd be a reasonable solution to avoid duplication I think.

But unless there's an explicit guarantee of this behavior the debug client cannot be sure nothing happened between the two messages. So should the description of Pause say that "the client is guaranteed of getting a Status before the Pause notification"? That might seem a bit odd to guarantee, but if we don't there's a natural tendency to replicate Status fields into the notify messages.

Implementing a guaranteed Status+Pause sequence is not a problem, because we can just trigger a status send and then immediately send the notify without anything going on between them (luckily there are no threading issues or the like).

fatcerberus:
Now that AppRequest is in master, I'm going to take another look at this and try to come up with a more extensible method that doesn't involve adding fifty new commands. :)

harold-b:
Hmm, can you guys explain how the status notify is coming up short in terms of breaks? ( Other than not telling you the reason why )

I'm not sure I'm following the issue here. Why the need to add another notify?

I personally found the status based approach quite simple to track the state. The client simply doesn't assume any state until confirmed by the server.

fatcerberus:
The issue is that the Status notify doesn't tell _which_ breakpoint was hit (as in, its breakpoint index as returned by AddBreak).  In, e.g. GDB it will tell you which breakpoint was hit when the program stops.

fatcerberus:
To clarify, a Paused state can be reached by several methods right now: Thrown error, Breakpoint, initiated-by-client, etc. and there is at present no way to know the reason from the client side.

harold-b:
Yes, that's why I qualified with "other than it not telling you the reason why", as in why it stopped.

But in that case, wouldn't it be simpler to add that to the status notify instead of adding another notification, as mentioned above?

I agree that getting the reason for the pause would very useful, but wouldn't it be easier if it were appended into the status notify, when the state is paused?

At least I find that it would be less complex than handling notification. Since the state is then handled from a single place. That's just my input :)

fatcerberus:
Perhaps illustrating the problem would help.  Here's the code I had to write to handle a Status notify properly:
https://github.com/fatcerberus/minisphere/blob/master/src/plugin/Debugger/DebugSession.cs#L182-L226

That logic is way too convoluted for handling what should be a stateless notify.  If I had separate "Pause" and "Resume" notifies it would be much simpler.

If I were designing the debugger from scratch, this is probably the setup I would use:
* Status gives you stateless updates on the state of the program: Current filename and line number, whether or not it's running or not, etc.
* When the program is paused, you get a Paused notify telling you the reason for the pause.  The next Status then tells you where exactly you stopped.
* Likewise when the program resumes, you get a Resume notify.  This lets you refocus the debuggee on resume, etc. - things you wouldn't want to do with every Status notify because it would be disruptive to the user.

This seems much cleaner and more event-driven to me than forcing the client to maintain its own state while observing status notifys.

fatcerberus:
There's also a logistical issue: It would be difficult to include the pause reason directly in the Status notify since they are sent out statelessly by Duktape in the debugger message loop based on a boolean 'dirty' flag.  It might be possible to rework the handling so that instead of a single 'dirty' flag, there was a structure that stored the reason for updating the status that Duktape could refer back to when generating the notify.

svaarala:
There are some fundamental differences in programming preferences I think - I don't like event driven (edge triggered) programming when there's a state based (level triggered) alternative because event driven code often runs into various corner cases and annoying bugs when it loses or mishandles events. But I emphasize this is just a preference, not a statement of one thing being better than another.

@fatcerberus Just commenting on your "ideal setup", why would there need to be a Resume notify in this setup, assuming the client always resumes and the only cause for a resume is "the client requested it"?

It might be useful to use more concrete example sequences here for various alternatives, it's difficult for me at least to otherwise comment what solutions seem minimal compared to the functionality they provide. (For its shortcomings, the single Status notify is IMO very effective in providing a chunk of useful functionality for a very minimal code and conceptual footprint.)

svaarala:
Re: sending the pause reason, I don't see why that would be difficult? A call site setting the pause would conceptually:

- Set the `dbg_paused` internal flag to non-zero value indicating the cause. Now we're just using 0 (running) and 1 (paused), but it could equally well have non-zero values for debugger statement, breakpoint, initial pause, etc.
- Set the `dbg_state_dirty` flag to ensure a status is sent when we go paused - it is always sent right after we pause anyway.
- If the pause reason required another field (say breakpoint index) that'd also be recorded, say `heap->dbg_paused_breakpt_index`.

This would cause the Status to go out with a paused field indicating the cause and whatever extra fields were required.

harold-b:
Hmm I'm not sure I'm tracking. The points you mentioned if you were to design it from scratch is pretty much the way mine works right now with a single status notify.

Here's my approach:

Client requests attach and a connection attempt is made, we wait for a correct signature. Upon confirmation, we check some attach options, if the user requested stop on attach or not, we send a resume/pause command accordingly and we then wait upon the initial status emitted.

From there we either have 2 states, client-controller ( paused ) or target controlled ( running ). Any changes in the client are made from a single status notify and a state is never assumed on the client until we get confirmation by a status notification. 

When we get a status with a paused state, we clear our client-side state ( breakpoints, stackframes, vars... ) and start requesting a new state. Then the GUI is updated appropriately.

If the status is resume, we simply resume, that is assume the target is running and only have the option of detach and pause buttons.


The *only* thin I can think of, which I'm not sure if this is what you refer to, is that there's no guarantee that the next status received after I requested a pause, for example, is the status I requested ( there may have been an incoming status on the wire already ). But I guess it doesn't matter, since I'm expecting a paused state to come, I simply wait until we receive any pause state.

svaarala:
@harold-b And indeed, if you request a Pause, it may take potentially seconds before Duktape has a chance to react, and a bunch of Status (and other notifys) may precede the Status that finally confirms the target is paused.

harold-b:
For me, that only thing I guess I would ask of the status is the reason for the break. That is, breakpoint,  user requested pause, debugger statement or throw.

We can find the breakpoint in question by checking our client side state for a breakpoint for the given file/line.

svaarala:
@harold-b Adding a breakpoint index to further describe a "paused because breakpoint" would also not be conceptually difficult. There's always a well defined call site inside Duktape which decides "OK I'm going to pause", and the breakpoint related call sites know what breakpoint they hit.

There is a versioning concern re: tacking on new values for the Status message but that's, IMO, unrelated to more philosophical issues regarding edge triggered events vs. sending a level based Status.

fatcerberus:
I guess my thinking was more along the lines of genericity, i.e. While it's true right now that resuming is always initiated by the client, this might not always be the case (of course that's entirely theoretical because it wouldn't be a very good debugger if it restarted execution behind your back!)  I try to think in generic terms always--that's why AppRequest was such a natural addition to me.

A more practical concern is that the Resume command is lost, but the client assumes it was delivered, it's now out of sync.  Of course if this is the case then Status notifys could just as well be lost, so that's not really an argument. :-)

I tend to prefer event-driven designs because, as long as you can assume event delivery is guaranteed (a reasonable assumption in most environments, admittedly not so much with Duktape in the general case), you can piece together needed state from past events.  The other way around (ascertaining edge transitions from state you wouldn't otherwise need to save) is much more difficult.

Admittedly if messages have a possibility of being lost, this doesn't buy you much.  So it's a trade-off either way.  I guess the level-based approach does make more sense from Duktape's point of view.

svaarala:
So just for some concreteness, here's the current Status format:

    NFY <int: 1> <int: state> <str: filename> <str: funcname> <int: linenumber> <int: pc> EOM

State is either 0 (running) or 1 (paused).

This could be extended to:

    NFY <int: 1> <int: state> <str: filename> <str: funcname> <int: linenumber> <int: pc> <int: pause reason> <int: breakpt index or -1> EOM

Pause reason could enumerate the various reasons, e.g.:

- Initial pause
- Breakpoint
- Debugger statement
- User requested via Pause command

And breakpoint index would be given when breakpoint is the cause (-1 or perhaps 'undefined' otherwise).

It'd also be possible to extend the state instead of adding a pause reason instead; this was the original intent for the "state" field.

harold-b:
> Set the dbg_paused internal flag to non-zero value indicating the cause. Now we're just using 0 (running) and 1 (paused), but it could equally well have non-zero values for debugger statement, breakpoint, initial pause, etc.

That's a very nice solution. :)

svaarala:
Handling lost commands etc is currently a rather theoretical concern because the debugger protocol is reliable (transport-wise) and assumes full sync to match up requests and reply/error messages in any case. So a lost message is, intentionally, a catastrophe not to be recovered from except by detaching :)

svaarala:
@fatcerberus I see we have quite the opposite views on event vs. state based programming :) In my experience it's more reliable to compare previous state to current state than writing event handlers which never, ever make any mistake whatsoever. Any small bug in event handling can leave the state permanently out of sync - whereas in a level based system the next status update will bring you up to sync.

Again, I'm most definitely not trying to say either view is more correct. It's truly a matter of preference and past experiences. :)

harold-b:
> I guess my thinking was more along the lines of genericity, i.e. While it's true right now that resuming is always initiated by the client, this might not always be the case (of course that's entirely theoretical because it wouldn't be a very good debugger if it restarted execution behind your back!) I try to think in generic terms always--that's why AppRequest was such a natural addition to me.

Well, in my case I don't even resume the client does the resumes, the client simply serves as a dumb terminal, the state is driven completely from the status notifies. If it somehow sends us a running status even though we didn't request it, then we adopt that state. To me, that's what's nice about the state-based approach.

harold-b:
> Again, I'm most definitely not trying to say either view is more correct. It's truly a matter of preference and past experiences. :)

Yes this is what I mean as well. Not trying to knock your view either, I'm expressing my point of view and why I think it works nice and clean. No harm meant!, Just discussion!

fatcerberus:
@svaarala Not to say that the level-based design can't make detrimental mistakes either - see the progression of fixes for #597, in particular the one where it stopped sending status notifys entirely :)  (just pulling your chain here, no harm meant :).

svaarala:
@fatcerberus It would probably be useful if you could propose a concrete message sequence example for discussing the alternative(s) to extending Status.

For example, I don't necessarily object to adding a Paused notify (which I'd probably prefer over separate notifys every cause) which precedes a Status. But I'd like to see what it'd look like and what the pros and cons of that would be.

By the way, in working with previous systems, one case where the state based approach works poorly is where the transitions are extremely important and it's possible for some transitions to be "ironed out" because a transition and its reverse happen before a status gets sent.

I tend to think like @harold-b that the debug client should just display the state the target is in, so that I don't care about the specific transitions; at least as far as the execution status is concerned. There are other things which are more events, for example, if the target is about to throw an error of if it printed out a message using `print()`. Those can't easily be modelled as states of course and are separate notifies now for that very reason.

svaarala:
@fatcerberus Sure it's possible to write bugs, also so for events. A similar bug for event-based status would have left the debug client in equally bad state: it would never get the event and be out of sync forever.

But note that in the bug in #597 that you refer to, adding an *extra* Status resend (even a duplicate one) would solve the problem perfectly: repeating a Status just in case is a valid (if not optimal) solution. That's not the case for events: you must always correctly send them and always correctly process them or we're potentially forever out of sync. That's why it's much easier to get a working solution in state based systems: you don't need to be 100% accurate as long as you err on the safe side of sending maybe one status too many.

fatcerberus:
Another potential issue is that sometimes your state space topology overlaps in places and edges intersect, leading to problems similar in structure to the point-in-polygon problem where you can't reliably determine edge transitions from current state:
https://en.wikipedia.org/wiki/Point_in_polygon#Special_cases

Like I said, it's a trade-off in either case. :)

Anyway, like I said, I'll look into it some more and see if I can come up with a clean solution that doesn't add needless bloat.

fatcerberus:
Regarding this:
> I tend to think like @harold-b that the debug client should just display the state the target is in, so that I don't care about the specific transitions

Just keep in mind that that's not always viable--for example, in the minisphere GUI debugger it automatically refocuses the game window when resuming execution.  You wouldn't want to do that with every single "running" notify because it would be quite annoying to the user. ;-)

svaarala:
Yeah, there are definitely issues like that in state based approaches in general. But for the Duktape execution status that's most definitely not a problem:

- There's always a clear call site which pauses the target. This moves Duktape from RUNNING to PAUSED-BY-XXX.
- There's always a clear request to resume which moves Duktape from PAUSED-BY-XXX to RUNNING.

Because the request to Resume is explicit, the transitions cannot be "coalesced" accidentally.

svaarala:
> Just keep in mind that that's not always viable--for example, in the minisphere GUI debugger it automatically refocuses the game window when resuming execution. You wouldn't want to do that with every single "running" notify because it would be quite annoying to the user. ;-)

Right, you'd do that exactly when "old_state == PAUSED && new_state == RUNNING".

harold-b:
> Just keep in mind that that's not always viable--for example, in the minisphere GUI debugger it automatically refocuses the game window when resuming execution. You wouldn't want to do that with every single "running" notify because it would be quite annoying to the user. ;-)

That's write but since we keep a local state we can just ignore the resumes when we're already in resume, that's what I do hehe :) Since resume in this case all it really does is enable the pause button and disable the stepping controls.

I only adopt the latest state when it's a paused state, in which case we are responsible for asking duk for the current info.

fatcerberus:
I'll close this pull.  If a better solution is found, I'll open another one.

PR #572, opened 2016-02-02
Add trivial readFile() and writeFile() calls to 'duk' command line for easier testing

svaarala:
These are not official bindings (and not enabled for "duk" by default either); just something to read/write e.g. JSON files when testing.


svaarala:
Seems to work on Windows with MSVC too (and it's optional so this is not critical in any case).

Issue #130, opened 2015-03-09
API to replace duk_put_function_list() and duk_put_number_list()


svaarala:
There are two API calls to conveniently initialize objects from C:

- http://duktape.org/api.html#duk_put_function_list
- http://duktape.org/api.html#duk_put_number_list

There are a few downsides to these:

- Separate calls are needed to initialize functions and numbers
- There's no support for all common types (e.g. booleans)
- There's no support for lightfunc initializers

Also, the current approach exposes two specific C structs used as the initializer, which was probably a mistake: C structures are difficult to extend in a binary compatible fashion.  A more extensible approach would be to use macros for both declaring the initializer array and for initializer entries; this would allow the underlying structures to be extended to e.g. support lightfuncs.  Also the same API call could provide initializers for both functions and numbers (and other constants).

svaarala:
One problem with using initializer macros is that the most natural initializer would involve a type/flags field and a union of some sort for the value alternatives. But union initializers are not portable to pre-C99 compilers.

svaarala:
Having the ability to provide property attributes too (if user wants to use non-default values) would also be useful.

svaarala:
One option for portable initialization is to:

- Use unions as the property list value type
- Use named union initializers in C99 compilers, as they're quite widely supported
- Fall back to using helper functions to fill out the entries for other compilers

Example of a helper:

```
duk_proplist_union duk_make_proplist_number(const char *key, duk_double_t numval) {
    duk_proplist_union u;
    u.key = key;
    u.number = numval;
    return u;
}
```

C99 compilers would just use the initializer:

```
    { .key = (key), .number = (numval); }
```

For both the initializer list entry would look identical, e.g.:

```
    ...
    DUK_PROP_NUMBER("myprop", 123.0),
    ...
```

svaarala:
Another portable option which is a lot messier is to use a struct initializer, but share the fields so that e.g. both void and string pointers can share the same field in the struct. This doesn't result in as optimal a layout as the union approach because there would be both data and function pointers in the struct etc.

svaarala:
Here's a very basic sketch of how that union initializer/helper call model would work:

- https://github.com/svaarala/duktape/compare/api-new-list-inits

This is just to document what the most potential approach seems to be so far; I'll return to this later on for Duktape 1.4.

svaarala:
Helper functions may be an issue for global values.

darkuranium:
[...]

PR #574, opened 2016-02-06
put_function_list with magic

lzubiaur:
Hi, I find it convenient to be able to specify the `magic` flag when bulk loading functions using `duk_put_function_list`. As an example, I opened this pull request which adds the function `duk_put_function_list_magic` and struct `duk_function_list_magic_entry` with the field `magic`.

One could bulk load functions in a module and specify magics like this:

```
const duk_function_list_magic_entry module_funcs[] = {
    { "text",                  fun_text,    DUK_VARARGS,    1 },
    { "textColored",     fun_text,    DUK_VARARGS,     2 },
    { "textDisabled",    fun_text,    DUK_VARARGS,    3 },
    { "textWrapped",   fun_text,    DUK_VARARGS,     4 },
    { NULL, NULL, 0 }
};

duk_put_function_list_magic(ctx, -1, module_funcs);
```

### DIFF
--- a/src/duk_api_object.c
+++ b/src/duk_api_object.c
@@ -475,6 +475,24 @@ DUK_EXTERNAL void duk_put_function_list(duk_context *ctx, duk_idx_t obj_index, c
 	}
 }
 
+DUK_EXTERNAL void duk_put_function_list_magic(duk_context *ctx, duk_idx_t obj_index, const duk_function_list_magic_entry *funcs) {
+	const duk_function_list_magic_entry *ent = funcs;
+
+	DUK_ASSERT_CTX_VALID(ctx);
+
+	obj_index = duk_require_normalize_index(ctx, obj_index);
+	if (ent != NULL) {
+		while (ent->key != NULL) {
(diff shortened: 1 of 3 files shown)


svaarala:
The main problem with adding another API is that eventually there'll be another property that would be convenient to add, leading to yet another API call, etc. Also, it'd be nice to be able to specify functions and constants in the same list.

So the next planned step in this direction is to add an API call where the struct initializers are macros so that it's possible to add/remove fields without changing the external API contract, see discussion in: https://github.com/svaarala/duktape/issues/130.

svaarala:
As a concrete example of the planned approach: https://github.com/svaarala/duktape/blob/741c60262155e3d7e2f9bd45436cc98884628e09/tests/api/test-put-prop-list.c

lzubiaur:
An union + macro initializers approach makes more sense indeed. Thanks for your reply.

svaarala:
Other than this the pull would have been nice to merge with testcase updates etc.

That new API call has been on hold for a while but I think I could merge an initial version into master at this point and we can then provide a function+magic initializer through that.

PR #575, opened 2016-02-07
Add duk_put_prop_list() API call with more flexible and versionable initializers

svaarala:
Implement the new API call described in #130. Basic approach is to hide the property initializer structure (so it's not a public part of the API) and use initializer macros to define property keys and values. This makes the API easier to version; new fields and macros can be added without breaking syntax compatibility as is the case with `duk_put_function_list()` and `duk_put_number_list()`. The initializer macro approach uses union initializers when possible (C99 and some C++ compilers) and falls back to portable struct initializers.

Tasks:

- [ ] Add DUK_USE_UNION_INITIALIZERS config option and baseline detection for supported compilers
- [ ] Add public API struct/union format and initalizer macros
- [ ] Add public API call `duk_put_prop_list()` and implement the initialization code
- [ ] API test case
- [ ] API documentation
- [ ] Deprecation note for duk_put_function_list() and duk_put_number_list(), ensure these calls are in Duktape 2.x removal list
- [ ] Placeholder in 1.5.0 release note for noting that users should migrate to the new call
- [ ] Open issue: should property attributes be included in the property description? If so, proper API call name would maybe be `duk_def_prop_list()`
- [ ] Open issue: ability to provide property keys through value stack for special values?


svaarala:
Still pretty much a placeholder, I'll rework the function helper initializers into a union+struct approach based roughly on: https://gist.github.com/darkuranium/565a8526abd2de078371

svaarala:
So far the property key/value initializer size is:

- x64 with union initializers: 40 bytes
- x86 with union initializers: bytes
- x86 without union initializers: 36 bytes

The struct contains a type/flags, key, and value. The most complicated value so far is a native function which needs a function pointer, nargs, length, and magic.

svaarala:
Here's the initializer struct so far:

- https://github.com/svaarala/duktape/blob/0c27ad940847f5d7a6120af309fad978581cc9bc/src/duk_api_public.h.in#L89-L125

svaarala:
Pushed some cleanups:

- The union/struct difference no longer duplicates code, value choice is hidden behind `DUK__PROP_SELECT()` local macro.
- Added undefined, null, boolean, pointer, function, lightfunc.

Accessors next, but that means defining them using `duk_def_prop()` so this API call should probably be renamed `duk_def_prop_list()`. It also means property attributes could be specified explicitly which would also be nice.

If it works out, I'll also add an object value type which allows a value to point to another property initializer so that you can describe nested structured directly.

svaarala:
The x64 code size for `duk_put_prop_list()` is 330 bytes now; `duk_put_function_list()` and `duk_put_number_list()` are about 140 bytes so after deprecating them this branch will add 200-250 bytes of footprint. Not bad for a pretty generic feature :)

svaarala:
Switched to using `duk_def_prop()` to set the properties; attributes are automatic for now and hidden inside the macros because they're not always needed. Adding separate macros with explicit attributes (e.g. `DUK_PROP_STRING_ATTR(...)`) will be possible later.

There's also support for subobjects:

```
const duk_prop_list_entry subprops[] = {
    DUK_PROP_STRING("name", "sub object"),
    DUK_PROP_END()
};
const duk_prop_list_entry props[] = {
    DUK_PROP_STRING("name", "main object"),
    DUK_PROP_PROPLIST("child", subprops),
    DUK_PROP_END()
};
```

This creates the structure:

```
{
    name: "main object",
    child: {
        name: "sub object"
    }
}
```

Not sure if this should stay in (or be renamed). If it stays, it'd also be nice to define arrays, which would need a `duk_prop_list_value` structure (one without a key) and its own set of macros (maybe too awkward to support).

svaarala:
I actually have VS2013 so tested that union initializer thing: at least for me union-inside-struct works.

fatcerberus:
I didn't see anything above about this, will this support defining accessors?  I've started to use `duk_put_function_list()` to construct native modules, but there's there's no equivalent call for adding accessor properties.

svaarala:
Yes, accessors will look something like:

```c
DUK_PROP_ACCESSOR("accessorValue", my_getter, my_setter)
```

The nice thing with this pull is that the struct holding the initialization values is hidden so that it can be extended and new initializer macros added without breaking compatibility (at least in almost all cases).

There's still one major issue to consider which is why I dropped this from 1.5.0: should there be an initializer for defining arrays (rather than key/value objects)? If so, the initializer format needs to be split into a key and a value part, with an array defined solely as a list of values.

fatcerberus:
Hm, I personally haven't had any cases where I wanted to initialize an array in bulk AND all the values were known in advance.  Normally if you're pushing an array onto the value stack you're going to fill it with values from some native array, which this call isn't suited for.

svaarala:
One case where this can appear is an array used to convert integer error codes (or other enum values) to strings.

svaarala:
The property initializer size is still relatively large in this pull. This is the case because a single union structure represents all the possible value types, and the largest value types like setter/getter pairs require several pointers.

I'll try one more approach based on the initializer list consisting of an array of parts, with each part a union of a pointer, a function pointer, a `duk_uint_t`, and a `duk_double_t`. One property initializer macro would then emit one `duk_uint_t` initializer to encode type, flags, and attributes, another initializer to encode the property key, and then 0-N initializer parts depending on the value. This allows e.g. a double constant to be encoded in 3 parts which would usually come out as 3x8 = 24 bytes; on x64 the current approach would take 40 bytes even with union initializers.

The downsides of this approach are that:
- The initializer macros cannot be placed in parentheses because they emit multiple items per macro.
- The array size of the initializer sequence is not known beforehand and doesn't match apparent property count.

svaarala:
For x86 with union initializers the initializer size is 24 bytes at present and wouldn't be smaller with the alternative approach. However, if a `duk_double_t` wasn't a part of the union, the union could be 4 bytes and a typical integer/string property could be represented using 12 bytes which is close to optimal.

Doubles could be split into two component parts and combined during initialization (taking alignment into account), but splitting a double with a static initializer may be a little bit tricky.

PR #579, opened 2016-02-10
Add Windows debug transport example

svaarala:
Add a Windows variant of the debug transport example and ensure `duk --debugger` works on Windows.

Tasks:

- [x] Add Windows debug transport example
- [x] Update internal and external documentation
- [x] Debugger Makefile trivia for Windows (still Cygwin but support e.g. spaces in Node.js path)
- [x] Test on Windows with Linux debug client
- [x] Test on Windows with Windows debug client
- [x] Releases entry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -171,7 +171,7 @@ DUKTAPE_CMDLINE_SOURCES = \
 	dist/examples/alloc-logging/duk_alloc_logging.c \
 	dist/examples/alloc-torture/duk_alloc_torture.c \
 	dist/examples/alloc-hybrid/duk_alloc_hybrid.c \
-	dist/examples/debug-trans-socket/duk_trans_socket.c
+	dist/examples/debug-trans-socket/duk_trans_socket_unix.c
 
 # Compiler setup for Linux
 CC	= gcc
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1381,6 +1381,8 @@ Planned
   type is indicated using a fake directive rather than a comment; for example,
   'function () {"ecmascript"}' (GH-554)
 
+* Add Windows version of the debugger example TCP transport (GH-579)
+
 * Fix potentially memory unsafe behavior when a refcount-triggered finalizer
   function rescues an object; the memory unsafe behavior doesn't happen
   immediately which makes the cause of the unsafe behavior difficult to
--- a/debugger/Makefile
+++ b/debugger/Makefile
@@ -14,14 +14,14 @@ all: run
 
 .PHONY: run
 run: node_modules static/socket.io-1.2.0.js static/jquery-1.11.1.min.js static/reset.css static/jquery-ui.min.js static/jquery-ui.min.css static/images
-	$(NODE) duk_debug.js --source-dirs=$(SOURCEDIRS)
+	"$(NODE)" duk_debug.js --source-dirs=$(SOURCEDIRS)
 
 rundebug: node_modules static/socket.io-1.2.0.js static/jquery-1.11.1.min.js static/reset.css static/jquery-ui.min.js static/jquery-ui.min.css static/images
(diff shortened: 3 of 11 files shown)


svaarala:
Work in progress still, I'll try to get this working today.

svaarala:
Ok, this works for me at least on MSVC 2013.

Tried both a Windows and a Linux debug client.

svaarala:
Any comments on Windows best practices from more experienced Windows programmers welcome :-)

fatcerberus:
I assume you mean in general?

One thing to watch out for is paths - Windows itself will happily parse paths using `/` separators, however the default shell (cmd.exe) will sometimes misinterpret them as command options.  If you're invoking the shell for any reason, better use backslashes.

svaarala:
I mean if the `duk_trans_socket_windows.c` socket code is doing something contrary to best WIndows practices.

fatcerberus:
I don't think it's necessary to check whether _WIN32_LEAN_AND_MEAN is defined before defining it: as far as I'm aware, in no case will MSVC define it by default.

The socket code looks good.  Winsock isn't all that different from UNIX/BSD sockets, so nothing much to go wrong there.  The only thing I notice is that you're using AF_INET, which implies IPv4 and will fail on IPv6-only systems.  Probably not an issue in practice though, and it's just example code anyway.

The one thing that legitimately bugs me is this:
 ```c
#pragma comment (lib, "Ws2_32.lib")
```

That probably won't work with any compiler other than MSVC.  MinGW uses UNIX-style library names, for example (libfoo.a)

svaarala:
Both the _WIN32_LEAN_AND_MEAN (though not checking before defining it, that was from some other example) and the pragma are from this example: https://msdn.microsoft.com/en-us/library/windows/desktop/ms737593(v=vs.85).aspx

Should AF_INET be AF_UNSPEC instead?

I'll place that pragma inside a `#if defined(_MSC_VER)`.

fatcerberus:
AF_UNSPEC is probably the best choice.  Beware that, if the hostname is NULL, it will sometimes prefer IPv6 if both protocols are available (speaking from experience :) ) But this can easily be overridden by using a "catch-all" hostname for the desired protocol (`0.0.0.0` or `::`)

svaarala:
Hmm, so would you suggest using "0.0.0.0" and AF_UNSPEC?

svaarala:
Minor fixes here (I'll squash this later): https://github.com/svaarala/duktape/commit/15704aef53e91b4474429fba0f2d9cecddaecc1e

svaarala:
To clarify, would you suggest the default to be "0.0.0.0" and AF_UNSPEC, and the user can then override DUK_DEBUG_ADDRESS to something else if necessary?

fatcerberus:
Just for the record, I think that would have the same outcome as AF_INET if the system has no IPv4 support.  The only way to support v4 and v6 simultaneously is to open two listening sockets.

For this example code dual sockets is probably complicating things too much.  The DUK_DEBUG_ADDRESS with a default hostname is a good compromise. :)

svaarala:
Ok, I'll leave that in and retest, rebase, and merge unless something comes up. Thanks! :)

fatcerberus:
The reason for the above is that Windows implements IPv4 and IPv6 as completely separate networking stacks.  I don't think this is the case in, e.g. Linux.

svaarala:
Right, AF_UNSPEC is then maybe preferable because one can then just `-DDUK_DEBUG_ADDRESS=xxx` to make the transport bind to IPv6. If AF_INET was in place you'd need to tweak the source.

Quite a minor detail of course.

svaarala:
Ok, things still seem to work fine with VS2013.

svaarala:
For MinGW needed to apply this workaround to get around "missing getaddrinfo()" during linking: http://programmingrants.blogspot.fi/2009/09/tips-on-undefined-reference-to.html

svaarala:
Ok, with the fix both VS2013 and MinGW work correctly. Also added example command lines into the source comment to illustrate correct MinGW link options. I'll merge once the tests pass.

PR #580, opened 2016-02-10
Add inline macros for MSVC

svaarala:
- [x] Add DUK_NOINLINE, DUK_INLINE, and DUK_ALWAYS_INLINE for Windows
- [x] Figure out correct condition for MSVC version - use VS2008 as a limit for now
- [x] Test on VS2013 at least (what I have installed) - works
- [x] Check effect on binary size - adds around 22kB to unstripped x64 binary
- [x] Check effect on performance - only relevant effect is in dump/load right now

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1487,6 +1487,8 @@ Planned
 * Portability improvement for VS2012 C++ compilation: avoid double definition
   of replacement double constants (GH-595)
 
+* Portability improvement for Windows MSVC: add inline macros (GH-580)
+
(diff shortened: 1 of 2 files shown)


svaarala:
Based on AppVeyor tests works at least from VS2008 so I'll make that the check for these. For older VS versions the inline attributes won't be present but they're not mandatory so this is not a big issue.

fatcerberus:
I believe VS2005 just hit EOL, so it shouldn't be a problem if Duktape doesn't support it.

PR #582, opened 2016-02-13
Use raw value stack accessors internally when it's safe to do so

svaarala:
For example, when internal code pushes a value and then does:

```
tv = duk_get_tval(ctx, -1);
```

it can be changed to:

```
tv = DUK_GET_TVAL_NEGIDX(ctx, -1);
```

which ultimately compiles to:

```
tv = ((duk_hthread *) ctx)->thr_valstack_top - 1;
```

This leads to both smaller and (slightly) faster code. The effects are not very large for the changes in this pull request (about 100 bytes smaller gcc -Os build and 800 bytes smaller gcc -O2 build).

The internal accessors are unsafe so the call site must be 100% certain that no validation is needed for the index.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1405,6 +1405,9 @@ Planned
 * Change OS string (visible in Duktape.env) from "ios" to "osx" for non-phone
   targets (GH-570, GH-571)
 
+* Internal performance improvement: use raw value stack accessors internally
+  when it's safe to do so (GH-582)
+
 2.0.0 (XXXX-XX-XX)
 ------------------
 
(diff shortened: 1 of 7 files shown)


svaarala:
Mostly the performance changes are marginal, these seem to be positively affected:

```
test-assign-add.js                  : duk.O2  4.90 duk.O2.master  5.22
test-assign-addto.js                : duk.O2  4.93 duk.O2.master  5.24
test-assign-proprhs.js              : duk.O2  4.09 duk.O2.master  4.47
test-mandel.js                      : duk.O2  4.27 duk.O2.master  4.53
test-prop-read.js                   : duk.O2  4.68 duk.O2.master  5.08
test-prop-write.js                  : duk.O2  4.21 duk.O2.master  4.13
```

fatcerberus:
Actually it looks like `test-prop-write` either took a hit or at least is the same (4.21 <-> 4.13 seems like it might be within the margin of error for benchmark results)

svaarala:
It's within the usual variance - it's not a very accurate measurement.

svaarala:
What's more, there are no code path changes for property write and all changed paths contain less code in any case. So logically there should be no adverse impact other than functions getting aligned differently etc which is not really worth worrying about (it changes on every code change).

PR #590, opened 2016-02-16
Add a debug JSON proxy written using DukLuv

svaarala:
Many users have wanted to use the JSON proxy for debugger integration but the current JSON proxy example is written using Node.js which, while portable, is sometimes a too heavy dependency for a debug client.

This pull adds a JSON proxy written using DukLuv which is much smaller.

Tasks:

- [x] Resolve FIXMEs and clean up
- [x] Command number mappings, metadata source file in JSON?
- [x] Virtual notifications for `_TargetConnected` etc
- [x] Clean up virtual notifications to clearly separate notifications related to target connection as opposed to JSON proxy connection
- [x] Add command line options for target host/port, metadata files, etc; depends on https://github.com/creationix/dukluv/pull/17, but if the support is not present will just skip command line parsing
- [x] Options for output verbosity
- [x] Add to dist
- [x] Add a README note, explain DukLuv setup on Windows/Unix
- [x] Test on Linux and Windows
- [x] Update internal documentation
- [x] Update external documentation
- [x] Split client and target commands in command list metadata
- [x] Makefile target to build joint metadata JSON and add that to dist
- [x] Releases


svaarala:
Still quite a bit of cruft but already functional to some extent.

JSON proxy session for running mandelbrot (just Resume immediately):

```
sva@keppihevonen:~$ telnet localhost 9093
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
{"notify":"_Connecting","args":["127.0.0.1",9091]}
{"notify":true,"command":1,"args":[1,"test-dev-mandel2-func.js","global",58,0]}
{"request":19}
{"reply":true}
{"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","global",58,0]}
{"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","mandel",88,54]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["........................................#.......................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["....................................####........................................\n"]}
{"notify":true,"command":2,"args":["....................................####........................................\n"]}
{"notify":true,"command":2,"args":["..............................#..##########.....................................\n"]}
{"notify":true,"command":2,"args":["..............................#################.................................\n"]}
{"notify":true,"command":2,"args":["............................###################.................................\n"]}
{"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","mandel",88,54]}
{"notify":true,"command":2,"args":["...........................#####################................................\n"]}
{"notify":true,"command":2,"args":[".................#######..######################................................\n"]}
{"notify":true,"command":2,"args":["................#########.######################................................\n"]}
{"notify":true,"command":2,"args":[".#############################################..................................\n"]}
{"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","mandel",88,54]}
{"notify":true,"command":2,"args":["................#########.######################................................\n"]}
{"notify":true,"command":2,"args":[".................#######..######################................................\n"]}
{"notify":true,"command":2,"args":["...........................#####################................................\n"]}
{"notify":true,"command":2,"args":["............................###################.................................\n"]}
{"notify":true,"command":2,"args":["..............................#################.................................\n"]}
{"notify":true,"command":2,"args":["..............................#..##########.....................................\n"]}
{"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","mandel",88,54]}
{"notify":true,"command":2,"args":["....................................####........................................\n"]}
{"notify":true,"command":2,"args":["....................................####........................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["........................................#.......................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":2,"args":["................................................................................\n"]}
{"notify":true,"command":6,"args":[0]}
{"notify":"_Disconnecting"}
```

And the output from the DukLuv JSON proxy:

```
duktape@keppihevonen:~/duktape/debugger$ dukluv duk_debug_proxy.js 
Listening for incoming JSON debug connection on 0.0.0.0:9093
*** JSON proxy client connected, connecting to target...
PROXY --> CLIENT: {"notify":"_Connecting","args":["127.0.0.1",9091]}
*** Connected to target
*** Target handshake: {"line":"1 10499 v1.4.0-132-g4ed0c7e duk command built from Duktape repo","protocolVersion":1,"text":"10499 v1.4.0-132-g4ed0c7e duk command built from Duktape repo","dukVersion":"1","dukGitDescribe":"10499","targetString":"v1.4.0-132-g4ed0c7e"}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |81|
PROXY <-- TARGET: |81|
PROXY <-- TARGET: |78746573742d6465762d6d616e64656c322d66756e632e6a73|
PROXY <-- TARGET: |66676c6f62616c|
PROXY <-- TARGET: |ba|
PROXY <-- TARGET: |80|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":1,"args":[1,"test-dev-mandel2-func.js","global",58,0]}
PROXY <-- CLIENT: {"request":19}
PROXY --> TARGET: |01|
PROXY --> TARGET: |93|
PROXY --> TARGET: |00|
PROXY <-- TARGET: |02|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"reply":true}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |81|
PROXY <-- TARGET: |80|
PROXY <-- TARGET: |78746573742d6465762d6d616e64656c322d66756e632e6a73|
PROXY <-- TARGET: |66676c6f62616c|
PROXY <-- TARGET: |ba|
PROXY <-- TARGET: |80|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","global",58,0]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |81|
PROXY <-- TARGET: |80|
PROXY <-- TARGET: |78746573742d6465762d6d616e64656c322d66756e632e6a73|
PROXY <-- TARGET: |666d616e64656c|
PROXY <-- TARGET: |c058|
PROXY <-- TARGET: |b6|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","mandel",88,54]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["........................................#.......................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["....................................####........................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["....................................####........................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232e2e232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["..............................#..##########.....................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e23232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["..............................#################.................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["............................###################.................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |81|
PROXY <-- TARGET: |80|
PROXY <-- TARGET: |78746573742d6465762d6d616e64656c322d66756e632e6a73|
PROXY <-- TARGET: |666d616e64656c|
PROXY <-- TARGET: |c058|
PROXY <-- TARGET: |b6|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","mandel",88,54]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2323232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["...........................#####################................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232323232323232e2e232323232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":[".................#######..######################................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2323232323232323232e232323232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................#########.######################................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2323232323232323232323232323232323232323232323232323232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":[".#############################################..................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |81|
PROXY <-- TARGET: |80|
PROXY <-- TARGET: |78746573742d6465762d6d616e64656c322d66756e632e6a73|
PROXY <-- TARGET: |666d616e64656c|
PROXY <-- TARGET: |c058|
PROXY <-- TARGET: |b6|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","mandel",88,54]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2323232323232323232e232323232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................#########.######################................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232323232323232e2e232323232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":[".................#######..######################................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2323232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["...........................#####################................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232323232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["............................###################.................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e23232323232323232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["..............................#################.................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232e2e232323232323232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["..............................#..##########.....................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |81|
PROXY <-- TARGET: |80|
PROXY <-- TARGET: |78746573742d6465762d6d616e64656c322d66756e632e6a73|
PROXY <-- TARGET: |666d616e64656c|
PROXY <-- TARGET: |c058|
PROXY <-- TARGET: |b6|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":1,"args":[0,"test-dev-mandel2-func.js","mandel",88,54]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["....................................####........................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232323232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["....................................####........................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e232e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["........................................#.......................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |82|
PROXY <-- TARGET: |1200512e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e0a|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":2,"args":["................................................................................\n"]}
PROXY <-- TARGET: |04|
PROXY <-- TARGET: |86|
PROXY <-- TARGET: |80|
PROXY <-- TARGET: |00|
PROXY --> CLIENT: {"notify":true,"command":6,"args":[0]}
*** Target disconnected
PROXY --> CLIENT: {"notify":"_Disconnecting"}
```

Pretty verbose at the moment, and there are no command line options etc. This kind of output is useful for working with the binary protocol because it shows the JSON and binary translation back and forth.

svaarala:
@creationix Some Dukluv fun in this pull :)

svaarala:
@JoshEngebretson This might also interest you if you are still considering debugger options.

fatcerberus:
What exactly is Dukluv, is it like Node.js but using Duktape instead of V8?

svaarala:
http://dukluv.io/ provides LibUV bindings for Duktape, which allow Node.js like programming (async callbacks) but the details are a bit different. Works great out of the box, and has support for timers, sockets, etc.

svaarala:
Also tested this on Windows, seems to work fine with the following steps:

- Ensure you have cmake installed
- Clone DukLuv, `git submodule init; git submodule update`
- `make` in DukLuv directory running in e.g. VS2013 command prompt
- Go to `duktape/debugger`, copy `modules/` from DukLuv into that directory
- Run `...\dukluv\build\Debug\dukluv duk_debug_proxy.js`

The proxy is now running. You can then compile `duk` with Windows debug transport and:

- Run `duk --debugger mandel.js`

Finally, connect to `localhost:9093` using telnet or netcat and do debug stuff through the proxy.

svaarala:
From a debug client packaging perspective it'd now be enough to package the DukLuv exe(s) into the debug client and ensure `duk_debug_proxy.js` and the `modules` directory are present for including `Tcp` module. That should be enough to run the proxy with no further dependencies.

fatcerberus:
What's the license on DukLuv?

svaarala:
MIT: https://github.com/creationix/dukluv/blob/master/LICENSE

fatcerberus:
Good to know.  Just wanted to make sure packaging the dukluv executable with the client wouldn't be a licensing issue. :)

svaarala:
License shouldn't be an issue, and DukLuv works on Linux, OSX, and Windows so it's pretty great :)

creationix:
This is very cool, let me know if you need any changes to dukluv to make this work!

svaarala:
Removed the dependency on DukLuv modules/classes.js (which provides e.g. a `Tcp` wrapper), so that only a DukLuv binary and the proxy source file are needed.

fatcerberus:
I'm tempted to try this out now.  It'd probably be much cleaner to use the JSON protocol for the C# debugger, but I'm reluctant to blow away all the binary dvalue parsing I wrote...

For the console debugger which is written in C, I actually prefer the binary protocol though.

svaarala:
Just note that this branch is still under development: it's functional but error handling is mostly missing, there may be bugs in dvalue details, etc. But it does work for prototyping :-)

Longer term it might actually make more sense to make the JSON interface the more important one and make the binary a much more low level option. Some convenience operations could be implemented only in the JSON protocol/proxy rather than in the low level protocol - for example, breaking down an error into its components could be a convenience method implemented using the low level Eval mechanism operating on the error.

This would allow the target to remain minimal but make the debug protocol more convenient to use. The downside would be that there would be two conceptual protocols though, and the proxy functionality would become a part of the official implementation (because it'd be the only place implementing those convenience operations). In any case I'd bet that the majority would prefer the JSON mapping if it was "promoted" as the default interface.

fatcerberus:
I guess that would essentially make the JSON "proxy" more of just another layer in the protocol stack.  Currently there are actually two levels, as I came to realize while refactoring SSJ: the low-level dvalue protocol, with a messaging protocol on top of it.  JSON would make it three. :)

svaarala:
That's one way of looking at it - though I'd rather think it would be layer 2+: it'd conceptually have the same dvalue/dmessage structure but it'd just extend the available command set.

It would more clearly be a separate layer if the format was also different, say it supported structured values, nested structures, etc. That would also be a useful architecture but quite different from what's there now.

fatcerberus:
The dual-level nature of the binary protocol is a big reason why I had so much trouble understanding it at first BTW.  It's theoretically possible to write a client which only deals at the dvalue level, but very difficult - it's much cleaner if the high level client only deals in "messages".

svaarala:
The intent is actually that clients would process full messages at a time instead of basically staring at a stream of dvalues. So I think a client should normally queue finished dvalues until an EOM is found, and then process the message. If the message is not supported or ignored otherwise, just drop the dvalues and move on. This is how I'd implement a C client unless something forced me to do otherwise.

This is one reason why there is an unambiguous, message independent EOM: it allows the client to queue up dvalues with knowing the message type and only start processing at the EOM.

svaarala:
Duktape itself avoids queueing at all costs which is why that implementation is based on parsing dvalues on the fly. Otherwise it wouldn't work on very low memory targets. For that implementation the key factor is the ability to parse dvalues with minimal overhead and being able to skip over a message without knowing its structure (again relying on EOM).

fatcerberus:
I agree - all I'm saying is that conceptually, it's actually 2 stacked protocols rather than a single one as the documentation seems to imply.  A Duktape debug peer can either deal with dvalues directly or with whole dmessages, in much the same way that it's possible to write an HTTP peer that deals in raw sockets vs. using something like libcurl

Just something to keep in mind to perhaps improve the documentation. :)

svaarala:
Hmm. I still don't see them as separate protocol layers. Or at least the relationship is similar to what exists between a tokenizer and a compiler - one deals in characters and the other in tokens, but conceptually they are part of the same syntax. I see your point though, perhaps the document should be structured more clearly along describing a dvalue stream and then how that is parsed into messages.

svaarala:
Pushed a few more changes:

- Fixed a few bugs in handling of e.g. number and pointer dvalues.
- Added option `--readable-numbers` to add `.value` for IEEE double values: it's human readable only and ignored otherwise, but makes it easier to read the dumps. Note that command line parsing relies on a yet-to-be-merge DukLuv pull.
- Fixed request/notify name/number parsing to match the Node.js proxy. Concretely: if "request" is a name it is resolved. If that resolution fails the "command" key can provide the raw command number. "request" key is required but it can also be just `"request": true` with the raw command number in "command".

fatcerberus:
I think the main thing that differentiates the two parts of the protocol for me is that dvalues have intrinsic meaning (whereas tokenizer output in isolation is meaningless).  Eval, for example, could simply respond with a dvalue holding the result, but for the fact it has to follow the rules for messaging, and thus is forced to wrap the result in a `REP EOM`.  Obviously this would be a versioning nightmare, but it would be entirely possible in principle.

Anyway, I think I've done enough issue-derailing for one day, this isn't really related to the DukLuv proxy at all now is it? :)

svaarala:
It's indeed a bit off topic but I can't help wanting to respond :-)

I think the tokenizer/compiler analogy is actually pretty accurate. Tokens also carry type like dvalues, e.g. "identifier", "equal sign", "number", "semicolon". That token sequence would then make up the statement "x = 123;". Individual tokens do carry type and value but can't be used in isolation but must instead be part of a expression/statement structure.

Anyway, I see your point and I'll try to see if I can improve the debugger document structure a little bit when I next need to touch it in a major way :)

creationix:
I saw concerns about JSON and bloat size.  I would not be opposed to bundling msgpack bindings in with dukluv. I feel msgpack fits very nicely with the goals of dukluv.  It is semantically very much like JSON, but more compact and also support binary values which JSON does not.

fatcerberus:
Huh, I looked up messagepack, it's very similar in spirit and execution to Duktape's dvalue protocol.  Might be worth looking into in Duktape's future to align the dvalue representation with it.

svaarala:
In addition to MsgPack there's CBOR which is quite similar: https://tools.ietf.org/html/rfc7049, and I'm sure several others. The basic idea is to have very short encodings for small integers and short strings while retaining space for a wide set of tagged types. The main difference are Duktape specific types (which are a must to some extent but can be represented in some compatible way too) and dvalues not providing a key/value map.

A key/value map is not used because parsing it on the Duktape side would very easily lead to buffering which is avoided at all costs. However, that's not actually a reason not to have key/value pairs in the low level format: they could still be used in responses sent by Duktape because that doesn't have any buffering concerns and would make it easier to model a few commands.

Changing the dvalue format to something else is definitely a possibility if it has concrete benefits. I guess the main benefit would be if you could pick up a library and parse the data with the library with minimal adaptation. If you couldn't do that there wouldn't be that much of an advantage because these compact formats are more or less equally tedious to parse.

svaarala:
@creationix I'm not sure what bloat you're referring to regarding bloat issues.

If it was about the debugging model, the reason why JSON is not used directly on the target is memory bloat issues. It would be next to impossible to run the debugger with a fixed <1kB memory overhead (necessary on very low memory targets) with a JSON based protocol. This is why the target always talks binary only and the JSON part is proxied from that.

JSON on the proxy is not an issue, and other engines (like V8) use JSON for their debug protocol too. It's trivial to parse in dozens of languages and also nicely readable, so I can't really think of a better base protocol when there are no memory concerns. For debugging purposes I don't think JSON encoding/decoding on an ordinary laptop/desktop is a relevant concern - maybe there are exceptions to that.

svaarala:
@creationix As a sidenote, DukLuv is quite nice to work with and I don't actually find the `uv` bindings to be too low level. On the contrary, it's nice to be able to deal with the event loop directly without abstracting classes in the middle. The most annoying part so far has been mixing different buffer types which is still a bit in awkward in Duktape 1.4.0, but that's not really a DukLuv issue ;-)

svaarala:
Opened #593 for that debugger dvalue discussion. It's a nice design topic because it's relatively contained and there's clear room for improvement.

svaarala:
Tested all the dvalue parsing for dvalues coming from the target. There was one more bug but now all types seem to work correctly.

Also tested all dvalue encoding for dvalues going to the target (this can be done using PutVar). Duktape doesn't parse all dvalue types for inbound PutVar values but I could verify that at least the values were dvalue converted.

svaarala:
Alright, this is pretty close to complete, it's still missing better error handling and I'll run a few more tests but functionally this branch is pretty much complete.

fatcerberus:
> it's nice to be able to deal with the event loop directly

If this means what I think it does, it's indeed a very nice design.  minisphere (and the Sphere engine it's based on) uses a similar design, the simplest possible minisphere program is:

```javascript
while (true) FlipScreen();
```

Which just displays a black screen until you close the engine.  Being from a C background, being able to write programs like this feels very natural.

creationix:
@svaarala 

> it's nice to be able to deal with the event loop directly without abstracting classes in the middle

Yes, I love using the libuv API directly when I'm in lua.  I'm not happy with the current bindings in dukluv because of the way it encourages you to wrap things in classes.  If you have suggestions to making the API easier for direct use I'm all ears.

svaarala:
One thing in Javascript in general is the notorious "this" binding. I ended up using the bind() approach, for example:

```js
uv.tcp_connect(targSock.handle, targetHost, targetPort, targSock.onConnect.bind(targSock));
```

Some example code approaches this another way.

fatcerberus:
@svaarala We're getting off-topic again it seems. :)

I used to hate the semantics of `this` in JS and it indeed can be confusing, but having gotten used to it I actually rather like the way it works.  It's quite useful to be able call any function with an arbitrary `this`, and you can easily build delegates, etc. on top of that.

svaarala:
Memory usage for the DukLuv based proxy seems pretty reasonable. This is for running through hundreds of test cases:

```
    MB
7.809^                           #                                            
     |                           #                                            
     |                          @#                                            
     |                          @#                   ::  ::                   
     |                       @  @#                   :   :                    
     |                       @::@#                   :   :                    
     |                    @@ @: @#                  ::   :                    
     |                    @ :@: @#              :  ::: : :                    
     |                    @ :@: @#              :  ::: :::                    
     |                    @ :@: @#              :  ::: :::                    
     |                    @ :@: @#              :  ::: :::                    
     |                 :::@ :@: @#        :: ::::::::: :::                    
     |                ::: @ :@: @#        : @:: :: ::: :::                    
     |                ::: @ :@: @#::  :: :: @:: :: ::: :::                    
     |            ::  ::: @ :@: @#:   : ::: @:: :: ::: :::                    
     |            : ::::: @ :@: @#: ::: ::: @:: :: ::: :::   ::               
     |        @@ :: : ::: @ :@: @#: : : ::: @:: :: ::: ::: ::: : ::::  :::: ::
     |       @@ ::: : ::: @ :@: @#: : : ::: @:: :: ::: ::: : : :@:: :::: ::@::
     |    :::@@ ::: : ::: @ :@: @#: : : ::: @:: :: ::: ::: : : :@:: :::: ::@::
     | ::::: @@ ::: : ::: @ :@: @#: : : ::: @:: :: ::: ::: : : :@:: :::: ::@::
   0 +----------------------------------------------------------------------->Gi
     0                                                                   47.52
```

The memory usage varies quite a bit, I'm assuming because of collectable garbage which gets freed by mark-and-sweep.

fatcerberus:
Is this more stable now?  If so I'll test it out.  I was too lazy to build dukluv yesterday...

fatcerberus:
@creationix I was unable to build dukluv using VS2015.  I get the following syntax error:
```
4>c:\src\dukluv\src\duv.h(12): error C2059: syntax error: 'constant'
```

On this line:
```c
typedef enum { false, true } bool;
```

svaarala:
It should work pretty well now, except for error handling, i.e. what happens when a socket is disconnected etc. I think I'll add both a single shot and a persistent mode to the proxy so that it will either exit when the first (and only) connection finishes, or that is stays running and serves more clients.

You should probably open that DukLuv build error as an issue in the DukLuv repo. For what it's worth, I managed to build it with VS2013 although there were some warnings or ignored errors in the build. (Or maybe it was that install failed, not sure.)

svaarala:
Ok, FIXMEs cleaned up:

- Error handling should now be mostly correct; connection finish handling refactored into a single function for both JSON and debug target connection handlers to ensure it goes the same way each time.
- Added single connection vs. persistent mode: default is persistent where the proxy keeps on accepting new connections and connects to the target each time. Single mode is where the proxy accepts only one connection and exits when that's finished.
- Ran torture testing for `onRead()` calls so that all data was fed in one byte at a time (things worked fine still).
- Refactored some buffer helpers.

This branch is done except for a final readthrough before a merge.

svaarala:
Well, there'd be room for work on improving the proxy code. For example, the dvalue encode/decode sections are adapted from `duk_debug.js` and are mostly manual bit shifting etc, when they could just use DataView. But I don't want to start fixing/breaking that anymore in this branch :)

svaarala:
Another thing which is a bit messy now is the file organization and naming:

- `duk_debug.js` provides the packaged Node.js client and the web UI. It also provides the Node.js proxy. It's a bit confusing that the same file provides both.
- `duk_debug_proxy.js` provides the DukLuv proxy.

But I don't think I'll touch these in this pull.

svaarala:
Did a final pass using `gjslint`, fixed some missing var declarations, and re-smoked-tested the thing. Merging as soon as tests pass.

PR #594, opened 2016-02-17
Add some benchmarking notes

svaarala:
Add some rough notes on benchmarking tips.

### DIFF
--- a/doc/benchmarking-notes.rst
+++ b/doc/benchmarking-notes.rst
@@ -1,0 +1,222 @@
+==================
+Benchmarking notes
+==================
+
+This document provides some notes on how to benchmark performance or memory
+consumption so that you'll get the most relevant results for an actual
+target device.
+
+Memory benchmarking
+===================
+
+Enable Duktape low memory options
+---------------------------------
+
+Enable Duktape low memory config options for benchmarking if the target
+would actually be running with these options enabled.  For example, if
+the target has 128-256 kB of system RAM, low memory options would be very
+recommended:
+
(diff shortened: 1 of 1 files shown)


fatcerberus:
Regarding functions not being collected by refcounting, would it be possible to add a special case to detect the loop (since the exact properties involved in the loop are known), and free the function anyway?  I do something similar in minisphere when freeing socket objects to detect if the socket is piped into itself (an echo server), and if so, free it at refcount = 1.

fatcerberus:
I guess it's not a huge issue though because an emergency GC would free it anyway.

svaarala:
Some kind of special case hack would be possible, but not sure if worth it.

The main change I'm planning to do is to make the default prototype object virtual, so that it only gets created if you actually access `.prototype` (which user code doesn't do at all for most functions). This avoids the reference loop for most functions and also reduces memory usage in general.

fatcerberus:
Yeah, like I said the emergency garbage collector will get it anyway if needed, so it's not really an issue in practice.  Just a random thought that occurred to me.

svaarala:
Yeah, it's not an issue when a pool allocator is used. When a pool allocator is not used, that memory won't be available for other processes in the system, and if *they* run out of memory they can't cause an emergency GC to free up the memory tied up in circular references. For some low memory targets this might be an issue (similarly for any other garbage in reference loops). In practice very low memory targets tend to use pool allocators anyway so it's not really an issue for them.

In any case, memory churn would be reduced if the reference loop was broken by default.

PR #595, opened 2016-02-17
Fix double definition of `duk_double_nan` (and others) in a single file build

svaarala:
Reported on IRC, not diagnosed yet.

Happens in VS2012, `duktape.c` renamed to `duktape.cpp`, options:

    WIN32;_DEBUG;_CONSOLE; DUK_OPT_CPP_EXCEPTIONS;%(PreprocessorDefinitions)

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1484,6 +1484,9 @@ Planned
 
 * Portability improvement for Atari Mint: avoid fmin/fmax (GH-556)
 
+* Portability improvement for VS2012 C++ compilation: avoid double definition
+  of replacement double constants (GH-595)
+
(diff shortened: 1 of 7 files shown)


svaarala:
Build error: http://codepad.org/vnc3UC0G

svaarala:
Should have the same treatment as static const: http://stackoverflow.com/questions/936446/is-it-possible-to-forward-declare-a-static-array

I'm going to apply the same fix as for static consts and update code issues.

svaarala:
Confirmed that this doesn't happen on VS2013 but does on VS2012. AppVeyor test to cover the VS2012 C++ build issue added.

PR #598, opened 2016-02-18
Add an "extra" for non-BMP string handling

svaarala:
Some user code needs to deal with standard Ecmascript strings where non-BMP characters are represented as surrogate pairs. Add some utility functions to decode surrogate pairs into raw UTF-8 and vice versa so that C code can work with the raw UTF-8 form conveniently.

### DIFF
--- a/extras/nonbmp-string-util/README.rst
+++ b/extras/nonbmp-string-util/README.rst
@@ -1,0 +1,5 @@
+========================
+Non-BMP string utilities
+========================
+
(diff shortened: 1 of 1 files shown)


svaarala:
There is already an example for code page conversion: `examples/codepage-conv`. Maybe that should be promoted into an officially supported "extra" too.

svaarala:
TextEncoder/TextDecoder does much of what's intended here but there's not yet a C API for it.

PR #600, opened 2016-02-19
API and website documentation and minor cleanups for AppRequest/AppNotify

svaarala:
- [x] Debugger document improvements
- [x] Minor debug print improvements for debugger AppRequest/AppNotify
- [x] Add 'duk' example for AppRequest, provide a custom "CommandLine" AppRequest
- [x] Add 'duk' example for AppNotify, provide custom "DebuggerHandleFile" when running files
- [x] Add debugger error code metadata into the debugger build (not used yet)
- [x] Add API document for `duk_debugger_attach_custom()` and `duk_debugger_notify()`
- [x] Add note to `duk_debugger_attach()` to API document
- [x] Discuss Duktape 2.x deprecation of one or the other (most likely `duk_debugger_attach_custom()` staying and renamed to something else)
- [x] Add one paragraph summary of the feature to website Debugger section
- [x] API testcase for `duk_debugger_notify()` checking that it returns false and consumes arguments; also check API error for invalid nvalues
- [x] Releases entry for AppRequest/AppNotify

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1383,6 +1383,9 @@ Planned
 
 * Add Windows version of the debugger example TCP transport (GH-579)
 
+* Add support for application specific debugger commands (AppRequest) and
+  notifications (AppNotify) (GH-596, GH-563)
+
 * A DukLuv-based JSON debug proxy is now included in the dist package;
   it should allow much easier and more flexible packaging of a JSON debug
   proxy into a debug client (GH-590)
--- a/debugger/Makefile
+++ b/debugger/Makefile
@@ -57,7 +57,8 @@ duk_debug_meta.json:
 	python merge_debug_meta.py --output $@ \
 		--class-names duk_classnames.yaml \
 		--opcodes duk_opcodes.yaml \
-		--debug-commands duk_debugcommands.yaml
+		--debug-commands duk_debugcommands.yaml \
+		--debug-errors duk_debugerrors.yaml
 
 static/socket.io-1.2.0.js:
 	wget -O $@ http://cdn.socket.io/socket.io-1.2.0.js
--- a/debugger/duk_debugerrors.yaml
+++ b/debugger/duk_debugerrors.yaml
@@ -1,0 +1,6 @@
+error_codes:
+  - Unknown
+  - UnsupportedCommand
+  - TooMany
+  - NotFound
+  - ApplicationError
--- a/debugger/merge_debug_meta.py
+++ b/debugger/merge_debug_meta.py
@@ -11,6 +11,7 @@ if __name__ == '__main__':
 	parser.add_option('--output', dest='output', default=None, help='output JSON filename')
 	parser.add_option('--class-names', dest='class_names', help='YAML metadata for class names')
 	parser.add_option('--debug-commands', dest='debug_commands', help='YAML metadata for debug commands')
+	parser.add_option('--debug-errors', dest='debug_errors', help='YAML metadata for debug protocol error codes')
 	parser.add_option('--opcodes', dest='opcodes', help='YAML metadata for opcodes')
 	(opts, args) = parser.parse_args()
 
@@ -23,6 +24,7 @@ if __name__ == '__main__':
 
 	merge(opts.class_names)
 	merge(opts.debug_commands)
+	merge(opts.debug_errors)
 	merge(opts.opcodes)
 
 	with open(opts.output, 'wb') as f:
--- a/doc/code-issues.rst
+++ b/doc/code-issues.rst
@@ -153,6 +153,20 @@ result to minimize error proneness::
    * similar reasons.
    */
 
+Labels are intended by one space relative to the parent tab depth::
+
+  DUK_LOCAL void duk__helper(duk_context *ctx) {
+          if (!ctx) {
+                  DUK_D(DUK_DPRINT("ctx is NULL"));
+                  goto fail;
+          }
+
+          return;
+
+   fail:
+          DUK_D(DUK_DPRINT("failed, detaching"));
+  }
+
 Local variable declarations
 ---------------------------
 
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -21,11 +21,12 @@ Duktape provides the following basic debugging features:
 
 * Get/put variable at any callstack level
 
+* A mechanism for application-defined requests (AppRequest) and notifications
+  (AppNotify)
+
 * Forwarding of print(), alert(), and logger writes
 
 * Full heap dump (debugger web UI converts to JSON)
-
-* Bindings for application-defined commands (AppRequest and AppNotify)
 
 Duktape debugging architecture is based on the following major pieces:
 
@@ -79,9 +80,14 @@ To integrate debugger support into your target, you need to:
   etc.  (A detach can also occur if explicitly requested by the debug client
   or if Duktape detects a debug stream error.)
 
-* **If you have an eventloop**: optionally call ``duk_debugger_cooperate()``
+* **If you have an event loop**: optionally call ``duk_debugger_cooperate()``
   once in a while when no call to Duktape is in progress.  This allows debug
   commands to be executed outside of any Duktape calls.
+
+Duktape comes with a bundled debug client which supports a plain TCP transport.
+There are also several third party debug clients which may be adapted to talk
+to your target: they share the same debug protocol so only the transport will
+need adaptation.
 
 You can also write your own debug client by implementing the client side of
 the debug protocol.  The debug client is intended to adapt to the target
@@ -92,8 +98,8 @@ with the same semantic versioning principles as the Duktape API.
 You can implement the binary debug protocol directly in your debug client,
 but an easier option is to use the JSON mapping of the debug protocol which
 is much more user friendly.  Duktape includes a proxy server which converts
-between the JSON mapping and the binary debug protocol (which actually runs
-on the target).
+between the JSON mapping and the binary debug protocol which actually runs
+on the target.
 
 Example debug client and server
 -------------------------------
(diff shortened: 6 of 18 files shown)


svaarala:
@fatcerberus The minor code changes in https://github.com/svaarala/duktape/commit/284a29959c8abd9e27624676bfe86f905b3b5c19 are essentially:

- Debug log (DUK_DPRINT) when application calls `duk_debugger_notify()`. This matches a DUK_DPRINT level debug log for attach/detach, and also for received AppRequest.
- Return after DUK_ERROR_API(); the code is not consistent w.r.t. this however. It generates better code (and less warnings) when "noreturn" is not available.
- NULL the request_cb in detach1(), regardless of detach reason. This should have no practical effect, but it's best to NULL on detach.
- NULL the request_cb along with I/O callbacks on transport errors. This should have no practical effect, but the best state to be in after an I/O error is that we only have the detached callback non-NULLed.
- More strongly worded debug log when the AppRequest nrets doesn't look sane: when asserts are disabled it's not necessarily safe to continue (even with the if-condition) so this documents the issue in code and debug logs.

svaarala:
As for the debugger documentation, proposed changes in https://github.com/svaarala/duktape/commit/aac9ef58c9539490f4005264f7b976164997b8c1:

- Feature AppRequest/AppNotify more prominently in the intro blurp, moving it up a bit
- Clarify in AppNotify how an application error is created (the previous wording was e.g. for "target" which is a bit ambiguous), also make the example error message clearly application specific
- Reword the example callback request example a little bit; add a small note that JSON/JX conversion should avoid errors (might not be apparent, e.g. for looped objects), not sure if it's improved or not
- Add an explicit nvalues check for the callback request example to highlight it's critical to check (it was present in the other section but not here), also fix handling when argument 0 is not a string (the example used duk_get_string() which might return NULL and then strcmp()'d it)
- Emphasize in the callback explanation section that `nvalues` may have any value, including zero, so there isn't necessarily even a string command name there, otherwise it'd be easy to write incorrect code

There are also unrelated changes and cleanups in the commits:

- Some more notes on TCP transport
- Future work note on error codes

svaarala:
Here's the "duk" example for AppRequest and AppNotify: https://github.com/svaarala/duktape/commit/d0fb89c96e107fb9afd1d12e1e44028a809fe4bd.

It'd be nice to illustrate how the AJS heap info could be dumped but it'd need to access AjsHeap internals so I'll probably delete that example; the `CommandLine` AppRequest is pretty good I think, it's more or less realistic and has no portability impact like some other examples like getting current time or something.

fatcerberus:
I like the CommandLine example, it might be useful for the debugger to see the target's command line arguments.

fatcerberus:
There's an issue that just caught me up that is probably impossible to fix, but should be noted in the documentation: Integers pushed to the value stack for requests and notifys will be encoded as "number" (i.e. IEEE double) dvalues.  This came as quite a surprise at first but when I thought about it it made sense why that would be the case.

svaarala:
It would be fine to send them as plain integers when there's no loss of precision. This is what debugger.rst says:

Plain integers are guaranteed for e.g. command numbers:

>* When not sending a ``duk_tval`` value, integer number values must always be encoded as plain integers (not the IEEE double encoding).

duk_tval may use either:

>* When parsing a ``duk_tval`` value, both plain integers and IEEE double values must be accepted.  The plain integers map uniquely to IEEE doubles so there's no loss of information.  Note that a negative zero must be represented as an IEEE double to preserve the sign.

I haven't done that yet but it'd be quite easy to change. It might break some debug clients in a minor way if they played fast and loose with the types though, but shouldn't be difficult to fix either.

fatcerberus:
For now I just made a minor change to `dvalue_as_int()` in the SSJ dvalue code that casts the float to an int for number dvalues.  Like I said, it's not really surprising once I thought about it, since by spec all numbers in Ecmascript are floating point.

But that's true that it shouldn't be too hard to fix, since you already have double->int downgrade code in the form of fastint support. :)

svaarala:
I'll add that downgrade code in a separate branch just in case. I have a hunch the change may reveal some incorrect debug client assumptions so it's good to have a bit more advance warning.

svaarala:
Ok, so this is what it looks like when request_cb returns a broken nrets with asserts enabled:

```
[D] duk_debugger.c:1481 (duk__debug_handle_apprequest): calling into AppRequest request_cb with nvalues=1, old_top=7, top=8
[D] duk_debugger.c:1484 (duk__debug_handle_apprequest): returned from AppRequest request_cb; nvalues=1 -> nrets=5, old_top=7, top=11
PANIC 54: assertion failed: duk_get_top(ctx) >= old_top + nrets (duk_debugger.c:1486) (segfaulting on purpose)
Segmentation fault
```

and without asserts:

```
[D] duk_debugger.c:1481 (duk__debug_handle_apprequest): calling into AppRequest request_cb with nvalues=1, old_top=7, top=8
[D] duk_debugger.c:1484 (duk__debug_handle_apprequest): returned from AppRequest request_cb; nvalues=1 -> nrets=5, old_top=7, top=11
[D] duk_debugger.c:1492 (duk__debug_handle_apprequest): AppRequest callback doesn't match value stack configuration, top=11 < old_top=7 + nrets=5; this might mean it's unsafe to continue!
```

fatcerberus:
I like that it's explicitly noted in the debug log that it might be unsafe to continue.

fatcerberus:
@svaarala By the way, thanks for the quick response on the AppRequest pull, because of that and the very functional heap walking I was able to release the second beta of minisphere 3.0 with a vastly improved debugger. :-)

svaarala:
@fatcerberus Ok, so I'm done with this pull; the relevant changes which you may want to check are:

- Minor changes to the AppRequest implementation: https://github.com/svaarala/duktape/commit/ea8866562a1f830ebc600a0bcaf75ab1054776bb. One thing I don't like there is the naming of `duk__debug_null_most_callbacks()` but I couldn't come up with a short alternative for "NULL everything but detached callback".
- API docs: https://github.com/svaarala/duktape/commit/1b155525b6cc5d8f4a0779e292f3b7cb94a73ec4
- Minimal API testcases (what's possible without attaching): https://github.com/svaarala/duktape/commit/b49053d3b6721c7597c7754267c361f17a00822d
- Website updates: https://github.com/svaarala/duktape/commit/c236fa0eed4e72c020949340a4e71dd55de395b7

fatcerberus:
No issues with the diffs, I'll test the branch a bit later.

fatcerberus:
Everything seems to be working properly still:

```
D:\src\spectacles-i>ssj specs-bs.spk
SSJ 3.0a3 minisphere CLI debugger (x64)
A powerful JavaScript debugger for minisphere
(c) 2016 Fat Cerberus

starting D:/src/spectacles-i/specs-bs.spk... OK.
connecting to 127.0.0.1:1208... OK.
verifying... OK.
    inferior: minisphere 3.0a3
    duktape v1.4.0-167-gc2ecdc4

querying target... OK.
    game title: Spectacles: Bruce's Story
    author: Fat Cerberus
    source: D:/src/spectacles-i/

=> # 0: eval() at scripts/main.js:6
6: const DBG_DISABLE_TEXTBOXES = false;

scripts/main.js:6 eval()
(ssj) c
t: Spectacles: Bruce's Story [miniRT]
t: Sphere v2.0 (minisphere 3.0a3)
t:
t: Initializing Specs Engine test harness
t: Loading testcases from 'brucesStory.js'
t: Added battle test 'rsb1'
t: Added battle test 'rsb2'
t: Loading testcases from 'lastLucidan.js'
t: Added battle test 'temple'
t: Added battle test 'scott'
t: Initializing new game session >>diff lv: 2
t: Created party manager
t: Created new PC Scott >>lvl: 50
t: PC Scott learned skill Sword Slash
t: PC Scott learned skill Quickstrike
t: PC Scott learned skill Charge Slash
t: PC Scott learned skill Flare
t: PC Scott learned skill Chill
t: PC Scott learned skill Lightning
t: PC Scott learned skill Quake
t: PC Scott learned skill Hellfire
t: PC Scott learned skill Windchill
t: PC Scott learned skill Electrocute
t: PC Scott learned skill Upheaval
t: PC Scott learned skill Necromancy
t: PC Scott learned skill Crackdown
t: Added PC Scott to party
t: Initializing time-of-day manager
FATAL UNCAUGHT - Error: scott was eaten by maggie
    at: scripts/maps/Testville.js:72
=> # 0: [anon] at scripts/maps/Testville.js:72
72:                     throw new Error(food + " was eaten by " + muncher);

scripts/maps/Testville.js:72 [anon]
(ssj) v
var muncher = "maggie"
var food = "scott"
var maggieX = 847
var maggieY = 783
var pose = "south"
var foodX = 847
var foodY = 796
var name = "scott"

scripts/maps/Testville.js:72 [anon]
(ssj) bt
=> # 0: [anon] at scripts/maps/Testville.js:72
   # 1: MapEngine() at undefined:0
   # 2: game() at scripts/main.js:53

scripts/maps/Testville.js:72 [anon]
(ssj) c
SSJ session has been detached.
```

The bit on startup where it says "querying target"?  That was all done with AppRequests. :)

svaarala:
Alright, I think the small tweaks are finally finished and this feature is complete :)

svaarala:
Querying the game title etc is a nice example of what custom integration is possible. If someone were to connect to the game using another debug client things would still work. Also I hope your debug client would be able to connect to another target, notice there's no game title, but be otherwise functional :)

fatcerberus:
In theory SSJ could connect to any Duktape target (assuming a V1 protocol), although it probably wouldn't be able to locate source code since the target wouldn't respond to the "source path" request.  I personally find symbolic debuggers useless without the original source. ;)

fatcerberus:
You know what else is awesome?  Now that I have AppRequests at my disposal, I no longer need Duktape to be able to track source code - I can just query minisphere for the source.  Truly this is one of the best additions to Duktape ever. :)

svaarala:
That'd be worth mentioning in the documentation too I think, reading sources from the target.

PR #604, opened 2016-02-21
Use plain integer dvalues instead of IEEE double when serializing duk_tval numbers in debug protocol

svaarala:
When the debugger serializes `duk_tval` numbers it currently always sends them as "number" dvalues which sends the IEEE double value as is. Change the serialization so that plain integers are used when the value being serialized can be safely expressed as a plain integer: this is shorter and easier to read.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1401,6 +1401,10 @@ Planned
   it should allow much easier and more flexible packaging of a JSON debug
   proxy into a debug client (GH-590)
 
+* Use plain integer dvalues when serializing duk_tval numbers in the debugger
+  protocol when it's safe to do so, i.e. when the plain integer converts back
+  to an identical IEEE double with no loss of precision (GH-604)
+
 * Fix potentially memory unsafe behavior when a refcount-triggered finalizer
   function rescues an object; the memory unsafe behavior doesn't happen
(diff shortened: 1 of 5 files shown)


svaarala:
Some examples from JSON proxy dumps.

Evaluate 1+2 = 3:

```
{"request":"Eval","args":["1+2"]}
{"reply":true,"args":[0,3]}
```

Evaluate 1+2.5 = 3.5, answer is in IEEE double format:

```
{"request":"Eval","args":["1+2.5"]}
{"reply":true,"args":[0,{"type":"number","data":"400c000000000000","value":3.5}]}
```

Zero can be encoded as a plain integer without loss of information:

```
{"request":"Eval","args":["0"]}
{"reply":true,"args":[0,0]}
```

But negative zero cannot, so it comes out as an IEEE double:

```
{"request":"Eval","args":["-0"]}
{"reply":true,"args":[0,{"type":"number","data":"8000000000000000","value":0}]}
```

svaarala:
This change is within the protocol guarantees: https://github.com/svaarala/duktape/pull/600#issuecomment-186809491

However, it may affect some debug clients which don't parse numbers as described in debugger.rst, so the change is worth an update note.

PR #610, opened 2016-02-24
Add option callstack level or object argument to GetBytecode

svaarala:
Extend GetBytecode behavior to match e.g. GetLocals by adding an additional argument which be either:

- A negative integer to indicate a callstack level
- An object dvalue to specify an explicit target function

Tasks:

- [x] Change GetBytecode behavior
- [x] Test manually: no argument, valid integer, out-of-bounds integer, valid object pointer, invalid object pointer, invalid type
- [x] Test PutVar
- [x] Test AppRequest
- [x] Fix bug in `duk_debug_read_tval()` call sites when read fails and we detach: nothing gets pushed but some call sites will assume so: AppRequest, PutVar
- [x] Debugger documentation
- [x] Releases

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1391,6 +1391,9 @@ Planned
 
 * Add sizeof void pointer to the BasicInfo debugger command (GH-611)
 
+* Extend debugger GetBytecode command to accept an optional callstack level
+  or direct heap object argument (GH-610)
+
 * A DukLuv-based JSON debug proxy is now included in the dist package;
   it should allow much easier and more flexible packaging of a JSON debug
   proxy into a debug client (GH-590)
@@ -1424,6 +1427,10 @@ Planned
 * Fix debugger transport write error bug which could cause Duktape to call
   the debug transport write callback after it had already returned an error
   (GH-599)
+
+* Fix debugger PutVar command bug where a failure to read the PutVar variable
+  value (e.g. due to a transport detach) could lead to memory unsafe behavior
+  (GH-610)
 
 * Portability improvement for Atari Mint: avoid fmin/fmax (GH-556)
 
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -765,11 +765,13 @@ types in the text)::
     REP
     ERR
     NFY
-    <int: field name>    e.g. <int: error code>
-    <str: field name>    e.g. <str: error message>
-    <buf: field name>    e.g. <buf: buffer data>
-    <ptr: field name>    e.g. <ptr: prototype pointer>
-    <tval: field name>   e.g. <tval: eval result>
+    <int: field name>      e.g. <int: error code>
+    <str: field name>      e.g. <str: error message>
+    <buf: field name>      e.g. <buf: buffer data>
+    <ptr: field name>      e.g. <ptr: prototype pointer>
+    <tval: field name>     e.g. <tval: eval result>
+    <obj: field name>      e.g. <obj: target>
+    <heapptr: field name>  e.g. <heapptr: target>
 
 When a field does not relate to an Ecmascript value exactly, e.g. the field
 is a debugger control field, typing can be loose.  For example, a boolean
@@ -1847,18 +1849,42 @@ GetBytecode request (0x21)
 
 Format::
 
-    REQ <int: 0x21> EOM
+    REQ <int: 0x21> [<int: level> OR <obj: target> OR <heapptr: target>] EOM
     REP <int: numconsts> (<tval: const>){numconsts}
         <int: numfuncs> (<tval: func>){numfuncs}
         <str: bytecode> EOM
 
-Example::
+Example without argument, gets bytecode for current function::
 
     REQ 33 EOM
     REP 2 "foo" "bar" 0  "...bytecode..." EOM
 
-Bytecode endianness is target specific so the debug client needs to get
-target endianness and interpret the bytecode based on that.
+Callstack level can be given explicitly, for example -3 is the third callstack
+level counting from callstack top::
+
+    REQ 33 -3 EOM
+    REP 2 "foo" "bar" 0  "...bytecode..." EOM
+
+An explicit Ecmascript function object can also be given using an "object" or
+"heapptr" dvalue::
+
+    REQ 33 {"type":"object","class":6,"pointer":"00000000014839e0"} EOM
+    REP 2 "foo" "bar" 0  "...bytecode..." EOM
+
+An error reply is returned if:
+
+* The argument exists but has an invalid type or points to a target value
+  which is not an Ecmascript function.
+
+* Callstack entry doesn't exist or isn't an Ecmascript activation.
+
+Notes:
+
+* Bytecode endianness is target specific so the debug client needs to get
+  target endianness and interpret the bytecode based on that.
+
+* Minor change from Duktape 1.4.0: when the callstack entry doesn't exist
+  Duktape 1.5.x and above will return an error rather than an empty result.
 
 .. note:: This command is somewhat incomplete at the moment and may be modified
    once the best way to do this in the debugger UI has been figured out.
--- a/src/duk_debugger.c
+++ b/src/duk_debugger.c
@@ -425,7 +425,7 @@ DUK_LOCAL duk_double_t duk__debug_read_double_raw(duk_hthread *thr) {
 	return du.d;
 }
 
-DUK_INTERNAL void duk_debug_read_tval(duk_hthread *thr) {
+DUK_INTERNAL duk_tval *duk_debug_read_tval(duk_hthread *thr) {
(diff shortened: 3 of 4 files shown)


fatcerberus:
How would one get the bytecode for top-level code (outside of any function)?

svaarala:
Everything ultimately gets executed as a function, even top level code, so as long as you can pause there or get a reference to the function somehow GetBytecode will work.

For example, if you run `./duk --debugger` to enter interactive mode, the "duk" command will the greeting line by eval'ing some code. With the `--debugger` option it will start in a paused state, just before it executes the first opcode of that eval function.

svaarala:
Tto clarify a bit: if you're currently executing the top level code (say you're paused at a breakpoint), it will be on the call stack and that top level code is executing from a function object in Duktape internally.

When you're not even yet executing the top level code, you'd need some reference to the function representing the program. If you store the result of `duk_compile()` somewhere, that will be the function object representing the code before it executes. There's no way to do that from Ecmascript code though, because the only primitive is an atomic eval which both compiles and runs.

fatcerberus:
That makes sense, explains why SSJ lists one function call in the backtrace when first attached (`eval()`).  That always confused me because as far as I was concerned there was no function.  Boy was I wrong! :-)

svaarala:
This pull also fixes a minor issue in AppRequest: if `duk_debug_read_tval()` fails it doesn't push anything on the value stack (it just marks the connection detached). Since AppRequest handler doesn't check for that, it will read dvalues and assume something got pushed on each read. This may cause it to enter the callback with a certain `nvalues` not matching the stack configuration.

This is one more case where it'd be better to use normal error handling: `duk_debug_read_tval()` could just throw on error; error would cause detach, and there'd be no risk of doing anything stupid on the success path.

svaarala:
PutVar had a similar bug: if `duk_debug_read_tval()` failed to parse the variable value nothing would get pushed on the stack and things could then go wrong.

svaarala:
The AppRequest issue is a real one and reproducible e.g. by sending a dvalue `duk_debug_read_tval()` does not support:

```
{"request":"AppRequest","args":["CommandLine",{"type":"unused"}]}
```

The "unused" dvalue is parsed which causes a detach and a segfault follows:

```
[D] duk_debugger.c:1459 (duk__debug_handle_apprequest): debug command AppRequest
[D] duk_debugger.c:522 (duk_debug_read_tval): debug connection error: failed to decode tval
[D] duk_debugger.c:45 (duk__debug_do_detach1): debugger transport detaching, marking transport broken
[D] duk_debugger.c:250 (duk_debug_read_bytes): attempt to read 1 bytes in detached state, return zero data
[D] duk_debugger.c:1482 (duk__debug_handle_apprequest): calling into AppRequest request_cb with nvalues=2, old_top=7, top=8
Segmentation fault
```

The PutVar case causes a detach in master but couldn't reproduce actual memory unsafe behavior. (Doesn't mean it's not there though.)

Both issues are fixed by the pull.

fatcerberus:
I wonder why only it's only AppRequest that segfaults... both should be the same basic symptom of an inconsistent value stack.

svaarala:
PutVar depends quite a bit less on the shape of the value stack; what most likely happens is that variable name also becomes its value because nothing was pushed for the value but the variable name was read in beforehand. So it may not actually never yield unsafe behavior; best not to assume that of course.

AppRequest will assume the shape holds for the application call, and it's reset afterwards etc.

The inconsistency can be seen from the debug log BTW:

```
[D] duk_debugger.c:1482 (duk__debug_handle_apprequest): calling into AppRequest request_cb with nvalues=2, old_top=7, top=8
```

old_top was 7, nvalues is 2, so top should be 9.

svaarala:
In general it's not a big deal if a crafted debugger command causes unsafe behavior on the target: that's always possible via made-up pointers etc. What is significant is if the situation can be caused accidentally; here it could be caused by a detach occurring in just the right place between dvalues which is possible by accident. The trigger I used for testing ("unused" dvalue with no inbound parsing support) would not alone be really a significant issue because there's no reason to send such commands.

svaarala:
Just as a reminder for future work, the callstack index approach works for most code, but not when the code involves coroutine resume chains. There's no way to identify the thread whose callstack is being inspected, so you can't get the callstack of the coroutine which resumed the current thread for example.

I think the best step forward to do that is to provide more comprehensive generic inspection primitives for the callstack, the scope objects, heap objects, etc, instead of extending the individual commands piece by piece. We can hopefully then remove some overlapping functionality.

svaarala:
@fatcerberus I'm done with this branch now; I'll merge it in an hour or so after I've played a little Alpha Centauri :)

fatcerberus:
I agree that a more robust inspection primitive is preferable, however the callstack index approach I feel is a "good enough" solution for most cases.  Once the first incompatible break is made however, it'd probably be worth revisiting.

fatcerberus:
Maybe it would be worth adding an assert in the AppRequest code to verify that `nvalues` were actually pushed before calling request_cb?  That's an assumption that, as we see here, is easy to get wrong even internally to Duktape.

svaarala:
It's good enough if you don't work with coroutines, I agree :) But because Duktape supports coroutines it's not very reasonable the debugger doesn't support looking at the whole execution state, which includes coroutine resume chain.

The callstack index approach is quite simple internally so it's not that much of a versioning baggage (compared to not supporting it at all). But in hindsight maybe even that effort would have been better spent in figuring out a solution for the whole execution state. It's not a big deal in any case.

svaarala:
Agree about that assert, added and pushed an update.

fatcerberus:
It'd actually be very difficult for me to support coroutines in minisphere I think since the map engine is implemented as a native function that simply runs a game loop until termination is signaled by the game (via ExitMapEngine API).  As I understand it it's impossible to yield with a Duktape/C function on the callstack.

I didn't choose this design, unfortunately that's what I had to work with to be compatible with the Sphere engine.  I would really like to support coroutines because I think they'd be quite useful for game programming, but it's just not practical with the existing design being what it is.

svaarala:
One quite simple approach to that resume chain thing is just to add a "resumer index" so that you can request the Nth resumer and then use the callstack index to enumerate that. But I'd much rather ensure there are primitives in place to:

- Inspect the value, call, and catch stacks of any thread in the heap
- Get the currently active thread

With these the debug client can just get the current thread, inspect the stacks of that thread. It can then look at the thread inspection results to see if there's a "resumer", and if so, repeat the procedure for that. This will be much more flexible I think.

One nice thing about the debugger is that it's quite easy to enable and disable commands. So once enough inspection features are in place, it's not necessary to wait for a binary break to provide an option to disable commands that will eventually be removed.

svaarala:
A native event loop is not an obstacle to using coroutines: you'd just use that native loop to resume pure Ecmascript coroutines. They can then yield back to the event loop coroutine without issues. That's how coroutine based event loop are quite typically written.

What you can't do is yield out of a thread which has native call stack entries. But that's not the case here.

fatcerberus:
Actually it is: A game will typically render its title screen, etc. outside of the map engine, then call MapEngine() to start the game loop.  MapEngine() is now part of the Ecmascript callstack until the game terminates, preventing coroutine use.  I guess nothing is stopping a game from implementing a custom game loop in script, but the built-in facilities are not coroutine friendly. :)

svaarala:
What I meant is that you have thread A, which runs the native event loop until the game ends. That thread cannot yield "across" the event loop for several reasons.

But that does not prevent thread A from maintaining a bunch of other threads, implemented in pure Ecmascript, whose callstacks don't contain native entries. Thread A is then free to:

- Detect that some thread needs to run, and call resume() on it.
- Once the thread is ready and needs to wait for something, it yields. This works because *that* thread doesn't have any native calls in its callstack.
- Once the thread has yielded we're back in thread A, the event loop. Note that the yield doesn't propagate further in thread A, and never crosses the native call(s) thread A callstack.

This is what's quite typical for a coroutine based eventloop. But I might be misunderstanding your situation though.

fatcerberus:
Hm, I'll have to look at the docs again, you may be right and I just misunderstood how it all works.

Should I test this branch before merge?

svaarala:
Just a quick addition to that coroutine description: so, thread A cannot yield, but that wouldn't make sense as it would be yielding out of the eventloop. (The technical problem preventing a yield is that we'd then fall out to e.g. a top level protected call or a previous thread, and be unable to resume because we'd need to "reconstitute" native stack frames which is difficult and more or less equivalent to actual thread switching.)

Re: testing, I don't that should be necessary unless you have extra time on your hands :) I'll rebase the inspection branch against master once this is merged so if you test that after the merge, you'd be effectively ensuring nothing was broken here.

svaarala:
Well, I'll merge this now - if you happen to test the branch (or master right after) let me know if you find something :) I'll rebase and push the inspect branch in a few minutes.

PR #615, opened 2016-02-25
Implement "pause debugger now" API

fatcerberus:
The target can call `duk_debugger_pause()` to initiate an immediate debugger pause, enabling implementation of a "break into debugger now" hotkey or similar mechanism.

The intent is that `duk_debugger_pause()` will pause execution at the exact point where it was called.  I'm not sure if this is true in practice due to the way the interrupt counter works.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1389,6 +1389,9 @@ Planned
 * Add support for application specific debugger commands (AppRequest) and
   notifications (AppNotify) (GH-596, GH-563)
 
+* Add duk_debugger_pause() API which allows the target to quickly pause
+  Ecmascript execution and break into the debugger (GH-615)
+
 * Add sizeof void pointer to the BasicInfo debugger command (GH-611)
 
 * Extend debugger GetBytecode command to accept an optional callstack level
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -320,6 +320,25 @@ blocking), you can also use it like this::
         /*...*/
     }
 
+duk_debugger_pause()
+--------------------
+
+The target may call this at any time to pause Ecmascript execution and turn over
+control to the attached debug client.  This is safe to call if the debugger is
+not attached, in which case it has no effect::
+
+    duk_debugger_pause(ctx);
+
+A pause so requested may not happen immediately; for example if a long-running
+native call such as a Duktape/C function is in progress. The target will pause
+once the bytecode executor is re-entered, either by the native call returning
+into Ecmascript code or calling e.g. ``duk_eval()`` or ``duk_call()``.
+
+A common use case for this call is to bind it to a hotkey, which allows the
+user to break out of and debug infinite loops.  However, like all Duktape API
+calls, the call is not thread safe and must be called from the same thread used
+to run the Ecmascript code being debugged.
+
 Debug transport
 ===============
 
--- a/src/duk_api_debug.c
+++ b/src/duk_api_debug.c
@@ -181,6 +181,29 @@ DUK_EXTERNAL duk_bool_t duk_debugger_notify(duk_context *ctx, duk_idx_t nvalues)
 	return ret;
 }
 
+DUK_EXTERNAL void duk_debugger_pause(duk_context *ctx) {
+	duk_hthread *thr;
+
+	DUK_ASSERT_CTX_VALID(ctx);
+	thr = (duk_hthread *) ctx;
+	DUK_ASSERT(thr != NULL);
+	DUK_ASSERT(thr->heap != NULL);
+
+	DUK_D(DUK_DPRINT("application called duk_debugger_pause()"));
+
+	/* Treat like a debugger statement: ignore when not attached. */
+	if (DUK_HEAP_IS_DEBUGGER_ATTACHED(thr->heap)) {
+		DUK_HEAP_SET_PAUSED(thr->heap);
+
+		/* Pause on the next opcode executed.  This is always safe to do even
+		 * inside the debugger message loop: the interrupt counter will be reset
+		 * to its proper value when the message loop exits.
+		 */
+		thr->interrupt_init = 1;
+		thr->interrupt_counter = 0;
+	}
+}
+
 #else  /* DUK_USE_DEBUGGER_SUPPORT */
 
 DUK_EXTERNAL void duk_debugger_attach_custom(duk_context *ctx,
@@ -229,6 +252,12 @@ DUK_EXTERNAL duk_bool_t duk_debugger_notify(duk_context *ctx, duk_idx_t nvalues)
 	/* No debugger support, just pop values. */
 	duk_pop_n(ctx, nvalues);
 	return 0;
+}
+
+DUK_EXTERNAL void duk_debugger_pause(duk_context *ctx) {
+	/* Treat like debugger statement: nop */
+	DUK_ASSERT_CTX_VALID(ctx);
+	DUK_UNREF(ctx);
 }
 
 #endif  /* DUK_USE_DEBUGGER_SUPPORT */
--- a/src/duk_api_public.h.in
+++ b/src/duk_api_public.h.in
@@ -959,6 +959,7 @@ DUK_EXTERNAL_DECL void duk_debugger_attach_custom(duk_context *ctx,
 DUK_EXTERNAL_DECL void duk_debugger_detach(duk_context *ctx);
 DUK_EXTERNAL_DECL void duk_debugger_cooperate(duk_context *ctx);
 DUK_EXTERNAL_DECL duk_bool_t duk_debugger_notify(duk_context *ctx, duk_idx_t nvalues);
+DUK_EXTERNAL_DECL void duk_debugger_pause(duk_context *ctx);
 
 /*
  *  Date provider related constants
--- a/website/api/duk_debugger_pause.yaml
+++ b/website/api/duk_debugger_pause.yaml
@@ -1,0 +1,32 @@
+name: duk_debugger_pause
+
+proto: |
+  void duk_debugger_pause(duk_context *ctx);
+
+summary: |
+  <p>Trigger a debugger pause as soon as possible.  In most cases Duktape pauses
+  immediately; however if a native call such as a Duktape/C function is in
+  progress or no Ecmascript code is currently executing, the pause will be
+  triggered the next time Ecmascript bytecode is executed.  This may take some
+  time e.g. if the native call in progress is long-running.</p>
+
+  <p>If debugger support has not been compiled in, or if the debugger is not
+  attached, the call is a no-op.  This mimics the semantics of the Ecmascript
+  <code>debugger</code> statement.</p>
+
+  <div class="note">
+  Like all Duktape API calls, this call is not thread safe.  It may be tempting
+  to want to trigger a pause from a thread different than one running Ecmascript
+  code for the context, but this is unsafe and must be avoided.
+  </div>
+
+example: |
+  /* In your event loop: */
+  if (key_pressed == KEY_F12) {
+      duk_debugger_pause(ctx);
+  }
+
+tags:
+  - debugger
+
+introduced: 1.5.0


svaarala:
If this is called from outside of the debugger message loop, it could reset the interrupt counter so that the pause gets noticed right away.

If this can be called from inside the debugger message loop (hopefully not!) then that should hopefully not be necessary at all, but care would have to be taken that we don't change the interrupt counter if it has been already primed for the next execution segment. I don't think this is the case; the final interrupt counter value is selected after the debugger message loop exits.

fatcerberus:
I don't envision this being callable from the debugger message loop, no.  The intent is that a running target would call this when a certain key is pressed to allow the user to break out of e.g. infinite loops.

Concretely, what would have to be done to implement the interrupt counter reset?

svaarala:
If we're outside the debugger code, it's basically this with `ctr == 1`:

https://github.com/svaarala/duktape/blob/master/src/duk_js_executor.c#L1764-L1766

fatcerberus:
So, like this then:

```c
DUK_EXTERNAL void duk_debugger_pause(duk_context *ctx) {
	duk_hthread *thr = (duk_hthread *) ctx;
	
	if (DUK_HEAP_IS_DEBUGGER_ATTACHED(thr->heap)) {
		DUK_HEAP_SET_PAUSED(thr->heap);
		thr->interrupt_init = 1;
		thr->interrupt_counter = 0;
	}
}
```

Incidentally, is it safe to call `DUK_HEAP_SET_PAUSED()` while already paused, or should I check for that first?

svaarala:
Right now it is: https://github.com/svaarala/duktape/blob/master/src/duk_heap.h#L277-L281.

fatcerberus:
I implemented that interrupt counter thing, and added a note to the top of the function that it can't safely be called from inside the debugger.

svaarala:
That could maybe be just an assert, e.g. `DUK_ASSERT(thr->heap->dbg_processing == 0)`. This won't be the strongest possible assert but enough to catch most abuses.

fatcerberus:
Does this need any documentation other than website changes?

fatcerberus:
I took a stab at adding a website page, let me know if I did it right. :)

svaarala:
Usually I try to have all of the following in pulls if possible:

- Testcases
- Website changes e.g. guide mentions / links for high level feature changes
- If the feature needs or benefits from Wiki changes, ensure there's a Wiki issue/pull (doesn't have to be completed right away or in sync with the pull)
- Website API documentation changes
- Releases entry

Looking at the diff quickly looked good to me, I'll take a closer look tomorrow.

fatcerberus:
I didn't add a releases entry, I'll do that tonight.

svaarala:
I'll fill in whatever's missing but much appreciated if you go through the motions :)

fatcerberus:
Okay, I added the feature to RELEASES.rst.

fatcerberus:
Alright, I added a note on the new function to the Debug API section of debugger.rst.  This is merge-ready pending nitpicks. :)

svaarala:
It might be worth emphasizing in the documentation that no calls into Duktape API are thread safe, so that it's not safe to call `duk_debugger_pause()` unless: (a) there is no call into Duktape in progress, or (b) the `duk_debugger_pause()` call comes from the same thread that is currently running.

Although this is the case for all API calls, it may be especially tempting with this API call I think.

fatcerberus:
I agree.  The obvious use case is for debugging infinite loops, and it'd be very tempting to try to initiate the pause from another thread.  Where do you want me to add the note?

svaarala:
I guess the API document (website/api/duk_debugger_pause.yaml) would be the logical place.

You can use a `<div class="note">...</div>` (see https://github.com/svaarala/duktape/blob/master/website/api/duk_to_pointer.yaml#L23-L29) if that feels appropriate, it'll look like this: http://duktape.org/api.html#duk_to_pointer. The **NOTE:** text is added automatically.

svaarala:
As such this calls would in 99.99% cases work correctly from another thread, but it wouldn't be 100% without some form of locking.

It'd actually be quite easy to make this call thread sage: `duk_debugger_pause()` would be changed to just write a volatile "please pause" flag to 1, and let the executor interrupt read the flag and do the rest when it notices the flag is != 0 (and then reset the flag to 0). The downside is that the interrupt wouldn't be immediate, as updating the interrupt counters concurrently is definitely not safe because it's not atomic (and they're also not declared volatile).

But it'd be a bit odd for one API call to have threading support while others didn't.

svaarala:
Another API call which would be useful is to force throw an uncatchable error like the script timeout. It'd be useful to be able to call that from another thread too. I'll make a note of these two calls on the "relax threading" work list :)

fatcerberus:
For future work it would actually be useful if Duktape had some way for the user to plug in their own locking primitives, through preprocessor magic perhaps, which Duktape could then use to ensure synchronization.  Then it wouldn't matter if the underlying threading framework were pthreads, Win32, or whatever.

svaarala:
Yeah, adding those macros would be easy but figuring out all the places where they're needed (each individual API call?) is not so easy :) An explicit duk_lock() and duk_unlock() would be an easy solution but awkward to use.

fatcerberus:
Okay, note added to both the API entry as well as debugger.rst.

fatcerberus:
I don't mind so much the core API calls being non-thread-safe (since JS itself is single-threaded), but the debugger is one case where running it on a separate native thread would be useful.  Anyway we can figure that out later. :)

svaarala:
Hmm, one sequence what needs a bit of checking is where:

- The user requests a pause
- The debugger gets detached before the pause can take effect (either user requests the detach or it's caused by another trigger)

I think this should work out fine, because we don't check for paused state or debugger handling in general, if the debugger is not attached. So, any changes to `dbg_paused` wouldn't matter.

Also a corner case: what happens when this API call happens when detach1() has executed by we haven't detached yet fully. This case might happen e.g. when the user calls `duk_debugger_notify()` which causes debugger protocol writes which trigger detach1() but we're still outside the debugger message loop. Then, before we get to finish the detach in detach2(), the user calls `duk_debugger_pause()`.

fatcerberus:
You're right - the user might request a pause while an `duk_debugger_cooperate()` event loop is running and would expect it to pause the next time some JS code is run.  So the current behavior is okay.

As for the detach handling, the call should be safe even in that case - The only state changes as far as I know are `heap->dbg_paused = 1` and setting the interrupt counter to trigger on the next opcode.  That shouldn't interfere with detach handling, should it?

svaarala:
Logically no, but the only way to be sure is to read the related code sections. I can do that before merging.

fatcerberus:
Changes made in the last push:
* Remove the "immediate" references and clarify what happens if a native call is in progress
* Move the thread safety note into the main text of the API entry
* Add a debug print when `duk_debugger_pause()` is called
* Fix spacing in documentation (use 2 spaces after a period)

fatcerberus:
Travis must be having difficulties, the test builds never seem to start...

fatcerberus:
Okay, tests passed.  This looks good from my end, just waiting on a final review.

fatcerberus:
So I was getting mixed signals - do you want me to remove the assert?

fatcerberus:
I guess the only thing that could possibly go wrong with pausing from a debugger loop is the interrupt counter manipulation.  I think interrupt handling is already fully completed by the time the debugger message loop is entered.

Come to think of it, would a debugger pause even work during Eval?  Message processing can't be entered into recursively (barring bugs), so assume everything is working properly the call should have no effect.

fatcerberus:
I'm going to go out for a walk, I'll do some in-house testing on this when I get back to try various scenarios.  That might help us decide what to do about that assert. :-)

svaarala:
Actually the interrupt counter is reset when we come out of the debugger message loop. So here's what I think would happen if `duk_debugger_pause()` was called from the message loop -- regardless of whether that message loop is executing because we're paused or because we peeked and noticed an inbound Eval while running:

- The executor enters the debugger message loop which (for argument's sake) reads and handles an Eval.
- The Eval code triggers a call to `duk_debugger_pause()`. This sets `dbg_paused=1` and `dbg_state_dirty=1`.
- We then exit the "handle one message" call and resend Status based on the dirty flag: https://github.com/svaarala/duktape/blob/master/src/duk_debugger.c#L2097
- On the next message loop we'd notice we're paused and want to read in blocking mode, regardless of why we originally entered the loop: https://github.com/svaarala/duktape/blob/master/src/duk_debugger.c#L2091-L2093
- We can only exit the loop via a Resume or a step command.
- We end up returning to the executor interrupt and initializing the interrupt counter for the next segment of opcode execution: https://github.com/svaarala/duktape/blob/master/src/duk_js_executor.c#L1754-L1767
- The "immediate" flag controls what value is ultimately used; immediate is set by the debugger integration helper e.g. when we're stepping.

So based on this I think it should work correctly within the current handling structure.

svaarala:
So I think that scenario should be tested in practice, and I think the `duk_debugger_pause()` code block which writes the interrupt counter etc should have a comment that this works also in the message loop because the interrupt counter is reinitialized (overwritten) before returning to the executor.

fatcerberus:
Just to make sure, it won't actually pause DURING the Eval, right?  As we've seen in the past, bad things happen when an Eval is halted. :)

svaarala:
No, it will just flag to we want to remain paused and this affects the current message loop as it needs to.

svaarala:
One solid test for this would be:

- Implement an AppRequest "CallPause" which just unconditionally causes `duk_debugger_pause()`.
- Execute an infinite loop in Ecmascript with debugger attached and running.
- While Duktape is in running state, execute the "CallPause" command.
- Duktape's periodic transport peek would cause that call to execute with the effect that we should remain paused and responsive in the message loop, just as if we paused via an explicit request.
- From this state, the critical tests are to see that (1) Resume works, and (2) e.g. StepInto works. These two exercise the two different counter reinitialization alternatives.

svaarala:
A little bit better test would be to do the same as above but:

- Implement a native `callPause()` Duktape/C binding which calls `duk_debugger_pause()`.
- In the same setup as above (Duktape running, infinite loop) send an Eval for "callPause()".
- See that the Eval runs and exits, and then stops in the message loop for further commands.

svaarala:
Btw, I can do one or both tests before the merge (with the assert removed), but I wouldn't want to merge the change until it's tested in practice.

But assuming it works the way I described above, the changes needed for this pull would be:

- Remove the assert.
- Add a comment to `duk_debugger_pause()` that it's safe to call both from outside and inside the debugger message loop. The message loop will correctly react to getting paused e.g. during Eval or AppRequest, and the executor interrupt will rewrite the counter when we've exited the message loop and are about to resume execution.
- Ensure documentation is in line with this. I think both debugger.rst and the API document are fine as is, because there are no surprising limitations with the assert removed.

fatcerberus:
I set up a test harness using minisphere and SSJ to test this.  It was actually more difficult than originally anticipated because SSJ is designed to lock the user out of the command line while the target is running.  I had to refactor it a bit to show the command prompt while not paused.

Here's the test setup:
https://github.com/fatcerberus/minisphere/blob/duktape-pause-harness/src/ssj/inferior.c#L248-L259
https://github.com/fatcerberus/minisphere/blob/duktape-pause-harness/src/minisphere/debugger.c#L299-L301
https://github.com/fatcerberus/minisphere/blob/duktape-pause-harness/src/minisphere/api.c#L844-L849

Test 1: While target is running, enter `pause` at SSJ command line.  This triggers an AppRequest which ultimately calls `duk_debugger_pause()` on the target.
Result: Target pauses.  Stepping commands work correctly, and `continue` causes execution to resume as normal.

Test 2: While target is running, enter `e DebugPause()`.  This calls a Duktape/C function which calls `duk_debugger_pause()`.
Result: Target pauses. Stepping commands work correctly.  `continue` resumes normal execution.

So yes, it looks like the tests checked out and the assert can be removed. :)

svaarala:
Ok, sounds good :)

fatcerberus:
I removed that assert, wrote a comment explaining why the interrupt counter manipulation is safe, and made some edits to the documentation to clarify some things.  In particular: I made the thread safety notice a footnote again (as it's simply a reminder of the contract already in place), and added mentions that, if a native call is in progress, execution will halt once -EITHER- the native call returns -OR- it calls back into Duktape.

I also gave the new code a quick review and fixed a few nits I noticed.

Let me know if you want me to do anything else.  If you want to run those two tests yourself that's okay by me, otherwise this is merge-ready.

svaarala:
Looks good to me, except clarifying when we'll become paused: the current text was a bit misleading.

I'll take your word for the tests, thanks for going through them :)

fatcerberus:
How's this?  If that's good I'll squash the fixup commit before merge.

svaarala:
Looks good to me!

fatcerberus:
Okay, squashed.  Thanks for the help with this, it's not always obvious what the various corner cases are.

It seems like the simplest changes in theory end up being the most difficult to get right.

svaarala:
Yeah, anything to do with the execution state needs careful thought because there are natural dependencies between a lot of different mechanisms and states. It's not always obvious what's easy and what's hard.

For example, adding a debugger command to "resume until expression is true", where an expression would be Eval'd before every opcode executed, would have less corner cases than this API call, and would be pretty easy to implement :)

fatcerberus:
I should be done with pulls for 1.5.x - this makes everything I need for minisphere 3.0.  I'm angling for a March 28 release, by the way. ;-)

svaarala:
That sounds possible for Duktape 1.5.0 release too.

PR #576, opened 2016-02-07
Add support for debugger heap walking

svaarala:
Add support for safe debugger heap walking:

- Add a primitive to inspect an object and its properties and flags without side effects. This allows the debug client to enumerate properties, look up the prototype chain, walk referenced objects, etc.
- Add a mechanism for some kind of heap locking to avoid garbage collection while walking the heap.

Tasks:

- [x] Add a new debugger command for inspecting an object
- [x] Add new debugger commands to "lock" the heap for walking (see discussion in #358) => For now, uses automatic locking during paused state; future work is to add an explicit GC during paused state if that is necessary for some reason
- [x] Add `sizeof(void *)` to BasicInfo
- [x] Add optional heapptr target to GetBytecode?
- [ ] Debugger protocol documentation
- [ ] Wiki page for debugger heap walking with more verbose discussion and example command sequences - https://github.com/svaarala/duktape-wiki/pull/81
- [ ] Some minimal implementation in the web UI
- [ ] Document garbage collection behavior during paused state into debugger documentation; refzero cases go through mark-and-sweep now
- [ ] Read through debugger.rst and update design notes and future work re: inspect
- [ ] Releases

Low level issues to resolve:

- [x] What to do about virtual properties - i.e. properties logically part of the object and visible to Ecmascript code, but not concretely part of the object's internal property table? For comparison, enumeration code needs to "fake" these properties. => For now, virtual properties are not returned in inspection; future work.
- [ ] Make the `InspectHeapObject` command optional because of the relatively high footprint.
- [ ] Should property keys be object references to heap strings? If yes, what to do about artificial keys? If no, how would user code be able to dump all heap strings? Should there be a LookupString command to find an interned string?

Consider overlaps with existing commands:

- [ ] `InspectHeapObject` and `DumpHeap`: Could DumpHeap simply provide a list of heap pointers which the debug client would then inspect using `InspectHeapObject` so that there'd be no formatting overlap?
- [ ] `InspectHeapObject` and `GetBytecode`: GetBytecode could be replaced by simply getting a reference to the currently executing function and then adding the bytecode, constants, and inner functions into the InspectHeapObject result. There's a small issue in representing structured values there; temporary objects may need to be created to hold the constants, etc.

Issue #585, opened 2016-02-14
Incompatible debugger changes for 2.x release


svaarala:
Some possibilities:

- Change Eval error from a plain string to a raw dvalue which user code can then inspect. Or perhaps provide both a string and an object.

fatcerberus:
As you note, doing a full heap object inspection can often be an involved process involving multiple requests, state tracking, etc.  One possibility for reducing that pain when dealing with Error objects received by the client would be to provide a convenience command that took the Error's heapptr and returned back the typical values used in error handling: filename, line number, etc.

fatcerberus:
We'll call this hypothetical command 'ParseError'.

svaarala:
Adding an "eval with arguments" would allow the user to do that parsing relatively conveniently (see #586) without adding an overlapping mechanism so I'd somewhat prefer that.

fatcerberus:
Just to verify, semantic versioning guarantees apply only to source-level compatibility, not binary, correct?  If so one way to improve `duk_debugger_attach()` would be to have it accept a struct as argument describing all the callbacks.

We'd need to stress in the documentation that the caller must zero the struct first so that when new fields are added in later versions, they don't get garbage values.

svaarala:
That's one way to approach extensibility - but it prevents literal initializers which is a shame. Further, to be strict, there are no guarantees that zeroing memory initializes pointers to NULL because the NULL bit pattern is not necessarily all zeroes.

svaarala:
Named struct whose members are not public, with macro initializers would be a source compatible approach. It's used for the `duk_def_prop_list()` API call and it works quite well for that. The downside is that it's not as direct as intuitive as a plain call with plain arguments.

fatcerberus:
Huh.  Are there really platforms where NULL is not all zero bits?  It's a very, very common pattern to initialize struct memory containing pointers using memset so I assumed it was guaranteed by the C standards.

Anyway this kind of thing is why object literals are in my top 3 favorite JS features. :)

svaarala:
It's not guaranteed; what is guaranteed is that converting an integer 0 to a pointer results in a NULL pointer, but there are no guarantees (AFAIK) for the memory pattern holding a NULL pointer.

svaarala:
[...]

PR #617, opened 2016-02-26
Add debugger object inspection support

svaarala:
This is the alternative approach to object inspection support, consisting of three command (to be renamed):

- InspectHeapObject: same as in #576 but only returns artificial properties without filtering.
- InspectPropertyKey: get attributes, key, and value for a named property.
- InspectPropertyRange: get attributes, keys, and values for a range of properties.

The property range is given as `[idx_start,idx_end[` with the index mapping to a virtual range containing:

- Array part entries `[0,a_size[`, including all allocated entries (including those above `.length`). Unmapped entries are returned as dvalue "unused" to distinguish them from "undefined".
- Entry part entries `[0,e_next[`, with unused (deleted) entries returned as dvalue "unused".

The virtual index range allows a client to fetch properties incrementally; it can request index ranges of a certain size (for example: [0,10[, [10,20[, [20,30[ ...) until a partial range or empty range is returned which indicates the index sequence has been fully enumerated.

Because the artificial properties returned by InspectHeapObject returns a_size, the client can directly choose to enumerate either array part entries or entry part entries, or it can choose to remain oblivious to the distinction.

Tasks:

- [x] Add a placeholder flags field for each command? If so, it saves a "get optional int" check in the future. => There are no placeholders for other commands either, so not added.
- [x] Drop class number or name? => Follow-up pull
- [x] Prune artificial properties?
- [x] Mark pruning in debugger document
- [x] Share error writers? => Had no effect on footprint, not enough sharing right now
- [x] Document limitation when garbage generated with mark-and-sweep disabled from config options
- [x] Ensure refzero GC disabled check is a bit or bitmask test, e.g. by promoting dbg_paused to a heap flag or by renaming "mark and sweep running" flag to "no gc" flag and sharing it => separate perf branch
- [x] Command naming
- [x] Future work issue for improving the GC behavior (relevant for example if strings get link pointers)
- [x] Future work issue fix unsigned 32-bit flags


fatcerberus:
Does InspectHeapObject return the number of properties?  This way the client could still pull all the properties in one go if it wanted to.

svaarala:
This branch is in pretty raw shape still. I reworked the InspectHeapObject from the previous branch a bit, added InspectPropertyRange, and a placeholder InspectPropertyKey.  The result is minimally functional:

```
// Eval for a test array
{"request":"Eval","args":["['foo', 'bar', 'quux', 'baz']"]}
{"reply":true,"args":[0,{"type":"object","class":2,"pointer":"00000000018a5ad0"}]}

// Request index range [0,10[: 0-3 are array part entries, 4 is 'length'.
{"request":"InspectPropertyRange","args":[{"type":"object","class":2,"pointer":"00000000018a5ad0"},0,10]}
{"reply":true,"args":[7,0,"foo",7,1,"bar",7,2,"quux",7,3,"baz",1,"length",{"type":"number","data":"4010000000000000","value":4}]}

// Request index range [1,10[, skips the first array part entry.
{"request":"InspectPropertyRange","args":[{"type":"object","class":2,"pointer":"00000000018a5ad0"},1,10]}
{"reply":true,"args":[7,1,"bar",7,2,"quux",7,3,"baz",1,"length",{"type":"number","data":"4010000000000000","value":4}]}
```

There's no documentation etc, and much cleanup ahead.

svaarala:
@fatcerberus It does return the array part size and entry part sizes. But they're quite unnecessary, you can just request the index range [0,0x7fffffff] instead (signed range for now; fixing it later).

fatcerberus:
Ah, that's true, I hadn't considered you could just request a larger range than is available.  I'll try this out soonish.

svaarala:
That's also how you know when to end enumerating if you do it in pieces. There are two equivalent alternatives:

- Keep on enumerating successive ranges ([0,10[, [10,20[, etc) until you come up short. You'll know you're finished.
- Even simpler, keep on requesting ranges until you get an empty response.

The debugger documentation should describe the two use cases: (1) how to enumerate all properties, and (2) how to enumerate in pieces and when to stop.

svaarala:
@fatcerberus Just note I just wrote this up, compiled and minimally tested; there may be trivial bugs, there's no documentation yet etc. I'll get the branch into more functional form soon.

svaarala:
Just as I suspected this branch is larger, at the moment about +220 bytes over the previous approach and about +3000 bytes over master (x64). I can probably reduce that a bit, maybe 10%. That's one reason I'd like the target to avoid optional arguments, peeking, etc, because it all adds up to footprint for often very little benefit.

In any case it's important to find the most useful, and reasonably versionable model for object inspection. Hopefully this is closer to the mark than the previous one :)

fatcerberus:
I agree w.r.t. avoiding optional arguments.  That's the price we paid to avoid bumping the debugger protocol version. :)

fatcerberus:
Do note that a peek primitive is still needed for certain commands though, e.g. AppRequest.

svaarala:
I know, I just mean that when new commands are added, it's best to avoid the temptation of adding convenient defaults or alternative forms. Instead, ideally the target would provide minimal, orthogonal functionality without trying to be too pretty about it.

For the AppRequest case it's fine as it's a variable argument list. One could express that also with a length prefix which would then avoid that peeking (a similar approach is used for other lists in the command set right now btw). A length prefix would also make AppRequest more easily extensible because it would then be possible to append additional arguments if necessary. As it is now, it will be impossible to add any fields to AppRequest (flags or whatever). Not sure which approach is better on the whole though.

To revisit the binary protocol discussion: if the binary protocol was changed into something more structured like a CBOR subset, Duktape wouldn't be capable of parsing arbitrary structured values but we *could* support flat arrays which would be much cleaner. For example something conceptually like this would be possible:

    REQ "foo" "bar" ["my","app","request","keys"] ["another","list"] "quux" EOM

This would replace the manual convention of a length dvalue followed by N dvalues.

For replies arbitrary structure would be possible because it's not a problem to generate such structure without an intermediate representation.

fatcerberus:
I think AppRequest is fine as-is because the application defines the meaning of the contents.

Anyway I guess we can revisit this once Duktape 2.x work has started.

svaarala:
Agreed, not sure if it'll be possible to change the binary transport yet in Duktape 2.x (I don't want that release to become too large) but at the very least we can drop some optional fields and make the commands more regular.

harold-b:
I'll see if I can test out this branch today. So all 3 commands are in (albeit rough form)?

svaarala:
The InspectPropertyKey command doesn't yet do anything except send out an error. The shortest form for that needs a helper which is currently local to `duk_hobject_props.c` which I'll probably expose as an internal helper.

harold-b:
OK, got it. I'll keep an eye out for updates then.

svaarala:
Ok, now there's minimal documentation and preliminary implementation of all the commands. Documentation:

- https://github.com/svaarala/duktape/blob/b9d94063c9a886e7dc869a5684875330f1ffba58/doc/debugger.rst#inspectheapobject-0x23
- https://github.com/svaarala/duktape/blob/b9d94063c9a886e7dc869a5684875330f1ffba58/doc/debugger.rst#inspectpropertykey-0x24
- https://github.com/svaarala/duktape/blob/b9d94063c9a886e7dc869a5684875330f1ffba58/doc/debugger.rst#inspectpropertyrange-0x25

Because the artificial properties take quite a lot of space I suggest we:

- Identify the central artificial properties which need to be explicitly documented, such as the internal prototype and class name.
- Remove the explicitly decoded header flag properties for flags that don't fit the above criteria. Instead, provide the bitmask and allow the debug client to decode the bitmask based on debugger metadata if it wants to. This is analogous to what's done with opcodes.
- Specific bitmask flags can be decoded into artificial properties if they're directly useful, e.g. maybe "constructable" flag for a function. But internal GC flags (like "reachable", "temproot", etc) can be left as part of the bitmask and not decoded explicitly.

This would impact very few debug clients but would probably save close to 0.5-1kB of the feature footprint. It's always possible to decode more flags in the future if that turns out to be preferable.

fatcerberus:
I just refactored the entire SSJ codebase again and had to throw away all my Eval code in the process, this is a good opportunity to test this branch I think. :)

fatcerberus:
By the way how much of the additional footprint is actual code as opposed to strings?

svaarala:
I'd have to measure it to be sure but I'd guess 50-50 for artificial properties: the call sites for adding properties are maybe ~12-15 bytes and strings are on average shorter.

fatcerberus:
It looks like internal properties aren't being flagged correctly.  I set up SSJ to skip internal properties:

```c
			if (prop_flags & 0x100) {
				idx += is_accessor ? 4 : 3;
				continue;
			}
```

But they are still being displayed.

svaarala:
I rewrote the property emit code and that dropped out as an oversight, I'll add it back.

fatcerberus:
Other than that issue with internal properties, InspectPropertyRange at least seems to work correctly.  I still need to test InspectHeapObject, though.

svaarala:
@fatcerberus How does this approach feel to you? Do the primitives seem right?

Merging the entry and array parts into a conceptual index space turned out to be relatively straightforward after all: I used a helper which does the conceptual index to entry/array part mapping statelessly. This makes the outer loop unaware of the mapping so that straddling the parts requires no special handling (which would easily become error prone).

fatcerberus:
I'm failing to envision a use case for InspectPropKey, but other than that this definitely seems more natural: One command to get information about the object (metadata), then another to pull in the concrete property table.  It's a much more defined separation of concerns than before.

svaarala:
I think one concrete use case was described by @harold-b: to get the `.length` of an array before pulling in the array entries (note that `.length` may be != `a_size` so `a_size` is not enough to know the user-visible length). But that'd depend on the use case.

fatcerberus:
By the way the "artificial" flag (0x200) is no longer needed with this approach.

svaarala:
That's true, I'll fix that. The internal flag is still a bit 50-50 in my mind: internal properties may be added also by user code so they're not really "internal" in that sense. But I'll add back the flagging for them.

svaarala:
If `.length` is the only reasonable case for ReadObjectKey, we could just stick that into the metadata instead. Overall it wouldn't really make a difference though, as ReadObjectKey is very small: it shares the same helper for serializing the property.

fatcerberus:
Just for the sake of argument re: internal properties: I view them as basically the same case regardless of whether it's Duktape or the application adding them, as the intent is for Ecmascript code not to be able to see them (the sandboxing isn't perfect of course, but that's the theory).  So it's useful to allow the client to easily filter them out without relying on implementation details like e.g. the initial 0xff byte.

svaarala:
The situation is a bit in a grey area because:

- Ecmascript code can currently access them given the right key (or constructing one). This is currently a feature, though maybe disabled for Duktape 2.x.
- ES6 introduces Symbols which will quite possibly be implemented as internal properties. They *are* accessible to Ecmascript code.

In addition to internal keys, there are other invalid UTF-8 keys which are equally difficult to access using Ecmascript code, but which are not flagged as internal keys because that's reserved to the 0xFF prefix only.

So, it's not a problem to flag the 0xFF prefix, but the meaning (especially going forward) may be a bit vague.

svaarala:
For InspectHeapObject there'd actually be no need for a per-property flags field either. But I have a feeling it would be needed later, so I'd be inclined to leave it in.

svaarala:
Ok, artificial property flag removed, internal property flag added back.

fatcerberus:
So far this is working well:
```
maps/Testville.rmp/maggie/genCommands:1 global()
(ssj) f 2
=> # 2: game() at scripts/main.js:53
53      MapEngine('Testville.rmp', 60);

scripts/main.js:53 game()
(ssj) x game
= {
    ---  "name" : "game"
    w-c  "fileName" : "scripts/main.js"
    ---  "length" : { float:0 }
    w--  "prototype" : { obj:"0000015e0df97cd0h" }
}

scripts/main.js:53 game()
(ssj) x manifest
= {
    wec  "name" : "Spectacles: Bruce's Story"
    wec  "author" : "Fat Cerberus"
    wec  "summary" : "Follow Scott Starcross in his quest to stop the Primus."
    wec  "resolution" : "320x240"
    wec  "script" : "scripts/main.js"
    wec  "minimumPlatform" : { obj:"0000015e122675c0h" }
    wec  "frameRate" : { float:60 }
    wec  "sceneRenderPriority" : { float:99 }
    wec  "logPath" : "~usr/Spectacles Saga/Console Log.txt"
    wec  "consoleBuffer" : { float:1000 }
    wec  "consoleLines" : { float:10 }
    wec  "disableAnimation" : false
    wec  "disableBattles" : false
    wec  "disableSplash" : true
    wec  "disableTitleScreen" : true
    wec  "directory" : "C:/src/spectacles-i/dist/"
}
```
```

**Comment by svaarala** [comment]
Rebased after merging the change to use integer dvalues when the `duk_tval` is compatible. You should now see integer dvalues for those properties that can be safely (without loss of information) be represented as signed 32-bit integers. This is only a representation change, conceptually the values are the same as before.

Example:

```
{"notify":"_TargetConnecting","args":["127.0.0.1",9091]}
{"notify":"_TargetConnected","args":["1 10499 v1.4.0-227-g9452a0f duk command built from Duktape repo"]}
{"notify":"Status","command":1,"args":[1,"dist/examples/cmdline/duk_cmdline.c","eval",1,0]}
{"request":"Eval","args":["[1,1.5,2,2.5,3]"]}
{"reply":true,"args":[0,{"type":"object","class":2,"pointer":"0000000001ddf9f0"}]}
{"request":"InspectPropertyRange","args":[{"type":"object","class":2,"pointer":"0000000001ddf9f0"},0,10]}
{"reply":true,"args":[7,0,1,7,1,{"type":"number","data":"3ff8000000000000","value":1.5},7,2,2,7,3,{"type":"number","data":"4004000000000000","value":2.5},7,4,3,7,5,{"type":"unused"},1,"length",5]}
```

Pretty printed and commented, the property range response is:

```js
[
    // 7=writable, enumerable, configurable
    // key: 0
    // value: 1 (integer)
    7,
    0,
    1,

    // key: 1
    // value: 1.5 (ieee)
    7,
    1,
    {
        type: "number",
        data: "3ff8000000000000",
        value: 1.5
    },

    // Key: 2
    // Value: 2 (integer)
    7,
    2,
    2,

    // Key: 3
    // Value: 2.5 (ieee)
    7,
    3,
    {
        type: "number",
        data: "4004000000000000",
        value: 2.5
    },

    // Key: 3
    // Value: 4 (integer)
    7,
    4,
    3,

    // Unused slot
    7,
    5,
    {
        type: "unused"
    },

    // 1 = writable
    // Key: "length"
    // Value: 5
    1,
    "length",
    5
]
```

fatcerberus:
SSJ's `examine` command is pretty much directly mapped to InspectPropertyRange now, and it's awesome. :)

```
scripts/main.js:6 eval()
ssj: x CoffeeScript.require
---- name            "require"
-wc- fileName        "~sys/coffee-script.js"
---- length          1
-w-- prototype       { obj: "00000156d663c130h" }
ewc- ./helpers       { obj: "00000156d663beb0h" }
ewc- ./rewriter      { obj: "00000156d663cd60h" }
ewc- ./lexer         { obj: "00000156d6639ac0h" }
ewc- ./parser        { obj: "00000156d84e1e50h" }
ewc- ./scope         { obj: "00000156d85893f0h" }
ewc- ./nodes         { obj: "00000156d8589d00h" }
ewc- ./sourcemap     { obj: "00000156d85d6c60h" }
ewc- ./coffee-script { obj: "00000156d85ca400h" }
ewc- ./browser       { obj: "00000156d85ca310h" }
```

I took inspiration from the output format of `ls -l` with attributes displayed like Unix permissions.

Playing around with this revealed a surprising thing about minisphere's API: All my system calls were enumerable, so doing `eval global` in SSJ spewed out a huge listing of properties (the entire minisphere API, plus whatever game code added).  Heap inspection allowed me to see this fact in front of me and then go about fixing it (by using `duk_def_prop()` to register system calls instead of `duk_put_prop_string()`).

fatcerberus:
My only complaint, and this is more personal preference than anything, is that array part entries are listed first:

```
scripts/main.js:6 eval()
ssj: x CoffeeScript.FILE_EXTENSIONS
ewc- 0      ".coffee"
ewc- 1      ".litcoffee"
ewc- 2      ".coffee.md"
-w-- length 3
```

I'd prefer the array part to be listed last, just in case an array has a bunch of add-on properties, I can see the numbered entries at the bottom of a long listing and immediately know it's an array.  That's all hypothetical of course, I doubt most JS developers go around adding a bunch of extra properties to their arrays!

So yeah, an inconsequential thing really, but I had to mention it anyway since it was bugging me. :)

fatcerberus:
On the other hand, it might be more useful in a GUI client to list the numbered array entries _first_, and it can do that without having to pull the rest of the properties.  So on second thought the current order does make the most sense and I was just being shortsighted as usual.  Ignore me. :)

If it bugs me enough I can always re-sort the keys myself on the client side.

svaarala:
I placed the array items first because they're also enumerated first:

```
duk> a = [1,2,3]
= 1,2,3
duk> a.foo = 'bar'
= bar
duk> for (var k in a) { print(k); }
0
1
2
foo
```

But matching order to enumeration order is not guaranteed in the debugger protocol on purpose because there may be changes later (ES6 will require some changes to enumeration, some of them looking quite weird and engine specific). I should actually change that text so that the conceptual order is not guaranteed, but *current* order is array part first, etc.

Assuming the order isn't guaranteed to be enumeration order here, the debug client should then just reorder the properties however it sees fit :)

svaarala:
That `ls -l` inspired syntax is quite nice for a console client by the way :)

svaarala:
Re: that enumeration ordering: https://esdiscuss.org/topic/nailing-object-property-order

Implementing that order which is now mandatory (and very odd in my view because integer-like properties enumerate first *even* for non-array objects!) will require either:

- Object changes so that integer keys are inserted in the middle of the entry part instead of being appended only (which is much simpler).
- Enumeration changes to reorder properties on the fly instead of just dumping them from the entry part. I guess the change would be to walk the keys, and notice if there are any integer-like properties in the entry part. If so, perform an in-place sort for the enumeration order.

If the latter approach is used, there'll be memory churn when the enumeration order is determined, and I wouldn't want that churn for the debugger implementation. Which is why I wouldn't want to guarantee the property ordering to be exact enumeration ordering.

fatcerberus:
That *is* very odd.  The standard is basically contradicting itself: On one hand saying integer indices aren't special and are simply shorthand for the equivalent string key, but on the other hand treating them as a special case anyway.

fatcerberus:
> That `ls -l` inspired syntax is quite nice for a console client by the way :)

Thanks, I thought so too.

Originally my `e` (eval) and `x` (examine) commands were the same operation except that `x` was more verbose when printing dvalues.  Now they are quite distinct: `e` prints the result in a friendly object literal syntax and omits non-enumerable properties; while `x` shows all properties and their attributes in the long form shown above.  It's quite useful to see what's going on internally sometimes. :)

svaarala:
Pushed a bunch of fixes and cleanups:

- Force mark-and-sweep twice when resuming: this ensures finalizers queued up during pause state are executed (should be quite rare but probably nice that they execute before execution resumes). This is not very critical though, nothing will leak (freeing will just be postponed). Tested manually.
- Add perftest for object garbage to estimate how much the extra "is paused" check in refzero impacts concrete performance. Cannot really measure the difference but being able to do a single bit test would be ideal.
- Go through FIXMEs, relevant FIXMEs are in pull description
- Missing debugger document parts, security note in debugger document

fatcerberus:
I noticed you changed the signed/unsigned issue from a FIXME to an XXX note - I guess you're postponing the fix for that? ;-)

fatcerberus:
Looks like some of the finalizer tests failed.  Regression caused by the refzero change maybe?

svaarala:
The signed/unsigned integer issue: there's no unsigned 32-bit integer dvalue type so the only compatible fix would be to use an IEEE double (number dvalue) to represent such values. Adding a new dvalue type for unsigned 32-bit integer would be an incompatible change so it can't be done in this pull. It's noted in #585 for a later fix.

svaarala:
Right, so the alternatives are:

1. Send as signed, and comment in debugger.rst that the field will change to unsigned later (as part of an incompatible break).
2. Use the "number" dvalue for the field which covers 53-bit signed range accurately. Or a slight variation, use the "number" dvalue only when the highest bit is set and a signed 32-bit won't do.

The latter is correct even now, and could then compatibly get encoded as an unsigned integer later. After all, all the number dvalues are equivalent - the only guarantee is that e.g. command and error numbers are not in the "number" dvalue encoding (although even that would work perfectly provided the debug client doesn't unnecessarily assume a specific encoding).

svaarala:
I'm a bit on the fence what to do about the GC behavior, some alternatives are:

- Leave as is: forced mark-and-sweep when resuming/detaching. Causes garbage created (but not yet collected) before getting paused to also get collected on resume/detach.  This is an unexpected side effect for the application garbage.
- Remove the forced GC entirely. Causes no changes to how actual application garbage is treated compared to master, but any garbage created during the paused state (even if not in a reference loop, i.e. normally refcount collectable) will remain in the heap until a mark-and-sweep.
- Try to work out some hybrid where objects and buffers created during paused state are refzero-queued and strings are freed later by mark-and-sweep (because they can't be queued, at least at the moment).

Each has pros and cons, but I think I'm in favor of the second choice because it has the least side effects for the application. Also, not much garbage is normally created when paused, only if you e.g. Eval structured values. Some strings might be created as a result of how commands operate, e.g. if a value is string coerced.

fatcerberus:
From a theoretical point of view I like the third option--delayed refzero processing seems to match what my expectations would be if I weren't familiar with Duktape.

Practically speaking, however, that's probably more trouble than it's worth.  I agree that little garbage is typically created during debugging, the majority of operations (even Evals) will be working on existing objects.  And running a mark-and-sweep after every pause isn't desirable in any case, since as you mentioned it triggers after every step when stepping.

svaarala:
Ignoring implementation difficulty, the ideal behavior would be to somehow detect garbage created during the paused state so that only garbage created for the debugger, even garbage in reference loops, would be collected. Application objects would not be touched. But this is pretty complicated in practice.

The next best thing would be to refzero queue all objects, strings also. But even in this case garbage in reference loops would remain until the next mark-and-sweep.

The hybrid is a bit more confusing, because strings (which never participate in reference loops) will be handled by mark-and-sweep later, but buffers and objects would be refzero collected, except for objects in circular reference loops. That's a bit much to comprehend :)

So maybe the most practical option now is to just let mark-and-sweep handle the garbage without doing anything artificial to speed it up. It's then a future work issue to improve the behavior.

I'll remove the forced GC and push.

fatcerberus:
I'd agree with that decision.

Honestly though I'm not sure it even needs to be fixed: Garbage collection is designed to be invisible to the application, freeing it of the need to manage allocation patterns as in e.g. C, and in practice is nondeterministic (at least at a high level).  So why should that app care if the GC runs sooner than it was "supposed to"?  The side effects only matter to Duktape.

svaarala:
GC may trigger finalizers, and interfere with memory pool measurements, which does matter to some use cases.

svaarala:
I've been trying to come up with better command names with little success.

InspectHeapObject doesn't actually inspect the whole heap object so it'd probably be better renamed to "get heap object metadata". It would be nice to have some shorter version for HeapObject, maybe Heapobj or HeapObj? So maybe GetHeapObjMeta. The 'Get' prefix is already used for reading bytecode which is internal data, so maybe it'd be good to stick to that. The downside is that "get" may imply side effects.

The property read commands now operate only on actual Ecmascript objects. They don't work on e.g. plain strings or buffers. So while you can read e.g. `.length` of a plain string in Ecmascript code due to "automatic object promotion" semantics, you can't do that now in the debug protocol property commands. So maybe GetObjPropKey, GetObjPropRange?

So what would this sound:

- `GetHeapObjMeta`
- `GetObjPropKey`
- `GetObjPropRange`

Better suggestions welcome :)

fatcerberus:
I would prefer `GetHeapObjData` or if that's too vague ("data" could be anything), `GetHeapObjInfo`.  "Meta" feels awkward.

The other two seem fine, and match the existing naming pattern of GetVar, GetLocals, etc.  Maybe GetObjPropKey could just be `GetObjProp`?

svaarala:
Ok, sounds good to me, so:

- GetHeapObjInfo
- GetObjProp
- GetObjPropRange

I'll do the renames shortly.

svaarala:
@fatcerberus I'm ready to merge this to master: this seems like the most viable approach so far.

There is definitely room for improvement e.g. in the documentation, but it'd be nice to have the bulk of the work in master (it's quite functional after all) and work on polishing the documentation etc in a separate pull. The most important thing with respect to merging is the basic design approach which I think we've reasonably verified now.

fatcerberus:
@svaarala I agree, this seems like the best approach.  Despite my objections to `GetObjProp` originally, the split is very similar in spirit to the split GetVar/GetLocals.  I don't have much use for the former personally, but I'm sure there are use cases for it. :)

One additional command that might be worth adding in the future (for symmetry) is PutObjProp.

Thanks for this by the way, object inspection was the one major weakness in minisphere's debugger.  It's so easy in JS to end up with a reference loop, and once that happens it completely defeats JSON-based solutions.

svaarala:
For symmetry it'd actually be more "DefObjProp" because we're dealing with property attributes and possibly setting up an accessor. That's one reason why "get" wasn't on obvious choice for me because the counterparts are logically get/put and something/define (inspect/define, read/define, peek/define, something like that).

In the Ecmascript APIs the counterpart to defineProperty() is getOwnPropertyDescriptor(), so there's no single verb for the "define" counterpart. Define/inspect seemed like a reasonable pair in some way (and in the Ecmascript APIs defineProperty() / inspectProperty() would be better in my opinion). I'd be happy to rename the commands once more if a more logical naming is found... :)

Glad to hear this is filling a purpose, this needed a few iterations :)

By the way, one more high level design alternative occurred to me yesterday: when the debugger is paused, the debug client could maybe interact with the value stack of the target similar to how API calls do. If this were the case, Eval would push a value on the stack so that it'd be reachable and no GC lockouts would be necessary. The other upside is that it'd mirror the API and be somewhat logical in that way. However, there are also downsides and my first impression is that it wouldn't be better overall.

fatcerberus:
I think getting the valstack involved would add a lot more complexity for the client and add transport overhead as well, since more back-and-forth communication will be needed in order to carry out an operation.  The current debugger setup is very elegant in that regard.

By the way I just added GetObjProp support to my GUI debugger, now it has a nice expandable treeview of properties instead of an ugly JSON dump in a textbox.  This way you can drill down and only see what you need to instead of sifting through a potentially huge JSON listing. :smile:

fatcerberus:
One thing I have to say I don't like about this is the array semantics - it returns "unused" values past the `.length` of the array.  I don't mind if it returns `unused`s for "gappy" arrays, but getting values past the end is unexpected and difficult for the client to deal with without explicitly code to parse out the `.length` property.

fatcerberus:
Re: Naming: I wouldn't necessarily mind `InspectObjProp` and `InspectObjPropRange`, although they are kind of verbose.  'GetHeapObjInfo` is fine as-is of course.

However in the end I think the current names will be most intuitive for users of the debugger protocol so I'd be hesitant to change them.  As long as there's a note in the docs that the command doesn't cause side effects like getter calls, I think it'll be okay.

svaarala:
Regarding `.length`, the internal structure would allow for objects with an array part but without a `.length` property at all, which is why I wouldn't want to tie the debugger protocol to assume `.length` exists, look it up, and have it affect how the array part is enumerated. Note that in the Ecmascript specification, and even inside Duktape, the array property handling is separate from managing `.length` which is just automatically updated on certain operations.

Anyway, I don't understand why the elements beyond `.length` needs specific handling? You can treat them exactly the same as entry part entries which are missing, i.e. skip them because their value is "unused".

In fact, the only difference between having an array part property beyond `.length` and an unused entry in the beginning of a property part is that the former is returned with an integer key -- both will have an "unused" dvalue and can be skipped based on that information alone. Even if the array part terminated at `.length` you could still have an arbitrary number of unused property slots in the entry part.

Also note that some arrays may not have an array part at all, and all the array entries will enumerate as string-keyed properties. You can create such an array easily as follows:

    var arr = [1, 2];
    arr[1000] = 2;  // array part abandoned
    arr.length = 2;  // back to length == 2, but still no array part

For a debug client to hide the array nature (if that's what is desired), it'd need to display this identically to an array `[1,2]` which had an array part.

svaarala:
Names matching Ecmascript would probably be GetObjPropDesc and GetObjPropDescRange... :)

Anyway, I'll merge this and open a cleanup pull for any tweaks and changes.

svaarala:
How about a screenshot of the tree view? :)

svaarala:
One more observation re: command naming: GetObjProp won't invoke side effects which means e.g. getters don't get called (which is good default behavior for a debugger). This alone wouldn't be a reason to add a separate command because it's not a big deal for the debug client to invoke the getter explicitly if needed (there's no command for that yet though).

However, the target may also be a Proxy instance. In this case the only properties you'll see are the internal properties pointing to the target and the handlers table. If the debug client wanted to invoke the side effects to see what the actual property value is for a certain property (as seen by user code), it'd need to emulate the entire Proxy semantics which is a bit more work than emulating accessors (which just means: call the getter).

This, I think, may motivate adding a "get object property with side effects" command. What would that be named? A logical name for that would be... GetObjProp :-)

But I'll merge this now, and open a separate pull for trivia :)

harold-b:
I have to get up to speed with this thread. I've been busy and haven't been able to track it! Looking forward to implementing it's use on my client.

fatcerberus:
@svaarala About that sparse array example: That's quite surprising to me!  I was under the impression (mistaken I see) that, if you have a normal array of length one and then assign a value to the 1000th index, not only will `.length` change to 1001, but all the intervening indices will then show up when the array is enumerated.

If this is not the case then obviously I have no qualms about just skipping the "unused" values whereever they occur. :)

svaarala:
Yes, the entries would be skipped in enumeration (regardless of the existence of an array part):

```
duk> a = [1,2]; a[100] = 1; for (var k in a) { print(k, a[k]); }
0 1
1 2
100 1
= undefined
```

However, because JSON doesn't support unmapped entries, it will fill in `null` for such entries (and any other nonsupported types like functions too):

```
duk> JSON.stringify(a)
= [1,2,null,null,[...],null,1]

duk> JSON.stringify([1,2,function foo(){},3])
= [1,2,null,3]
```

fatcerberus:
Huh.  I guess that's yet another great reason why everyone recommends against using `for...in` on arrays.  Thanks, here I was thinking I knew JS inside and out and you go and teach me something new :)

I'll get you a screenshot of the treeview in a bit.

svaarala:
The other reason is that technically there's no guarantee in ES5/ES5.1 for the index ordering. In practice there are pretty good (but not 100%) guarantees. Duktape uses the index ordering but *not* for sparse arrays:
```
duk> a = [1,2]; a[1000]=3; a.length=2;
= 2
duk> a[4] = 'foo'
= foo
duk> a[3] = 'bar'
= bar
duk> for (var k in a) { print(k, a[k]); }
0 1
1 2
4 foo
3 bar
= undefined
```

This is required to be in index order (as opposed to insert order) in ES6.

fatcerberus:
@svaarala Here you go:
SSJ2

It's not perfect; my original preference was for the key and value in separate columns, but apparently WinForms has no hierarchical listview control, so this was the best I could do without bringing in external dependencies.  But this is still a huge improvement over the JX solution I had in place before.

svaarala:
Quite interestingly, `Array.forEach()` is required to enumerate in index order, so the internal enumeration primitive supports index sorting via an internal flag: https://github.com/svaarala/duktape/blob/master/src/duk_hobject_enum.c#L34-L159. So fixing it for ES6 is not a big deal, but just annoying because it affects performance.

svaarala:
@fatcerberus Cool, so you can launch an object viewer from e.g. a variable value I guess?

fatcerberus:
Yeah.  It should also trigger that window if an Eval produces an object result.

Note: The `{ obj: 'Object' }` text changes depending on the class number. :)

PR #622, opened 2016-02-27
Debugger pause trivia

svaarala:
- Minor documentation edits

### DIFF
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -323,16 +323,14 @@ blocking), you can also use it like this::
 duk_debugger_pause()
 --------------------
 
-The target may call this at any time to pause Ecmascript execution and turn over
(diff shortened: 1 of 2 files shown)


svaarala:
@fatcerberus Just a few edits I didn't want to bother you with - the main thing being to clarify that (1) the call returns immediately (which may not be apparent) and (2) that it's safe to call and a no-op when already paused.

fatcerberus:
Looks good.  It was difficult to get the API documentation right - I wanted to communicate as much of the semantics as possible without making the summary overly long.  Your rewrite did a good job I think. :)

svaarala:
I tried rewording the pause description a bit more; I removed some of the overlap from the debugger.rst (it doesn't need to be as detailed as the API documentation).

fatcerberus:
Looks much better now. :)

PR #629, opened 2016-03-04
Minor cleanups for debugger heap object inspection support

svaarala:
- [x] Minor code comment improvements
- [x] Document guaranteed artificial properties which will have version guarantees
- [x] Add explicit note on how Proxy objects are treated, i.e. traps are not invoked, how to detect a Proxy
- [x] Note on sparse arrays, missing elements, and potentially unordered indices
- [x] Command naming, downside of using "get" is that "get with side effects" (normally called "get") clashes
- [x] Comment on crossed indices in GetObjPropRange
- [x] Test command names with debug proxy

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1371,8 +1371,8 @@ Planned
   (read-only data section) which reduces the startup RAM usage of a low
   memory build to ~3kB (from ~27kB) (GH-559)
 
-* Add debugger heap object inspection commands GetHeapObjInfo, GetObjProp,
-  and GetObjPropRange, which allow a debug client to inspect heap objects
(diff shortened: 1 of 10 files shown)


fatcerberus:
By the way, following up on https://github.com/svaarala/duktape/pull/617#issuecomment-192196028:
I would be okay with it if you decided to rename `GetObjProp` to `GetObjPropDesc`.

svaarala:
There's no functional change here: just documentation clarifications and renaming internal defines.

fatcerberus:
Oh good, now Duktape master matches the names in my debug client code. ;)

Issue #637, opened 2016-03-13
Error thrown during CJS module cause debugger pause in wrong location


fatcerberus:
An unhandled Error thrown during processing of `require()` will cause the debugger to halt at the `require()` call, outside the module.  The expected behavior is for a pause to occur at the throw site inside of the module.

fatcerberus:
Nevermind, false alarm - turns out there's no Duktape bug, it was caused by my GUI client automatically sending a Step Out command when it can't find source code.  SSJ downloads source code directly from minisphere so it correctly stopped inside the module.

fatcerberus:
There is an issue though, not a bug but an annoyance: Duktape automatically uses the module ID as the filename, which makes it difficult for the client to find the source code (since it can't differentiate a module ID from a real filename)

svaarala:
The function executing when a module is being loaded is not actually the source provided by modSearch() directly. Instead, a function wrapper is automatically generated around the source code, so all line numbers will reference that function (because the wrapper has now newlines line numbers will match; column numbers wouldn't match for the first line). This is an artifact of the current implementation.

Currently modSearch() will just return a source file with no filename information so the filename is not actually available for compiling the wrapper. One simple compatible workaround would be to allow modSearch() to assign `module.fileName` so that when a source file is returned, that would be set as the fileName of the function wrapper. The property could be initialized with the module ID as it is now.

svaarala:
I'll move the .fileName discussion to #639 (placeholder pull request).

PR #639, opened 2016-03-14
Add support for module.fileName and module.name

svaarala:
Currently Ecmascript modules are loaded using a wrapper function whose `.fileName` is automatically set to the module ID. This is a useful default, but in some situations user code may want to use a different fileName, e.g. mapping to a concrete file when one exists. See #637.

Change current behavior so that:

- `module.fileName` is initialized to the resolved module ID, i.e. the current .fileName used
- `module.name` is initialized to the last component of the resolved module ID (e.g. `foo/bar` -> `bar`)
- Allow user modSearch() to replace `module.fileName` and/or `module.name`
- Use `module.fileName` for the wrapper function `.fileName` so that it shows up in tracebacks etc
- Similarly `module.name` is used for the wrapper function `.name`

Tasks:

- [x] Change `require()` behavior to use module.fileName
- [x] Change module wrapper function .name also to match module.fileName
- [x] Add module.name handling
- [x] Add #defines for stack indices in require() implementation
- [x] Rewrite `q_last` tracking to avoid explicit scan
- [x] Add .name for fresh `require()` function
- [x] Testcase for setting module.fileName in modSearch()
- [x] Testcase for module wrapper function name, affects stack traces generated during module load
- [x] Testcase for initializing a logger within a module with default fileName (= module resolved ID) and modified module.fileName
- [x] Testcase for ensuring that module wrapper function's `.name` does not create a capturing function name binding
- [x] Testcase for module.name overriding
- [x] Testcase for fresh require() .name
- [x] Stress testcase for last component scanning (truncated IDs etc)
- [ ] Assertion run
- [x] Update documentation
- [x] Wiki modules howto update: https://github.com/svaarala/duktape-wiki/pull/104
- [x] Releases entry

### DIFF
--- a/src/duk_bi_global.c
+++ b/src/duk_bi_global.c
@@ -580,7 +580,7 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_eval(duk_context *ctx) {
 	DUK_DDD(DUK_DDDPRINT("eval -> lex_env=%!iO, var_env=%!iO, this_binding=%!T",
 	                     (duk_heaphdr *) outer_lex_env,
 	                     (duk_heaphdr *) outer_var_env,
-	                     (duk_tval *) duk_get_tval(ctx, -1)));
+	                     duk_get_tval(ctx, -1)));
 
 	/* [ source template closure this ] */
 
@@ -846,6 +846,7 @@ DUK_LOCAL void duk__bi_global_resolve_module_id(duk_context *ctx, const char *re
 	duk_uint8_t buf_out[DUK_BI_COMMONJS_MODULE_ID_LIMIT];
 	duk_uint8_t *p;
 	duk_uint8_t *q;
+	duk_uint8_t *q_last;  /* last component */
 
 	DUK_ASSERT(req_id != NULL);
 	/* mod_id may be NULL */
@@ -907,9 +908,17 @@ DUK_LOCAL void duk__bi_global_resolve_module_id(duk_context *ctx, const char *re
 	for (;;) {
 		duk_uint_fast8_t c;
 
-		/* Here 'p' always points to the start of a term. */
+		/* Here 'p' always points to the start of a term.
+		 *
+		 * We can also unconditionally reset q_last here: if this is
+		 * the last (non-empty) term q_last will have the right value
+		 * on loop exit.
+		 */
+
 		DUK_DDD(DUK_DDDPRINT("resolve loop top: p -> '%s', q=%p, buf_out=%p",
 		                     (const char *) p, (void *) q, (void *) buf_out));
+
+		q_last = q;
 
 		c = *p++;
 		if (DUK_UNLIKELY(c == 0)) {
@@ -959,6 +968,9 @@ DUK_LOCAL void duk__bi_global_resolve_module_id(duk_context *ctx, const char *re
 				*q++ = c;
 				c = *p++;
 				if (DUK_UNLIKELY(c == 0)) {
+					/* This was the last term, and q_last was
+					 * updated to match this term at loop top.
+					 */
 					goto loop_done;
 				} else if (DUK_UNLIKELY(c == '/')) {
 					*q++ = '/';
@@ -980,8 +992,17 @@ DUK_LOCAL void duk__bi_global_resolve_module_id(duk_context *ctx, const char *re
 		}
 	}
  loop_done:
-
+	/* Output #1: resolved absolute name */
+	DUK_ASSERT(q >= buf_out);
 	duk_push_lstring(ctx, (const char *) buf_out, (size_t) (q - buf_out));
+
+	/* Output #2: last component name */
+	DUK_ASSERT(q >= q_last);
+	DUK_ASSERT(q_last >= buf_out);
+	duk_push_lstring(ctx, (const char *) q_last, (size_t) (q - q_last));
+
+	DUK_DD(DUK_DDPRINT("after resolving module name: buf_out=%p, q_last=%p, q=%p",
+	                   (void *) buf_out, (void *) q_last, (void *) q));
 	return;
 
  resolve_error:
@@ -990,6 +1011,19 @@ DUK_LOCAL void duk__bi_global_resolve_module_id(duk_context *ctx, const char *re
 #endif  /* DUK_USE_COMMONJS_MODULES */
 
 #if defined(DUK_USE_COMMONJS_MODULES)
+/* Stack indices for better readability */
+#define DUK__IDX_REQUESTED_ID   0  /* Module id requested */
+#define DUK__IDX_REQUIRE        1  /* Current require() function */
+#define DUK__IDX_REQUIRE_ID     2  /* The base ID of the current require() function, resolution base */
+#define DUK__IDX_RESOLVED_ID    3  /* Resolved, normalized absolute module ID */
+#define DUK__IDX_LASTCOMP       4  /* Last component name in resolved path */
+#define DUK__IDX_DUKTAPE        5  /* Duktape object */
+#define DUK__IDX_MODLOADED      6  /* Duktape.modLoaded[] module cache */
+#define DUK__IDX_UNDEFINED      7  /* 'undefined', artifact of lookup */
+#define DUK__IDX_FRESH_REQUIRE  8  /* New require() function for module, updated resolution base */
+#define DUK__IDX_EXPORTS        9  /* Default exports table */
+#define DUK__IDX_MODULE         10  /* Module object containing module.exports, etc */
+
 DUK_INTERNAL duk_ret_t duk_bi_global_object_require(duk_context *ctx) {
 	const char *str_req_id;  /* requested identifier */
 	const char *str_mod_id;  /* require.id of current module */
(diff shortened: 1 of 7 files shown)


fatcerberus:
I like this, it seems like a very clean solution.  My idea was for modSearch to return an object like:

```javascript
return { source: "...", fileName: "commonjs/module.js" }
```

But that's not compatible of course, unless the implementation still allows a string to be returned.  `module.fileName` is a better solution.

svaarala:
Independent of using `module.fileName` it would probably be better to return an object instead of a string from modSearch() (or whatever the equivalent is in the future). It's a place in the API where it's quite easy to see the need for some backward compatible changes and with a string value there's no "space" for them.

But for this particular change I think `module` might be good enough. The final `module` object gets cached in `Duktape.modLoaded[]` so it's not appropriate for passing any temporary control information (which is not needed after loading is complete). Keeping `module.fileName` around after loading is complete is OK I think because it's related to where the module came from (i.e. it's actual metadata rather than just some temporary control information).

fatcerberus:
Regarding the function wrapper used for modules: Is there a reason why Duktape can't simply compile the module code using the internal equivalent of `DUK_COMPILE_FUNCTION` instead of modifying the source string directly?

svaarala:
`DUK_COMPILE_FUNCTION` expects to see the function argument list (`function (x, y, z) ...`) which is used to get a list of bound variables. A module source has local bindings for `module`, `exports`, etc, which are not declared in the module source at all. So there needs to be some way of compiling a function body but declaring the bindings (here argument names) from outside the source itself. As there's no internal API to do that, using a function wrapper is the shortest way to accomplish the goal.

But it's definitely a feature to keep in mind next time the compiler gets some attention. It'd be nice to be able to compile sequences of statements more flexibly, even outside modules.

svaarala:
With the updated behavior, when modSearch() sets `module.fileName = "my_source.js";` an error traceback inside the module looks like this:

```
Error: aiee
    at my_source.js (my_source.js:2) preventsyield
    at require () native strict preventsyield
    at test (tests/ecmascript/test-commonjs-module-filename.js:61)
    at global (tests/ecmascript/test-commonjs-module-filename.js:82) preventsyield
```

If `module.fileName` is not set, it defaults to the resolved module ID (here `test/foo2`):

```
Error: aiee
    at test/foo2 (test/foo2:2) preventsyield
    at require () native strict preventsyield
    at test (tests/ecmascript/test-commonjs-module-filename.js:51)
    at global (tests/ecmascript/test-commonjs-module-filename.js:82) preventsyield
```

Before the change, in Duktape 1.4.x and below:

```
Error: aiee
	anon test/bar2:2 preventsyield
	require  native strict preventsyield
	test tests/ecmascript/test-commonjs-module-filename.js:61
	global tests/ecmascript/test-commonjs-module-filename.js:82 preventsyield
```

The two improvements are:

- The function .name in the traceback was "anon" but now matches module.fileName, e.g. "my_source.js" or the resolved module ID if nothing is explicitly set. It's a bit odd as a function name but better than the anonymous default I think.
- The function .fileName in the traceback is now module.Filename, e.g. "my_source.js" or the resolved module ID if nothing is explicitly set. This is the main change requested by @fatcerberus.

svaarala:
I'm not sure which behavior is better for the module wrapper function's `.name` property. The current behavior in this branch leads to the stack trace:

```
Error: aiee
    at my_source.js (my_source.js:2) preventsyield
    at require () native strict preventsyield
    [...]
```

The alternative is not to set `.name` so that the function appears anonymous, as the `.fileName` property will still reflect the module name or filename:

```
Error: aiee
    at [anon] (my_source.js:2) preventsyield
    at require () native strict preventsyield
    [...]
```

Finally, it could be set independently of `.fileName` to match the resolved module ID:

```
Error: aiee
    at test/bar2 (my_source.js:2) preventsyield
    at require () native strict preventsyield
    [...]
```

This could be made user controllable by initializing `module.name` to the resolved module ID (`test/bar2`) but allow it to be overridden in modSearch() so that both the function .name and function .fileName appearing in tracebacks would be fully controllable.

fatcerberus:
Personally, the most reasonable defaults to me would be:
* `.fileName`: the full resolved module ID (like it is now)
* function wrapper `.name`: The final component of the resolved module ID.  e.g. If you have a module `foo/bar/baz`, the function name would be `baz`.  That seems the most consistent with the rest of Duktape, e.g. `Array.prototype.forEach()` shows up in a stack trace as "forEach"

Using the full module ID for `.name` seems awkward since a function can't normally have slashes in its name (unless you manually change it after definition).

svaarala:
I agree those would be the best defaults. I'll see if I can manage to add scanning of the last component cheaply ;)

svaarala:
Logger default name comes actually from the caller `.fileName` so that won't default to the last component without some change in that behavior (in a separate pull, if at all).

svaarala:
Ok, so now the default traceback looks like:

```
Error: aiee
    at foo2 (test/foo2:2) preventsyield
    at require () native strict preventsyield
    at test (tests/ecmascript/test-commonjs-module-filename.js:44)
    at global (tests/ecmascript/test-commonjs-module-filename.js:76) preventsyield
```

The .name defaults to "foo2". It can be overridden in modSearch via `module.name` to get e.g.:

```
Error: aiee
    at my_module (my_source.js:2) preventsyield
    at require () native strict preventsyield
    at test (tests/ecmascript/test-commonjs-module-filename.js:55)
    at global (tests/ecmascript/test-commonjs-module-filename.js:76) preventsyield
```

Here modSearch() has set `module.name = 'my_module'` and `module.fileName = 'my_source.js'`.

svaarala:
@fatcerberus Want to give this branch a spin re: your debugger issue? I might tinker a bit with it still but it should be fully functional now.

fatcerberus:
Sure, I'll test it out a bit later, I'll have to update my modSearch to support this first.

src/duk_bi_global.c
```diff
@@ -980,8 +981,24 @@ DUK_LOCAL void duk__bi_global_resolve_module_id(duk_context *ctx, const char *re
...
  loop_done:
-
+	/* Output #1: resolved absolute name */
 	duk_push_lstring(ctx, (const char *) buf_out, (size_t) (q - buf_out));
+
+	/* Get last component of resolved ID: backtrack to previous slash or
+	 * start of buffer.
+	 */
+	q_last = q;
+	while (q_last > buf_out && *(q_last - 1) != '/') {
```
fatcerberus: Is the backtrack scan necessary?  The ID string was already scanned once, presumably the resolver could just note the position of the final slash and then we'd use that here.

fatcerberus:
Looks like it works, using this branch and a matching change to `modSearch`, my GUI debugger is now able to find the source files for modules.

I noticed a bug which I assume is unrelated to this change, but figured I'd mention it: When calling `require()` from inside of a module, it shows up as an anonymous function.  Here's an SSJ backtrace to illustrate the issue:
```
~sys/commonjs/link.js:10 link()
(ssj) bt
   # 0: Abort() <system call>
=> # 1: link() at ~sys/commonjs/link.js:10
   # 2: [anon] <system call>     <-- this is a require() call
   # 3: console() at ~sys/commonjs/miniRT/console.js:12
   # 4: require() <system call>
   # 5: eval() at scripts/main.js:6
```

svaarala:
I first implemented it by tracking the last slash but there are multiple places where they may be encountered (and I actually managed to introduce a bug in doing so).

But I think it should suffice to set `q_last = q` at the start of the scanning loop unconditionally: the scanning loop always begins at the first character of a new term (a NUL at the top of the loop is a resolution error). I'll try this out.

svaarala:
That require() thing is related to the fresh require lacking a name; I'll fix that too.

svaarala:
There's actually no footprint difference (at least for x64) between updating q_last in the loop vs. scanning for it explicitly (the scanning compiles to very dense code anyway, and the q_last assigned in the loop doesn't need to be saved past the first duk_push_string()). But I'll leave the current version (= assigning it at the top of the loop) in.

fatcerberus:
Thanks, the `require()` issue is fixed and the branch still seems to be working correctly:
```
uncaught error! - Error: *munch*
    at: ~sys/commonjs/link.js:10
=> # 0: link() at ~sys/commonjs/link.js:10
10 throw new Error("*munch*");
~sys/commonjs/link.js:10 link()
(ssj) bt
=> # 0: link() at ~sys/commonjs/link.js:10
   # 1: require() <system call>
   # 2: console() at ~sys/commonjs/miniRT/console.js:12
   # 3: require() <system call>
   # 4: eval() at scripts/main.js:6
```

svaarala:
Good to hear :) Added minimal documentation and a separate pull for a Wiki update. I'll re-run the assertion test having touched the last component scanning, and merge in a bit.

fatcerberus:
It seems this isn't so outlandish after all: Node.js has `module.filename` (and it's available to the module itself!):
https://nodejs.org/api/modules.html#modules_module_filename

svaarala:
It would probably make sense to use `module.filename` instead of `module.fileName` to match Node.js behavior. The casing is inconsistent with other uses of `.fileName` unfortunately :/

PR #640, opened 2016-03-15
Add a minimal multi-platform testrunner with push webhook script jobs

svaarala:
Add a minimal multi-platform test runner which integrates to Github webhook "push" events. For each authorized push, run a set of script test cases using a client/server approach. The goal is to include current makefile tests as script jobs, and run them on multiple platforms (in particular OS X and Windows) automatically for each commit.

This is very early work but still functional and will be enabled for the main repo incrementally. Expect to see funny Github status lines :)

### DIFF
--- a/testrunner/Makefile
+++ b/testrunner/Makefile
@@ -1,0 +1,15 @@
+# Some useful development targets.
+
+.PHONY: dist
+dist:
+	rm -f testrunner-server-node.tar.gz
+	rm -f testrunner-client-simple-node.tar.gz
+	tar cvfz testrunner-server-node.tar.gz \
+		server-node/*.js \
+		server-node/package.json \
+		server-node/config.yaml.template
+	tar cvfz testrunner-client-simple-node.tar.gz \
+		client-simple-node/*.js \
+		client-simple-node/*.sh \
+		client-simple-node/package.json \
+		client-simple-node/config.yaml.template
--- a/testrunner/README.rst
+++ b/testrunner/README.rst
@@ -1,0 +1,312 @@
+==================
+Duktape testrunner
+==================
+
+Overview
+========
+
+Duktape testrunner provides multi-platform client-server test execution
+which is integrated with Github webhooks.  Current functionality:
+
+* Accepting webhook POSTs from Github
+
+* Managing the execution of a set of tests associated with commits
+  ("push" events) together with test clients; each test is a simple
+  script with a textual output and a success/failure indication
+
+* Integrating test execution status with Github status API
+
+* Providing a placeholder web UI for showing pending tests
+
+The main goals for the client-server model are:
+
+* The ability of execute tests on a wide variety of platforms, including
+  exotic platforms which have no emulators and sometimes no TCP/IP networking.
+
+* Allow easy integration of simple test scripts to make it easy to extend test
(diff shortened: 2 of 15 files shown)


fatcerberus:
Huh, when I click on "Details" on the new tests I get a browser warning saying the certificate is invalid.

svaarala:
It's a self signed certificate - it's "invalid" on purpose.

fatcerberus:
I figured that was the case, just wanted to point it out since you have to jump through hoops to get Firefox to load the log page (adding an exception, etc.).  Chrome makes it a bit easier since you can just click through.

svaarala:
The first things I want to add coverage-wise is OS X and Windows builds - at first simply compilation tests using various compilers and options.

svaarala:
Yeah, I'll probably make the links plain HTTP later, there's really no point in using HTTPS for that.

svaarala:
Another coverage thing which I'd like to add is compiling a few largeish applications using Duktape to see changes don't break compilation (which is the version compatibility promise given).

@fatcerberus How complicated would it be to compile Minisphere+Duktape as a test job? It doesn't need to be a headless Windows machine, I can just run the test client in Cygwin or command prompt and leave it running.

fatcerberus:
minisphere isn't terribly difficult to compile.  It depends on Allegro, libmng and zlib.  For Windows, the latter two are included as static libraries in the repo, and the former is pulled using NuGet, so it's easy to automate.  For your part you'd just have to copy duktape.c, duktape.h and duk_config.h into src/shared.

On Ubuntu you can get all the dependencies from the package manager, so that's easy to automate too.

svaarala:
A minor improvement over a "compile works" test case is to run some sort of "hello world" script which verifies that the binary doesn't e.g. crash on startup. Is there a way to e.g. run a "mini game" which would start up and pretty much immediately exit? :)

fatcerberus:
Yeah, as of version 2.0 you can pass the engine a `.js` script on the command line and it will run it.  And you're in control of the event loop, so unless you deliberately code a FlipScreen/DoEvents loop, the script will run and exit.

fatcerberus:
The command line you want by the way is `spherun foo.js`.  If you want logging (shouldn't be necessary just to test Duktape) use `--verbose <1-4>`.

svaarala:
Ok, I'll see about adding that as a compilation test later on. Most likely I'll use a pinned minisphere version and avoid downloading packages for the test (depending on how that works e.g. a caching proxy works for wget). But this is much later, once the basics are in place and working better :)

fatcerberus:
By the way I just wanted to ask if Duktape 1.5.0 is in sight?  I'm releasing minisphere 3.0 on the 28th.  If it won't be available by then that's okay--I'll just use Duktape master. :)

svaarala:
There are a couple of open pulls which have an API or compatibility impact which I'd like to resolve (e.g. the new API call for property list initializers). Improving automatic test coverage is also something I'd like to improve on before the release - this pull and a followup cleanup pull should go a long way in getting a better baseline in place.

All in all, two weeks might be a bit tight, but master should be in pretty good shape by 28th anyway.

fatcerberus:
@svaarala minisphere 3.0 has been released, so you can use that as a baseline for testing once you get around to that. :)

PR #641, opened 2016-03-15
Cleanups for minimal webhook testrunner

svaarala:
- [x] Add test client name (e.g. "OS X testclient") which can be shown for pending jobs #653
- [x] Include test client name in Github status for "Running..."
- [ ] Include test client name in barebones web UI front page
- [ ] Minimal web UI front page styling, zebra patterns, margins, etc
- [ ] Add an explicit ID for simple commit job tracking states, so that get/finish primitives can reference a specific state without guesswork (improves handling of re-delivered webhooks)
- [ ] Add plain HTTP port for the server, provide web UI also there; change data links to plain HTTP to avoid issues with self-signed certificate
- [ ] Add explicit script indication of "unknown error" vs. "certain testcase error"; a certain testcase error never needs a retry, but other errors warrant one or two automatic server-driven retries (e.g. OS X github.com DNS issues when user logged out)
- [x] Avoid Makefile targets, as they depend on writing and cleaning `/tmp` which conflicts with parallel runs (e.g. test262test is broken for this reason now)
- [x] Linux x64 coverage
- [x] Linux x86 coverage
- [x] OS X x64 coverage - skipped due to OS X wifi issues
- [x] Windows x64 coverage (concrete issue is running a Cygwin bash script; Node.js doesn't execute BAT files) - same as OS X (VM runs on OS X), use AppVeyor to get some Windows coverage
- [x] From release checklist: dddprint coverage
- [x] From release checklist: separate sources build
- [x] From release checklist: packed duk_tval by default on x86
- [x] From release checklist: genconfig from distributable
- [ ] Note compiler warnings in e.g. duk-clang build (can mark as success but use description to indicate warning count)
- [ ] Scan-build automatically, at least to see results if known issues can't be suppressed
- [ ] Coverage for Array fast paths
- [ ] Coverage for fastints
- [ ] Show testcase diffs

### DIFF
--- a/testrunner/Makefile
+++ b/testrunner/Makefile
@@ -7,7 +7,8 @@ dist:
 	tar cvfz testrunner-server-node.tar.gz \
 		server-node/*.js \
 		server-node/package.json \
-		server-node/config.yaml.template
+		server-node/config.yaml.template \
+		server-node/style.css
 	tar cvfz testrunner-client-simple-node.tar.gz \
 		client-simple-node/*.js \
 		client-simple-node/*.sh \
--- a/testrunner/server-node/server.js
+++ b/testrunner/server-node/server.js
@@ -2,6 +2,13 @@
  *  Testrunner server.
  */
 
+// FIXME: separate contexts in web UI; identify client by human readable name (e.g. "OS X testclient"
+// FIXME: minimal web UI styling (tables mainly, zebra pattern, margins)
+// FIXME: add webhook identifier to allow retriggering
+// FIXME: prefer newest webhook when possible (useful for successive forced pushes)
+// FIXME: plain http port
+// FIXME: separate script exit code for "certain test failure" vs "other failure"; automatic server-driven retry
+
 var fs = require('fs');
 var yaml = require('yamljs');
 var express = require('express');
@@ -60,8 +67,11 @@ function main() {
     var logRequest = expressutil.makeLogRequest();
 
     // URI paths served.
-    app.get('/', logRequest,
-            webuiutil.makeIndexHandler(state));
+    var indexHandler = webuiutil.makeIndexHandler(state);
+    app.get('/', logRequest, indexHandler);
+    app.get('/index.html', logRequest, indexHandler);
+    app.get('/style.css', logRequest,
+            webuiutil.makeStaticFileHandler(state, 'style.css', 'text/css'));
     app.get('/out/:sha', logRequest,
             webuiutil.makeDataFileHandler(state));
     app.post('/github-webhook', logRequest, githubJsonBodyParser,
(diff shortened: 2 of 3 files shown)


svaarala:
Having quite a rough going with getting my Mac Mini to work reliably with all users logged out but `screen` running tests in the background. For some reason DNS resolution eventually starts failing even with all power saving disabled (but the client successfully communicates with the test server).

On Windows launching Bash scripts from Node.js (which doesn't have a Cygwin version AFAIK) is rather painful and I may need to implement the test scripts in Python instead of bash.

So, weird statuses will be around for a while :)

fatcerberus:
I've noticed Travis themselves often have connectivity issues with their OS X builds... I guess that operating system just doesn't like running headless. :-/

As for Node.js and Cygwin I believe you're right.  I once tried to compile Node myself to get a Cygwin-native version but couldn't get it working.  So I was locked out of a good part of the Duktape makefile as a result.

svaarala:
My minimal goal for Windows now is to get MinGW + MSVC compilation + hello world tests working (including a DLL build). This covers quite a lot of ground. Adding a test for warning free compile would also be useful but not part of the minimally useful Windows coverage. But even this goal needs Node.js and bash scripts to work together, unless I rework the bash scripts.

fatcerberus:
You _can_ use the Windows version of Node.js with Cygwin, but you need to convert any paths to Windows format using `cygpath -w /path/to/file` (or `cygpath -u` to convert in the other direction).  Which would probably be easy enough to automate, assuming there is a way to reliably detect Cygwin vs. a native Unix environment.

svaarala:
Yes, I can run Node.js from Cygwin or bash scripts from command prompt (by providing an explicit bash command to move into Cygwin for script execution) - but it gets very messy with paths, robust escaping, etc.

svaarala:
Ok, now with the Python script (Node.js test client launches a Python script) tests run both in Cygwin and outside Cygwin. It's important to avoid using e.g. using the "tar" command and use the "tarfile" Python module instead: otherwise you may be running Python in command prompt but "tar" is provided by Cygwin which again leads to a lot of path headaches.

Still having issues with OS X networking when I leave a screen running in the background :/ I'll probably need to give up on that for now and focus on the 1.5 issue list. But at least there's something slightly better in place for testing, and I'll try to add most of the release checklist items as test runs so that the release-worthiness of each merge is tested more often.

fatcerberus:
On the plus side, all this additional test coverage will hopefully lead to making things easier for Duktape developers maining Windows (like me) :)

svaarala:
That's true - the automatic tests are still only enabled for white listed authors (I can add you to that list naturally) but they're not enabled for arbitrary pull requests because there's some potential for abuse (you can essentially compile and run arbitrary C code in the test VMs).

svaarala:
Closing as unmerged.

PR #642, opened 2016-03-16
Drop clen16 field from duk_hstring in low memory builds

svaarala:
This change reduces `duk_hstring` size from 12 to 8 bytes for low memory builds. Character lengths will be computed on-the-fly instead for non-ASCII strings (this has some footprint cost but is more than offset by the reduced string size).


fatcerberus:
I'm confused, how do you save 4 bytes per string (12 -> 8) by only dropping a 2-byte field?

svaarala:
Because the struct has a 4-byte field (heaphdr_flags) it was 10 bytes + 2 bytes padding before.

Issue #516, opened 2016-01-05
Change duk_(p)eval() .fileName default from __FILE__ to "eval"


svaarala:
Using `__FILE__` for the eval source .fileName leads to file/line combinations in stack traces where the filename is from the C file (`__FILE__`) but the line number is from the eval'd Ecmascript code, which is quite misleading.

If this default is not appropriate, calling code can even now use `duk_(p)compile()` + `duk_(p)eval()` to provide an explicit filename.

svaarala:
To elaborate a bit, consider the following call site:

```
/* test.c, eval is on line test.c:123 */
duk_peval_string(ctx, "\n\naiee");  /* ReferenceError on line 3 */
```

As things are now (up to 1.4.0) the temporary function compiled for the eval input has `.fileName` set to `__FILE__` of the call site, i.e. `test.c` here. As a result, the stack trace for the ReferenceError indicates the error occurred on file/line `test.c:3` which is quite misleading.

If the `.fileName` for the eval code were `eval` the error would be blamed on `eval:3` which is correct but doesn't identify the C file with the eval call site.

The `.fileName` could also be set to:

- `test.c/eval` so that stack trace would blame `test.c/eval:3`
- `test.c:123/eval` so that stack trace would blame `test.c:123/eval:3`

fatcerberus:
I'm confused... I know with the changes in 1.4 Duktape walks the callstack looking for an entry with a valid filename, but I was under the impression that it always used the line number from the same stack frame as the filename.  So if the duk_eval() code has no filename, the error should be correctly blamed on `test.c:123` and line 3 should never come into the equation?

svaarala:
`duk_eval()` creates a temporary function whose filename is `__FILE__` of the `duk_eval()` call site.

svaarala:
Also note that `test.c:123`, the call site of `duk_eval()`, is not in the call stack at all so it never directly enters into the file/line decision.

svaarala:
Eval code will now get .fileName "eval" and other function code compiled from a user string gets .fileName "input" so that file/line combinations will now make more sense, see #644.

PR #644, opened 2016-03-18
Change automatic filename for eval/compile from __FILE__ to "eval" / "input"

svaarala:
In Duktape 1.4.x API calls like `duk_eval_string()` and `duk_compile_string()` will use `__FILE__` for the automatic .fileName property associated with the function compiled. As a result file/line pairs reported for errors inside the eval/function code can be misleading: the filename refers to a C file but the line refers to a line inside the source compiled: #516.

Change this behavior so that eval API call without an explicit filename use `"eval"` for the automatic .fileName so that errors get associated with e.g. `eval:17` (17th line in eval source code). For functions use `"input"`.

Tasks:

- [x] API change
- [x] Testcases for new behavior
- [x] Assertion run
- [x] Website and API doc updates
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1391,6 +1391,11 @@ Planned
 * Make .stack format a bit more user friendly; if you are parsing the stack
   trace format this may need changes in your parser (GH-588, GH-592)
 
+* Change automatic filename of compiled functions and eval code from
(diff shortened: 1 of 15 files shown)


svaarala:
The new API behavior is documented by this testcase:

https://github.com/svaarala/duktape/pull/644/commits/df19af42750d2d0ec2a63ed044e1697b9276124a

PR #646, opened 2016-03-18
Make things work on systems defaulting to python3

jaseg:
There is a significant fraction of systems running with python symlinked to python3. To aid portability to these systems, use explicit python2 instead of generic python. I hope I did not forget or break anything. At least the build now runs on my system (archlinux).

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -28,6 +28,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Bruce Pascoe <fatcerberus1@gmail.com>
 * René Hollander <rene@rene8888.at>
 * Julien Hamaide (https://github.com/crazyjul)
+* Sebastian Götte (https://github.com/jaseg)
 
 Other contributions
 ===================
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ NODE:=$(shell which nodejs node | head -1)
 WGET:=$(shell which wget)
 JAVA:=$(shell which java)
 VALGRIND:=$(shell which valgrind)
-PYTHON:=$(shell which python)
+PYTHON:=$(shell which python2 python|head -1)
 
 # Scrape version from the public header; convert from e.g. 10203 -> '1.2.3'
 DUK_VERSION:=$(shell cat src/duk_api_public.h.in | grep define | grep DUK_VERSION | tr -s ' ' ' ' | cut -d ' ' -f 3 | tr -d 'L')
@@ -551,16 +551,16 @@ endif
 
 .PHONY: matrix10
 matrix10: dist
-	cd dist; python ../util/matrix_compile.py --count=10
+	cd dist; $(PYTHON) ../util/matrix_compile.py --count=10
 .PHONY: matrix100
 matrix100: dist
-	cd dist; python ../util/matrix_compile.py --count=100
+	cd dist; $(PYTHON) ../util/matrix_compile.py --count=100
 .PHONY: matrix1000
 matrix1000: dist
-	cd dist; python ../util/matrix_compile.py --count=1000
+	cd dist; $(PYTHON) ../util/matrix_compile.py --count=1000
 .PHONY: matrix10000
 matrix10000: dist
-	cd dist; python ../util/matrix_compile.py --count=10000
+	cd dist; $(PYTHON) ../util/matrix_compile.py --count=10000
 
 regfuzz-0.1.tar.gz:
 	# https://code.google.com/p/regfuzz/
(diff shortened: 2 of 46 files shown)


svaarala:
For portability the most important thing that needs to work is `util/make_dist.py`: does that work for you (before the change) if you execute it as follows (other parts of the Makefile are not really cross platform)?

    $ python2 util/make_dist.py

Is "python2" universally available on systems which only have Python 2 installed?

svaarala:
I guess `python2 make_dist.py` would fail because it executes further Python scripts and assumes "python" is the interpreter to use.

jaseg:
```python2 util/make_dist.py``` works with my patch applied, but not without due to hardcoded ```python``` invocations somewhere.

The python2 symlink is standardized, see https://www.python.org/dev/peps/pep-0394/ . It is supported at least on debian, ubuntu, fedora, arch and cygwin.

svaarala:
Ok, sounds like assuming "python2" is a good approach (especially since Cygwin provides it too). Any systems missing it will be relatively old I guess.

svaarala:
My OS X install doesn't seem to have "python2", it does have "python2.7".

fatcerberus:
Would it be possible to default to "python2" and then if that's not found use "python"?

svaarala:
Sure, there are just quite a lot of places where the 'python' (or 'python2' in this branch) appear literally, and all would need to be changed like that.

svaarala:
The other option I've considered is making all the scripts python2 and python3 compatible. I tried that quickly a while ago and it turned out to be much messier than I had hoped.

jaseg:
@fatcerberus That kind of logic would not be possible as this is used in all the shebang lines (and these can only contain one hard-coded name). To which distribution of python on Windows are you referring? As mentioned, cygwin provides the ```python2``` symlink.
@svaarala Are you sure that the absence of ```python2``` is not due to some packaging snafu? Otherwise this would be pretty weird. Unrelated to this PR you might want to give a heads-up to your upstream about that.

As far as porting goes, **YES**, that would be *the* better solution. ```2to3``` could probably do many things, also there is a bunch of support libs for just that use case. Just changing the shebangs was just my lazy way to (for now) achieve the same result.

jaseg:
dtgo

svaarala:
My OS X install is just out of the box, with Brew installed.

fatcerberus:
@jaseg I was referring to the native Windows build, not Cygwin.  Given that this build is maintained by Python themselves, that surprised me, since they're the ones championing the `python2` command.

Either way the point is moot since Windows can't run the shell scripts; it was just a curiosity I wanted to mention that Windows Python doesn't have `python2`.

svaarala:
If "python2" is not available, that'll be a problem even when running the Python utilities, as they launch python using a fixed command name now ("python" in master, and "python2" in this pull).

fatcerberus:
You mean Python doesn't have a built-in facility to fork a new interpreter?  That's a bit surprising.

svaarala:
@fatcerberus It's probably clearest if you look at the diff in this pull. Things can be changed but the diff shows what the current issues are.

fatcerberus:
The main issue (besides the shebangs) looks to be make_dist.py, which calls Python directly at several points (the makefile isn't as big a problem since it can just use a variable--and does in this pull).  That's why I wondered if there's a way to fork a new interpreter instance directly rather than invoking the python command.

The python 2/3 split really is a hassle...

jaseg:
@fatcerberus In ```make_dist.py```, you could probably use Python's [sys.executable](https://docs.python.org/2/library/sys.html#sys.executable) to get the running python interpreter's path.

jaseg:
Ok, I just updated the PR with that.

fatcerberus:
`make_dist.py` works in native Windows using `sys.executable`.  So now the only potential issue is the shebang lines on Linux systems without a "python2" command.

svaarala:
The most critical thing to work is `make_dist.py` - other stuff is not even guaranteed to be portable to OS X or Cygwin. With more automatic testing in place I'll most likely prune the top level Makefile because most of the targets there are to make development-time testing easier.

fatcerberus:
Yeah, I was referring to the shebangs in the Python scripts.  On Windows these are a moot point, but on Linux they can still be an issue if the platform doesn't provide a `python2` command.  Might be worth looking into that before merging this pull if there are any interesting Linux distros where that command isn't available.

svaarala:
@jaseg Could you rebase this pull so it can be merged without conflict? Also please add yourself to AUTHORS (see CONTRIBUTING.md).

Just to recap, the situation after this merge would be:

- Executing `python util/make_dist.py` will work as long as "python" is Python2.
- Executing `python2 util/make_dist.py` on systems where "python" is Python3.
- `util/make_dist.py` will use `/usr/bin/env python2` which won't work on some systems, but you can run `make_dist.py` with explicit Python interpreter command. This is not easily fixable because of shebang limitations.
- Saying `make dist` will assume "python2" exists; if it doesn't, you can run `make_dist.py` manually.

The Makefile fix could maybe be improved to say:

    which python2 python

That way if the system only has "python" which is Python2, it'd work without manual disting.

@markand is working on a CMake top level Makefile which would be applied on top of this pull. The core dist tools would remain the same so CMake wouldn't be needed to run the raw dist step.

jaseg:
@svaarala I have updated and rebased the commit, also including the "which python2 python" hack.

Your observations should be correct.

svaarala:
@jaseg Thanks, merged :)

PR #648, opened 2016-03-19
Minor footprint optimization for module ID normalization in require()

svaarala:
Minor footprint optimization:

- Simpler setup for initial module resolution input
- Use the same buffer for resolution input and output: possible because output is always at most as long as input
- Change `module.fileName` and `module.name` handling so that the properties are not present at all by default and the default values are used if modSearch() doesn't explicitly set the properties; this is better for RAM usage and has slightly shorter code

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1396,12 +1396,11 @@ Planned
   explicit filename is known; this makes file/line information thrown from
   such code more useful in practice (GH-516, GH-644)
 
-* Add support for non-standard module.fileName (initialized to resolved
-  module ID) and module.name (initialized to last component of resolved
-  module ID) which are used for the internal module wrapper function's
-  .fileName and .name properties; they show up in stack traces, debugger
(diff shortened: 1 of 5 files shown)


fatcerberus:
> output is always at most as long as input

Are you sure?  If you have a module `foo/bar` which requires `./baz`, then `./baz` will resolve to `foo/baz`, which is longer.

svaarala:
That comment (output is always at most as long as input) refers to the input which is used for the scanning process - it's not the same as the requested ID. In your example the resolution input would be:

    foo/bar/.././baz

svaarala:
The resolution input is prepared here: https://github.com/svaarala/duktape/blob/master/src/duk_bi_global.c#L881-L885

fatcerberus:
Oh I see, the user-provided ID is appended to the current module ID.  I figured that'd be the last step, not the first. :)  Now your comment makes more sense.

PR #650, opened 2016-03-21
Document DUK_(V)SNPRINTF() portability requirements

svaarala:
The explicit portability requirements for C standard library wrappers like `DUK_(V)SNPRINTF()` are not documented explicitly. This is important especially for library functions which have concrete variance across platforms (e.g. snprintf() NUL termination and return value variance, malloc() alignment guarantees, free() handling of NULL argument, etc).

Add self tests for the important guarantees where possible.

Duktape internals also don't make the same assumptions everywhere. For example, some code assumes that if the return value from `DUK_SNPRINTF()` is smaller than the buffer size given, the buffer is correctly NUL terminated (which is in practice guaranteed by all platforms). However, other code has explicit NUL termination even for these cases which is unnecessary.

### DIFF
--- a/doc/code-issues.rst
+++ b/doc/code-issues.rst
@@ -1274,6 +1274,8 @@ Examples of snprintf() calls which don't NUL terminate on truncation:
 
 * Windows ``_snprintf()``: http://msdn.microsoft.com/en-us/library/2ts7cx93.aspx
 
+FIXME: return value differences on truncation.
+
 s(n)printf %s and NULL value
 ::::::::::::::::::::::::::::
 


fatcerberus:
Regarding free(): It's my understanding that `free(NULL);` is guaranteed to be a no-op by ISO C.  It surprises me that there are platforms for which this doesn't hold, and developing for such a platform would make me nervous: If it got something this basic wrong, what other surprises are there?

svaarala:
It's not necessarily always a question of being "wrong": not all platforms are C99 or even C89 for that matter. For example, some low memory targets don't have a standard library at all, but only something vaguely similar.

This is the case even on major platforms sometimes: older MSVC versions which are still in use lack `snprintf()` and provide a similar `_snprintf()` which doesn't provide C99 compatible guarantees. Right now Duktape can work with the MSVC `_snprintf()` because it doesn't assume C99 NUL termination guarantees for truncated strings, but this is not apparent in documentation.

In any case, I'll try to document the assumptions *Duktape* makes, so that if you're porting or changing the macro wrappers, there's a consistent description of what's expected and what's not.

PR #652, opened 2016-03-22
Note .fileName and how to control it in duk_dump_function() API document

svaarala:
Slight improvement based on an IRC discussion re: .fileName being present in dumped function bytecode.

### DIFF
--- a/doc/bytecode.rst
+++ b/doc/bytecode.rst
@@ -122,6 +122,26 @@ However, when doing so, you should note the following:
   debugging.  In other words, **obfuscation is not a design goal for the
   bytecode format**.
 
+That said, concrete issues to consider when using bytecode for obfuscation:
+
+* Variable names in the ``_Varmap`` property: this cannot be easily avoided
+  in general but a minifier may be able to rename variables.
+
+* Function name in the ``name`` property: this can be deleted or changed
+  before dumping a function, but note that some functions (such as
+  self-recursive functions) may depend on the property being present and
+  correct.
+
+* Function filename in the ``fileName`` property: this can also be deleted
+  or changed before dumping a function.  You can avoid introducing a filename
+  at all by using ``duk_compile()`` (rather than e.g. ``duk_eval_string()``)
+  to compile the function.
+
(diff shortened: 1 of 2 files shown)


demothorg:
Thanks. That's exactly what I meant.

If I understand correctly, the bytecode by default contains information about line numbers. Therefore, information about pc2line from [here](https://github.com/svaarala/duktape/blob/master/doc/bytecode.rst#config-option-specific) may be useful too. But this will require modifying the duk_config.h...

svaarala:
@demothorg Dropping pc2line information will also make stack traces much less useful, so it's always a trade-off.

By the way, is your motivation to obfuscate the bytecode or to minimize its size? For size minimization compressed source code is almost always better.

demothorg:
@svaarala Of course, I understand that without pc2line, the stack traces will be less informative.

Yes, my motivation is obfuscate the bytecode.

svaarala:
Ok - please note that obfuscation is not a design goal for the dump/load format as such. If that's a real concern you should probably also use a minifier to handle variable renaming etc.

demothorg:
@svaarala Сode minification I implement myself, of course.
I just think, that 'strip' argument of [lua_dump](http://www.lua.org/manual/5.3/manual.html#lua_dump) from Lua 5.3 is good idea. :)

This is nice but not major feature. Just a note for the future.

svaarala:
@demothorg Yes, but note that Duktape doesn't intentionally store anything in the function object unless it's needed (or potentially needed) at run time. For example, pc2line is needed for tracebacks and it depends on the application whether it can be stripped without affecting functionality. Same goes for `.fileName` and `.name`. So there's no "pure" debug information in the function objects as such and each stripping operation has some functional impact. As a result a single "strip" flag might not work well.

That's the big picture anyway, with one exception: when debugger support is enabled the variable map is kept even if the function itself doesn't directly need it. This is so that a debug client can enumerate variables. This particular case is an exception in that the variable map is retained only for debugging, assuming Duktape compiler knows for certain it wouldn't otherwise be needed.

PR #657, opened 2016-03-25
Add more testclient jobs, minor fixes

svaarala:
More testclient jobs to improve coverage, minor bug fixes.

### DIFF
--- a/testrunner/client-simple-node/client.js
+++ b/testrunner/client-simple-node/client.js
@@ -194,7 +194,7 @@ function processSimpleScriptJob(rep) {
 
     var tmpDir = tmp.dirSync({
         mode: 0750,
-        prefix: 'testrunner-' + context + '-',
+        prefix: 'testrunner-' + (rep.context || 'undefined') + '-',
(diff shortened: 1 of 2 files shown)


fatcerberus:
How do you plug into GitHub with custom tests like this?  I can't figure out how you did it just by looking at the diffs.

svaarala:
I've configured a Github webhook to inform the `testrunner/server-node` of pushes and commits; it records the commits to be tested in a database (file based, NeDB). Testclients then pull jobs from the server and report back. Finally, the server uses the Github status API to update the Github status based on test results.

It's fairly simple and limited at the moment, being restricted to "script jobs". I actually had a working prototype where the server maintained a large test matrix containing platforms, compilers, config options, testcases, etc, and tests would be assigned and executed in a very fine grained manner (which parallelizes much better and provides automatic test coverage more effectively). But that tooling is more complicated and even if that was in place I'd want script job support so that doing third party tests etc would be more feasible. So this is why I'm trying to get the script jobs working reasonably well first before expanding the coverage more.

PR #660, opened 2016-03-25
Rename module.fileName to module.filename to match Node.js behavior

svaarala:
- https://github.com/svaarala/duktape/pull/639#issuecomment-201111627
- https://nodejs.org/api/modules.html#modules_module_filename

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1401,7 +1401,7 @@ Planned
   explicit filename is known; this makes file/line information thrown from
   such code more useful in practice (GH-516, GH-644)
 
-* Add support for non-standard module.fileName and module.name used for the
+* Add support for non-standard module.filename and module.name used for the
   internal module wrapper function's .fileName and .name properties
   respectively (if unset, defaults are resolved module ID and last component
   of the resolved module ID, respectively); these properties affect e.g.
--- a/doc/modules.rst
+++ b/doc/modules.rst
@@ -13,7 +13,7 @@ built into Duktape:
   - http://wiki.commonjs.org/wiki/Modules/1.1.1
 
   - Duktape supports also ``module.exports`` and a few Duktape specific
-    properties (``module.fileName`` and ``module.name``)
+    properties (``module.filename`` and ``module.name``)
 
 * The user must provide a *module search function* which locates a module
   corresponding to a resolved module ID, and can register module symbols
@@ -234,10 +234,12 @@ Logger names and tracebacks
 Logger name defaulting uses the calling function's ``fileName`` property.
 The ``fileName`` of the internal module wrapper function is set to the
 resolved module identifier to make the logger default name come out right.
+Application code can control logger name by overwriting ``module.filename``
+in modSearch().
 
 Tracebacks show both ``name`` and ``fileName`` of the internal wrapper
-function.  The ``name`` property is currently not set, so the wrapper
-function appears anonymous.  It could also be set to the module name.
+function.  Application code can control these by overwriting
+``module.filename`` and/or ``module.name``.
 
 module.exports
 ==============
@@ -253,19 +255,25 @@ Duktape supports ``module.exports`` since Duktape 1.3, see:
 
 * ``test-commonjs-module-exports-repl.js``
 
-module.fileName and module.name
+module.filename and module.name
 ===============================
 
-The ``module.fileName`` and ``module.name`` properties are Duktape specific
-and allow modSearch() to control the ``.fileName`` and ``.name`` properties
-of the module wrapper function used to implement module loading.  This is
-useful because they appear in e.g. tracebacks for errors created from the
-module, see: https://github.com/svaarala/duktape/pull/639.
+The ``module.filename`` and ``module.name`` properties are Duktape specific
+but also supported by e.g. Node.js.  They allow modSearch() to control the
+``.fileName`` and ``.name`` properties of the module wrapper function used
+to implement module loading.  This is useful because they appear in e.g.
+tracebacks for errors created from the module.  See discussion in:
+https://github.com/svaarala/duktape/pull/639.
 
 The properties are missing by default.  If modSearch() doesn't set them,
-module.fileName defaults to the full resolved module ID (e.g. ``foo/bar``)
-and ``name`` defaults to the last component of the resolved module ID
-(e.g. ``bar``).
+the module wrapper function .fileName defaults to the full resolved module ID
+(e.g. ``foo/bar``) and .name defaults to the last component of the resolved
+module ID (e.g. ``bar``).
+
+The capitalization of ``module.filename`` is chosen to match Node.js (see
+https://nodejs.org/api/modules.html#modules_module_filename); the form
+``.fileName`` is used elsewhere in Duktape.  See discussion in:
+https://github.com/svaarala/duktape/pull/660.
 
 C modules and DLLs
 ==================
@@ -348,6 +356,8 @@ Module loading APIs or "formats":
 * NodeJS, more or less CommonJS:
 
   - http://nodejs.org/docs/v0.11.13/api/modules.html
+
+  - https://nodejs.org/api/modules.html
 
 * ES6:
 
--- a/src/duk_bi_global.c
+++ b/src/duk_bi_global.c
@@ -1114,8 +1114,8 @@ DUK_INTERNAL duk_ret_t duk_bi_global_object_require(duk_context *ctx) {
 	 * - module.exports: initial exports table (may be replaced by user)
 	 * - module.id is non-writable and non-configurable, as the CommonJS
 	 *   spec suggests this if possible
-	 * - module.fileName: not set, defaults to resolved ID if not explicitly
-	 *   set by modSearch()
+	 * - module.filename: not set, defaults to resolved ID if not explicitly
+	 *   set by modSearch() (note capitalization, not .fileName, matches Node.js)
 	 * - module.name: not set, defaults to last component of resolved ID if
 	 *   not explicitly set by modSearch()
 	 */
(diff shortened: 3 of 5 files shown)


fatcerberus:
I don't know what the current behavior is, but I'd probably also expect `module.filename` to be undefined if modSearch doesn't explicitly set it (since in that case Duktape doesn't know what the filename is, or if it even originated from a file at all).  Tracebacks of course should default to the module ID in that case (as they do now).

fatcerberus:
By the way, unrelated but does Duktape currently support `module.loaded`?  That seems like a useful thing to have that would be trivial to implement.

svaarala:
Short answer: I've been trying to avoid feature creep ;)

Long answer:

For now I've tried to add a minimum of custom fields beyond CommonJS based on concrete needs arising so that footprint and feature creep is avoided for very low memory targets. Unlike several other components it's possible (and quite straightforward even) to replace the entire module loading mechanism with a custom one - so every feature should defend its place, being weighed against the option of using a replacement loader.

The current plan has been for the built-in module loading mechanism to be as barebones and abstracted as possible, but that a separate "deluxe" module loader would also be included in the distributable. The deluxe loader wouldn't port to all environments but would include, for example, a native module DLL loader for all major platforms. It could then also provide a much larger set of module properties to match the behavior of e.g. the Node.js loader.

In fact, I'd actually prefer removing the built-in module loader entirely and only ship a few optional loaders in the distributable because not all environments need module loading. But because ES6 has a module mechanism, Duktape does need some module mechanics to be built in.

The first step in this process would be to figure out what features can be *removed* from the current loader (keeping ES6 in mind), and then implement the first versions of those optional loader(s), at least the deluxe loader with DLL support which is often requested.

svaarala:
Adding to the above, I think `.fileName` (or `.filename` if renamed) and `.name` defend their place for being included by default, because they concretely affect at least tracebacks and debugger integration. Especially tracebacks are quite important for developers so it seems useful to include them despite the (relatively small) footprint cost.

fatcerberus:
Yeah, I know you're trying to keep this as minimal as possible. :)  I mentioned `module.loaded` specifically because it seemed like it would be practically free to implement, both in implementation cost as well as footprint.  You're just setting a property to false when loading the module and flipping it to true afterward, so it shouldn't amount to more than a few bytes of additional code (in the single digits, most likely).

That said, while it _would_ be possible to replace the built-in require() with a polyfill which set `module.loaded` itself, with the current architecture (single entry point, modSearch), the entire loader would have to be replaced wholesale in user code, which isn't really ideal (you lose automatic module ID resolution, among other things) and results in a LOT more code than if Duktape just sets the flag itself.

This I see is another strength of the planned changes for Duktape 2.0, with split modResolve/modLoad.  That will make it easier to plug in pre-loading logic such as this without having to reinvent the wheel.

svaarala:
> I mentioned module.loaded specifically because it seemed like it would be practically free to implement, both in implementation cost as well as footprint. You're just setting a property to false when loading the module and flipping it to true afterward, so it shouldn't amount to more than a few bytes of additional code (in the single digits, most likely).

The cost is very small but probably around 60-80 bytes on a 32-bit platform for one module: two calls, a new built-in string, and RAM overhead of storing the loaded flag for each module which is 13 bytes * module count. Almost nothing is single digits (9 bytes), that's barely enough to make one function call. 60-80 bytes taken alone is around 5-7 fewer user strings (for example) for a low memory target: very small, but of course the creep adds up over time.

> That said, while it would be possible to replace the built-in require() with a polyfill which set module.loaded itself, with the current architecture (single entry point, modSearch), the entire loader would have to be replaced wholesale in user code, which isn't really ideal (you lose automatic module ID resolution, among other things) and results in a LOT more code than if Duktape just sets the flag itself.

Yeah, but the idea would be to include a deluxe replacement directly in the Duktape distributable. The user wouldn't implement it from scratch naturally. I don't mind *that* implementation providing a lot of features because you can easily replace it with a barebones one on a low memory target.

> This I see is another strength of the planned changes for Duktape 2.0, with split modResolve/modLoad. That will make it easier to plug in pre-loading logic such as this without having to reinvent the wheel.

Depending on how that goes, it may be the case that Duktape only provides a `require()` function which calls into user code to do both resolution and loading. As a result, user code will need to implement the ID resolution (which you mentioned above as a drawback :).

Duktape could also ship with a default resolver, but because not all environments need module loading or a proper resolver even when using modules, so my gut feeling is that while one or two full featured loaders should be included with Duktape, the hardcoded built-in feature should just be the loader framework without an implementation.

fatcerberus:
That's true.  Being primarily a C programmer, I was estimating the footprint cost in terms of:
```c
module->loaded = false;
/* load module */
module->loaded = true;
```

Which of course is pretty much free.  But I forgot that the Ecmascript environment necessarily adds a lot of overhead to that.  60-80 bytes is of course rather significant when you're fighting for kilobytes.

fatcerberus:
By the way, the reason user code needing to do module ID resolution is a pain is because string parsing isn't really one of C's strong suits.  If `require()` is implemented in JS that's less of an issue since you can use regular expressions etc. to make things easier, but that also requires file bindings so it's usually more convenient to implement the module loading directly in C.

svaarala:
Commenting a bit more on what could be included by default, assuming a split resolve/load model:

- Duktape provides nothing. User code must provide a `require()` to get CommonJS and ES6 modules working. Distributable includes a default portable `require()` and a deluxe `require()` with DLL support.
- Duktape provides a skeleton `require()` which relies on user resolve/load functions but neither is provided automatically. Distributable includes a resolver and a few loaders.
- Duktape provides a skeleton `require()` and a default resolver but no loader. Distributable includes a few loaders and user can override the resolver for non-CommonJS behavior (like understanding file paths).
- Duktape provides a `require()`, a default resolver, and a default loader which works on mainstream platforms (perhaps including DLL support); config options are needed for non-mainstream platforms to disable the loader and use a custom one. Upside is that no additional files need to be included in a project.

Assuming there's nothing magic about `require()` right now I'd be leaning towards the first option: it's best for portability and footprint, and because you need to implement a modLoad() and possibly a modResolve() anyway (for non-CommonJS semantics which are nevertheless quite common), why not simply provide the resolver and loader(s) in the distributable?

Another upside of that would be that the resolve/load logic (i.e. what split is used) wouldn't concern Duktape at all: it would only assume the user provides a `require()` with much simpler guarantees and what goes on after that is up to the user. That's what DukLuv actually already does: it implements the resolve/load on top of Duktape 1.x by replacing `require()`. If Duktape came with a default loader, it'd be pretty trivial to use, and you could choose from a portable loader and a DLL-supporting deluxe loader.

Yet another reason which makes me lean this way (at least now) is that users would like Duktape to ship with a loader that supports DLLs for Linux, OS X, and Windows, so that it'll be possible to easily write native loadable modules. That loader can't realistically be inside the Duktape core because it'd be a portability issue - so it will be external but possibly shipped with Duktape. But if that's the default loader for most people, and it's already external, why include anything in the Duktape core?

svaarala:
I have similar feelings about the logging framework by the way: it was probably a mistake to include it by default. It'd be better to define the API, and ship the default implementation as a separate source file because many enviroments don't use the logging framework or may replace big parts of it. So while it's good to have a shared API you, it'd probably be best if the implementation was always external (with a default implementation included in the distributable).

While this is not ideal for embedding Duktape, instead of a single source file you have maybe 2-3, the advantage is that these optional modules don't have the same footprint/feature pain as all features in the core parts.

fatcerberus:
I thought ES6 didn't have `require()` at all but instead used `import`/`export` keywords?  i.e. it's built into the language and therefore _is_ magic.

svaarala:
Yeah, that's I meant above is that the key question is how `require()` and ES6 modules work together - and what primitive is the smallest possible interface to make both work. Basically having a `require+` which handled both while making minimal assumptions about anything else (such as how ID resolution and loading is structured) would be ideal.

svaarala:
So while Duktape needs to deal with CommonJS `require()` calls and ES6 import/export, it could call into user code as early as possible. Offloading complicated resolution logic there shouldn't be a big issue considering (1) a default implementation would be included, and (2) you need to be pretty involved anyway to get something like DLL loading to work as is. So why not push the resolution/loading algorithm also into user code?

In concrete terms, including a default loader with DLL support and an ordinary filesystem search path would probably make it easier to get into using modules compared to the current situation of explaining modSearch() and coming up with one :) So this is quite likely more important than what the division of work is between require, modResolve, modLoad, etc.

fatcerberus:
Yet another thing to add to the Duktape 2.x to-do list I guess. :)

svaarala:
I'm actually pretty much ready to start work on Duktape 2.x once 1.5 is done :) 1.5.x maintenance will then go on for a long while, and I think it's quite OK as a baseline until 2.0.

fatcerberus:
I agree, AppRequest/AppNotify will go a long way in making the debugger extensible for me so I don't need to ask for more debugger features in 1.x. :)

svaarala:
The Node.js semantics for `.filename` are:

> The fully resolved filename to the module.

This is not exactly the same as Duktape `.fileName` whose semantics are:

> Filename which should be used for the internal module wrapper function and appears in tracebacks, etc.

In particular, the `.fileName` doesn't have to be an absolute path, and may be "censored" (for example: `[app]/foo/index.js`).

In any case it's probably best to rename the Duktape key. If the semantic differences are an issue going forward, we can add an explicit `.moduleWrapperFilename` for the module wrapper specific semantics (and if missing, default to `.filename` which is a good default).

fatcerberus:
I think the semantics not matching is fine.  Node.js differs in other ways too, for example `require()` module resolution is file-based and so on.  Since Duktape can run even on systems without a file system, it seems okay if `module.filename` is, well, not a filename. :)

svaarala:
As an example where this difference would matter: suppose you wanted to add extra properties to the `module` table so that it'd be closer to Node.js behavior. In doing so, you'd like to add `.filename` with Node.js semantics, i.e. provide the full file path. However, you wouldn't be able to do that without affecting how the module looks like in tracebacks. Ideally these two issues wouldn't need to be tied.

PR #661, opened 2016-03-27
Use shared error macros for error codes

### LINKED ISSUE #456: Use type/context specific error throw macros to reduce footprint
For example:

```
DUK_ERROR(thr, DUK_ERR_API_ERROR, "api error xyz");
```

Can be converted to:

```
DUK_ERROR_API(thr, "api error xyz");
```

This requires a helper which makes the actual error throw call, but saves one argument from the call site. Because there are a lot of error call sites, this approach can save some footprint quite easily. The call sites will be smaller, which is also good for instruction cache performance as they're mostly cold code.

See discussion in https://github.com/svaarala/duktape/pull/441.

svaarala:
Saves a little bit on footprint because internal call sites will be smaller.

Some error types also changed internally from internal error to range error. This is more consistent with the idea that an internal error indicates an actual error (perhaps a bug) while a range error indicates an internal limit being hit (range error is also used for executor timeouts etc).

Fixes #456.

Tasks:

- [x] Implement the error macros and change call sites
- [x] Remove (now) unused strings
- [x] Rework error macro and helper argument order to avoid 'thr' shuffling
- [x] Remove variadic macros (portability) and vararg function call for non-formatted call sites (vararg call site is larger than non-vararg)
- [x] Pack error code (8-bit) and internal line number into a single argument to remove one unnecessary constant load
- [x] Add automatic regression coverage for all error variants: (1) verbose and paranoid, (2) verbose and non-paranoid, (3) non-verbose
- [x] Fix testcases affected by error code changes
- [x] Check documentation for range error vs. internal error discussion
- [x] Check code issues for log vararg discussion
- [x] Check documentation for macro vararg limitations, e.g. threading (no such issues for internal errors anymore, only for debug macros and public API `__FILE__` and `__LINE__` capture)
- [x] Releases: note on error code changes

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1432,6 +1432,12 @@ Planned
   protocol when it's safe to do so, i.e. when the plain integer converts back
   to an identical IEEE double with no loss of precision (GH-604)
 
+* Change the error type for some current internal errors to RangeError when
+  the underlying cause is an implementation limit (like compiler temp limit)
+  rather than an unexpected internal situation (GH-661)
+
+* Minor changes to error message strings (GH-661)
+
 * Fix potentially memory unsafe behavior when a refcount-triggered finalizer
   function rescues an object; the memory unsafe behavior doesn't happen
   immediately which makes the cause of the unsafe behavior difficult to
@@ -1492,6 +1498,8 @@ Planned
 
 * Internal performance improvement: single step encoding for JSON values in
   the JSON slow path (GH-447)
+
+* Internal footprint improvement: reduce error call site size (GH-661)
 
 2.0.0 (XXXX-XX-XX)
 ------------------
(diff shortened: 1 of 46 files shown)


svaarala:
Most of the changes are done, the final effects on footprint (from multiple small changes) are roughly:

For x64:
- -2.5kB for "make duk"
- -4.3kB for compiling with default options but with `-flto` and removing unused symbols at link time
- -0.8kB for compiling with low memory options but with `-flto` etc

For x86:
- -0.6kB for compiling with low memory options but with `-flto` etc

For x32:
- -0.8kB for compiling with low memory options but with `-flto` etc

The savings for default options are between 2-4kB roughly, for several reasons: specialized helper calls for specific error types, removal of vararg helpers (whose call sites are larger than fixed argument call sites), argument reordering, etc.

The savings for low memory options are much less because the error call sites for them are smaller to begin with (the only arguments used are `thr` and error code).

PR #669, opened 2016-04-02
Suppress some harmless scanbuild warnings, cmdline realloc() leak fix

svaarala:
Fix duk_cmdline.c incorrect realloc() error handling where original buffer being reallocated might leak. Other scanbuild warnings are harmless; there are a few remaining issues which are intentional and harmless (e.g. intentional segfault on panic).

### DIFF
--- a/examples/alloc-logging/duk_alloc_logging.c
+++ b/examples/alloc-logging/duk_alloc_logging.c
@@ -90,8 +90,8 @@ void *duk_realloc_logging(void *udata, void *ptr, duk_size_t size) {
 		old_size = hdr->u.sz;
 
 		if (size == 0) {
-			free((void *) hdr);
 			write_log("R %p %ld NULL 0\n", ptr, (long) old_size);
+			free((void *) hdr);
 			return NULL;
 		} else {
 			t = realloc((void *) hdr, size + sizeof(alloc_hdr));
--- a/examples/cmdline/duk_cmdline.c
+++ b/examples/cmdline/duk_cmdline.c
@@ -318,14 +318,16 @@ static int handle_fh(duk_context *ctx, FILE *f, const char *filename, const char
 		avail = bufsz - bufoff;
 		if (avail < 1024) {
 			size_t newsz;
+			char *buf_new;
 #if 0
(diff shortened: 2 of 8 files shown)


svaarala:
Remaining errors on Linux x64:

Intentional dereference of a NULL pointer when intentionally causing a segfault on panic:
```
dist/src-separate/duk_error_macros.c:151:2: warning: Dereference of null pointer
        DUK_CAUSE_SEGFAULT();  /* SCANBUILD: "Dereference of null pointer", normal */
        ^~~~~~~~~~~~~~~~~~~~
dist/src-separate/duk_config.h:2435:70: note: expanded from macro 'DUK_CAUSE_SEGFAULT'
#define DUK_CAUSE_SEGFAULT()  do { *((volatile duk_uint32_t *) NULL) = (duk_uint32_t) 0xdeadbeefUL; } while (0)
                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
dist/src-separate/duk_debugger.c:180:9: warning: Access to field 'heap' results in a dereference of a null pointer (loaded from variable 'thr')
        heap = thr->heap;
               ^~~~~~~~~
1 warning generated.
```

Potential NULL pointer in debugger code; NULL pointer is not actually possible because debugger code is only entered from bytecode executor where `thr` is always non-NULL (not sure why scanbuild flags only this one place in the debugger code, but it seems to be harmless in any case):
```
dist/src-separate/duk_debugger.c:180:9: warning: Access to field 'heap' results in a dereference of a null pointer (loaded from variable 'thr')
        heap = thr->heap;
               ^~~~~~~~~
1 warning generated.
```

Harmless read of garbage data in Date handling. This is caused by a shared helper which is intentionally given a dummy uninitialized array by some call sites when a certain output is not needed. The access is memory safe (i.e. within the allocated stack array) but the data is uninitialized and any results from dealing with uninitialized data are ignored:
```
dist/src-separate/duk_bi_date.c:781:5: warning: Assigned value is garbage or undefined
                d = dparts[i];
                  ^ ~~~~~~~~~
1 warning generated.
```

PR #671, opened 2016-04-03
Shared decoding for opcode 'a' field

svaarala:
Decode the 'a' opcode field right after instruction fetching. This has the benefit of removing duplicate decoding in opcode case clauses, but also has potential downsides for spilling and performance.

In concrete testing this seems to have footprint benefits but a very slight negative performance impact. As such this might be useful for reducing footprint if it could be worked in an easy to maintain option.


svaarala:
No longer relevant, only two call sites which decode 'A' into a variable now.

PR #672, opened 2016-04-03
Check bytecode flags directly without decoding opcode fields

svaarala:
Check opcode flags directly from the instruction without decoding the field first. This has lower footprint, performance impact to be verified.

Also change some `DUK__REGCONSTP(b)` to `DUK__REGCONSTP_B(ins)`, etc.

Tasks:

- [ ] Change flag decoding
- [ ] Releases entry


svaarala:
After the opcode rework no longer seems to be a net positive change, so closing without merge.

PR #673, opened 2016-04-03
Add Kraken benchmark support files

svaarala:
Minimal Kraken support:

- Add a "shell" which accepts Kraken `sunspider` command argument format and runs `duk` with converted arguments.
- Add `kraken_harness.js` to provide a few bindings expected by Kraken

No Makefile or runner files etc.


svaarala:
Sunspider-0.9.1 and Ubench are runnable using the "sunspider" command, kraken-1.0 and kraken-1.1 currently fail due to "register limit" internal error.

PR #674, opened 2016-04-03
Use single shift+mask in executor reg/const access

svaarala:
Instead of decoding a field (say A) and then computing DUK__REGP which involves further shifting of the value, add macros such as DUK__REGP_A(ins) which operate directly on the 32-bit 'ins' and use a single shift+mask to get the byte offset (rather than element offset) for the target reg/const. Performance and footprint impact not yet evaluated.

- [x] Config option DUK_USE_EXEC_REGCONST_OPTIMIZE for enabling/disabling the optimization; it assumes sizeof(duk_tval) is 8 or 16 which is almost always (but not always) the case
- [x] Heap alloc check for sizeof() assumption (even with assertions disabled) because sizeof() cannot be detected by preprocessor
- [x] Unit test coverage for DUK_USE_EXEC_REGCONST_OPTIMIZE, otherwise cold macro path
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1693,6 +1693,9 @@ Planned
 * Fix -Wshift-sign-overflow warnings on some Clang versions for signed left
   shifts whose result was used as unsigned (GH-812, GH-813)
 
+* Internal performance improvement: avoid one extra shift when computing
+  reg/const pointers in the bytecode executor (GH-674)
+
 * Internal change: rework shared internal string handling so that shared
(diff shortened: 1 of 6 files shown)


fatcerberus:
1.5.0 must be getting close, you're into low-level performance stuff now. :)

fatcerberus:
Does this improve performance noticeably?

svaarala:
It did affect dispatch and reg/const heavy testcases a little bit.

PR #678, opened 2016-04-13
Replace readline with Linenoise in examples

svaarala:
Replace readline with https://github.com/antirez/linenoise in the 'duk' command line example (readline was an optional dependency, same for Linenoise). Linenoise is compiled in so it has less of portability trivia and also seems to be valgrind pure.

This change has no effect unless you're explicitly enabling DUK_CMDLINE_FANCY which used to depend on readline and now depends on linenoise.

Because linenoise is not C++ compatible, DUK_CMDLINE_FANCY no longer works when compiled with a C++ compiler.

Tasks:
- [x] C compile warnings, see if they can be eliminated (pull upstream for obvious fixes)
- [x] C++ compile errors => don't need to support C++ at this point, which means this won't work with Windows and MSVC yet
- [x] Switch to using linenoise fork with warnings suppressed


fatcerberus:
Another nice thing is that linenoise is BSD-licensed.  Readline is GPL which gets in the way sometimes.

svaarala:
There are several alternatives to Linenoise, so now would be a good time to figure out what works best for Duktape:

- https://github.com/yhirose/cpp-linenoise
- https://github.com/arangodb/linenoise-ng
- https://github.com/troglobit/editline

Ideally something that works on Linux, Windows, OS X and using C/C++ compilers out of the box.

fatcerberus:
Linenoise looks nice, but it doesn't compile under MSVC as it includes POSIX headers.  This disappointed me because I was going to use it in SSJ. :(

linenoise-ng looks like it supports Windows fully, but needs C++ which is a bit of a turn-off (for me anyway).

fatcerberus:
Actually it looks like most line editing libraries are UNIX-only, which is kind of a bummer.  I guess people figure Windows is suitably covered since its terminal does history, completion, etc. for you.

svaarala:
So what'd be the best choice for Windows (ignoring Linux for a while)?

fatcerberus:
If you don't mind the C++ dependency cpp-linenoise looks like a good choice.  It's all in a single .hpp file which makes it very easy to integrate.

However like I said above Windows' default terminal emulator already does history and I think some basic completion (I might be wrong about this part though...) even if you just use `getchar()`, so fancy line editing is not as important there.

svaarala:
C++ dependency is probably a bit too much (Duktape is pretty C oriented otherwise). But maybe we could just use Linenoise now and disable it automatically if compiling with a C++ compiler or outside a POSIX environment. It's only a convenience after all.

svaarala:
Fixed compile warnings by using a Linenoise fork in the top level Makefile. Similar fixes are already pending for linenoise master so this can be removed once linenoise master is clean.

PR #679, opened 2016-04-13
Add minimal completion to Linenoise integration in 'duk'

svaarala:
Having a minimal completion callback to e.g. provide global names would be useful for playing around with scripts and bindings. A more useful minimal functionality would be:

- Scan backwards from end of line, identifying property path compatible text (`foo.bar.quux`). Terminate on start-of-line and unknown characters (also whitespace).
- Having done that, process the property path component by component, looking up objects starting from the global object. For the last component lookup all properties and filter by already entered prefix.

This should provide useful completion for globals and anything reachable via global variables.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1511,6 +1511,9 @@ Planned
   is in the include path and that linenoise.c is compiled with the command
   line tool (or linked from a library) (GH-678)
 
+* Add minimal tab completion support for duk_cmdline.c ('duk') using Linenoise
+  completions (GH-679)
+
 * Internal performance improvement: use raw value stack accessors internally
   when it's safe to do so (GH-582)
 
--- a/examples/cmdline/duk_cmdline.c
+++ b/examples/cmdline/duk_cmdline.c
@@ -9,6 +9,7 @@
 /* Helper define to enable a feature set; can also use separate defines. */
 #if defined(DUK_CMDLINE_FANCY)
 #define DUK_CMDLINE_LINENOISE
+#define DUK_CMDLINE_LINENOISE_COMPLETION
 #define DUK_CMDLINE_RLIMIT
 #define DUK_CMDLINE_SIGNAL
 #endif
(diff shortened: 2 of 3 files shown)


svaarala:
There's a minimal implementation now in the pull. Consider the input:

     x = foo.bar.Q<TAB>

The minimal implementation will first extract the property path from the input, here `foo.bar.Q`, which is identified simply as a period delimited sequence of identifier-like parts (it's limited to ASCII now).

Having done that, `O` is initialized to the global object and then, for each component:

- For each full component ("foo" and "bar" in the example), identified as ending in a period, set `O = O[comp]`, i.e. follow the property chain. This will now invoke getters which is not ideal.
- For the final component, if it matches a property of `O` exactly, add a period to the completion. For example, if `foo.bar.Q` was a full property name, complete `foo.bar.Q` -> `foo.bar.Q.`.
- For the final component, if it matches a prefix of a `O` property name, emit the full property name as a completion.  For example, if `O` had properties `Quux`, `Quuux`, and `Baz`, emit completions for `Quux` and `Quuux`.

The completion offered to linenoise must be for the full line, e.g. `x = foo.bar.Q` might offer completions:

     x = foo.bar.Quux
     x = foo.bar.Quuux

svaarala:
Some shortcomings of the minimal implementation:

- Getters are invoked when following the property path. This might be considered a feature or a bug.
- Identifiers are limited to ASCII only.
- Only period syntax is support for properties; `foo.bar.Q` works, but `foo['bar'].Q` does not.

As the next step it might be useful to write the completion handling code in Ecmascript: it'd allow much easier manipulation of syntax, and `Object.getOwnPropertyDescriptor()` could be used to follow the property path to avoid side effects.

svaarala:
Fixed compile warnings by using a Linenoise fork in the top level Makefile. Similar fixes are already pending for linenoise master so this can be removed once linenoise master is clean.

PR #692, opened 2016-04-20
Minimal operator overloading test

svaarala:
Operator overloading has been requested multiple times. The basic downside of adding it to Duktape is that it's not available in (all) other engines which causes a lock-in which would be nice to avoid.

On the other hand, Duktape is used in some environments where the ability to manipulate domain specific objects using operator overloading is convenient (for example, vector and matrix calculation).

This pull contains an example of how operator overloading could be added to Duktape with a minimal impact. The example approach is to add an operator overload check to arithmetic bytecode opcodes (ADD, etc) which is done after fast path checks (e.g. number+number addition for ADD) and only if operator overloading config option is enabled (disabled by default).

To avoid getting drawn into the specifics of how operator overloading is handled (for which there are multiple options) the example simply checks if `Duktape.add()` exists and if so, offloads the operation to that function.

This pull is mostly for discussion, it's not merge ready but provides some concreteness on a possible approach.


svaarala:
At this point the branch just demonstrates how to hook overloading into the executor:

```
((o) Duktape [linenoise] 1.4.99 (v1.4.0-397-g4435a65-dirty)
duk> "foo" + 1
= foo1
duk> Duktape.add = function add(x,y) { print('ADD', x, y); return Number(x) + Number(y); };
= function add() {"ecmascript"}
duk> "foo" + 1
ADD foo 1
= NaN
duk> "123" + 1
ADD 123 1
= 124
```

It would be nice if the overload callback could somehow indicate it's not interested in the arguments given so that Duktape would fall back to standard semantics. As things are now in the example, if an overload function is present it must implement the entire semantics.

svaarala:
One option for providing operator overloading would be source-to-source translation - if anyone has done that it'd be great to point to it.

sant123:
What about something regard with this:
http://javascriptplayground.com/blog/2013/12/es5-getters-setters/
This is an example of how to implement getter and setter in JS, if we're able to overload operators we'd do this:

```javascript
class Polygon {
    constructor(height, width) {
        this.height = height;
        this.width = width;
    }

    get area() {
        return this.calcArea();
    }

    calcArea() {
        return this.height * this.width;
    }

    operator + (polygon1, polygon2) {
        return polygon1.calcArea() + polygon2.calcArea();
    }

    operator - (polygon1, polygon2) {
        return polygon1.calcArea() - polygon2.calcArea();
    }
}

var pol1 = new Polygon(150, 42);
var pol2 = new Polygon(72, 22);

console.log(pol1 + pol2); // 7884 
console.log(pol2 + pol1); // 7884 

console.log(pol1 - pol2); // 4716 
console.log(pol2 - pol1); // -4716 
```

But we need to be aware of the type comming into the function, in this example both need to be Polygon.

svaarala:
Typically an object could provide some method (say `__add`) to indicate it handles the overloaded operation. The function handling overloading would check if that method exists on the left argument (for example; there are other alternatives too) and if so, call that with both arguments. Sometimes if the left argument doesn't have the method it can be checked from the right argument too; this makes sense if the `__add` provider handles upgrading or can take multiple types. `instanceof` can be used to check for instance typing, as usual.

In any case, I'd prefer a solution where Duktape doesn't force a specific overloading model but would rather hand over the objects to a callback which can implement whatever strategy is most appropriate.

PR #703, opened 2016-04-26
Add an internal duk_harray type for faster array handling

### LINKED ISSUE #804: Duktape.info() for duk_hbufferobject returns incorrect struct size
Duktape.info() returns the wrong struct (heap header) size for duk_hbufferobject. The if-else ladder is missing `duk_hbufferobject` check which causes the size for a generic `duk_hobject` to be returned. This has no functional impact but the information is wrong.

svaarala:
Tasks:
- [x] Add an internal `duk_harray` type (similar to `duk_hbufferobject`)
- [x] Garbage collector changes for the new type
- [x] Property handling changes for the new type: `.length` is virtual, no need for `.length` lookups anymore
- [x] Specific testcase file: writable .length, non-writable .length, Object.defineProperty() operations on .length, write to object inheriting from array (both length and indices), dense/sparse array, write to array length (integer, string, etc) both putprop and defprop
- [x] JSON encode fast path for `new Array(N)` where `.length` > array part length (valid but not handled in fast path now)
- [x] Direct array part creation for tracebacks
- [x] Direct array part creation for `new Array(...)` constructor
- [x] ROM built-in support for `duk_harray` (needed for Array.prototype which is also an Array)
- [x] If there's space, maybe embed all three property attributes? => No, future work
- [x] Document that `duk_def_prop()` `DUK_DEFPROP_FORCE` flag cannot always change properties which don't have normal backing (= are virtualized somehow); e.g. it's possible to force Array `.length` writable, but not enumerable, because there's no internal state to track Array .length enumerability
- [x] Document enumeration behavior change for abandoned array: because .length is virtual, it is enumerated right after array part but always before entry part; this means that an abandoned array enumerates e.g. as `length,0,1,2` where it previously enumerated as `0,1,2,length`
- [x] Duktape.info() support for `duk_harray`
- [x] Fix for Duktape.info() handling of `duk_hbufobj`

With this change an Array instance would typically have no need for concrete properties. So it might make sense to separate the entry and array part allocations; it would also simplify access to entries for both.

Fixes #804.

Future work notes:

- Make `length_nonwritable` field of `duk_harray` a flag in the `duk_hobject` flags field (there's currently no space)
- Debugger inspection support for `duk_harray`; especially `.length` of an array needs documentation


svaarala:
This change will be roughly footprint neutral (maybe save 100-200 bytes), will reduce RAM usage for arrays by one property (net effect is 13 - 4 = 7 bytes per array, plus saved overhead of usually not needing a property table at all), and will improve performance for some array operations. For example, array writes benefit a lot because Array `.length` needs to be updated on writes:

```
test-array-write.js                 : duk.O2.master  2.71 duk.O2.harray  1.79
```

The effect for most tests is negligible however.

The main reason for this change is to prepare for better Array operation fast paths. The fast paths apply to Array instances with their array part intact; with this change in place, those will always be `duk_harray` objects with an array part, so that both array elements and `.length` can be read/written without going through actual property lookups. Previously the `.length` property would need to be located which was a bit awkward for the fast path.

fatcerberus:
If the array part is ever discarded, the object reverts to a normal `duk_hobject` I assume?

Re: Documentation changes for `duk_def_prop()`, I think that's already covered, because the docs already say this:

> In some cases even a forced change is not possible and will cause an error to be thrown. For instance, properties implemented internally as virtual properties (such as string "length" and index properties) cannot be modified because they don't have concrete internal storage and thus no way of recording changes.

svaarala:
Objects can't change internal type once they've been allocated as their heap pointers must remain stable while they're reachable. So while the array part values will be moved to the entry part, the `.length` will remain virtual and be backed by the `duk_harray->length` field.

svaarala:
Re: duk_def_prop() documentation, seems it's mostly covered. Maybe worth mentioning Array `.length` as an example too. It's the first virtual property which can accept some attribute changes (writability) but not all (enumerability and configurability are fixed).

svaarala:
Another category of performance tests benefiting from this change is compilation, which uses temporary arrays internally:

```
test-compile-mandel-nofrac.js       : duk.O2.harray 11.87 duk.O2.master 13.67 duk.O2.150json 13.69
test-compile-mandel.js              : duk.O2.harray 15.22 duk.O2.master 16.85 duk.O2.150json 16.91
test-compile-short.js               : duk.O2.harray  7.44 duk.O2.master  9.00 duk.O2.150json 10.21
```

fatcerberus:
Compiler performance improvements are always welcome, no matter how small. :)

fatcerberus:
Speaking of array performance: When building an array, which of these is faster in Duktape, or are they about the same?

```javascript
for (...) arr[i] = whatever;

// OR

for (...) arr.push(whatever);
```

I usually use `.push()` because it makes it clearer that the array is being added to, but if the former is much faster I'd be willing to use it in more performance-sensitive scripts.

svaarala:
The index assignment is much faster (around a factor of 5-6) because pushing involves a full function call sequence.

fatcerberus:
Thanks, that's what I suspected.  But 5-6x is quite a huge difference, I'll definitely be sure to use the direct assignment in my default engine modules!

svaarala:
It's a huge difference if you do a million assignments. If you do a thousand, I don't think you'll notice unless you do it every frame :) (People often focus on relative figures when they should be looking at absolute figures and context... :)

fatcerberus:
One thing I'm using this for is to build vertex lists for ellipses and circles every frame, so a 5-6x difference is potentially quite significant especially if a lot of them are drawn:

```javascript
function ellipse(surface, x, y, rx, ry, color, color2)
{
	color2 = color2 || color;

	var numSegments = 10 * Math.sqrt((rx + ry) / 2.0);
	var vlist = [ { x: x, y: y, color: color } ];
	var pi2 = 2 * Math.PI;
	var cos = Math.cos;
	var sin = Math.sin;
	for (var i = 1; i <= numSegments; ++i) {
		var phi = pi2 * i / numSegments;
		var c = cos(phi);
		var s = sin(phi);
		vlist[i] = {
			x: x + c * rx,
			y: y - s * ry,
			color: color2,
		};
	}
	vlist[numSegments + 1] = {
		x: x + rx,  // cos(0) = 1.0
		y: y,       // sin(0) = 0.0
		color: color2,
	};

	var shape = new Shape(vlist, null, ShapeType.Fan);
	shape.draw(surface);
}
```

fatcerberus:
@svaarala To be clear, when I said a "huge difference" I just meant 5-6x difference was more of a discrepancy than I was expecting, i.e. from an entirely theoretical standpoint.  I would have guessed a factor of 2, maybe 3 due to the `.push` lookup and subsequent function call, but not 5-6x.

svaarala:
Note that one is comparing:

- Fast path array write

to:

- Slow path, inherited `.push` lookup
- Function entry
- Native C function implementing a property push (more complicated than an assignment because it's generic)
- Function exit

If the difference were 2-3x it'd be quite surprising IMO :) Well, at least considering there's no inlining or fast paths for the built-in Array.prototype functions (they must deal with all object types).

fatcerberus:
That's true, although I thought it might not be fast pathed since`.length` has to be updated at each step.  Anyway, yeah, I'll make sure to use direct indexing in performance-sensitive code (like the ellipse drawing function above), thanks for the help. :)

fatcerberus:
By the way, at the start you said you thought this change would be more or less footprint-neutral.  Is that in fact the case in the final result?

svaarala:
At an earlier stage it reduced footprint a little bit - now it's practically +/- 0. I think I can improve on that but I want this large change in master first and then I'll try to figure out some improvements.

PR #715, opened 2016-04-29
API changes to allow user control over error __FILE__/__LINE__ blaming

### LINKED ISSUE #714: Avoid __FILE__ / __LINE__ augmentation in `duk_error()`
I sometimes need to throw an error from a Duktape/C function but want it to be augmented with the caller's filename and line number.  Currently `duk_error()` uses `__FILE__` and `__LINE__, which has forced me to implement a hack using `Duktape.act()`:

https://github.com/fatcerberus/minisphere/blob/master/src/minisphere/api.c#L377-L413

Additionally this forces me to keep the `Duktape` object visible (I would prefer to hide it) since minisphere needs to access it by name internally.

If there were a variant of `duk_error_*` that didn't augment, this wouldn't be needed, since Duktape would then default to the first callstack entry with a file/line (in practice, the Ecmascript caller).

svaarala:
Fixes #714.

### DIFF
--- a/src/duk_api_public.h.in
+++ b/src/duk_api_public.h.in
@@ -412,6 +412,12 @@ DUK_EXTERNAL_DECL duk_idx_t duk_push_thread_raw(duk_context *ctx, duk_uint_t fla
 #define duk_push_thread_new_globalenv(ctx) \
 	duk_push_thread_raw((ctx), DUK_THREAD_NEW_GLOBAL_ENV /*flags*/)
 
+/* FIXME: user control for error blaming; add a flags field for error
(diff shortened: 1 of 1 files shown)


mxmauro:
Hi.

Should the `duk_error` macro use `DUK_FILE_MACRO` and `DUK_LINE_MACRO` instead of `__FILE__` and `__LINE__`?

Regards,
Mauro.

svaarala:
@mxmauro It should - they weren't originally available for the public header and thus this leftover. I'll fix it, thanks!

svaarala:
Fix is in #897.

mxmauro:
Cool :)

PR #717, opened 2016-05-01
Avoid circular reference for eval-compiled functions

svaarala:
When code is `eval()`'d an internal function is created to represent the eval code. That function had an automatic `.prototype` object which made the eval function part of a reference loop so that it wasn't immediately collected after eval finished (it would be collected in the next mark-and-sweep pass).

Improve this behavior by removing the automatic `.prototype` object which allows eval functions to be collected by reference counting, and also avoids creating the unreferenced `.prototype` object unnecessarily.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1493,6 +1493,11 @@ Planned
 
 * Minor changes to error message strings (GH-661)
 
+* Avoid reference loop for eval() code (previously caused by an automatic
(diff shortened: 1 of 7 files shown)


fatcerberus:
Nice, now we'd just need a solution for function objects in general and we'd be set.  No idea how that could be done without breaking compliance though.

PR #727, opened 2016-05-03
Add a userdata argument to duk_safe_call()

### LINKED ISSUE #277: Add duk_safe_call() variant with a void pointer userdata
Add a duk_safe_call() variant with a void pointer userdata, say `duk_safe_call_udata()`. Such an API call would allow user code to avoid passing C values through the value stack (which is done even internally in a few places). Instead, the caller could just:

```
typedef struct {
    int a;
    int b;
} my_args;

duk_ret_t my_safe_call(duk_context *ctx, void *udata) {
    my_args *args = (my_args *) udata;
    printf("a=%d, b=%d\n", args->a, args->b);
}

/* ... */
struct my_args args;
args.a = 123;
args.b = 234;
rc = duk_safe_call_udata(ctx, (void *) &args);
/* ... */
```

Packing C values to a struct is also not the most convenient thing but it scales to arbitrary numbers and types of arguments.

Unfortunately this cannot be easily implemented so that the previous `duk_safe_call()` would just be a dummy wrapper for the udata variant - the target C function prototype is different (typedef `duk_safe_call_function`).

(This has been suggested before, but I couldn't find the issue, so opening this new one.)

If implemented, also rework `duk_pnew()` and other internal call sites to take advantage of the new call.

svaarala:
- [x] Add a userdata argument to duk_safe_call()
- [x] Fix internal call sites
- [x] Fix API testcases
- [x] Fix examples
- [x] Grep all duk_safe_call() strings and check they've been updated
- [x] Update API documentation
- [x] Update website
- [x] Release notes migration instructions
- [x] Releases entry

Fixes #277.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,7 +270,7 @@ CCOPTS_SHARED =
 CCOPTS_SHARED += -D_POSIX_C_SOURCE=200809L  # to avoid linenoise strdup() warnings
 CCOPTS_SHARED += -pedantic -ansi -std=c99 -fstrict-aliasing
 # -Wextra is very picky but catches e.g. signed/unsigned comparisons
-CCOPTS_SHARED += -Wall -Wextra -Wunused-result
+CCOPTS_SHARED += -Wall -Wextra -Wunused-result -Wdeclaration-after-statement
 CCOPTS_SHARED += -Wcast-qual
 CCOPTS_SHARED += -Wunreachable-code  # on some compilers unreachable code is an error
 # -Wfloat-equal is too picky, there's no apparent way to compare floats
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1593,7 +1593,9 @@ Planned
 2.0.0 (XXXX-XX-XX)
 ------------------
 
+* Incompatible change: add a userdata argument to duk_safe_call() to make it
+  easier to pass C pointers to safe functions (GH-277, GH-727)
+
 * Incompatible change: rename duk_debugger_attach_custom() API call to
   duk_debugger_attach() to eliminate an unnecessary API call variant
   (GH-735, GH-742)
-
--- a/doc/release-notes-v2-0.rst
+++ b/doc/release-notes-v2-0.rst
@@ -21,6 +21,18 @@ migrate from 1.5.x to 2.0.0.  There are also bug fixes and other minor
 behavioral changes which may affect some applications, see ``RELEASES.rst``
 for details.
 
+Supporting Duktape 1.x and Duktape 2.x simultaneously
+-----------------------------------------------------
+
+You can use the ``DUK_VERSION`` define to support both Duktape 1.x and 2.x
+in the same application.  For example::
+
+    #if (DUK_VERSION >= 20000)
+    rc = duk_safe_call(ctx, my_safe_call, NULL, 1 /*nargs*/, 1 /*nrets*/);
+    #else
+    rc = duk_safe_call(ctx, my_safe_call, 1 /*nargs*/, 1 /*nrets*/);
+    #endif
+
 DUK_OPT_xxx feature option support removed
 ------------------------------------------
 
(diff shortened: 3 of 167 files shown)


svaarala:
The upside of this change is that many C call sites will be more natural when a pointer can be passed via the userdata argument or a set of C values by passing a pointer to a stack-allocated struct. Code footprint is also reduced slightly because internal safe call sites don't need to pass C values through the value stack.

One downside of the change is that Duktape/C function and "safe call" function signatures are no longer the same, as Duktape/C functions don't take a userdata argument. Previously one could call the same function using both `duk_safe_call()` and `duk_(p)call()`.

Adding a userdata to Duktape/C functions would be possible but:

- A userdata argument is not needed for Duktape/C functions as commonly as for safe-called functions.
- Allocating a userdata field for every `duk_hnativefunction` would be a bit wasteful. It could be stored as an internal property of the called function however.

Overall, I don't think adding a userdata for Duktape/C functions would be an improvement.

fatcerberus:
Userdata for Duktape/C functions could be a useful in some circumstances, e.g. to emulate closures.  Think of functions like, say, `require()` that need to carry metadata around with them in order for the "magic" to work.

fatcerberus:
Another potential benefit to Duktape/C function userdata is cases where multiple related Ecmascript functions might want to share a native function.  Think of a Keyboard class with accessors for each key that return true/false depending on whether the key is pressed.  Each accessor could have the same native function with the userdata specifying the key it checks.

Of course this can be done even in Duktape 1.x by tacking on a hidden property to the function and reading that when it's called using `duk_push_current_function()`.  But that's less convenient to set up because it involves value stack shuffling.

The footprint argument is a sound one though.  It's difficult to say if the convenience would outweigh the cost, but my intuition says no, it would not.

svaarala:
One more difference between the two call types is that a Duktape/C function may be called much later after the function is created, which means the C call site may no longer exist when the function gets called.

This in turn means you can't pass multiple values conveniently by stashing them into a temporary struct allocated from the C stack as you can with safe calls. If you need more than a single void pointer, a heap allocated structure or multiple properties are needed.

svaarala:
> Think of a Keyboard class with accessors for each key that return true/false depending on whether the key is pressed. Each accessor could have the same native function with the userdata specifying the key it checks.

There is actually a way to do this already: a function carries a 16-bit magic value which comes "free" with the function. This of course adds 2 bytes to `duk_hnativefunction` but it's more or less paid for because Duktape internals use magic values quite heavily to reduce footprint.

svaarala:
Note that there are two different issues with respect to a userdata argument:

- The mechanism to remember such a value for a call: for safe calls the call is immediate, for Duktape/C functions it's stored into the function object.
- The mechanism to access such a value: the value can be a direct argument to the native C function call (as userdata is in this pull for safe calls), or you may need to use Duktape API to read the value (as is the case for magic values).

Adding an argument to the C function prototype (as for userdata here) affects every call/function so it needs to be justified very carefully. For safe calls carrying a userdata is close to idiomatic because there's quite often the need to pass one or more arguments (pointers, value stack indices, etc) from the C call site into the safe call. And because a single pointer suffices to pass multiple values (by placing them into a C struct in the call site) the userdata is useful for most of these cases.

If this weren't the case it'd probably be better for the userdata to be accessible with an explicit function call (`duk_get_current_userdata()`) rather than it being a part of the C prototype for all safe calls. This is still worth considering because using a separate accessor function would make the API change backwards compatible.

For Duktape/C calls carrying a userdata is maybe not as idiomatic as it is for safe calls, especially because the userdata/context can't be stack allocated in the general case, so that there's always more mechanics to deal with the context.

fatcerberus:
I just looked at the diff for this change - It seems it was a bit more involved than I expected, Duktape apparently does a lot of safe calls internally. :)

svaarala:
Not really, there are less than 10 internal call sites. However, the change affects all C API testcases which are almost entirely executed within safe calls.

PR #742, opened 2016-05-04
Merge duk_debugger_attach_custom() and duk_debugger_attach() API calls

### LINKED ISSUE #735: Merge duk_debugger_attach() and duk_debugger_attach_custom() calls
There are currently two debugger attach API calls; the `_custom()` variant was added to support application specific notifies and methods and needed to be separate for API portability.

As a minimum change, rename `duk_debugger_attach_custom()` to `duk_debugger_attach()` so that only a single API call is needed.

Because there are conceivably more flags and/or callbacks in the future, a more versionable API would be nice; not sure what would be better however (a C struct and a macro initializer hiding the C struct versioning will be used for duk_def_prop() but that might be overkill here).

svaarala:
Rename `duk_debugger_attach_custom()` to `duk_debugger_attach()` to remove an unnecessary API call variant. The variant was added in 1.x when AppRequest was implemented and API compatibility couldn't be broken by adding the necessary AppRequest callback to `duk_debugger_attach()`.

Fixes #735.


fatcerberus:
While this is open, maybe add a clarification somewhere in the documentation that the request callback should avoid blocking, since the client is waiting for a reply and might time out.

svaarala:
Ok, I'll add a note - though that would be client specific and there might be situations where blocking (e.g. waiting for a button press or whatever) might be appropriate.

fatcerberus:
That's why I said "should" instead of "must", I figured there might be cases where blocking is unavoidable.  For example the target might need to block on a file system access if the client requests source code.

svaarala:
Sounds good. There's still a small semantic difference between an operation taking some time (like I/O) vs. genuinely blocking waiting for some event (like user pressing a button on the hardware); both are fine as long as the behavior is intentional.

svaarala:
@fatcerberus Re: the request_cb blocking, I added this: https://github.com/svaarala/duktape/pull/742/commits/6371470684fab1dfb33f3dc58b20c53c56b3573b.

Saying "should not block" is IMO still a bit incorrect because it's perfectly fine to block as long as the implications of that are understood. So I tried to explain the implications instead of saying what "should" be done.

fatcerberus:
That text looks good.

Basically that's what I was getting at, the client and target should "agree" that blocking/delays may happen and thus request_cb shouldn't block indiscriminately.  For example it would probably be bad to block if the client is just asking for a game title. :)

svaarala:
API documentation issue re: what "introduced" version to use for version incompatible calls: #749.

fatcerberus:
First incompatible change merged to master, that has to be some sort of milestone :stuck_out_tongue_closed_eyes:

PR #745, opened 2016-05-05
Remove print() and alert() default bindings

svaarala:
While `print()` and `alert()` are useful for getting started and testing scripts, they're often a portability issue because they depend on stdout/stderr being available. They can be disabled entirely using config options, but not easily redirected to a custom mechanism. They can be replaced altogether of course.

Remove the `print()` and `alert()` default bindings, and provide them instead as an optional extra in the distributable. The "duk" command line tool should provide the bindings because tests depend on at least `print()` being present.

Tasks:

- [x] Remove print and alert bindings
- [x] Fix cmdline and other examples to include their own minimal print and/or alert bindings
- [x] Check consistency of print() providers
- [x] Add print and alert bindings as an extra
- [x] Include print/alert provider into `duk` to make tests run
- [x] Remove dependency on print/alert from `duk`, only needed for "greet line", convert to C code
- [x] Add `DUK_CMDLINE_PRINTALERT_SUPPORT` to testrunner scripts to fix test errors
- [x] Migration instructions for 2.0
- [x] Debugger document update re: no print/alert forwarding (application specific messages instead)
- [x] Website: short description why Duktape lacks print() etc by default (embedded engine, avoid platform dependencies)
- [x] Website: update getting started and programming to describe a minimal print() example (and point to better alternatives)
- [x] Website: update index page example
- [x] Releases entry


fatcerberus:
Hm, guess this is a good time to remove my internal dependency on the `print()` binding.

PR #746, opened 2016-05-05
Move Duktape.Logger framework into an optional extra

svaarala:
The built-in logger framework is not always needed, and even when logging is used there may be an existing (sometimes platform provided) logging framework present which can't be easily adapted to work with Duktape.Logger.

Remove the default Duktape.Logger binding and move them into an optional extra.

- [x] Move Duktape.Logger init to an extra
- [x] Implement removed C API calls in the extra too (duk_log(), duk_log_va())
- [x] Getting current time: duk_get_now()
- [x] Formatting current time: duk_time_to_components() + sprintf()
- [x] Fix logger name automatic default
- [x] Add logging to command line, `DUK_CMDLINE_LOGGING_SUPPORT`
- [x] Testcase fixes; for API test to work, the extra would need be in runtests
- [x] Website changes: keep logging section, note that it's an extra, point to extra and wiki
- [x] Wiki update: note that logging is an extra in Duktape 2.x - https://github.com/svaarala/duktape-wiki/pull/114
- [x] 2.0 migration note
- [x] Releases entry


fatcerberus:
These changes will be good for lowering default footprint if nothing else. :)

svaarala:
Footprint impact is about -1.4kB (x64) which is pretty large compared to some other features (and 1% of total footprint for a low memory build). The other alternative for moving the logging stuff to extras would be to make it optional, but since it relies on relatively little internal magic and there is no clear logging "standard" it seems more appropriate to make it external instead.

fatcerberus:
I'd agree with that.  The only thing that would be close to a "standard" would be `console.log` which is available in many environments including Node.js.  Even so I think that's usually provided by the application, not the JS engine.

svaarala:
That's true - it probably would have been a better decision to provide a built-in barebones `console` binding rather than a logging framework. Which reminds me, it'd be nice to provide that `console` binding as an extra because it's so commonly requested. I could maybe bundle it with the external `print` and `alert` provider.

svaarala:
The implementation will depend on #771 to get the current time for printed log lines (as Duktape 1.x logging framework does).

svaarala:
Ok, functionally this is complete for an initial version. Opened a few API convenience issues while doing the conversion too. Next up is fixing the website, wiki, and other documentation and this can be merged.

svaarala:
Ok, this is finally complete. Final cleanups (like using `duk_def_prop_list()` for inits) are tracked by #766.

PR #747, opened 2016-05-05
Require callstack index for debug commands

fatcerberus:
I remember you saying you wanted to remove the checks in 2.x, so here's a prototype change.  The change is not yet complete however, there were a couple issues:
- What to do for Eval?  If we make the callstack index mandatory, with corresponding sanity check, then it becomes impossible for a client to issue an Eval with nothing on the callstack.  Should we just leave Eval as-is?
- There is a very heavyweight check after the `duk_debug_peek_byte()` in the handler for GetBytecode that I wasn't able to fully decipher: https://github.com/svaarala/duktape/blob/master/src/duk_debugger.c#L1830-L1856.  I left it alone because I didn't want to accidentally break anything.

### DIFF
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -1689,12 +1689,12 @@ GetVar request (0x1a)
 
 Format::
 
-    REQ <int: 0x1a> <str: varname> [<int: level>] EOM
+    REQ <int: 0x1a> <int: level> <str: varname> EOM
     REP <int: 0/1, found> <tval: value> EOM
 
 Example::
 
-    REQ 26 "testVar" EOM
+    REQ 26 -1 "testVar" EOM
     REP 1 "myValue" EOM
 
 Level specifies the callstack depth, where -1 is the topmost (current) function,
@@ -1706,12 +1706,12 @@ PutVar request (0x1b)
 
 Format::
 
-    REQ <int: 0x1b> <str: varname> <tval: value> [<int: level>] EOM
+    REQ <int: 0x1b> <int: level> <str: varname> <tval: value> EOM
     REP EOM
 
 Example::
 
-    REQ 27 "testVar" "newValue" EOM
+    REQ 27 -1 "testVar" "newValue" EOM
     REP EOM
 
 Level specifies the callstack depth, where -1 is the topmost (current) function,
@@ -1738,18 +1738,17 @@ GetLocals request (0x1d)
 
 Format::
 
-    REQ <int: 0x1d> [<int: level>] EOM
+    REQ <int: 0x1d> <int: level> EOM
     REP [ <str: varName> <tval: varValue> ]* EOM
 
 Example::
 
-    REQ 29 EOM
+    REQ 29 -1 EOM
     REP "x" "1" "y" "3.1415" "foo" "bar" EOM
 
 List local variable names from specified activation (the internal ``_Varmap``).
 Level specifies the callstack depth, where -1 is the topmost (current) function,
--2 is the calling function, etc.  If not provided, the topmost function will be
-used.
+-2 is the calling function, etc.
 
 The result includes only local variables declared with ``var`` and locally
 declared functions.  Variables outside the current function scope, including
@@ -1765,25 +1764,26 @@ Eval request (0x1e)
 
 Format::
 
-    REQ <int: 0x1e> <str: expression> [<int: level>] EOM
+    REQ <int: 0x1e> <int: level | null> <str: expression> EOM
     REP <int: 0=success, 1=error> <tval: value> EOM
 
 Example::
 
-    REQ 30 "1+2" EOM
+    REQ 30 null "1+2" EOM
     REP 0 3 EOM
 
 Level specifies the callstack depth, where -1 is the topmost (current) function,
 -2 is the calling function, etc.  If not provided, the topmost function will be
 used (as with a real ``eval()``).  The level affects only the lexical scope of
 the code evaluated.  The callstack will be intact, and will be visible in e.g.
-stack traces and ``Duktape.act()``.
+stack traces and ``Duktape.act()``.  The level can also be null to perform an
+indirect Eval.
 
-The eval expression is evaluated as if a "direct call" to eval was executed
-in the position where execution has paused, in the lexical scope specified by
-the provided callstack index.  A direct eval call shares the same lexical scope
-as the function it is called from (an indirect eval call does not).  For
-instance, suppose we're executing::
+If a valid callstack level is given, the eval expression is evaluated as if a
+"direct call" to eval was executed in the position where execution has paused,
+in the lexical scope specified by the provided callstack level.  A direct eval
+call shares the same lexical scope as the function it is called from (an
+indirect eval call does not).  For instance, suppose we're executing::
 
     function foo(x, y) {
         print(x);  // (A)
@@ -1818,7 +1818,8 @@ so that the Eval statement would:
 When Eval is requested from outside any Duktape activation, e.g. while doing
 a duk_debugger_cooperate() call, there is no active Ecmascript activation so
 that a "direct" Eval is not possible.  Eval will then be executed as an
-indirect Eval instead.
+indirect Eval instead.  As noted above, you can request an indirect Eval
+explicitly by sending null for the callstack level.
 
 Current limitations:
 
@@ -1875,7 +1876,7 @@ GetBytecode request (0x21)
 
 Format::
 
-    REQ <int: 0x21> [<int: level> OR <obj: target> OR <heapptr: target>] EOM
+    REQ <int: 0x21> [<int: level | obj: target | heapptr: target>] EOM
     REP <int: numconsts> (<tval: const>){numconsts}
         <int: numfuncs> (<tval: func>){numfuncs}
         <str: bytecode> EOM
@@ -1955,7 +1956,7 @@ GetHeapObjInfo (0x23)
 Format::
 
     REQ <int: 0x23> <tval: heapptr|object|pointer> EOM
-    REP [int: flags> <str/int: key> [<tval: value> OR <obj: getter> <obj: setter>]]* EOM
+    REP [<int: flags> <str/int: key> [<tval: value> | <obj: getter> <obj: setter>]]* EOM
 
 Example::
 
@@ -2000,7 +2001,7 @@ GetObjPropDesc (0x24)
 Format::
 
     REQ <int: 0x24> <obj: target> <str: key> EOM
-    REP <int: flags> <str/int: key> [<tval: value> OR <obj: getter> <obj: setter>] EOM
+    REP <int: flags> <str/int: key> [<tval: value> | <obj: getter> <obj: setter>] EOM
 
 Example::
 
@@ -2079,7 +2080,7 @@ GetObjPropDescRange (0x25)
 Format::
 
     REQ <int: 0x25> <obj: target> <int: idx_start> <int: idx_end> EOM
-    REP [int: flags> <str/int: key> [<tval: value> OR <obj: getter> <obj: setter>]]* EOM
+    REP [<int: flags> <str/int: key> [<tval: value> | <obj: getter> <obj: setter>]]* EOM
 
 Example::
 
--- a/src-input/duk_debugger.c
+++ b/src-input/duk_debugger.c
@@ -1291,18 +1291,14 @@ DUK_LOCAL void duk__debug_handle_get_var(duk_hthread *thr, duk_heap *heap) {
 	DUK_UNREF(heap);
 	DUK_D(DUK_DPRINT("debug command GetVar"));
 
+	level = duk_debug_read_int(thr);  /* callstack index */
+	if (level >= 0 || -level > (duk_int32_t) thr->callstack_top) {
+		DUK_D(DUK_DPRINT("invalid callstack index for GetVar"));
+		duk_debug_write_error_eom(thr, DUK_DBG_ERR_NOTFOUND, "invalid callstack index");
+		return;
+	}
 	str = duk_debug_read_hstring(thr);  /* push to stack */
 	DUK_ASSERT(str != NULL);
-	if (duk_debug_peek_byte(thr) != DUK_DBG_IB_EOM) {
-		level = duk_debug_read_int(thr);  /* optional callstack level */
-		if (level >= 0 || -level > (duk_int32_t) thr->callstack_top) {
-			DUK_D(DUK_DPRINT("invalid callstack level for GetVar"));
-			duk_debug_write_error_eom(thr, DUK_DBG_ERR_NOTFOUND, "invalid callstack level");
-			return;
-		}
-	} else {
-		level = -1;
-	}
 
 	if (thr->callstack_top > 0) {
 		rc = duk_js_getvar_activation(thr,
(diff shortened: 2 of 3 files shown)


fatcerberus:
One solution for Eval would be to leave the check, but change the semantics:
* Eval with no callstack index is indirect, regardless of callstack content
* Eval with callstack index is direct, in the scope of the specified call (as now)

@svaarala What do you think?

svaarala:
Regarding eval, it might be easier for later extensions if the callstack level was present even when an indirect eval is needed, so that when new values are appended there wouldn't be confusion regarding their meaning. Perhaps it could be passed in as "undefined" to indicate an indirect eval? There's no existing command doing this so (AFAICR) so there's no existing model for this.

The GetBytecode check is a bit involved because the commands either a callstack level or a heap object (which can be any function, even outside the callstack). To support that, the command parses the argument as a `duk_tval` and then checks what it got. That probably shouldn't be changed in this pull because both use cases are needed at present.

I'd maybe prefer to simplify that by making GetBytecode only accept a heap object argument and provide a separate command to inspect the callstack to look up the function. Some other changes would also be needed for consistency I think - but that's not really an issue for this pull I think.

fatcerberus:
Alright, I'll leave GetBytecode alone.  For the other commands: Should I reorder it so the call stack index comes first?  It was appended onto the end to maintain compatibility but we can change it now. :)

I guess I should also bump the protocol version in this pull.

svaarala:
Hmm, I'll bump the protocol version in a separate pull so it's apparent from the 2.0.0 issue/pull list.

svaarala:
@fatcerberus I've started the incompatible 2.x changes from pending stuff that's been hashed over enough many times to be more or less obvious. For the debugger, we should review the open issues and previous discussions to figure out what a good, consistent set of improvements would be for 2.x.

fatcerberus:
@svaarala I updated the branch with the following changes:
* Callstack index is first argument for commands accepting one
* Eval accepts a null callstack level, indicating indirect eval

I also updated the documentation to match and added RELEASES entries.  Let me know what you think.

fatcerberus:
By the way I used `null` instead of `undefined` because traditionally in JS, if memory serves, you're supposed to use `null` to indicate deliberate omission of a value, leaving `undefined` with the semantics of "nobody put anything here yet".  You don't normally e.g. assign a variable `undefined` or pass it literally to a function.

fatcerberus:
@svaarala Any timetable on getting this merged into master?  I know there are more pressing concerns to deal with for 2.x first :)

svaarala:
It depends a bit on what other debugger stuff would make 2.x. If not much does, I think it may be necessary to revert the (very few) incompatible changes in master because breaking compatibility doesn't make sense unless the changes are big enough to warrant it. But instead of doing that, I hope we can do some other pending (incompatible) debugger changes so that the breakage is justified.

fatcerberus:
I think the only incompatible debugger changes so far have been API improvements, which shouldn't have to be reverted.  I'm almost positive the protocol is still compatible in master (other than the version number bump), since the only change I've had to make to SSJ to keep it 2.x compatible was to change the handshake code to accept V2.

svaarala:
Ah, you're right, I incorrect remembered that the argument ordering pull had been merged but it hasn't. API changes are fine, and the version bump can be reverted if necessary. But it'd of course be better to improve things instead :)

fatcerberus:
@svaarala I'll rebase this once I get home and see if any other changes are needed.  Then this can be merged in time for Duktape 2.0.0. :-)

fatcerberus:
It looks like the only thing that needs to be figured out before this can be merged is how to specify direct/indirect eval.  The current method used in this pull is:

Direct eval:
```
# idx = callstack index
REQ 0x1E <int: idx> "a+b" EOM
```

Indirect eval:
```
# idx of null = indirect
REQ 0x1E null "1+2" EOM
```

It seems kind of awkward to accept `null` (or even undefined) in place of an integer index like that.  So maybe instead it would be better to have separate debugger commands for direct and indirect eval?  The former would require a valid callstack index (and thus always be direct), while the latter would just take an eval string and be always indirect.  That seems overall cleaner and easier to explain in documentation than the current situation in master of, "sometimes it's direct, sometimes not; here are the conditions for that".

fatcerberus:
@svaarala Would you be opposed to the change described above (i.e. separate commands for direct and indirect eval)?

svaarala:
Not sure if that's much cleaner: even when a callstack index is specified the current implementation will use indirect eval when the callstack index points to a native function. If that's not changed, the "direct eval" command would still sometimes do an indirect eval.

Assuming the arguments are otherwise the same (beyond callstack level that is) it also seems a bit unclean to have multiple Eval variants. What if there's a new future flag? Would there then be 4 eval variants? If flags are likely, then maybe an explicit flags field would make more sense?

So overall not really sure what's best. I agree that reading multitype arguments (null or int) is currently a bit awkward, and is definitely something to consider when touching the debugger low level format the next time. As a cosmetic issue, for me `undefined` makes more sense that `null` (i.e. same behavior as if an argument to a function call was not given).

fatcerberus:
For a native function, the "direct" variant doing an "indirect" eval is not something I see as an issue w.r.t consistency, since "direct eval" just means "eval in the lexical environment of this call".  Native functions don't have a meaningful lexical environment (from an Ecmascript point of view), so there's actually no observable difference between direct and indirect in that case.

I do see where you're coming from with possibly needing more variants in the future, though.  That's a valid point.

Anyway, since you've expressed a desire to change the message framing at some point in the future, and with Duktape 2.0.0 getting closer, for this pull it's probably best to stick with the minimal change for now and just accept null (or undefined, if that's preferable) to mean "eval in global scope", or else a callstack index to eval in function scope.

fatcerberus:
Alternatively, I could just leave the Eval command as-is, i.e. with the callstack index being completely optional and given after the eval string.  That's more in line with usual JS conventions for optional function parameters, where the optional arguments come after all the required ones.

fatcerberus:
Regarding null vs. undefined: In this case there's a subtle difference in semantics that makes me prefer null: By passing `null` as a callstack index, you're being explicit: "Not in the scope of any active call".  `undefined` can be taken to mean "value not provided; use the default" - which, to someone looking only at debug client code, is equally likely to mean "the topmost call" as "the global scope".

fatcerberus:
Anyway, I'll leave this pull as-is.  I'm overall satisfied with the way things are here, so unless a clearly superior solution comes along I'm okay with it being merged in its present state.  A better setup can be worked out later, perhaps at the same time as the future message framing reworks.

svaarala:
Re: null vs. undefined, maybe the convention that `null` is used for explicitly provided arguments with the intention "no data", as you said. There's no concept of "no argument" yet for the debugger as the debugger code doesn't parse an argument list in the conventional sense. But if there were, `undefined` would then be logical for missing arguments if one draws an analogy to Ecmascript calls.

Out of the various alternatives, *if* all other callstack related commands have a callstack index it would seem cleanest to have Eval include a callstack index and then just accept a `null` also. It's not cleanest to implement though, but maybe that's secondary in this case. With the ability to peek a byte now, the `null` can be peeked and if it's not there, the code can read an integer without further awkward checks.

fatcerberus:
Right, the argument in question is not optional.  It comes before a required argument (the eval string) and is either a valid callstack index or `null` to mean "no callstack index, global scope".  Using `undefined` with that meaning doesn't seem right to me.

To be clear, the signature of the Eval command in this pull is:
```
REQ Eval=0x1e <int: stack_idx OR null> <str: eval_str>
```

`null` in this case meaning something closer to "not a stack index - do an indirect eval" rather than undefined's typical meaning of "argument omitted, use your best judgment".

fatcerberus:
All this consternation could have been avoided if JS had just had one "null" value like every other programming language in existence :)

svaarala:
Right, I understood the argument list - I still don't fully agree that `null` is most appropriate there and typical Ecmascript built-ins will treat an explicit `undefined` and `null` the same way.

But I'm fine with `null` too, we should then just be consistent in how it's used in debugger commands.

fatcerberus:
Like I said above, the analogy I was making was to `Object.setPrototypeOf()`, which uses `null` to mean "no object", as distinct from any other value.  That function doesn't accept `undefined` as an argument, but if it did I wouldn't be surprised if it were to map to a default value, say `Object.prototype` (or maybe to not change the prototype at all).

Anyway, I'm not planning to make any other changes to this pull, so whenever you get a chance to review is fine :)

svaarala:
Other bindings do accept undefined too, e.g.:

```
duk> JSON.stringify({foo:123}, null, 4)
= "{\n    \"foo\": 123\n}"
duk> JSON.stringify({foo:123}, void 0, 4)
= "{\n    \"foo\": 123\n}"
```

But, I'm fine with `null`, it just wouldn't be my choice :)

I'll review the pull later tonight.

fatcerberus:
I guess it's just my coding style - my rule of thumb for arguments is that a `null` argument is explicit: "there's nothing here on purpose"; and `undefined` is more of a placeholder: "I don't care; do whatever you want for this value"

But yeah, it doesn't really matter what convention is used as long as it's consistent.  No sense getting into "space or tab" style religious debates ;)

src-input/duk_debugger.c
```diff
@@ -1494,33 +1480,48 @@ DUK_LOCAL void duk__debug_handle_eval(duk_hthread *thr, duk_heap *heap) {
...
 	 * as seen by, e.g. Duktape.act() will be the same regardless.
 	 */
 
-	/* nargs == 2 so we can pass a callstack level to eval(). */
+	/* nargs == 2 so we can pass a callstack index to eval(). */
 	duk_push_c_function(ctx, duk_bi_global_object_eval, 2 /*nargs*/);
 	duk_push_undefined(ctx);  /* 'this' binding shouldn't matter here */
 
-	(void) duk_debug_read_hstring(thr);
-	if (duk_debug_peek_byte(thr) != DUK_DBG_IB_EOM) {
```
svaarala: I think it should be enough here to peek for a DUK_DBG_IB_NULL and if that's not present, read the int as before?

doc/debugger.rst
```diff
@@ -1765,25 +1764,26 @@ Eval request (0x1e)
 
 Format::
 
-    REQ <int: 0x1e> <str: expression> [<int: level>] EOM
+    REQ <int: 0x1e> <tval: level> <str: expression> EOM
```
svaarala: Saying tval: level is maybe a bit too broad - maybe: `<int: level | null>`?

src-input/duk_debugger.c
fatcerberus: Ah, right: `duk_tval` is a borrowed reference and must be left on the  value stack to prevent premature collection.

But the lookahead solution is a bit cleaner, so I'll just do that.

doc/debugger.rst
fatcerberus: I thought the same thing, but wasn't sure of a good notation for a union type.  There was no precedent for such a thing before this pull.

svaarala:
One small detail about Eval this pull changes is that prior to the pull, leaving the callstack index out would mean "evaluate in the topmost activation if it exists, and indirect eval otherwise". In essence this would be callstack index -1, but with leniency so that if the index doesn't exist, eval runs in the global scope.

The web UI debugger used this capability to allow one to just eval code which would happen inside a function if there was something in the callstack, or in the global scope otherwise, which is quite logical from an end user point of view. I can't substitute a -1 because it's strictly validated; nor can I substitute a `null` because it has a different meaning.

It's a minor issue, and the web UI can be changed to first query the callstack state before deciding on Eval arguments. This works when paused. When Duktape is running there's no atomic way to do it other to issue a Pause, GetCallstack, Eval, and Resume.

doc/debugger.rst
svaarala: GetBytecode does have a precedent but a bit different:

    REQ <int: 0x21> [<int: level> OR <obj: target> OR <heapptr: target>] EOM

src-input/duk_debugger.c
svaarala: Since it's ultimately coerced to an integer, it could be popped off once the level had been figured. But the lookahead solution is better.

fatcerberus:
Hm, I'm not sure if I'd ever want to fire off a direct eval into the wind like that, personally.  Weird, difficult-to-diagnose things may happen if one were to eval, say, `var x = 812` :)

That said, it's a valid point that this removes functionality that existed before.  It would be possible to special-case -1 specifically to behave like before, thoughts?

svaarala:
Well, then just don't say that? :)

svaarala:
If -1 was treated differently from other indices I'd find the very surprising so that feels quite awkward to me.

Treating the index leniently in all cases would be possible. It would then treat an invalid index the same way as an index pointing to a native function - i.e. evaluate it using indirect eval. Callers that care about this would need to inspect the callstack first using GetCallstack to make sure what indices can be used without invoking an indirect eval.

That's actually the case regardless, because any index pointing to a native function will silently result in an indirect eval, so if you want to avoid that you'll need to scan the callstack manually anyway. For example, you could locate the topmost Ecmascript activation and execute an Eval there. It depends on the debugger UI really and what Eval is being used for (manual eval, an automatic periodic watch expression, etc).

But this existing functionality may not be all that important. It just means I'll need to rethink the web UI Eval function a bit because it's intended to be user friendly and Eval in the topmost activation when that exists, and global scope otherwise.

fatcerberus:
Alright, so I guess leave this pull as is (except for fixing the review issues), and we can figure out the rest separately?

svaarala:
To keep master consistent I don't want to merge this until #1103 is also ready to be merged. The only hitch is Eval handling. I'll see if I can work around that in #1103; if not, we'll need to figure out something here.

svaarala:
Hmm, it seems that actually the debugger web UI in master has already been broken when the Eval callstack index was added. The Node.js code uses this for level:

    (typeof level === 'number' ? level : -1)

which means that level is always explicit. As a result, attempting an Eval with an empty callstack always fails in master because an explicit -1 is treated strictly while a missing argument is treated as a lenient -1.

I hadn't noticed this in manual testing: when you debug a program the debugger becomes paused at the first opcode of the program. But the global code of that program is already running so that there's a single level callstack. I hadn't tested the changed behavior with an empty callstack which makes a difference.

So hmm... :)

svaarala:
Given the current behavior of the debugger Web UI is already broken in master (for an empty callstack), I can merge this without making Eval changes (other than the suggested parsing changes) and figure out the next step separately. Doing so won't make master behave worse than it already does.

svaarala:
This might be enough of a fix, and to document a "how to" pattern for debug clients: https://github.com/svaarala/duktape/blob/abafb03abf0f2da4deced291f494f76d6e51b168/debugger/duk_debug.js#L1324-L1329.

fatcerberus:
This kind of thing is where some automated testing would come in handy.  Unfortunately it's difficult to automate :-(

svaarala:
Or rather, it's time consuming. I still intend to do it, but improving coverage for the non-debugger tests (more platforms, more variants, etc) has been more important so far.

fatcerberus:
Alright, I made the suggested changes, anything else you need me to do here?

svaarala:
I'll check the changes and merge a bit later. Thanks! :)

doc/debugger.rst
```diff
@@ -1875,7 +1876,7 @@ GetBytecode request (0x21)
 
 Format::
 
-    REQ <int: 0x21> [<int: level> OR <obj: target> OR <heapptr: target>] EOM
+    REQ <int: 0x21> [<int: level | obj: target | heapptr: target>] EOM
```
svaarala: These notation fixes are a bit inconsistent, couple of them are `<int: something | obj: something>` while others are `<int: something> | <obj: something>`. I'll fix it in a cleanup pull.

fatcerberus: The ones that are inconsistent is because the contract is "(this 1 value) OR (these 2 values)".  So it got a different notation than "(1 value with either this or that type)"

svaarala: That wasn't very readable to me, I'll see if something else might work better.

PR #756, opened 2016-05-07
Bump debug protocol version

svaarala:
With minor incompatible changes in Duktape 2.x, bump the debug protocol version to 2.

- [x] Update API constant (also controls Duktape version identification line)
- [x] Update debugger documentation for bumped version
- [x] Fix duk_debug.js to use the bumped version
- [x] Releases
- [x] 2.0 migration note


svaarala:
Migration notes for the protocol version entry are quite high level, but can't think of much more to say other than to update the debug client for the new protocol version and command changes. There should be separate migration entries for specific changes of course.

fatcerberus:
That looks good to me.  There's not much else to say until the concrete changes have been made.

PR #758, opened 2016-05-07
Add `ctx` argument to debugger detached callback

fatcerberus:
Immediately reattaching in the detached callback is a supported scenario.  This pull makes that easier by adding a `ctx` argument to detached_cb.


fatcerberus:
Oh hey, this got merged.  I wasn't expecting that, thanks. :-)

PR #760, opened 2016-05-07
Map shared string macros to plain string constants

svaarala:
Map shared string macros (e.g. `DUK_STR_UNSUPPORTED`) to plain strings rather than symbols (`duk_str_unsupported`). Except for very old compilers, string sharing will provide the same benefit, and maintaining the macro/symbol structure is awkward because it leads to "unused symbol" warnings when strings are not referenced with non-default config options.


fatcerberus:
I always wondered why there was this extra layer of indirection with these.  Most compilers will pool string literals so there's not much to be gained from going through a const.  And a lot of these only have one or two places where they're used anyway so there shouldn't be much duplication going on in any case.

svaarala:
Depending on what one considers significant, there's a reasonable amount of reuse with the strings. In fact, now that the strings are string literals and not explicit symbols, GCC is able to share strings better (the -flto enabled build is 193 bytes smaller for x64, this probably comes from sharing suffixes so that the string constant "foo bar\0" can also be used for "bar\0" by simply sharing the suffix).

In any case the compilers that don't do string sharing well are quite old and I don't know a practical case where they would now be used. It's possible to do something about the strings if such a concrete case surfaces.

PR #761, opened 2016-05-08
Remove file I/O support from config

svaarala:
Remove file I/O support such as: `duk_file`, `DUK_STDOUT`, `DUK_FWRITE`, etc. The libc sprintf/scanf bindings will remain. This can be merged once all the I/O using call sites have been removed: #781, #782, and #788.


svaarala:
Alright, now what's left is roughly:
- setjmp/longjmp
- snprintf/sscanf and variants
- strlen and variants
- memset, memcpy, etc
- malloc, realloc, free (for the default allocator)
- math functions

Issue #331, opened 2015-09-08
Allow JX-serialization of cyclic objects


fatcerberus:
I know JSON specifically doesn't allow cyclic objects to be serialized, but it would be useful if JX could do so.  Obviously round-trip serialization would be impossible, but maybe Duktape could put some sort of backreference in the output when it encountered a cycle, similiar to how it adds `{ _func:true }` tags for otherwise unserializable functions.

Currently variable viewing in the minisphere debugger is limited because there are no commands to walk objects, so I have to rely on `Duktape.enc` and an Eval command to display the full object.  This fails, however, in the case of a cyclic object (which it turns out happens _a lot_ in the game I'm making).

svaarala:
This should be relatively simple to implement. Here's a quick hack branch of one approach (it's a hack in that it breaks JSON.stringify() output for example):

- https://github.com/svaarala/duktape/compare/cyclic-jx-gh331

Example:

```
$ ./duk
((o) Duktape 1.2.99 (v1.2.0-583-g81c2e5d-dirty)
duk> obj = {};
= [object Object]
duk> obj.foo = {}
= [object Object]
duk> obj.foo.bar = {}
= [object Object]
duk> obj.foo.bar.quux = { ref: obj.foo }
= [object Object]
duk> Duktape.enc('jx', obj)
= {foo:{bar:{quux:{ref:{_loop:true}}}}}
```

The plumbing is a bit awkward so maybe reorganizing the helpers a bit would be needed.

fatcerberus:
Even more useful would be if there were some way for the _loop tag to specify what it looped back on, but that would probably be too difficult to implement and/or hurt performance.

As long as all the unique objects are present in the output I'll be happy. :)

fatcerberus:
Looked at the diff, I'm impressed at how few changes were needed for a minimal implementation.  I thought this would be a lot more difficult!

svaarala:
One possible format would be something like:

```
{_loop:(0xdeadbeef)} 
```

The pointer there could be the `duk_heaphdr` address of the object. But since it's not present in the serialized form of the target object, this only really serves to indicate if there are multiple separate loops.

A more fancy version would be to serialize the `_loop` tag and then perhaps serialize the target shallowly (just one level) which would mostly allow identifying the target. For example if the target object has an `.id` or `.type` property or similar, that would be shown. But this will essentially serialize certain objects twice so it seems quite messy to me.

fatcerberus:
[...]

PR #767, opened 2016-05-09
Add minimal console extra

svaarala:
Add a minimal (optional) console extra with:

- console.trace
- console.log (= console.debug)
- console.info
- console.warn
- console.error (= console.exception)
- console.assert
- console.dir

Object formatting is handled through `console.format` which uses JX by default, falling back to `ToString(v)` if JX serialization fails (e.g. due to a looped object). User code can easily replace `console.format` with something more appropriate.

Tasks:

- [x] Proxy semantics (required by specification to avoid errors on missing functions; not implemented like that by Node.js though)
- [x] Make Proxy semantics optional (init parameter)
- [x] Dense and multiline object formatting, maybe use multiline for console.dir() => Skip, Node.js also uses single line formatting
- [x] Dukweb console object
- [x] Add `DUK_CMDLINE_CONSOLE_SUPPORT` to enable console bindings for `duk`
- [x] Website: suggestion for I/O bindings (suggest where print/alert is mentioned)

These were marked as future work (XXX note in source):

- [x] Minimal level filtering
- [x] Compile time selection of stdout/stderr output, flushing behavior
- [x] Compare stdout/stderr behavior to Node (seems to use stderr for >= warn and trace, stdout for info and log)
- [x] Initialize using `duk_def_prop()`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,8 @@ DUKTAPE_CMDLINE_SOURCES = \
 	dist/examples/alloc-torture/duk_alloc_torture.c \
 	dist/examples/alloc-hybrid/duk_alloc_hybrid.c \
 	dist/examples/debug-trans-socket/duk_trans_socket_unix.c \
-	dist/extras/print-alert/duk_print_alert.c
+	dist/extras/print-alert/duk_print_alert.c \
+	dist/extras/console/duk_console.c
 LINENOISE_SOURCES = \
 	linenoise/linenoise.c
 
@@ -257,6 +258,7 @@ CCOPTS_FEATURES += -DDUK_OPT_FASTINT
 #CCOPTS_FEATURES += -DDUK_OPT_ROM_OBJECTS
 CCOPTS_FEATURES += -DDUK_OPT_JSON_STRINGIFY_FASTPATH
 CCOPTS_FEATURES += -DDUK_CMDLINE_PRINTALERT_SUPPORT
+CCOPTS_FEATURES += -DDUK_CMDLINE_CONSOLE_SUPPORT
 CCOPTS_FEATURES += -DDUK_CMDLINE_FANCY
 CCOPTS_FEATURES += -DDUK_CMDLINE_ALLOC_LOGGING
 CCOPTS_FEATURES += -DDUK_CMDLINE_ALLOC_TORTURE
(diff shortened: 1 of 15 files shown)


fatcerberus:
Speaking of JX serialization and loops, any progress on #331? ;)

svaarala:
@fatcerberus Not so far, I'm still working through removing built-in bindings dependent on stdout/stderr or file I/O.

fatcerberus:
That's an admirable pursuit, carry on. :)

svaarala:
Overall I think a good order will be to first look at the multi property get/set API calls. If those turn out to be good enough, then `duk_def_prop()` could maybe use an object syntax (and avoid the variable value stack shape approach which would be inconsistent/asymmetric with a `duk_get_prop_desc()`). And finally, JX serialization could then take an object providing formatting arguments such as loop support etc. But we'll see :)

fatcerberus:
Anyway this is nice, I'll keep this in mind for possible inclusion in minisphere 4.0.

svaarala:
@fatcerberus I think most applications with an existing console model should just implement `console` on their own. The extra provided is an easy-to-use minimal implementation which can work as a starting point (and won't port to embedded platforms with non-standard I/O setups etc).

svaarala:
Alright, this now looks something like this:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-85-g9b70f20)

duk> console.assert(true, 'no error');
= undefined

duk> console.assert(1 > 2, 'should not happen', { foo: 123 });
AssertionError: should not happen {foo:123}
    at [anon] (dist/extras/console/duk_console.c:51) internal
    at assert () native strict preventsyield
= undefined

duk> console.log(1, 2, 3, { foo: 'bar', quux: [1,2,3] });
1 2 3 {foo:"bar",quux:[1,2,3]}
= undefined

duk> console.trace();
Trace
    at [anon] (dist/extras/console/duk_console.c:51) internal
    at trace () native strict preventsyield
= undefined
```

Formatting uses `String(v)` for everything but objects. For objects JX is attempted and if that fails (e.g. looped object), the formatter falls back to `String(v)`. This is a reasonable initial behavior IMO:

```
duk> var a = {}; var b = {}; a.ref = b;
= [object Object]

duk> console.log(a, b);
{ref:{}} {}
= undefined

duk> b.ref = a;  // complete loop
= [object Object]

duk> console.log(a, b);
[object Object] [object Object]
= undefined
```

User code can change the object formatting behavior simply by replacing `console.format`:

```
duk> console.log(123, {foo: 123});
123 {foo:123}
= undefined

duk> console.format = function format(v) { return '<OBJECT>'; };
= function format() {"ecmascript"}

duk> console.log(123, {foo: 123});
123 <OBJECT>
= undefined
```

fatcerberus:
Like the JX formatting for objects.  Node.js does something similar in its REPL.

svaarala:
A better solution would have some sanity limits on object key/value lengths (e.g. truncating strings and buffers above, say, 256 bytes/chars), detect loops (indicating them using some markup), handle accessors, etc. But maybe we'll leave these for future work :)

svaarala:
The footprint of this extra is currently about 1.6kB on x64.

PR #771, opened 2016-05-12
First draft of time support API calls

svaarala:
Add API calls for minimal date/time support in C code:

- duk_get_now() API call to get Date.now() in C. This allows C code to easily get the same view of current time as Ecmascript code (which is not always the same as time seen by standard time API calls).
- duk_time_to_components(), duk_components_to_time(): convert between time value and components (year, month, etc)

These should be enough to get time values, modify them, convert time into local and UTC components which then allows custom formatting.


svaarala:
The Ecmascript API is somewhat annoying in my opinion in that month is zero-based and doesn't match what gets formatted. For example, to create `2016-01-02 03:04:05.006Z` one uses:

```
> var d = new Date(Date.UTC(2016, 0 /* ! */, 2, 3, 4, 5, 6));
undefined
> console.log(d.toISOString());
2016-01-02T03:04:05.006Z
undefined
```

Similarly, `d.getUTCMonth()` will return a zero-based month (0 for January etc). If you're formatting a timestamp, you need to `+ 1` for the month.

For me it'd make more sense that a component breakdown was directly in the format one uses for human readable time formatting, i.e. both month and day-in-month being one-based. So I'll draft the API initially like this.

svaarala:
Ok, as soon as the tests pass this initial merge is going in. API calls are marked experimental for now because significant changes may still be made; I need this early merge to progress on the logging framework pull. The basic limitations here now are:

- UTC only.
- No API call for "get local time offset at time value T"
- No formatting/parsing support in the API, but you can use the breakdown to format on your own (not sure if formatting should even be added)
- API call and struct member names liable to change etc.

svaarala:
Code footprint at the moment is about +300 bytes on x64; pretty high considering there are three API calls which don't do much except argument conversion. There are a few ways to improve on this, e.g. by avoiding the conversion by using the same struct internally, etc.

PR #781, opened 2016-05-13
Remove fatal/panic distinction and use fatal errors only

svaarala:
Remove the Duktape 1.x distinction between fatal errors and panic errors which had different handling paths. The main difference between these two was that fatal errors went through a user-registered callback associated with a Duktape heap, while panics could happen anywhere without context so that no user-registered callback could be looked up.

Simplify this as follows:

- Fatal errors with context are handled as previously: call `heap->fatal_func` which is either a default fatal error handler or a user-registered fatal error handler.
- Fatal errors without context call into a Duktape-wide default fatal error handler (also used if user code doesn't register a fatal error handler in heap creation) with `ctx == NULL`. By default, that handler segfaults and forever loops, and avoids any I/O for portability reasons. The default handler can be replaced via `duk_config.h` if that behavior is not good enough (but see below).
- Remove the error code argument from the `duk_fatal_function` signature (so that it now only takes a `duk_context *ctx, const char *msg`) because the code is pretty useless in practice.

Context-free fatal errors only happen in two places now: (1) when assertions fail and (2) when self tests fail in heap creation. In other words, if assertions and self tests are disabled, there are no context-free fatal errors at all. Replacing the default fatal handler is thus usually not necessary, but if replaced it must be done via `duk_config.h`; this matches the way panics worked in Duktape 1.x but is simpler conceptually.

- [x] Simplify `duk_fatal_function` signature by dropping the error code.
- [x] Change `ctx` in fatal function signature into (heap) `udata`: it's less confusing that an arbitrary context pointer, especially because the context pointer cannot be used e.g. in Duktape API calls safely.
- [x] Should `msg` be optional or not?
- [x] Config: remove panic options
- [x] Config: add option to replace default fatal error handler
- [x] Fix internal panic call sites to use fatal errors
- [x] Fix internal fatal error handling call sites for new signature
- [x] Fix other fatal error providers, like Dukweb, ajduk, etc.
- [x] Website update for discussion of errors, fatals, and panics
- [x] Wiki update for fatals and panics; maybe a new how to for fatal error handling with the bulk of the current (rather detailed) website content - https://github.com/svaarala/duktape-wiki/pull/116
- [x] Documentation: cover the default fatal error handler behavior (= intentional segfault) in tutorials; in particular, `duk_create_heap()` documentation should emphasize that unless a fatal error function is registered, the default one segfaults on purpose for uncaught errors
- [x] Test default and user-provided default fatal error handler
- [x] 2.0 migration notes
- [x] Releases entry


svaarala:
As things are now in this branch the fatal error function signature is:

```c
void func(duk_context *ctx, const char *msg);
```

That `ctx` argument is maybe somewhat pointless: it identifies a specific Duktape thread but the fatal error handler itself is heap-wide. Further, it's unsafe to actually use the `ctx` argument in the fatal error handler (for example to call into the Duktape API).

Couple of options:

- Keep the `ctx` argument as is (also same in Duktape 1.x):
- Replace `ctx` with the userdata given in heap creation. This is more clearly heap-wide and may be more useful to user code handling the fatal error than a certain Duktape thread in a Duktape heap.
- Replace `ctx` with a Duktape heap pointer; however, the heap type is not exposed in the API at present so that won't be possible unless a typing change is made.
- Remove the `ctx` argument altogether as more or less pointless.

For a context-free fatal error (assertion or self test error) the argument (if present) would always be NULL in any case, and the Duktape-wide fatal error handler would be used in that case because there's no context to look up a heap specific handler from.

Right now I'm leaning towards using the heap-initialization userdata because that seems more useful to a user handler than a Duktape thread pointer (noting that the thread is not necessarily the initial thread, i.e. the one returned from `duk_create_heap()`). After all, `ctx` is an opaque value, only meaningful if you use it in the Duktape API, but that's not safe in the fatal error handler.

svaarala:
Having gone through the alternatives using the udata instead of ctx seems the best approach. The signature of a fatal function will be:

```c
void func(void *udata, const char *msg);
```

where `udata` is the userdata given in heap creation. This also has the nice side effect that a self test failure in heap creation can call the fatal handler with the correct userdata argument (unlike when `ctx` is the first argument: there's no context to give because it hasn't been created when the self tests run).

fatcerberus:
I definitely agree with using `udata`.

`ctx` is completely useless in this context: You can't actually utilize it safely (via the Duktape API), and it's also useless for identification because it might not even be the original context returned by `duk_create_heap()`.  Which might be okay if only C code could create new threads (in which case `ctx` could be traced back easily), but Ecmascript code can also create threads, whose `ctx` the native code has no reliable way of tracking.

fatcerberus:
One minor (possible) issue: Would it be better for the udata parameter to come after the message?  That seems to be the norm in C for user data pointers, udata is typically the final argument.  However, Duktape has udata first for e.g. debugger callbacks, so I'm not actually sure what's best here.

svaarala:
It's first also in alloc callbacks etc and actually my experience is that it often comes first :)

svaarala:
At the moment the prevailing convention is that `ctx` comes first it if exists, and `udata` comes after that (or first if there is no `ctx`).

fatcerberus:
Hm, I guess this is a good opportunity to fix my uncaught error handling since I'm working on minisphere 4.0 anyway.  The way I use the fatal error handler in minisphere is... questionable. :o)

PR #786, opened 2016-05-15
Reorder duk_create_heap() arguments, userdata should be last

svaarala:
Now that the fatal error handler accepts a (heap) userdata argument, the userdata argument should be last in `duk_create_heap()`. This follows the prevailing convention that when callbacks are registered userdata follows the relevant callback(s) (and when the callbacks are called, userdata is first).


svaarala:
This may be better implemented by hiding some of the initialization variables into struct(s), with a macro to initialize a struct to a default state (ensuring portability for initializing NULLs). This makes the API more easily versionable. So, assigning off 2.0.0.

Issue #60, opened 2014-10-29
Proposal for new require system


creationix:
I've designed a new require system that duktape could provide that would be flexible for my needs and enough to allow the user to implement a node.js or lua style require system in full.

I don't know enough about duktape performance to know if using prototypes of closures is preferred.  This sample uses prototypes since that's way faster in V8 where most my experience is and I _think_ it would use less ram in duktape which I know is a big goal.

``` js
Module.cache = {};

function Module(id) {
  var cached = Module.cache[id];
  if (cached) { return cached; }
  Module.cache[id] = this;
  this.exports = {};
  this.id = id;
  this.load();
}

Module.prototype.require = function (id) {
  return (new Module(this.resolve(id))).exports;
};

Module.prototype.resolve = function () {
  throw new Error("Please replace Module.prototype.resolve(id) with your own implementation.");
};

Module.prototype.load = function () {
  throw new Error("Please replace Module.prototype.load() with your own implementation.");
};
```

A sample concrete integration with this could be a simple loader that resolves IDs starting with "." as relative to caller's directory and leaves others unchanged.  It's loader simply loads the files from disk and evaluates them using common-js semantics.

``` js
// Sample usage:
// Assume `readFile` and `pathJoin` are provided by user code not shown here`.
// function pathJoin(part1, part2, ...) {} -> path
// function readFile(path) {} -> buffer


// Simple resolve that maps relative requires relative to caller
// and leaves module requires as-is.
Module.prototype.resolve = function resolve(id) {
  if (id[0] === '.') {
    return pathJoin(this.id, '..', id);
  }
  return id;
};

// Simple loader that loads from disk and evals using new Function
Module.prototype.load = function () {
  var code = readFile(this.id);
  // Like `new Function`, but also allows setting filename for stack traces
  var fn = compile(code, this.id, "require", "module", "exports");
  var ret = fn(this.require.bind(this), this, this.exports);
  if (ret !== undefined) { this.exports = ret; }
};
```

The only thing I can't prototype from pure JS is the `compile` function.  The `new Function` builtin is close, but it doesn't provide a way to specify the filename for use in stack traces.  Also some people might prefer modules to each have their own global context so a new API would be required for that as well.  Personally I prefer when each module shares the same global, but has a new unique local scope (which is why I say `new Function` is very close to what I want).

I'm not sure the best way to implement this.  I looked at the source to `new Function` and it uses some APIs not available to me such as `duk_js_push_closure`.

[...]

PR #796, opened 2016-05-20
Add a Node.js-like module loader extra

fatcerberus:
A module loading framework and `require()` implementation based on Node.js.  Semantics match Node.js wherever possible and most features of the Node module system are provided.

https://nodejs.org/api/modules.html

Features:
* Application provides callbacks to perform module resolution and loading
* Module caching based on resolved module ID
* Tolerates circular `require()` chains
* Supports `module.exports`, `module.loaded`, and `module.require`
* Supports `require.cache`
* Supports `__filename` (but not `__dirname`)

Checklist:
- [x] Expose `duk__push_require_function()` to calling code? => `duk_module_node_init()` registers a global `require`, so not needed
- [x] Manual tests to ensure correctness of all features such as `module.loaded`, `require.cache`, etc.
- [x] Full documentation, in particular for callbacks
- [x] More/better sanity checks/guards around callback invocations? => callbacks are now Duktape/C functions
- [x] Rename extra to `module-node`
- [x] Match Node.js module ID semantics (`module.id === module.filename`)

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1669,6 +1669,9 @@ Planned
 * Add an extra module (extras/console) providing a minimal 'console' binding
   (GH-767)
 
+* Add an extra module (extras/module-node) providing a Node.js-like module
+  loading framework supporting require.cache, module.loaded, etc.
+
 * Add an extra module (extras/minimal-printf) providing minimal,
   Duktape-optimized sprintf(), snprintf(), vsnprintf(), and sscanf()
   implementations; the extra compiles to less than 1kB of code which is
--- a/extras/module-node/README.rst
+++ b/extras/module-node/README.rst
@@ -1,0 +1,84 @@
+=====================================
+Node.js-like module loading framework
+=====================================
+
+This directory contains an example module resolution and loading framework and
+``require`` implementation based on the Node.js module system:
+
+* https://nodejs.org/api/modules.html
+
+The application needs only to provide the module resolution and loading logic:
+
+* Add ``duk_module_node.c`` to list of C sources to compile.
+
+* Ensure ``duk_module_node.h`` is in the include path.
+
+* Include the extra header in calling code and initialize the bindings::
+
+      #include "duktape.h"
+      #include "duk_module_node.h"
+
+      /* After initializing the Duktape heap or when creating a new
+       * thread with a new global environment:
+       */
+      duk_push_object(ctx);
+      duk_push_c_function(ctx, cb_resolve_module, DUK_VARARGS);
+      duk_put_prop_string(ctx, -2, "resolve");
+      duk_push_c_function(ctx, cb_load_module, DUK_VARARGS);
+      duk_put_prop_string(ctx, -2, "load");
+      duk_module_node_init(ctx);
+
+  Do not call ``duk_module_node_init()`` more than once for the same global
+  environment.  Doing so is undefined behavior and may put the module system
+  in an inconsistent state.
+  
+  It is possible to replace the callbacks after initialization by setting the
+  following internal properties on the global stash:
+  
+  - `\xffmodResolve`
+  
+  - `\xffmodLoad`
+
+* The resolve callback is a Duktape/C function which takes the string passed
+  to ``require()`` and resolves it to a canonical module ID (for Node.js this
+  is usually the fully resolved filename of the module)::
+
+      duk_ret_t cb_resolve_module(duk_context *ctx) {
+          /*
+           *  Entry stack: [ requested_id parent_id ]
+           */
+
+          const char *requested_id = duk_get_string(ctx, 0);
+          const char *parent_id = duk_get_string(ctx, 1);  /* calling module */
+          const char *resolved_id;
+
+          /* Arrive at the canonical module ID somehow. */
+
+          duk_push_string(ctx, resolved_id);
+          return 1;  /*nrets*/
+      }
+
+  If the module ID cannot be resolved, the resolve callback should throw an
+  error, which will propagate out of the ``require()`` call.  Note also that
+  when the global ``require`` is called, the parent ID is an empty string.
+
+* The load callback is a Duktape/C function which takes the resolved module ID
+  and either returns the Ecmascript source code for the module, or populates
+  ``module.exports`` itself and returns undefined (useful for C modules)::
+
+      duk_ret_t cb_load_module(duk_context *ctx) {
+          /*
+           *  Entry stack: [ resolved_id exports module ]
+           */
+
+          /* Arrive at the JS source code for the module somehow. */
+
+          duk_push_string(ctx, module_source);
+          return 1;  /*nrets*/
+      }
+
+  As with the resolve callback, the load callback should throw an error if the
+  module cannot be loaded for any reason.
+
+* After these steps, ``require`` will be registered to the global object and
+  the module system is ready to use.
--- a/extras/module-node/duk_module_node.c
+++ b/extras/module-node/duk_module_node.c
@@ -1,0 +1,275 @@
+/*
+ *  Node.js-like module loading framework for Duktape
+ *
+ *  https://nodejs.org/api/modules.html
+ */
+
+#include "duktape.h"
+#include "duk_module_node.h"
+
+#if DUK_VERSION >= 19999
+static duk_int_t duk__eval_module_source (duk_context *ctx, void *udata);
+#else
+static duk_int_t duk__eval_module_source (duk_context *ctx);
+#endif
+static void duk__push_module_object (duk_context *ctx, const char *id);
+
+static duk_bool_t duk__get_cached_module(duk_context *ctx, const char *id) {
+	duk_push_global_stash(ctx);
+	(void) duk_get_prop_string(ctx, -1, "\xff" "requireCache");
+	if (duk_get_prop_string(ctx, -1, id)) {
+		duk_remove(ctx, -2);
+		duk_remove(ctx, -2);
+		return 1;
+	} else {
+		duk_pop_3(ctx);
+		return 0;
+	}
+}
+
+/* Place a `module` object on the top of the value stack into the require cache
+ * based on its `.id` property.  As a convenience to the caller, leave the
+ * object on top of the value stack afterwards.
+ */
+static void duk__put_cached_module(duk_context *ctx) {
+	/* [ ... module ] */
+
+	duk_push_global_stash(ctx);
+	(void) duk_get_prop_string(ctx, -1, "\xff" "requireCache");
+	duk_dup(ctx, -3);
+
+	/* [ ... module stash req_cache module ] */
+
+	(void) duk_get_prop_string(ctx, -1, "id");
+	duk_dup(ctx, -2);
+	duk_put_prop(ctx, -4);
+
+	duk_pop_3(ctx);  /* [ ... module ] */
+}
+
+static void duk__del_cached_module(duk_context *ctx, const char *id) {
+	duk_push_global_stash(ctx);
+	(void) duk_get_prop_string(ctx, -1, "\xff" "requireCache");
+	duk_del_prop_string(ctx, -1, id);
+	duk_pop_2(ctx);
+}
+
+static duk_ret_t duk__handle_require(duk_context *ctx) {
+	/*
+	 *  Value stack handling here is a bit sloppy but should be correct.
+	 *  Call handling will clean up any extra garbage for us.
+	 */
+
+	const char *id;
+	duk_ret_t nrets;
+	const char *parent_id;
+	duk_idx_t module_idx;
+	duk_idx_t stash_idx;
+	duk_int_t ret;
+
+	duk_push_global_stash(ctx);
+	stash_idx = duk_normalize_index(ctx, -1);
+
+	duk_push_current_function(ctx);
+	(void) duk_get_prop_string(ctx, -1, "\xff" "moduleId");
+	parent_id = duk_require_string(ctx, -1);
+
+	/* [ id stash require parent_id ] */
+
+	id = duk_require_string(ctx, 0);
+
+	(void) duk_get_prop_string(ctx, stash_idx, "\xff" "modResolve");
+	duk_dup(ctx, 0);   /* module ID */
+	duk_dup(ctx, -3);  /* parent ID */
+	duk_call(ctx, 2);
+
+	/* [ ... stash ... resolved_id ] */
+
+	id = duk_require_string(ctx, -1);
+
+	if (duk__get_cached_module(ctx, id)) {
+		goto have_module;  /* use the cached module */
+	}
+
+	duk__push_module_object(ctx, id);
+	duk__put_cached_module(ctx);  /* module remains on stack */
+
+	/*
+	 *  From here on out, we have to be careful not to throw.  If it can't be
+	 *  avoided, the error must be caught and the module removed from the
+	 *  require cache before rethrowing.  This allows the application to
+	 *  reattempt loading the module.
+	 */
+
+	module_idx = duk_normalize_index(ctx, -1);
+
+	/* [ ... stash ... resolved_id module ] */
+
+	(void) duk_get_prop_string(ctx, stash_idx, "\xff" "modLoad");
+	duk_dup(ctx, -3);  /* resolved ID */
+	(void) duk_get_prop_string(ctx, module_idx, "exports");
+	duk_dup(ctx, module_idx);
+	ret = duk_pcall(ctx, 3);
+	if (ret != DUK_EXEC_SUCCESS) {
+		duk__del_cached_module(ctx, id);
+		duk_throw(ctx);  /* rethrow */
+	}
+
+	if (duk_is_string(ctx, -1)) {
+		duk_int ret;
+
+		/* [ ... module source ] */
+
+#if DUK_VERSION >= 19999
+		ret = duk_safe_call(ctx, duk__eval_module_source, NULL, 2, 1);
+#else
+		ret = duk_safe_call(ctx, duk__eval_module_source, 2, 1);
+#endif
+		if (ret != DUK_EXEC_SUCCESS) {
+			duk__del_cached_module(ctx, id);
+			duk_throw(ctx);  /* rethrow */
+		}
+	} else if (duk_is_undefined(ctx, -1)) {
+		/* nop */
+	} else {
+		duk__del_cached_module(ctx, id);
+		duk_error(ctx, DUK_ERR_API_ERROR, "error in module load callback");
+	}
+
+	/* fall through */
+
+ have_module:
+	/* [ ... module ] */
+
+	(void) duk_get_prop_string(ctx, -1, "exports");
+	return 1;
+}
+
+static void duk__push_require_function(duk_context *ctx, const char *id) {
+	duk_push_c_function(ctx, duk__handle_require, 1);
+	duk_push_string(ctx, "name");
+	duk_push_string(ctx, "require");
+	duk_def_prop(ctx, -3, DUK_DEFPROP_HAVE_VALUE);
+	duk_push_string(ctx, id);
+	duk_put_prop_string(ctx, -2, "\xff" "moduleId");
+
+	/* require.cache */
+	duk_push_global_stash(ctx);
+	(void) duk_get_prop_string(ctx, -1, "\xff" "requireCache");
+	duk_put_prop_string(ctx, -3, "cache");
+	duk_pop(ctx);
+}
+
+static void duk__push_module_object(duk_context *ctx, const char *id) {
+	duk_push_object(ctx);
+
+	/* Node.js uses the canonicalized filename of a module for both module.id
+	 * and module.filename.  We have no concept of a file system here, so just
+	 * use the module ID for both values.
+	 */
+	duk_push_string(ctx, id);
+	duk_dup(ctx, -1);
+	duk_put_prop_string(ctx, -3, "filename");
+	duk_put_prop_string(ctx, -2, "id");
+
+	/* module.exports = {} */
+	duk_push_object(ctx);
+	duk_put_prop_string(ctx, -2, "exports");
+
+	/* module.loaded = false */
+	duk_push_false(ctx);
+	duk_put_prop_string(ctx, -2, "loaded");
+
+	/* module.require */
+	duk__push_require_function(ctx, id);
+	duk_put_prop_string(ctx, -2, "require");
+}
+
+#if DUK_VERSION >= 19999
+static duk_int_t duk__eval_module_source (duk_context *ctx, void *udata) {
+#else
+static duk_int_t duk__eval_module_source (duk_context *ctx) {
+#endif
+	/*
+	 *  Stack: [ module source ]
+	 */
+
+#if DUK_VERSION >= 19999
+	(void) udata;
+#endif
+
+	/* Wrap the module code in a function expression.  This is the simplest
+	 * way to implement CommonJS closure semantics and matches the behavior of
+	 * e.g. Node.js.
+	 */
+	duk_push_string(ctx, "(function main(exports, require, module, __filename, __dirname) { ");
+	duk_dup(ctx, 1);  /* source */
+	duk_push_string(ctx, " })");
+	duk_concat(ctx, 3);
+
+	/* [ module source func_src ] */
+
+	(void) duk_get_prop_string(ctx, 0, "filename");
+	duk_compile(ctx, DUK_COMPILE_EVAL);
+	duk_call(ctx, 0);
+
+	/* [ module source func ] */
+
+	/* call the function wrapper */
+	(void) duk_get_prop_string(ctx, 0, "exports");   /* exports */
+	(void) duk_get_prop_string(ctx, 0, "require");   /* require */
+	duk_dup(ctx, 0);                                 /* module */
+	(void) duk_get_prop_string(ctx, 0, "filename");  /* __filename */
+	duk_push_undefined(ctx);                         /* __dirname */
+	duk_call(ctx, 5);
+
+	/* module.loaded = true */
+	duk_push_true(ctx);
+	duk_put_prop_string(ctx, 0, "loaded");
+
+	/* [ module source retval ] */
+
+	duk_pop_2(ctx);
+
+	/* [ module ] */
+
+	return 1;
+}
+
+void duk_module_node_init(duk_context *ctx) {
+	/*
+	 *  Stack: [ ... options ] => [ ... ]
+	 */
+
+	duk_idx_t options_idx;
+
+	duk_require_object_coercible(ctx, -1);
+	options_idx = duk_require_normalize_index(ctx, -1);
+
+	/* initialize the require cache to a fresh object */
+	duk_push_global_stash(ctx);
+	duk_push_object(ctx);
+	duk_put_prop_string(ctx, -2, "\xff" "requireCache");
+	duk_pop(ctx);
+
+	/* stash callbacks for later use */
+	duk_push_global_stash(ctx);
+	duk_get_prop_string(ctx, options_idx, "resolve");
+	duk_require_function(ctx, -1);
+	duk_put_prop_string(ctx, -2, "\xff" "modResolve");
+	duk_get_prop_string(ctx, options_idx, "load");
+	duk_require_function(ctx, -1);
+	duk_put_prop_string(ctx, -2, "\xff" "modLoad");
+	duk_pop(ctx);
+
+	/* register `require` as a global function */
+	duk_push_global_object(ctx);
+	duk_push_string(ctx, "require");
+	duk__push_require_function(ctx, "");
+	duk_def_prop(ctx, -3, DUK_DEFPROP_HAVE_VALUE |
+	                      DUK_DEFPROP_SET_WRITABLE |
+	                      DUK_DEFPROP_SET_CONFIGURABLE);
+	duk_pop(ctx);
+
+	duk_pop(ctx);  /* pop argument */
+}
--- a/extras/module-node/test.c
+++ b/extras/module-node/test.c
@@ -1,0 +1,76 @@
+#include <stdio.h>
+#include <string.h>
+#include "duktape.h"
+#include "duk_module_node.h"
+
+static duk_ret_t cb_resolve_module(duk_context *ctx) {
+	const char *module_id;
+	const char *parent_id;
+
+	module_id = duk_require_string(ctx, 0);
+	parent_id = duk_require_string(ctx, 1);
+
+	duk_push_sprintf(ctx, "%s.js", module_id);
+	printf("resolve_cb: id:'%s', parent-id:'%s', resolve-to:'%s'\n",
+		module_id, parent_id, duk_get_string(ctx, -1));
+
+	return 1;
+}
+
+static duk_ret_t cb_load_module(duk_context *ctx) {
+	const char *filename;
+	const char *module_id;
+
+	module_id = duk_require_string(ctx, 0);
+	duk_get_prop_string(ctx, 2, "filename");
+	filename = duk_require_string(ctx, -1);
+
+	printf("load_cb: id:'%s', filename:'%s'\n", module_id, filename);
+
+	if (strcmp(module_id, "pig.js") == 0) {
+		duk_push_sprintf(ctx, "module.exports = 'you\\'re about to get eaten by %s';",
+			module_id);
+	} else if (strcmp(module_id, "cow.js") == 0) {
+		duk_push_string(ctx, "module.exports = require('pig');");
+	} else if (strcmp(module_id, "ape.js") == 0) {
+		duk_push_string(ctx, "module.exports = { module: module, __filename: __filename, wasLoaded: module.loaded };");
+	} else {
+		duk_push_string(ctx, "module.exports = undefined;");
+	}
+
+	return 1;
+}
+
+static duk_ret_t handle_print(duk_context *ctx) {
+	printf("%s\n", duk_safe_to_string(ctx, 0));
+	return 0;
+}
+
+int main(int argc, char *argv[]) {
+	duk_context *ctx;
+	int i;
+
+	ctx = duk_create_heap_default();
+	if (!ctx) {
+		return 1;
+	}
+
+	duk_push_object(ctx);
+	duk_push_c_function(ctx, cb_resolve_module, DUK_VARARGS);
+	duk_put_prop_string(ctx, -2, "resolve");
+	duk_push_c_function(ctx, cb_load_module, DUK_VARARGS);
+	duk_put_prop_string(ctx, -2, "load");
+	duk_module_node_init(ctx);
+	printf("top after init: %ld\n", (long) duk_get_top(ctx));
+
+	for (i = 1; i < argc; i++) {
+		printf("Evaling: %s\n", argv[i]);
+		(void) duk_peval_string(ctx, argv[i]);
+		printf("--> %s\n", duk_safe_to_string(ctx, -1));
+		duk_pop(ctx);
+	}
+
+	printf("Done\n\n");
+	duk_destroy_heap(ctx);
+	return 0;
+}
--- a/extras/module-node/Makefile
+++ b/extras/module-node/Makefile
@@ -1,0 +1,16 @@
+# For manual testing; say 'make' in extras/module-node and run ./test.
+
+.PHONY: test
+test:
+	gcc -o $@ -I../../src/ -I. ../../src/duktape.c duk_module_node.c test.c -lm
+	@printf '\n'
+	./test 'typeof require("pig") === "string" ? "require() OK" : "require() FAILED";'
+	./test 'require("cow").indexOf("pig") !== -1 ? "nested require() OK" : "nested require() FAILED";'
+	./test 'var ape1 = require("ape"); var ape2 = require("ape"); ape1 === ape2 ? "caching OK" : "caching FAILED";'
+	./test 'var ape1 = require("ape"); var inCache = "ape.js" in require.cache; delete require.cache["ape.js"]; var ape2 = require("ape"); inCache && ape2 !== ape1 ? "require.cache OK" : "require.cache FAILED";'
+	./test 'var ape = require("ape"); typeof ape.module.require === "function" ? "module.require OK" : "module.require FAILED";'
+	./test 'var ape = require("ape"); ape.module.exports === ape ? "module.exports OK" : "module.exports FAILED";'
+	./test 'var ape = require("ape"); ape.module.id === "ape.js" && ape.module.id === ape.module.filename ? "module.id OK" : "module.id FAILED";'
+	./test 'var ape = require("ape"); ape.module.filename === "ape.js" ? "module.filename OK" : "module.filename FAILED";'
+	./test 'var ape = require("ape"); ape.module.loaded === true && ape.wasLoaded === false ? "module.loaded OK" : "module.loaded FAILED";'
+	./test 'var ape = require("ape"); ape.__filename === "ape.js" ? "__filename OK" : "__filename FAILED";'
--- a/extras/module-node/duk_module_node.h
+++ b/extras/module-node/duk_module_node.h
@@ -1,0 +1,8 @@
+#if !defined(DUK_MODULE_NODE_H_INCLUDED)
+#define DUK_MODULE_NODE_H_INCLUDED
+
+#include "duktape.h"
+
+extern void duk_module_node_init (duk_context *ctx);
+
+#endif  /* DUK_MODULE_NODE_H_INCLUDED */
--- a/util/make_dist.py
+++ b/util/make_dist.py
@@ -159,6 +159,7 @@ def create_dist_directories(dist):
 	mkdir(os.path.join(dist, 'extras', 'console'))
 	mkdir(os.path.join(dist, 'extras', 'logging'))
 	mkdir(os.path.join(dist, 'extras', 'minimal-printf'))
+	mkdir(os.path.join(dist, 'extras', 'module-node'))
 	mkdir(os.path.join(dist, 'polyfills'))
 	#mkdir(os.path.join(dist, 'doc'))  # Empty, so omit
 	mkdir(os.path.join(dist, 'licenses'))
@@ -615,6 +616,14 @@ copy_files([
 	'Makefile',
 	'test.c'
 ], os.path.join('extras', 'minimal-printf'), os.path.join(dist, 'extras', 'minimal-printf'))
+
+copy_files([
+	'README.rst',
+	'duk_module_node.c',
+	'duk_module_node.h',
+	'Makefile',
+	'test.c'
+], os.path.join('extras', 'module-node'), os.path.join(dist, 'extras', 'module-node'))
 
 copy_files([
 	'Makefile.cmdline',


fatcerberus:
One thing that bugs me about this right now is the resolve_cb: It takes the target module ID and originating module ID on the stack (not so bad), but is expected to push the filename and resolved module ID in reply.  Returning two values from a callback this way seems awkward.  Maybe that's just me though.

load_cb is more straightforward: It takes the `module` object on the stack and is expected to push the source code in response (or return 0 if it loads `module.exports` itself)

svaarala:
The extra could maybe be renamed `module-xxx` as it's likely there will be several loader examples.

One of the loaders should match Duktape 1.x as closely as possible IMO (same is true for the other extras to ease migration). Another that would be useful would support DLL loading for Linux, Windows, OS X.

fatcerberus:
That's fine, so I can rename this one to `module-node` since it's meant to emulate Node.js semantics.  In that case I could probably remove the filename/module ID distinction since in Node.js `module.id === module.filename`.

Agree on a Duktape 1.x compatible loader, that's something for a separate pull. :)

fatcerberus:
Let me know if there are any problems with the code.  There might be sanity checks missing, I get overwhelmed after a while shuffling the value stack around. :)

saghul:
This is awesome, thanks for sharing!

fatcerberus:
@saghul No problem, I get a lot of mileage out of Duktape so I like to give back when I can. :)

svaarala:
`module-node` sounds good to me if it's indeed replicating Node.js semantics - something I think many users will find useful.

The main thing w.r.t. merging is that I'm not yet sure what form the 2.x module changes will concretely take, I'll need to think about the ES6 stuff a bit first. On the other hand, whatever changes will then be needed should be relatively minor, so maybe it's better to merge and look at these changes later.

A smaller consideration is the question: what should be distributed as extras? In general, stuff that can be added to an application can be part of the Duktape distributable or simply come from an external repo. Some thoughts on that:

- The extra is part of the release and is maintained with the release (and the release cycle: extras should be fixed for a new release etc, unless they are no longer applicable).
- They should also, ideally, have some form of version compatibility which tracks Duktape version semantics at least in most cases (otherwise they're not really part of the release cycle). In other words, if you're using `extras/xyz` and you upgrade Duktape to a new minor version, there should be a high probability the updated extra would work (source compatibility).
- The functionality provided is quite close to the core Duktape functionality. For example, it might not make sense to bundle something like DukLuv with the release because it goes way beyond core functionality. Module loading, on the other hand, is quite close to the core functionality.
- Right now extras have very sparse testing coverage, but my intention is that both examples/ and extras/ would have test coverage for CI.

Not really 100% clear guidelines, and there's also a distinction between examples/ and extras/ (examples not being "production quality") which is somewhat artificial. Doing changes here before 2.0 release may be useful. For example, maybe all extras could come from external repos and simply be included in the dist phase? That way the issue tracking related to them would be kept separate.

fatcerberus:
Yeah, that's my intent--to replicate Node semantics as far as possible.  Doing so requires two callbacks because Node.js maps concrete file names to module IDs, so it's up to the application to deal with the file system.

I agree re: versioning.  The readme for extras says that semantic versioning doesn't apply to them, but it's still nice to try to maintain compatibility where possible.  In practice that shouldn't be too difficult since Duktape itself is source-compatible between minor releases.

Note that this does still need some cleanup before you go merging it. :)

svaarala:
@fatcerberus Right, the extras don't strictly follow the versioning because that might be pretty difficult for release management. But even so, there should IMO be a pretty high expectation they're changed with versioning in mind. Otherwise it's somewhat confusing that they're part of the versioned release cycle IMO.

svaarala:
Added a separate issue for the extras/examples stuff: https://github.com/svaarala/duktape/issues/797.

fatcerberus:
I totally agree.  Like I said Duktape being source-compatible between minor versions helps a lot here. :)

fatcerberus:
Getting back to this pull in particular: The callbacks seem fragile.  They are called directly like the debugger callbacks and can do evil things to the value stack, such as popping off the `module` object under construction.

Would it be worthwhile to add sanity checks after the callback invocations to verify stack shape, or leave it as-is under the assumption that the callbacks are trusted since they are C functions?

fatcerberus:
I added a basic test harness.  Currently it tests for:

* Whether the ID arrived at by resolve_cb makes it into load_cb
* Proper caching behavior
* Correct behavior of `require()` from both global code and inside a module

Ideally the test should be more comprehensive, but I'm happy to report the basic tests described above pass. :-)

fatcerberus:
Thinking about the callbacks some more, it might be nice if `duk_safe_call()` gave the called function its own stack frame.  This would be useful to users as well as internally for callbacks: For example the debugger could call the app request callback isolating it from the rest of the value stack.

fatcerberus:
Alright, I improved the tests.  Coverage should be pretty comprehensive now.  This is merge ready for my part, pending code review of course.

svaarala:
@fatcerberus Cool, I'll take a look at this as soon as possible. One thing from a quick glance is the file/function naming: it would be good if the file and function names matched the directory name, i.e. duk_module_node.c, duk_module_node_init(), etc. For example duk_module_init() is rather generic and it's not apparent from an init call site which loader variant is being initialized.

fatcerberus:
Hm, looking over the Node Module API again, I could probably also implement support for `module.children` and `module.parent` easily enough, I'll look into that.

fatcerberus:
I decided not to implement `module.children` and `module.parent` as doing so is not trivial (`require()` tracks the parent ID, but not the `module` object) and these seem like pretty niche features.  The feature set implemented here is plenty for a Node-like loader, I think.

I did go ahead and change the init function's name to `duk_module_node_init()` though.

extras/module-node/README.rst
```diff
@@ -0,0 +1,54 @@
...
+       */
+      duk_module_node_init(ctx, resolve_cb, load_cb);
+
+* The resolve callback looks like this::
+
+      duk_ret_t cb_resolve_module(duk_context *ctx) {
+          const char *parent_id = duk_get_string(ctx, -2);  /* calling module ID */
+          const char *module_id = duk_get_string(ctx, -1);
+
+          /* Arrive at the canonical module ID in whatever way makes sense */
```
svaarala: Trivia: comments starting with capitals should end in a period (sentences).

extras/module-node/README.rst
```diff
@@ -0,0 +1,54 @@
...
+
+  Important note: When ``require()`` is called outside of any module, the
+  parent ID will be an empty string.
+
+* The load callback looks like this::
+
+      duk_ret_t cb_load_module(duk_context *ctx) {
+          /* `module` object on stack top, `module.id` is the resolved ID */
+
+          /* Retrieve the module's JS source code */
```
svaarala: Trivia: period.

extras/module-node/README.rst
```diff
@@ -0,0 +1,54 @@
...
+
+          duk_push_string(ctx, source_string);
+          return 1; /*nrets*/
```
svaarala: Trivia: typically there are two spaces between a semicolon and a same-line comment (just for some visual separation).

extras/module-node/test.c
```diff
@@ -0,0 +1,68 @@
...
+
+	duk_push_sprintf(ctx, "/hungry/%s.js", module_id);
+	printf("resolve_cb: id:'%s', parent-id:'%s', resolve-to:'%s'\n",
+		module_id, parent_id, duk_get_string(ctx, -1));
+
+	return 1;
+}
+
+static duk_ret_t cb_load_module(duk_context *ctx) {
+	const char* filename;
```
svaarala: Trivia: the star placement is `const char *xxx` in Duktape sources.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
+/*
+ * Node.js-like require() and module loader
+ *
+ * https://nodejs.org/api/modules.html
+ */
+
```
svaarala: Trivia: there are two basic comment styles in the Duktape code base. First style is "compact":

    /* Foo bar.
     * Quux.
     */

In this style there's a single space between the star and the text. The second style is a "banner":

    /*
     *  Two spaces.
     */

Here there are two spaces before the text. This is pretty arbitrary but that's the current convention in any case. The basic intent is to visualize separate the banner as a block separate, and the compact comment as applying to the next function, variable, etc (and there's usually no empty line between the compact comment and the function/variable).

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+static duk_int_t duk__eval_module_helper (duk_context *ctx);
+static void duk__push_module_object (duk_context *ctx, const char *id);
+
+static duk_resolve_module_function resolve_module_cb;
+static duk_load_module_function load_module_cb;
+
+static duk_bool_t duk__get_cached_module(duk_context *ctx, const char *id) {
+	duk_push_global_stash(ctx);
+	duk_get_prop_string(ctx, -1, "\xffModuleCache");
```
svaarala: The internal key here (and in other places) follows the format reserved for Duktape internals. Extras are maybe a bit of a grey area in this respect, so I'm not sure what'd be best.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+	if (duk_get_prop_string(ctx, -1, id)) {
+		duk_remove(ctx, -2);
+		duk_remove(ctx, -2);
+		return 1;
+	}
+	else {
```
svaarala: Trivia: else-style:

    } else {

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+		duk_pop_3(ctx);
+		return 0;
+	}
+}
+
+/*
+ * NOTE: duk__put_cached_module() leaves the module object on the
+ * stack as a convenience to the caller.
+ */
+static void duk__put_cached_module(duk_context *ctx) {
```
svaarala: Trivia: block comment style (two spaces).

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+	/* ... module */
+
```
svaarala: Trivia (very): in representing value stack shape I've usually used brackets to make the top and bottom explicit:

    [ foo bar quux]

and when the bottom is not known or isn't interesting:

    [ ... bar quux ]

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+	id = duk_require_string(ctx, 0);
+
+	duk_push_string(ctx, parent_id);
+	duk_push_string(ctx, id);
+	nrets = resolve_module_cb(ctx);
+	if (nrets > 1) {
+		/* Pop off extraneous return values. */
+		duk_pop_n(ctx, nrets - 1);
+	}
```
svaarala: This seems a bit awkward (I guess it's related to the safe call issue). Couldn't the resolve function simply be trusted to leave the stack in a certain shape (i.e. never return > 1). Or maybe throw an error for `nrets > 1`?

The popping behavior seems a bit odd because it doesn't actually match Duktape/C behavior because the topmost value is discarded.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+	 *  clean up the extra stack garbage for us.
+	 */
+
+	const char *id;
+	duk_ret_t nrets;
+	const char *parent_id = NULL;
+	duk_idx_t module_idx;
+
+	duk_push_current_function(ctx);
+	duk_get_prop_string(ctx, -1, "\xffModuleID");
```
svaarala: Cast to void to signal intentional return value ignore:

     (void) duk_get_prop_string(...)

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+	parent_id = duk_get_string(ctx, -1);
+
```
svaarala: Since this is not performance critical code, maybe use `duk_require_string()` here. It will cheaply ensure `parent_id` will be non-NULL without any further checks.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+	/* id require parent_id */
+
+	id = duk_require_string(ctx, 0);
+
+	duk_push_string(ctx, parent_id);
+	duk_push_string(ctx, id);
```
svaarala: Here maybe just use `duk_dup()`? Using `duk_push_string()` causes an unnecessary re-interning of the string already on the value stack?

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+		duk_error(ctx, DUK_ERR_API_ERROR, "error in module resolution");
+	}
+
+	/* ... resolved_id */
+
+	id = duk_get_string(ctx, -1);
+	if (duk__get_cached_module(ctx, id)) {
+		goto have_module;  /* use cached module */
+	}
+	else {
```
svaarala: Trivia: else style.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+		module_idx = duk_normalize_index(ctx, -1);
+
+		/* ... module */
+
+		nrets = load_module_cb(ctx);
+		if (nrets > 1) {
+			/* Pop off extraneous return values. */
+			duk_pop_n(ctx, nrets - 1);
+		}
+		if (nrets < 0 || (nrets > 0 && !duk_is_string(ctx, -1))) {
```
svaarala: Same nrets issue as above.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+			duk_error(ctx, DUK_ERR_API_ERROR, "error in module loading");
```
svaarala: Should there be a duk__del_cached_module() here too?

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+	 *  Stack: ... module source --> ... module (retval == 0)
+	 *         ... module source --> ... err    (retval != 0)
+	 */
+
+	/*
+	 * Synthesize a function to wrap the module code.  This is the easiest way
+	 * to implement CommonJS semantics and matches the behavior of Node.js.
+	 * The wrapper has no line breaks so it won't interfere with line
+	 * numbering.
+	 */
```
svaarala: Trivia: block comment style.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+	duk_pop(ctx);
+
+	/* module */
+
+	duk_push_true(ctx);
+	duk_put_prop_string(ctx, -2, "loaded");
+
+	return DUK_EXEC_SUCCESS;
+
+fail:
```
svaarala: Label indent by one space.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+ */
+
+#include "duktape.h"
+#include "duk_module_node.h"
+
+static duk_int_t duk__eval_module_helper (duk_context *ctx);
+static void duk__push_module_object (duk_context *ctx, const char *id);
+
+static duk_resolve_module_function resolve_module_cb;
+static duk_load_module_function load_module_cb;
```
svaarala: Having these as file locals means that different resolve/load functions cannot be used for different Duktape global environments or different Duktape heaps?

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,248 @@
...
+
+	/* push the default `require` builtin */
+	duk_push_global_object(ctx);
+	duk_push_string(ctx, "require");
+	duk__push_require_function(ctx, "");
+
+	/* global `require` is not enumerable */
+	duk_def_prop(ctx, -3, DUK_DEFPROP_HAVE_VALUE
+		| DUK_DEFPROP_SET_WRITABLE
+		| DUK_DEFPROP_SET_CONFIGURABLE);
```
svaarala: The indent style makes the flags a bit difficult to read. The typical style for this would be to tab indent for the base level (`duk_def_prop()`) and then space indent to align the rest:

    duk_def_prop(ctx, -2, DUK_DEFPROP_HAVE_VALUE |
                          DUK_DEFPROP_SET_WRITABLE |
                          DUK_DEFPROP_SET_CONFIGURABLE);

svaarala:
@fatcerberus Ok, some high level questions / comments from review:

- Out of curiosity, how does the resolve/load process here compare to #60 and DukLuv's loader mechanism? While it's not critical for extras to align, it'd also be good to avoid differences where they don't need to exist.
- The resolve/load callbacks are now called directly without protection. Could they be called via a combination of `duk_push_c_function()` + `duk_call()` instead? They only return (at most) one return value so that should be possible I think? (This is also related to the separate discussion on changing safe call semantics.)
- Overall using custom non-Duktape calls for resolve/load seems awkward. If they remain as is, maybe a short comment on the required call stack behavior for the resolve/load callbacks.
- The loader currently stores resolve/load function references to a file local variable which won't work with multiple different loader configurations (in separate global objects or Duktape heaps). On the other hand, stashing a custom function pointer is not supported by the Duktape API now (that's actually something to maybe fix as it's OK to cast between function pointer types, but function pointer size may differ from data pointer size, #808).
- The resolve/load callbacks cannot be Ecmascript functions at present. This would be fixed if they were provided as function references which could be replaced. But I'm guessing it's intentional they must be C functions? For me it'd be more natural if they were Duktape/C functions (C or Ecmascript) and stored in the stash to isolate them from Ecmascript code.
- What's the intended behavior when used from master as it is now, i.e. with `require()` still part of the built-ins? I'd guess it'd replace the built-in module loader? This probably doesn't warrant mentioning however, as it's likely the built-in require() will disappear altogether.

Formatting trivia etc in commit comments.

svaarala:
So actually most of the high level comments are about the resolve/load functions. If they were represented as Duktape/C functions and stashed as bona fide function references in the stash, several problems would be solved at once. Would there be some concrete issue in doing so?

extras/module-node/duk_module_node.c
fatcerberus: The idea was if the function claims to return multiple values we only want the first one pushed, but I agree the behavior is odd.  I guess I could change it to be a `void`-returning function, but for me personally the `duk_ret_t` return is a convenient indicator that Duktape expects something on the value stack.

I'll change it to throw an error if `nrets != 1` I guess.

extras/module-node/duk_module_node.c
fatcerberus: Whoops, I should know better than to edit Dujtape code in MSVC, it likes to do this automatically.  I like the style but it doesn't match Duktape conventions. :)

extras/module-node/duk_module_node.c
fatcerberus: Or maybe it should be a compact comment instead?

extras/module-node/duk_module_node.c
fatcerberus: Yes, good catch.

fatcerberus:
The callbacks ended up as pure C calls because, in my first draft, I was passing and returning values directly.  I ended up changing it to just use the valstack because that made it easier to deal with strings which would otherwise need to be dynamically allocated (awkward) or stored in a static buffer (also awkward, and a potential security issue).

And yes, the static pointers were because I couldn't figure out a safe way to stash them.  As you point out, it's not safe to cast function pointers to `void*` (not if you care about portability anyway :).

It should be possible to change them to be Duktape/C calls now, since their signatures match anyway.  I agree that's the best solution here.  It would also solve the issue of dealing with extra garbage values pushed by the callbacks, since Duktape call handling would clean that up automatically.

The `require` registered in `duk_module_node_init()` _should_ override Duktape's, assuming that's registered at heap creation.  If not (e.g. if it's actually registered on setting `Duktape.modSearch`) then there might be an issue.  But again, like you say, it might just be better to move `require` out of the core entirely.  It isn't part of Ecmascript spec, and ES6 modules will need a different mechanism anyway.

Regarding overlap with #60: The resolve/load mechanism was indeed inspired by that implementation, however this also provides several features from Node (like `module.loaded`) which aren't supported by the minimal system built into Duktape, so I think it justifies its existence.  One change that might be worth making (perhaps in a future pull) is to expose the callbacks to Ecmascript, i.e. `Duktape.modResolve` and `.modResolve`.

Anyway, I'll go through the code and fix the trivia first, then we'll see about maybe improving the semantics before the merge.

fatcerberus:
One issue with `module.loaded` handling is that if a module does `Object.freeze(module)` then the loader will throw an error when trying to set `.loaded` to true.  Node also throws an error in this circumstance though, so I think it's okay.  Maybe consider it a failed load and delete the module from the cache before rethrowing?

Anyway I did the mentioned fixes and made some enhancements, but I have no internet access right now due to a Comcast outage, so I'll push the changes when I can.  For now I need sleep. :)

extras/module-node/duk_module_node.c
saghul: Side question: is there a way to attach stuff to a Duktape heap? Like some user data `void*` for instance.

svaarala:
Re: #60, one important question is whether this pull addresses/implements the proposed design. It doesn't have to, but I'd like to understand what else is needed. (I'll go through the module loading issues as soon as I can; there are many pending issues and ideas to address; an extra with DLL loading support is one.)

fatcerberus:
The proposed design was to split `Duktape.modSearch` into modResolve and modLoad, in which case yes, this extra implements that--with the caveat that the callbacks are set from C once on initialization and are inaccessible from Ecmascript.

Honestly I think that makes more sense - if you're going to implement your module loader in JS anyway you might as well go the whole hog and replace `require` too.

fatcerberus:
One thing this doesn't have that the Dukluv implementation does is a concept of a "compile" callback, i.e. `Duktape.modCompile`.  I didn't implement that here, mainly because I couldn't figure out what problem it was supposed to solve. :)

svaarala:
@fatcerberus There are many cases where it makes quite a lot of sense for modResolve / modLoad to be implemented in Ecmascript (calling into C for low level operations like looking for DLLs, or loading stuff over the network), so being able to do that would be nice. Some applications are written with the style where all logic and state handling is in Ecmascript side and C calls are fully stateless helpers only, and don't handle stuff like search paths, etc.

However, being able to register the callbacks from Ecmascript code is a different issue, and I agree it might be best if that wasn't possible by default. For example, the handlers could be pushed on the value stack before calling `duk_module_node_init()` and then be registered into the global stash for later lookup. Ecmascript code couldn't access them directly, but they could still be implemented in Ecmascript.

fatcerberus:
Oh, I don't doubt the utility of implementing the callbacks in ES--I just meant that if you're already doing that, it's probably cleaner at that point just to replace `require()`.

fatcerberus:
@svaarala I pushed the fixes.

svaarala:
Sure you could write another provider. But what if you wanted Node.js module semantics?

It seems counterproductive for me if Duktape extras don't support writing stuff in Ecmascript :-)

svaarala:
As a general design point, callbacks/providers are represented as Duktape/C functions wherever possible: they have existing call mechanics, and can be implemented as Duktape/C functions, lightfuncs, or Ecmascript functions. The exceptions to this basic principle are callbacks which are closely tied to internals so that the side effects of call handling are not desirable - for example, debugger transport callbacks.

fatcerberus:
@svaarala Aw, crud.  I didn't have internet access so I couldn't get the new code into my Ubuntu VM.  It seems the current code is broken.  I'll fix it and re-push.

svaarala:
By the way, it might seem registering callbacks as arbitrary Duktape functions seems heavy but that's actually not the case. If you store a lightfunc property in the global stash for example, the cost of that is 25 bytes on a 64-bit platform and 13 bytes on a 32-bit platform. There's no Ecmascript object to represent the lightfunc; all the lightfunc information needed for a call are in the `duk_tval`. The cost is actually the same as storing a plain pointer.

So, unless the call mechanics themselves are undesirable, it's hard to justify not using a normal function call :-)

fatcerberus:
Well, it isn't so much the footprint that bothers me, but the fact that a callback which might be confined to C for sandboxing reasons can nonetheless show up in an Ecmascript stack trace.

fatcerberus:
Okay, I fixed it.  All the tests are okay, except for `require.cache`.  That test is failing, and I can't figure out why.  To elaborate, the full test is:

* `require()` a module
* Check whether it's visible in `require.cache`
* `delete require.cache[id];`
* `require()` the module again
* Check whether `mod1 !== mod2`

For some reason the last check is failing and I have no idea what's causing it.  The load callback is clearly being called twice.  Maybe you can figure it out.

svaarala:
Hmm, what's the harm of showing up in a stack trace? Are you worried about the .fileName/.lineNumber blaming or just in general C filenames showing up in stack traces? The blaming behavior can be fixed for 2.x.

svaarala:
Regarding the call style etc, I'm not actively trying to be difficult by the way :) I'm trying to figure out what kind of stuff is useful to be distributed as extras, and one criterion for that (for me) is that the extras are as generally useful as possible and play well with concepts native to Duktape. That doesn't mean other kinds of modules wouldn't be useful, just that they might not be ideal to be distributed with Duktape. Providing wiki links to useful stuff like module loaders, logging frameworks, etc, is another way to promote useful modules.

fatcerberus:
> Hmm, what's the harm of showing up in a stack trace?

To be honest, mostly just OCD on my part.  It just seems odd to have a function show up in the ES stack that isn't actually visible to Ecmascript in any way.  Of course if we change it to make the callbacks accessible to Ecmascript then that's a moot point. :)  That said, I did convert the callbacks to Duktape/C calls since it makes for more sane value stack semantics (the functions can use positive indices to access parameters).

The blaming is potentially an issue though - ideally an error thrown by one of the callbacks (assuming they are only accessible from C) would be blamed on the function calling `require()`.

Regarding inclusion: I think it does make sense to have an example module loader as part of the Duktape dist, especially if built-in `require()` is going to be axed (in which case we'll need a Duktape 1.x polyfill as well).  And since Node.js is the most well-known module system, it's worthwhile to emulate its semantics I think.

svaarala:
There's going to be at least a Duktape 1.x compatible module loader in the extras for migration (it will be minimal and match Duktape 1.x conventions, e.g. using `Duktape.modLoaded[]` etc). And there's been a lot of requests for a default loader supporting a search path and DLL loading (for DLLs which match the recommended "module conventions") for major platforms. DLL support would provide an easy default for module sharing; it's possible now but there's somewhat of a barrier without a DLL-capable loader in the distributable.

fatcerberus:
I can implement the Duktape 1.x loader (in a separate pull), but I'd need your help to make sure all the semantics are correct. :)

svaarala:
I was more or less going to just copy the existing `require()` implementation from duk_bi_global.c into an extras directory, and add a `DUK_CMDLINE_MODULE_SUPPORT` define to build it into "duk". The existing testcases would then have pretty good coverage for the loader.

fatcerberus:
Anyway, I rebased and squashed, waiting on the next round of review.

P.S. There have been a ton of `-Wshadow` warnings showing up in recent Travis runs.  Just thought I'd give you a heads-up in case you weren't aware.

svaarala:
Thanks, I wasn't aware: all of them seem to be:

    duk_error_macros.c:44:118: warning: declaration of ‘index’ shadows a global declaration [-Wshadow]

I don't get that locally which is probably a GCC version thing (`index` is a function provided by strings.h). It's probably best if I rename all `index` variables to something else.

fatcerberus:
I fixed the `require.cache` test.  It was a bug in the test: `pig` exports a string, which is compared by value, not by reference.  All the test cases pass now.

svaarala:
Ok, I'll give this a read-through and testing tomorrow. I'll compare against @creationix's proposal too to see if every issue there was addressed (or if not, note it for future reference).

fatcerberus:
One thing I'd also like to compare against is the property descriptors used by Node.js for e.g. `require.cache`.  Here everything is fully E/W/C (except the global `require` which is non-enumerable to match other built-ins).

svaarala:
What might matter is enumerability, and that matters mostly if there's a chance enumerable properties would somehow accidentally make their way into a for-in loop or JSON serialization. For `require.cache`, for example, that would probably not be a concern. Matching Node.js is still a good idea I agree.

svaarala:
For documentation it'd be useful to list the features not supported if that's an easy list to come by. I can think of:

- Loading `.json` and `.node` files (the latter being non-applicable anyway)
- The search path isn't provided out of the box, as it's offloaded to the resolve function
- Other stuff?

README.rst notes:

- Provide an explicit link to the Node.js modules page, maybe identify what version the extra tries to mimic?
- Link to wiki page, http://wiki.duktape.org/HowtoNodejsModules.html (?) for detailed step-by-step getting started stuff (pretty essential from practical experience :-)

Other useful documentation issues as a Wiki pull because it's "how to" stuff: https://github.com/svaarala/duktape-wiki/issues/124.

svaarala:
And maybe noting at least the ROM footprint ballpark would be nice - I can check that when I test this out.

svaarala:
Couple of dist-related issues:

- The extra should be added to `util/make_dist.py`
- The extras/module-node/Makefile test should except paths available in the dist package (it now assumes it's being run in the Duktape repo); pretty arbitrary, but that's how the test Makefiles in other extras are set up

extras/module-node/README.rst
```diff
@@ -0,0 +1,62 @@
...
+      #include "duktape.h"
+      #include "duk_module_node.h"
+
+      /* After initializing the Duktape heap or when creating a new
+       * thread with a new global environment:
+       */
+      duk_module_node_init(ctx, resolve_cb, load_cb);
+
+* The resolve callback looks like this::
+
```
svaarala: Here it might be good to say "The resolve callback is a Duktape/C function which [... does what ...], for example:" or something along these lines: it's good to say the function is a standard Duktape/C function so that the argument/return value semantics are clear, and summarizing the purpose ("... resolves a requested ID into a canonical module ID ...") is nice for the reader.

extras/module-node/README.rst
```diff
@@ -0,0 +1,62 @@
...
+          duk_push_string(ctx, resolved_id);
+          return 1;  /*nrets*/
+      }
+
+  Important note: When ``require()`` is called outside of any module, the
+  parent ID will be an empty string.
+
+* The load callback looks like this::
+
```
svaarala: Same here, mention that it's a Duktape/C function.

Maybe mention somewhere that both resolve and load can be Ecmascript functions too?

extras/module-node/README.rst
```diff
@@ -0,0 +1,62 @@
...
+* Include the extra header in calling code and initialize the bindings::
+
+      #include "duktape.h"
+      #include "duk_module_node.h"
+
+      /* After initializing the Duktape heap or when creating a new
+       * thread with a new global environment:
+       */
+      duk_module_node_init(ctx, resolve_cb, load_cb);
```
svaarala: If this convention was changed so that `duk_module_node_init()` would expect that callbacks in the value stack, it would then support Ecmascript resolve/load functions. If you want to make versioning easier, the value stack argument could also be an initialization parameter object providing the callbacks which would then be easier to extend:

    {
        "resolve": ...,
        "load": ...
    }

fatcerberus:
Technically this could be used to load `.json` files since the load callback doesn't need to return source code - it can assign to `module.exports` itself.

As for the Makefile, I originally copied the Makefile from another extra and actually assumed it was broken because it didn't work from the repo - I thought the repo and dist used to same relative paths--forgot that you can't "make dist" with the dist. :)

svaarala:
Ok, for `.json` files it's maybe a topic to cover in the wiki page: it's not directly supported because of the abstraction but you can implement it within the load callback.

Some extras used to have paths like `../../dist/src/` but I think I've fixed them all in master to be compatible with the dist layout instead.

extras/module-node/README.rst
fatcerberus: Assuming that either (or both) of #575 or the format string-based `duk_put_props()` will eventually make it into master, then I'm fine with using an object.

fatcerberus:
Yeah, I'll fix the path, and also see about adding to `make_dist.py`.  No promises on that latter one, Python isn't really my forte. :)

svaarala:
Regarding the init argument, I can symphatize with trying to avoid the rather verbose object parameter - on the other hand, it makes versioning much easier to deal with, and a fully featured loader like this will quite likely need additional flags or parameters going forward. It'd be annoying if that also meant a change in the value stack behavior or C signature - both violating source compatibility which would then potentially prevent making changes for a year or so, or would necessitate introducing multiple C initializer functions which then need deprecating etc. The cost of a verbose init sequence is far smaller in my view :)

Most extras don't do this for the inits now which is fine if there are no foreseeable changes needing extra parameters. I did add a placeholder flags field to the C signature of most init functions just to ensure there'd be no versioning issues in adding some trivial behavioral flags (for example, one may want to control flushing behavior of the logging framework or console.log writes). This isn't of course a perfect solution because the unforeseen thing may be a value or a function, etc. But versioning is more a matter of avoiding most pitfalls to minimize effort rather than being 100% right.

svaarala:
If you want I can make the dist changes too.

svaarala:
Also a releases entry is needed, I can add that separately if you wish.

fatcerberus:
While I find it annoying in general to work with objects on the value stack, in this case I'm okay with it because it only needs to be done once.

I think my main issue with how things are right now is that it's difficult when scanning value stack code to tell if you've balanced the stack properly.  Verifying correctness often requires reading over the code 4-5 times (this was the case for me with `duk__eval_module_source()` in this extra).  For Duktape/C functions it's not as important since call handling will pop any garbage values for you, but elsewhere you have to be very careful to make sure you balance the stack properly--which often requires keeping the Duktape API documentation open at the same time to check stack signatures.

If the situation is improved so that correctness is clear just from reading the code without needing to cross reference, I'd be much more receptive to Duktape APIs accepting objects on the stack.

fatcerberus:
Should `duk_module_node_init()` throw if the object doesn't have "resolve" and "load" properties (or they are not functions)?

svaarala:
That'd be my expectation - the other alternative would be for the initialization to finish but remain in a broken state which would be quite unintuitive. Also returning an error code basically replicates the error mechanism (and the caller will still need a protected call to catch any runtime errors which may always happen).

fatcerberus:
For future work it might good to eventually support `Duktape.modLoad`/`.modResolve` when using this extra.

Let me know what you think now.

svaarala:
Hmm, what do you mean with `Duktape.modLoad` and `Duktape.modResolve`? Looking up the callbacks from there?

I'd be fine with the resolve/load functions to be initialized from C code so that they're by default not exposed to Ecmascript code. There's a native init call site anyway. What I would like, though, is for it to be OK for them to Ecmascript functions and this should now be the case.

Switching the functions after initialization might be necessary in some situations and I guess in that case having them be addressable from Ecmascript code might be useful (but also a sandboxing concern). That case would now require a C helper, and there's no defined way of replacing the functions now; user code would need to interact with the global stash directly.

That's actually another design alternative: the module loader could just assume you register the load/resolve callbacks into the global stash using some documented name. They could then be changed at any time without any need for additional APIs provided by the module loader.

svaarala:
That design alternative is basically what Duktape.modSearch() tries to do; the only difference would be that the properties would be in the stash, i.e. hidden by default from Ecmascript code.

RELEASES.rst
```diff
@@ -1669,6 +1669,9 @@ Planned
 * Add an extra module (extras/console) providing a minimal 'console' binding
   (GH-767)
 
+* Add an extra module (extras/module-node) providing a Node.js-like module
+  loading framework supporting `require.cache`, `module.loaded`, etc.
+
```
svaarala: The formatting is incorrect for RST, should be double backticks for preformatted/code. Though I've mostly avoided formatting in the RELEASES file so that it reads better as plain text.

svaarala:
The releases file and README.rst both use single backticks for code/preformatted; for RST two backticks are needed.

RELEASES.rst
fatcerberus: Alright, fixed locally. :)

fatcerberus:
Hm, interesting thought on looking up the functions from the stash.  I guess I'll leave it as-is for now and if you decide you want to implement that later you can.

By the way is `Duktape.modSearch` an accessor?  I thought I remember that there was no global `require` function until I provided a module callback.

svaarala:
The global `require()` was always present in Duktape 1.x and `Duktape.modSearch` is just a normal property that you assign to. You could also define it as an accessor, though I'm not sure what that would accomplish.

svaarala:
Hmm, so with no action there's no easy way to replace the resolve/load callbacks at runtime I guess? That might seem odd, but I've more than once worked with a setup where an initial loader is used to pull in an actual module loader, which then loads the remaining modules.

svaarala:
I'm fine with leaving things as they're now, but it's maybe worth having a pull request / issue for Duktape 2.x release to figure out if minor changes are needed to round out the functionality.

fatcerberus:
Huh, that's weird.  When I was originally dabbling with the module system I remembered `require` not being defined until I provided a modSearch callback.  Oh well, it doesn't really matter and is unrelated to this pull, just a curiosity.

Anyway yeah, it's technically possible to replace the callbacks later on by replacing `_modLoad` and `_modResolve` in the global stash, but that's kind of "undocumented". :)

svaarala:
Ok, well it might be enough to just document that convention for now.

But then, with that documented, having the init function set them for you is somewhat unnecessary: the caller could just set them before (or after) initializing the module framework.

fatcerberus:
This might be a worthwhile direction in general, using the stash to share data with Duktape internals in the same way the value stack is used for transient stuff.  We'd need to come up with a formal convention though, to ensure application code using the stash doesn't accidentally step on the Duktape values.

svaarala:
Agree, though it probably doesn't need to be very complicated. For example, Duktape internals could just be internal keys on the global stash (`\xffFoobar`) which are anyway reserved for Duktape. The module loader here is not part of internals so it'd need to "compete" with other users, but I don't think that's much of a concrete problem in practice, as long as expressive names are used.

fatcerberus:
This is what I really like about Duktape by the way, Ecmascript and C code are generally interchangeable.

So changes to make before merge are:
* Document that callbacks can be swapped out by modifying the relevant properties in the global stash
* Avoid setting callbacks in `duk_module_node_init()`
* Rename stash keys for callbacks (e.g. "moduleLoad"), avoid internal keys

Anything else?

svaarala:
I'll re-read the current state in around 3-4 hours so maybe something comes to mind. But this is already in good enough shape for master I think. It's still useful to try to figure out how to minimize mechanisms and/or increase flexibility, etc :)

fatcerberus:
By the way it's included in `util\make_dist.py` now.

fatcerberus:
By the way, I don't know which Node.js version to say this is based on - their module API has been documented as "Locked" for quite some time. :)

svaarala:
Ok, fair enough :)

svaarala:
What happens if you re-initialize the module loader for the same global environment? Is this allowed; if so, what's the expected outcome. For example, are module caches reset.

Another question that come to mind is whether the extra works in Duktape 1.x. It's quite clear extras should work with the version they ship with, but there's no reason why they couldn't support other versions too.

The reason I ask is that I think there are Duktape 1.x users who might want to use this loader before migrating to Duktape 2.x.

fatcerberus:
This is what happens if you try to double initialize:
* A new global `require` is registered
* The resolve and load callbacks are changed to whatever you pass in
* The module cache is replaced with a fresh object.

That last thing makes it not so harmless: Modules which are already loaded will then have a stale `require.cache`.

And theoretically this should work just fine with Duktape 1.x (right now at least): I don't believe it calls any new/updated APIs. I guess this would be a candidate for merging into the 1.x maintenance branch if there is ever a Duktape 1.6.

svaarala:
Hmm, so should we just say "don't initialize twice" for a global environment and avoid providing any guarantees if that's done? If reinitialization is allowed, then the semantics matter and become a versioning concern.

Re: Duktape 1.6, I think it's quite possible such a release will happen as a final release in the 1.x branch with a minimal set of features improving usability for 1.x users (migration takes time and there's often no pressing need to do so).

So I guess we'd then try to keep this extra Duktape 1.x compatible, #ifdef'ing with DUK_VERSION if necessary. Hopefully the extra itself didn't need to be branched.

fatcerberus:
I agree, I'll add a note to `README.rst` to not initialize more than once for the same global environment.  Right now the only reason to do so is to replace the callbacks--and that can be done by modifying the stashed properties directly.

fatcerberus:
Kind of nitpicky but `duk_module_node_init` seems awkward - should we rename it to `duk_module_init_node()` instead?  I think it would read back better - i.e. "Initialize the module system using Node-like semantics"

svaarala:
That seems a bit weird to me, `duk_init_module_node()` would read more like a sentence and would keep the extra name intact (`module_node)`. The other extras use that `duk_xxx_init()` convention to ensure the module name is intact; they should be changed to match if we use a different convention here.

The C module convention uses `dukopen_foo_bar()`. I originally thought of using that convention, but because the initializers make changes to the global object etc (rather than just returning a value) that wouldn't be strictly speaking correct if they were considered modules.

Yet another option would be to model all of the extras as C modules matching the c-module-convention.rst document. In that case the `dukopen_foo_bar()` call would return an object containing methods and constants. You'd then call, say, `init()` from those, as if you'd `require('foo_bar').init()`.

But I suggest we revisit the naming and init conventions separately, as they should ideally be same or similar across the extras.

fatcerberus:
@saghul If you want to use this in sjs now, go right ahead - it's basically finished, we're just trying to hash out the little nitpicky details now. :)

svaarala:
@fatcerberus Let me know when you've pushed final changes for review :)

fatcerberus:
Hehe, sorry, I'm a perfectionist.

It should be good to go now.  I updated the test program to work with the new value stack semantics of `duk_module_node_init()` and everything still checks out.  No further changes should be needed.

saghul:
@fatcerberus Thanks mate! I'm following the discussion from the sidelines :-) I have a couple of things to iron out first, then I'll have a deep loop and integrate it.

svaarala:
@fatcerberus I'll float the idea of removing the resolve/load registration in the init function once more with a concrete example (for completeness if nothing else :-).

So this is the current init call site:
```
duk_push_object(ctx);
duk_push_c_function(ctx, cb_resolve_module, DUK_VARARGS);
duk_put_prop_string(ctx, -2, "resolve");
duk_push_c_function(ctx, cb_load_module, DUK_VARARGS);
duk_put_prop_string(ctx, -2, "load");
duk_module_node_init(ctx);
```

If the init function didn't set resolve/load but left the user to do that (which the calling code needs to do anyway if the functions are replaced), it would look like:
```
duk_push_global_stash(ctx);
duk_push_c_function(ctx, cb_resolve_module, DUK_VARARGS);
duk_put_prop_string(ctx, -2, "\xffmodResolve");
duk_push_c_function(ctx, cb_load_module, DUK_VARARGS);
duk_put_prop_string(ctx, -2, "\xffmodLoad");
duk_pop(ctx);
duk_module_node_init(ctx);
```

The difference isn't big and strips the init function of handling the resolve/load registration. But maybe we'll leave the init as it exists now, because:

- The current call site will look better with API primitives for multiple property set/get.
- There may be a need for other options besides the load/resolve functions so init arguments / options object will be needed anyway.

fatcerberus:
One thing I don't like about setting internal properties directly is that it's prone to error - if the name used starts with A-F the first character(s) will be interpreted as part of the character code (e.g. `"\xfffoo"` gets interpreted as /xfff + "oo").  I ran into that problem a few times in minisphere, so I ended up developing a convention to avoid it: `"\xff" "foo"`"

svaarala:
That's actually mentioned in the Guide too:

```
/* Create an internal string, which can then be used to read/write internal
 * properties, and can be passed on to Ecmascript code like any other string.
 * Terminating a string literal after a hex escape is safest to avoid some
 * ambiguous cases like "\xffab".
 */
duk_push_string(ctx, "\xff" "myprop");
```

extras/module-node/README.rst
```diff
@@ -0,0 +1,76 @@
...
+      duk_put_prop_string(ctx, -2, "resolve");
+      duk_push_c_function(ctx, cb_load_module, DUK_VARARGS);
+      duk_put_prop_string(ctx, -2, "load");
+      duk_module_node_init(ctx);
+
+  Do not call ``duk_module_node_init()`` more than once for the same global
+  environment.  Doing so will put the module system in an inconsistent state.
+
+* The resolve callback is a Duktape/C function which takes the string passed
+  to `require()` and resolves it to a canonical module ID (for Node.js this is
```
svaarala: Markdown quoting for require().

svaarala:
Hmm, it's probably best if I fix any of the tiny nits like RST markdown issues in a separate pull so we'll get the main functionality in master. I'll re-read the code and see if there's anything more substantial.

extras/module-node/README.rst
fatcerberus: Force of habit. :-/

svaarala: Yup, same here: it's somewhat annoying to have both Markdown and RST documents in the repo. I started with RST so I've only used Markdown for a few documents people usually expect to be Markdown (like README.md). (Though, Github does support RST for README, CONTRIBUTING, etc too.)

extras/module-node/README.rst
```diff
@@ -0,0 +1,76 @@
...
+          return 1;  /*nrets*/
+      }
+
+  If the module ID cannot be resolved, the resolve callback should throw an
+  error, which will propagate out of the ``require()`` call.  Note also that
+  when the global ``require`` is called, the parent ID is an empty string.
+
+* The load callback is a Duktape/C function which takes the resolved module ID
+  and either returns the Ecmascript source code for the module, or populates
+  ``module.exports`` itself and returns undefined (useful for C modules)::
```
svaarala: Just to be sure: it's also allowed to populate symbols into module.exports and *also* return a string (same as with Duktape 1.x)?

Also: I'm assuming it's allowed to replace `module.exports` in the load() function (as with the updated Duktape 1.x behavior)?

fatcerberus: Yes, both of those should work.  The `module` on the stack is the same one passed to the module to be loaded.

extras/module-node/test.c
```diff
@@ -0,0 +1,69 @@
...
+	module_id = duk_require_string(ctx, 0);
+	duk_get_prop_string(ctx, 2, "filename");
+	filename = duk_require_string(ctx, -1);
+
+	printf("load_cb: id:'%s', filename:'%s'\n", module_id, filename);
+	duk_push_sprintf(ctx,
+		"if (module.id == 'pig.js') module.exports = 'you\\'re about to get eaten by %s';"
+		" else if (module.id == 'cow.js') module.exports = require('pig');"
+		" else if (module.id == 'ape.js') module.exports = { module: module, __filename: __filename, wasLoaded: module.loaded, size: 0 };",
+		module_id);
```
svaarala: This has no functional impact, but wouldn't it be cleaner if `cb_load_module` contained an if-else ladder for the virtual filenames (in C) and would return the module sources for each? Now this returns a fixed string, regardless of module ID, which then exports different symbols.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+	duk_dup(ctx, 0);   /* module ID */
+	duk_dup(ctx, -3);  /* parent ID */
+	duk_call(ctx, 2);
+	if (!duk_is_string(ctx, -1)) {
+		duk_error(ctx, DUK_ERR_API_ERROR, "error in module resolution");
+	}
+
+	/* [ ... stash ... resolved_id ] */
+
+	id = duk_require_string(ctx, -1);
```
svaarala: A small nit here, can be fixed here or in a follow-up cleanup: because duk_require_string() is used, the duk_is_string() check directly above is not necessary. The error message will be a bit vague but I don't think it needs to be very explicit if the resolve() function is violating the basic contract? If the error stays in place, then the error message ("error in module resolution") should probably be made more explicit because it's actually quite vague now compared to the check which triggers it?

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+	if (duk__get_cached_module(ctx, id)) {
+		goto have_module;  /* use the cached module */
```
svaarala: Goto here is unnecessary I think, because the target will be right after the else-block.

OTOH if the got stays, then the else-block could be de-indented by placing it at the top level.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+		duk_dup(ctx, -3);  /* resolved ID */
+		(void) duk_get_prop_string(ctx, module_idx, "exports");
+		duk_dup(ctx, module_idx);
+		ret = duk_pcall(ctx, 3);
+		if (ret != DUK_EXEC_SUCCESS) {
+			duk__del_cached_module(ctx, id);
+			duk_throw(ctx);  /* rethrow */
+		}
+
+		/* [ ... module source? ] */
```
svaarala: Very small nit re: comment format: I've used "source?" to indicate that an element might be present or *absent*. Here it means there's either a source string or undefined, so maybe "source/undefined"?

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+
+		if (!duk_is_undefined(ctx, -1) && !duk_is_string(ctx, -1)) {
```
svaarala: This would have smaller footprint:

    if (!duk_check_type_mask(ctx, -1, DUK_TYPE_MASK_UNDEFINED | DUK_TYPE_MASK_STRING)) {
        ...
    }

svaarala: Another structure option would be:

    if (duk_is_string(ctx, -1)) {
        ...
    } else if (duk_is_undefined(ctx, -1) {
        /* nop */
    } else {
        error
    }

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+	return 1;
+}
+
+static void duk__push_require_function(duk_context *ctx, const char *id) {
+	duk_push_c_function(ctx, duk__handle_require, 1);
+	duk_push_string(ctx, "name");
+	duk_push_string(ctx, "require");
+	duk_def_prop(ctx, -3, DUK_DEFPROP_HAVE_VALUE);
+	duk_push_string(ctx, id);
+	duk_put_prop_string(ctx, -2, "\xffmoduleID");
```
svaarala: Capitalization trivia: would `\xffmoduleId` be more consistent with using `id` elsewhere?

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+static duk_int_t duk__eval_module_source(duk_context *ctx) {
+	/*
+	 *  Stack: [ ... module source ] => [ ... module ] (retval == 0)
+	 *         [ ... module source ] => [ ... err ]    (retval != 0)
+	 */
+
+	/* Wrap the module code in a function.  This is the easiest way to
+	 * implement CommonJS closure semantics and matches the behavior of e.g.
+	 * Node.js.
+	 */
```
svaarala: Maybe comment here that `__dirname` is not provided (it's the one real difference in the wrapping). Would it make sense for `__dirname` to exist but be `undefined`?

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+	/* module.require */
+	duk__push_require_function(ctx, id);
+	duk_put_prop_string(ctx, -2, "require");
+}
+
+static duk_int_t duk__eval_module_source(duk_context *ctx) {
+	/*
+	 *  Stack: [ ... module source ] => [ ... module ] (retval == 0)
+	 *         [ ... module source ] => [ ... err ]    (retval != 0)
+	 */
```
svaarala: This is a bit misleading because errors may be thrown and are not caught here (i.e. `duk_call()` is used to execute the wrapped module).

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+	duk_remove(ctx, -2);
+
+	/* [ ... module func_str ] */
+
+	(void) duk_get_prop_string(ctx, -2, "filename");
+	if (duk_pcompile(ctx, DUK_COMPILE_EVAL) != DUK_EXEC_SUCCESS) {
+		duk_remove(ctx, -2);  /* [ ... err ] */
+		goto fail;
+	}
+	duk_call(ctx, 0);
```
svaarala: This may error specifically.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+
+		/* [ ... module source? ] */
+
+		if (!duk_is_undefined(ctx, -1) && !duk_is_string(ctx, -1)) {
+			duk__del_cached_module(ctx, id);
+			duk_error(ctx, DUK_ERR_API_ERROR, "error in module loading");
+		}
+
+		if (duk_is_string(ctx, -1)) {  /* if we have source */
+			if (duk__eval_module_source(ctx) != DUK_EXEC_SUCCESS) {
```
svaarala: One approach here would be to allow `duk__eval_module_source()` to freely throw errors (it also needs to return code in that case) but use `duk_safe_call()` here to catch the errors more reliably.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+}
+
+void duk_module_node_init(duk_context *ctx) {
+	/*
+	 *  Stack: [ ... options ] => [ ... ]
+	 */
+
+	duk_idx_t options_idx;
+
+	duk_require_object_coercible(ctx, -1);
```
svaarala: This is perhaps an unnecessary check. If the argument is not object coercible (null or undefined) the first property read will throw anyway.

svaarala: Hmm, well, `\xffrequireCache` gets written before that so maybe this is useful after all.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,263 @@
...
+	options_idx = duk_require_normalize_index(ctx, -1);
+
+	/* initialize the require cache to a fresh object */
+	duk_push_global_stash(ctx);
+	duk_push_object(ctx);
+	duk_put_prop_string(ctx, -2, "\xffrequireCache");
+	duk_pop(ctx);
+
+	/* stash callbacks for later use */
```
svaarala: Maybe worth commenting here that user code can later modify the callbacks by overwriting them from the global stash.

svaarala:
Ok, I've re-read the code. Trivia stuff is in the commit comments, I can fix them in a cleanup pull if you prefer.

Two maybe not-so-trivia comments:

- Should `__dirname` be provided by the function wrapper with value set to `undefined`? If this is not done and a module reads `__dirname` it will lookup a global which doesn't exist and will trigger a `ReferenceError` which may not be the best behavior.
- An alternative to structuring the call to `duk__eval_module_source()` would be to allow `duk__eval_module_source()` to freely throw on errors and use a `duk_safe_call()` to call it. As things are now, there's at least one unprotected `duk_call()` in `duk__eval_module_source()` and potential for a few of the other operations to also throw.

extras/module-node/duk_module_node.c
fatcerberus: I didn't think there was anyway the duk_call could error?  It's the equivalent of `eval("(function() { ... })")`.  If it compiles successfully (and the compile *is* protected) I don't think it can throw?

extras/module-node/test.c
fatcerberus: That would probably be better, I agree.  This was kind of thrown together in 5 minutes, can you tell? :)

extras/module-node/duk_module_node.c
fatcerberus: My usual convention when using camelCase is to capitalize acronyms and initialisms unless it's the first term in the identifier.  So you'd have `idOfModule` but `moduleID`.  I don't know if there's an established convention for this in Duktape though, I don't object to changing it if so. :)

extras/module-node/duk_module_node.c
fatcerberus: Hm, I forgot that `duk_check_type_mask()` was a thing.

extras/module-node/duk_module_node.c
fatcerberus: In the early drafts for this pull (before the callbacks were converted to Duktape/C calls), that actually was the case, if the callback didn't return source there wouldn't be anything on the top of the stack except for the `module` object.  I'll fix the comment.

extras/module-node/duk_module_node.c
svaarala: Right, the call is evaluating the function expression. It could run out of memory when creating the function instance but that wouldn't be an "expected error" like something being wrong with the module.

fatcerberus:
I think I agree re: `__dirname` being present but undefined.  Getting an error thrown on attempting to access it (in strict mode) would be a nasty surprise and hurts Node.js compatibility.  It's trivial to add an argument to the wrapper, I'll add it.

extras/module-node/duk_module_node.c
svaarala: There's not really a very established convention for that but I've tended to capitalize only the first letter of an acronym (`requireId`). Similarly I'd probably favor `encodeJx()` over `encodeJX()`. It doesn't really matter though because it's entirely internal to the module loader.

fatcerberus:
w.r.t. `duk__eval_module_source()` throwing: Overall I agree it'd be cleanest to just wrap it in a safe call and let it go wild.  That'd probably help footprint too since there would be fewer explicit checks needed.

extras/module-node/duk_module_node.c
svaarala: That three-part if-statement structure would probably be the cleanest here though.

extras/module-node/test.c
svaarala: It's test code so it doesn't really matter other than if anyone looks, it might give the wrong idea of how to load built-in fixed strings :)

svaarala:
@fatcerberus I've got to sleep but I'll pick up tomorrow. Let me know if you're making any fixes - if not, I'll merge and open a trivia cleanup pull. :)

fatcerberus:
I need to update `duk__eval_module_source()` anyway so I'll address the trivia too tonight.  Then you can give it a final review tomorrow and merge to master.

extras/module-node/README.rst
fatcerberus: To clarify on the above point: `module.exports` isn't read until just before `duk__handle_require()` returns.  So even if `module.exports` is replaced wholesale by the load callback, `require()` handling will see the new value.

extras/module-node/duk_module_node.c
fatcerberus: I agree, the error message is too vague.  I removed the check.

extras/module-node/duk_module_node.c
fatcerberus: I left the goto because it makes it explicit that we're using the cached module.  I did de-indent the else block though, like you said.

fatcerberus:
@svaarala Alright, this is merge-ready again.  `duk__eval_module_source()` is now a safe call.

extras/module-node/duk_module_node.c
```diff
@@ -0,0 +1,275 @@
...
+	(void) duk_get_prop_string(ctx, module_idx, "exports");
+	duk_dup(ctx, module_idx);
+	ret = duk_pcall(ctx, 3);
+	if (ret != DUK_EXEC_SUCCESS) {
+		duk__del_cached_module(ctx, id);
+		duk_throw(ctx);  /* rethrow */
+	}
+
+	if (duk_is_string(ctx, -1)) {
+		duk_int ret;
```
svaarala: Typo in ret typing, `duk_int_t`.

svaarala:
There are some small bugs in the pull (e.g. `duk_int` instead of `duk_int_t`) but I'll merge and open a pull for the fixes because they're so small.

fatcerberus:
Sorry, I don't get up early enough or I would have fixed that `duk_int` bug before you merged.

svaarala:
No worries, I wanted this in master and I had already bugged you with trivia quite a bit :)

fatcerberus:
By the way I tested - this compiles successfully against Duktape 1.x :-)

PR #801, opened 2016-05-21
Add a minimal Duktape-optimized printf() provider to extras

svaarala:
This extra provides a Duktape-optimized printf() which only supports those format codes Duktape actually needs. It's distributed in extras because the format codes should be kept in sync with whatever Duktape needs.

This is useful for barebones platforms where printf() and other standard bindings need to be explicitly implemented (or scavenged). A typical printf() implementation can have e.g. a 20kB footprint, often due to the cost of IEEE double formatting. Sometimes the printf() implementation can be configured for smaller footprint; sometimes not. This extra compiles to less than 1kB.

Tasks:

- [x] Improve `test.c` coverage a bit for NULs etc (see FIXMEs)
- [x] Add `duk_minimal_sprintf()`, using the same helper with `SIZE_MAX` limit
- [x] Add `duk_minimal_sscanf()` for JX parsing, needs only `%p` parsing
- [x] `%lld` handling for JSON fast path => future work note
- [x] 64-bit Windows support (not very important in practice) => future work note
- [x] Manual testcase run with minimal printf in use
- [x] Fix `%09d` with negative argument, now produces `00000-321`; for space padding (`%9d`) correct output is `     -321` which works
- [x] Releases entry

### DIFF
--- a/extras/minimal-printf/README.rst
+++ b/extras/minimal-printf/README.rst
@@ -1,0 +1,110 @@
+==============================================
+Minimal sprintf/sscanf replacement for Duktape
+==============================================
+
+The ``duk_minimal_printf.c`` provides a portable provider for sprintf()/scanf()
+with a feature set matching minimally what Duktape needs.  The provider
+compiles to less than 1kB.  The functions provided are::
+
+    sprintf()
+    snprintf()
+    vsnprintf()
+    sscanf()
+
+Assumptions:
+
+* ``sizeof(void *) <= sizeof(long)``
+
+* ``sizeof(long) <= 8``
+
+Note that these assumptions don't hold e.g. on 64-bit Windows.  This printf
+provider is mostly useful for low memory targets where these assumptions are
+typically not an issue.  The limitations are easy to fix if one relies more
+on platform typing.
+
+Supported formatting
+====================
+
+sprintf()
+---------
+
+Duktape relies on a ``sprintf()`` provider which supports at least the
+following (this list is from Duktape 1.5.0)::
+
+    %c
+    %s
+    %p
+
+    %02d
+    %03d
+    %ld
+    %lld    (JSON fast path only)
+
+    %lu
+
+    %lx
+    %02lx
+    %08lx
+
+This minimal provider supports the following slightly different set:
+
+* Character format ``%c``.
+
+* String format ``%s``.
+
+* Pointer format ``%p``.
+
+* Integer formats ``%d``, ``%ld``, ``%lu`` with optional padding and
+  length modifiers.
+
+* Hex formats ``%x``, ``%lx`` with optional padding and length modifiers.
+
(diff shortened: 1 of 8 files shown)


extras/minimal-printf/README.rst
```diff
@@ -0,0 +1,51 @@
...
+Minimal printf replacement for Duktape
+======================================
+
+The ``duk_minimal_printf.c`` provides a portable provider for ``printf()``
+with a feature set matching minimally what Duktape needs.  The provider
+compiles to less than 1kB.
+
+Assumptions:
+
+* ``sizeof(void *) <= sizeof(long)``
```
fatcerberus: Note that this assumption is broken on, e.g. Windows x64 where `sizeof(void*) == 8` but `sizeof(long) == 4` (!)

svaarala: I know - this is mostly useful on low memory targets where this is not an issue.

svaarala:
Added a note on Windows x64 violating the sizeof() assumptions, and also explaining that these are intended for low memory targets.

svaarala:
A related issue is providing a minimal scanf(). That's considerably easier though, there's only one call site for parsing a `%p` in Duktape, for JX pointer parsing:

    src/duk_bi_json.c:        (void) DUK_SSCANF((const char *) js_ctx->p, DUK_STR_FMT_PTR, &voidptr);

It's probably best to bundle this scanf() provider into this same extra because the `%p` format output should also be parsed back so there's a formatting dependency.

With JX disabled no `scanf()` calls are used. Another option would be to avoid platform specific pointer formatting for JX or use a configurable provider for it, so that the entire `scanf()` dependency could be removed.

svaarala:
Now there's a minimal sscanf() in the extra too: it compiles to about 100 bytes.

svaarala:
Testing 32-bit "duk" with minimal printf/scanf:

```
duk> Duktape.enc('jx', {foo:Duktape.Pointer('foo'),bar:Duktape.Pointer(0)})
= {foo:(0x08f85af8),bar:(null)}
duk> obj = Duktape.dec('jx', '{foo:(0x08f85af8),bar:(null)}')
= [object Object]
duk> obj.foo
= 0x08f85af8
duk> obj.bar
= null
```

Date formatting uses `%02d` etc:

```
duk> new Date()
= 2016-05-22 03:33:32.397+03:00
```

svaarala:
Had to fix one more bug. Final sizes on x86:

- duk__format_long() helper: 137 bytes
- duk_minimal_vsnprintf(): 477 bytes
- duk_minimal_snprintf(): 43 bytes
- duk_minimal_sprintf(): 43 bytes
- duk_minimal_sscanf(): 90 bytes
- nybble lookup: 16 bytes

Total is 806 bytes.

PR #805, opened 2016-05-23
Add myself to AUTHORS.rst

dbohdan:


svaarala:
You should actually be on the first list (Authors) as you've contributed text to the Wiki :)

dbohdan:
Right. :-) Fixing that.

PR #807, opened 2016-05-23
Comment on minimal printf() and duk_push_sprintf()

svaarala:


fatcerberus:
I'm glad you added this note, I was wondering if the original change affected `duk_push_sprintf()` myself.

PR #809, opened 2016-05-24
Makefile program path fixes

megous:
Hi, these fixes help me build duktape on arch linux without the need to patch Makefile.

I haven't tested it elsewhere, but command -v should be more portable than which.


svaarala:
@megous Thanks, I'll check this out. Could you add yourself to AUTHORS.rst (see CONTRIBUTING.md) in this pull?

megous:
Done.

megous:
Thank you.

PR #810, opened 2016-05-24
Avoid variables named "index" which may cause -Wshadow warnings

svaarala:
- [x] Rename `index` -> `idx` in sources
- [x] Rename in website and API documentation
- [x] Code issues note
- [x] Regression testing; could codepolicycheck detect a blacklist of bad names?
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1678,6 +1678,9 @@ Planned
 * Fix a harmless compilation warning related to a shadowed variable (GH-793,
   GH-794)
 
+* Fix -Wshadow warnings on some GCC versions for variable/argument name
+  'index' by avoiding such identifiers in API and internals (GH-810)
+
 * Internal change: rework shared internal string handling so that shared
   strings are plain string constants used in macro values, rather than
   being declared as actual symbols; this reduces compilation warnings with
--- a/doc/code-issues.rst
+++ b/doc/code-issues.rst
@@ -275,6 +275,18 @@ The fix is::
           ...
   }
 
+Local variable naming
+---------------------
+
+Variables are generally lowercase and underscore separated, but no strict
+guidelines otherwise.
+
+Avoid local variable names which might shadow with global symbols defined in
+platform headers (not just one platform but potentially any platform).  For
+example, using ``alloc`` would be a bad idea, and ``index`` also causes
+concrete problems with some GCC versions.  There are a few blacklisted
+identifiers in the code policy check.
+
 Other variable declarations
 ---------------------------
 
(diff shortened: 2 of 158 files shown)


svaarala:
There are no issues on my local build or the testrunner runs (which run on some Ubuntu variant), but there are a lot of warnings in the Travis environment.

svaarala:
Ok, the build warnings seem to be fixed, though I get this weird empty error in Travis output:

```
duk_api_time.c: In function ‘duk_time_to_components’:

linenoise/linenoise.c: In function ‘linenoiseNoTTY’:
linenoise/linenoise.c:1015:13: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
linenoise/linenoise.c:1022:9: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
```

The linenoise warnings are expected but the duk_api_time.c seems like broken compiler output (I've seen similar issues with GCC warnings sometimes, and it's difficult to figure out why).

I'll see if I can add some check to codepolicycheck to catch this (it's pretty easy to imagine using a variable named "index" at some point in the future after all... :). Let's hope no platform defines, say, `i`.

fatcerberus:
> Let's hope no platform defines, say, i

Hehe.  But I thought the `index` being shadowed was in Duktape (not platform) code anyway?

svaarala:
No, Duktape just uses an argument/variable named `index` which shadows a function `index()` defined in strings.h.

svaarala:
Related: http://stackoverflow.com/questions/22028651/how-to-suppress-shadow-index-warning.

fatcerberus:
Ah.  That's pretty annoying then.

svaarala:
One platform I worked on a few years ago defined constants like `C1`, `X4`, etc in platform headers. This caused a lot of conflicts with e.g. MD5 implementations. C is not without its problems :)

fatcerberus:
Hm, so it's apparently related to older GCC versions because it was fixed in GCC 4.8.

svaarala:
Yeah, it's version specific quite clearly (I'm guessing this shadowing issue would turn up pretty often). But it'd no big deal avoiding "index" as a variable/argument name. I'll just have to figure out how to detect it for commit tests so I won't break it again at some later time :)

svaarala:
It seems codepolicycheck identifier pattern is good enough, so just blacklisting "index" works:

    vim +331   src/duk_api_public.h.in : invalid identifier 'index' (perhaps plain)  - DUK_EXTERNAL_DECL duk_idx_t duk_normalize_index(duk_context *ctx, duk_idx_t index);

svaarala:
Ok, now I got the time related shadow warning without it getting garbled:

```
duk_api_time.c:11:73: warning: declaration of ‘time’ shadows a global declaration [-Wshadow]
```

Fixed that too.

PR #813, opened 2016-05-25
Fix some signed shift warnings for clang

svaarala:
- [x] Example of signed shift issues
- [x] Fix internal signed shift issues, verify with clang warnings enabled
- [x] Fix some volatile NULL pointer issues, clang warnings
- [x] Add warning options for Makefile duk-clang target
- [x] Releases entry


svaarala:
@saghul If you have a chance it'd be good to hear if the shift-related sanitize warnings were fixed here for you too (ignore the align warnings :).

PR #815, opened 2016-05-25
Remove unused duk_hobject property layout(s)

svaarala:
Currently layouts 1 and 2 are used and layout 3 is not. Layout 3 would provide natural alignment without padding on a majority of common platforms, but is a bit slower than layout 1 and 2.

Another option would be to add direct property allocation pointers by default (to make the layout differences irrelevant) and drops these fields only for low memory builds. Yet another option would be to store direct pointers and use the pointers to dynamically compute values like `e_next` which are currently explicit fields.

### DIFF
--- a/src/duk_hobject.h
+++ b/src/duk_hobject.h
@@ -750,6 +750,9 @@ struct duk_hobject {
 	 *                                                      0xffffffffUL = unused, 0xfffffffeUL = deleted
 	 *    e_size * sizeof(duk_uint8_t)           bytes of   entry flags (e_next gc reachable)
 	 *
+	 *  FIXME: decide what layouts to keep; currently layout 3 is unused,
+	 *  performance test it to see if it could replace 1 or 2 (or both).
+	 *
 	 *  In layout 1, the 'e_next' count is rounded to 4 or 8 on platforms
 	 *  requiring 4 or 8 byte alignment.  This ensures proper alignment
 	 *  for the entries, at the cost of memory footprint.  However, it's


svaarala:
Footprints on x64:

```
duktape@keppihevonen:~/duktape$ size duk.1 duk.2 duk.3
   text	   data	    bss	    dec	    hex	filename
 274981	   1396	    200	 276577	  43861	duk.1
 274909	   1396	    200	 276505	  43819	duk.2
 275461	   1396	    200	 277057	  43a41	duk.3
duktape@keppihevonen:~/duktape$ size duk.O2.1 duk.O2.2 duk.O2.3
   text	   data	    bss	    dec	    hex	filename
 341334	   1436	    216	 342986	  53bca	duk.O2.1
 341254	   1436	    216	 342906	  53b7a	duk.O2.2
 341790	   1436	    216	 343442	  53d92	duk.O2.3
```

On x86:

```
duktape@keppihevonen:~/duktape$ size duk.x86.*
   text	   data	    bss	    dec	    hex	filename
 274949	    752	    156	 275857	  43591	duk.x86.1
 274773	    752	    156	 275681	  434e1	duk.x86.2
 275145	    752	    156	 276053	  43655	duk.x86.3
duktape@keppihevonen:~/duktape$ size duk.O2.x86.*
   text	   data	    bss	    dec	    hex	filename
 333963	    800	    168	 334931	  51c53	duk.O2.x86.1
 333851	    800	    168	 334819	  51be3	duk.O2.x86.2
 334155	    800	    168	 335123	  51d13	duk.O2.x86.3
```

svaarala:
Some perf tests:

```
test-buffer-object-read.js          : duk.O2.1  2.73 duk.O2.2  2.72 duk.O2.3  2.78 duk.O2.x86.1  3.65 duk.O2.x86.2  3.52 duk.O2.x86.3  3.52
test-buffer-object-write.js         : duk.O2.1  3.12 duk.O2.2  3.17 duk.O2.3  3.10 duk.O2.x86.1  3.74 duk.O2.x86.2  3.70 duk.O2.x86.3  3.69
test-object-garbage.js              : duk.O2.1  4.58 duk.O2.2  4.71 duk.O2.3  4.61 duk.O2.x86.1  5.09 duk.O2.x86.2  5.10 duk.O2.x86.3  5.22
test-object-literal.js              : duk.O2.1  2.45 duk.O2.2  2.53 duk.O2.3  2.61 duk.O2.x86.1  2.82 duk.O2.x86.2  2.79 duk.O2.x86.3  2.87
test-reg-readwrite-object.js        : duk.O2.1  3.24 duk.O2.2  3.25 duk.O2.3  3.14 duk.O2.x86.1  3.42 duk.O2.x86.2  3.43 duk.O2.x86.3  3.41
test-assign-proplhs-reg.js          : duk.O2.1  3.54 duk.O2.2  3.66 duk.O2.3  3.67 duk.O2.x86.1  4.83 duk.O2.x86.2  4.65 duk.O2.x86.3  4.73
test-assign-proprhs.js              : duk.O2.1  4.05 duk.O2.2  4.05 duk.O2.3  4.28 duk.O2.x86.1  5.85 duk.O2.x86.2  5.83 duk.O2.x86.3  5.87
test-prop-read.js                   : duk.O2.1  4.59 duk.O2.2  4.75 duk.O2.3  4.81 duk.O2.x86.1  6.38 duk.O2.x86.2  6.29 duk.O2.x86.3  6.57
test-prop-write.js                  : duk.O2.1  4.24 duk.O2.2  4.30 duk.O2.3  4.30 duk.O2.x86.1  5.44 duk.O2.x86.2  5.30 duk.O2.x86.3  5.41
```

svaarala:
It's measurable but small. I'm hoping to simplify property table handling a bit:

- For desktop and other non-memory-starved targets a naturally aligned layout would be used. Most likely the default configuration would also provide direct pointers to the various parts of the property table for faster common case access. The layout should emphasize quick access rather than being as packed as possible. I'll also be reworking the hash algorithm which is not very good now (it uses an arithmetic modulus for stepping which is sometimes expensive).
- For low memory targets the goal would be have a compact layout; staying within alignment restrictions means that there'd in practice be either a single naturally aligned layout or a few different layouts. Access to the various parts would use computed pointers. Add support for pointer compression of keys.

Also one thing worth trying is to drop the count fields (entry part size, array part size, hash part size, etc) and maintain pointers to the parts instead. The sizes, when necessary, can be computed from the pointers but for loops one can typically use the pointers for loop control.

With this overall picture in mind I'm not sure if this will be merged or not, but it's good to know what the impact would be.

fatcerberus:
Keep in mind that even systems with plenty of memory can still have performance issues with too sparse structures, causing cache thrashing and similar issues.

svaarala:
Of course.

svaarala:
With respect to cache behavior, the current solution is to avoid a hash table for relatively large objects, up to 64 properties, because it's cache friendly and relatively fast to scan successive keys. A hash table is also quite fast, but a dense hash table with a lot of collisions and a non-linear probing sequence "hops around" the hash table and can be cache unfriendly. This only matters if the property in question is accessed relatively rarely so that the hash table is not in the cache already. But even if the hash table is in the cache so that the probe sequence is already cached, it consumes more cache space than an object without a hash table, which may impact some other code in the calling application. The net effects are quite difficult to measure reliably because they ultimately depend on the context.

In any case, for performance a hash table is useful around 16 properties on this x64. That's for reads/writes on the object; there's also an additional cost in inserting new elements into the hash and rehashing in resizes, so it's of course dependent on the object access pattern -- i.e. how many times the same properties are read/written as opposed to creating new properties.

The current probe sequence is string hash dependent and non-linear (and involves a modulus which can be expensive). That isn't necessarily good in practice: for rarely accessed properties and a collision it means multiple cache lines are often fetched. A simple mask based modulus (`hash & mask`) and a linear scan seems to work better in practice, as long as the hash is relatively sparse so that collision sequences usually fit into the same cache line. That's a memory/performance trade-off obviously: a very sparse hash minimizes collisions at the cost of memory, and at the cost of hogging more of the cache *if* all properties are heavily accessed.

I'll of course try to do different measurements as I work on the options. But ultimately some use cases will benefit and others won't, it depends on what's the limiting factor in each context.

svaarala:
A quick example, I did a little experiment of using a much simpler hash table sizing and probing algorithm:
- Select hash table size as 2^(N+1) where N is smallest integer to ensure 2^N >= entry_count. This means initial hash table slot is simply `stringHash & mask`. Hash load will always be <50% because of that +1.
- Probe using a step of +1.

With this approach property read/write tests, reading/writing already existing properties so that there's no property add/remove load, with different object sizes (number indicates property count):

```
test-prop-read-1024.js              : duk.O2.hash  4.57 duk.O2.master  6.31
test-prop-read-16.js                : duk.O2.hash  4.59 duk.O2.master  5.02
test-prop-read-256.js               : duk.O2.hash  4.56 duk.O2.master  6.14
test-prop-read-32.js                : duk.O2.hash  4.50 duk.O2.master  5.51
test-prop-read-4.js                 : duk.O2.hash  4.57 duk.O2.master  4.48
test-prop-read-48.js                : duk.O2.hash  4.65 duk.O2.master  5.57
test-prop-read-64.js                : duk.O2.hash  4.54 duk.O2.master  5.52
test-prop-read-8.js                 : duk.O2.hash  4.56 duk.O2.master  4.71
test-prop-write-1024.js             : duk.O2.hash  4.14 duk.O2.master  4.98
test-prop-write-16.js               : duk.O2.hash  4.07 duk.O2.master  4.49
test-prop-write-256.js              : duk.O2.hash  4.13 duk.O2.master  4.45
test-prop-write-32.js               : duk.O2.hash  4.12 duk.O2.master  4.51
test-prop-write-4.js                : duk.O2.hash  4.19 duk.O2.master  4.05
test-prop-write-48.js               : duk.O2.hash  4.12 duk.O2.master  4.52
test-prop-write-64.js               : duk.O2.hash  4.03 duk.O2.master  6.50
test-prop-write-8.js                : duk.O2.hash  4.00 duk.O2.master  4.09
```

This is by no means conclusive (or necessarily fair because the load factor is not the same): the test set needs to include more variable scenarios (random accessing millions of properties so that the read/write won't always hit the cache, property add/remove load, not just 2^N property counts, etc), so I'll post a pull when more of that is in place. I especially don't want to move to using something advanced (like a cuckoo hash) without having a reasonable set of practical measurements in place.

Anyway, hopefully there'll be a roughly 10-20% improvement by changing the hash algorithm alone. This only applies to medium sized objects with >= 16 properties however. So it's usually not very critical.

PR #818, opened 2016-05-27
Minor cleanups to module-node extra

svaarala:
- [x] Fix value stack index use in eval module helper
- [x] Ensure test.c exitcode is != 0 when errors occur (also add this behavior to other extras)
- [x] Add assert() driven `make test`
- [x] __dirname note: not available due to being platform dependent
- [x] Other trivia

### DIFF
--- a/extras/module-node/Makefile
+++ b/extras/module-node/Makefile
@@ -2,15 +2,16 @@
 
 .PHONY: test
 test:
-	gcc -o $@ -I../../src/ -I. ../../src/duktape.c duk_module_node.c test.c -lm
+	gcc -Wall -Wextra -std=c99 -o $@ -I../../src/ -I. ../../src/duktape.c duk_module_node.c test.c -lm
 	@printf '\n'
-	./test 'typeof require("pig") === "string" ? "require() OK" : "require() FAILED";'
-	./test 'require("cow").indexOf("pig") !== -1 ? "nested require() OK" : "nested require() FAILED";'
-	./test 'var ape1 = require("ape"); var ape2 = require("ape"); ape1 === ape2 ? "caching OK" : "caching FAILED";'
-	./test 'var ape1 = require("ape"); var inCache = "ape.js" in require.cache; delete require.cache["ape.js"]; var ape2 = require("ape"); inCache && ape2 !== ape1 ? "require.cache OK" : "require.cache FAILED";'
-	./test 'var ape = require("ape"); typeof ape.module.require === "function" ? "module.require OK" : "module.require FAILED";'
-	./test 'var ape = require("ape"); ape.module.exports === ape ? "module.exports OK" : "module.exports FAILED";'
-	./test 'var ape = require("ape"); ape.module.id === "ape.js" && ape.module.id === ape.module.filename ? "module.id OK" : "module.id FAILED";'
-	./test 'var ape = require("ape"); ape.module.filename === "ape.js" ? "module.filename OK" : "module.filename FAILED";'
-	./test 'var ape = require("ape"); ape.module.loaded === true && ape.wasLoaded === false ? "module.loaded OK" : "module.loaded FAILED";'
-	./test 'var ape = require("ape"); ape.__filename === "ape.js" ? "__filename OK" : "__filename FAILED";'
+	./test 'assert(typeof require("pig") === "string", "basic require()");'
+	./test 'assert(require("cow").indexOf("pig") !== -1, "nested require()");'
+	./test 'var ape1 = require("ape"); var ape2 = require("ape"); assert(ape1 === ape2, "caching");'
+	./test 'var ape1 = require("ape"); var inCache = "ape.js" in require.cache; delete require.cache["ape.js"]; var ape2 = require("ape"); assert(inCache && ape2 !== ape1, "require.cache");'
+	./test 'var ape = require("ape"); assert(typeof ape.module.require === "function", "module.require()");'
+	./test 'var ape = require("ape"); assert(ape.module.exports === ape, "module.exports");'
+	./test 'var ape = require("ape"); assert(ape.module.id === "ape.js" && ape.module.id === ape.module.filename, "module.id");'
+	./test 'var ape = require("ape"); assert(ape.module.filename === "ape.js", "module.filename");'
+	./test 'var ape = require("ape"); assert(ape.module.loaded === true && ape.wasLoaded === false, "module.loaded");'
+	./test 'var ape = require("ape"); assert(ape.__filename === "ape.js", "__filename");'
+	./test 'var badger = require("badger"); assert(badger.foo === 123 && badger.bar === 234, "exports.foo assignment");'
--- a/extras/module-node/duk_module_node.c
+++ b/extras/module-node/duk_module_node.c
@@ -61,7 +61,6 @@ static duk_ret_t duk__handle_require(duk_context *ctx) {
 	 */
 
 	const char *id;
-	duk_ret_t nrets;
 	const char *parent_id;
 	duk_idx_t module_idx;
 	duk_idx_t stash_idx;
@@ -73,6 +72,7 @@ static duk_ret_t duk__handle_require(duk_context *ctx) {
 	duk_push_current_function(ctx);
 	(void) duk_get_prop_string(ctx, -1, "\xff" "moduleId");
 	parent_id = duk_require_string(ctx, -1);
+	(void) parent_id;  /* not used directly; suppress warning */
 
 	/* [ id stash require parent_id ] */
 
@@ -116,7 +116,7 @@ static duk_ret_t duk__handle_require(duk_context *ctx) {
 	}
 
 	if (duk_is_string(ctx, -1)) {
-		duk_int ret;
+		duk_int_t ret;
 
 		/* [ ... module source ] */
 
@@ -133,7 +133,7 @@ static duk_ret_t duk__handle_require(duk_context *ctx) {
 		/* nop */
 	} else {
 		duk__del_cached_module(ctx, id);
-		duk_error(ctx, DUK_ERR_API_ERROR, "error in module load callback");
+		duk_error(ctx, DUK_ERR_API_ERROR, "invalid module load callback return value");
 	}
 
 	/* fall through */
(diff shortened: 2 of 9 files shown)


svaarala:
Ok, my cleanups are complete. Good stuff, it's awesome there's now an off-the-shelf solution for people wanting to use a Node.js-like loader :)

On thing I'm not very happy about is the test coverage; a lot of basic mechanics are not covered (same applies to other extras too). But it's probably best not to try to fix it here (in `test.c`); it's better to figure out what to do about extras and examples w.r.t. testing. Currently the only examples/extras that get some coverage are those integrated into `duk` and covered by `tests/ecmascript` but that's mostly for legacy reasons. There needs to be a proper solution for doing a lot of tests not directly related to the core stuff, and so that every example/extra doesn't need a mini testing framework.

svaarala:
@fatcerberus I'd like to tackle the built-in 1.x loader next -- do you still want to take that on, or should I?

extras/module-node/Makefile
```diff
@@ -14,3 +14,4 @@ test:
 	./test 'var ape = require("ape"); ape.module.filename === "ape.js" ? "module.filename OK" : "module.filename FAILED";'
 	./test 'var ape = require("ape"); ape.module.loaded === true && ape.wasLoaded === false ? "module.loaded OK" : "module.loaded FAILED";'
 	./test 'var ape = require("ape"); ape.__filename === "ape.js" ? "__filename OK" : "__filename FAILED";'
+	./test 'var badger = require("badger"); badger.foo === 123 && badger.bar === 234;'
```
fatcerberus: Not sure what the purpose of this test is?  It will throw since everything other the pig/cow/ape exports undefined.

extras/module-node/duk_module_node.c
```diff
@@ -191,7 +191,7 @@ static duk_int_t duk__eval_module_source (duk_context *ctx, void *udata) {
 static duk_int_t duk__eval_module_source (duk_context *ctx) {
 #endif
 	/*
-	 *  Stack: [ module source ]
+	 *  Stack: [ ... module source ]
```
fatcerberus: I always forget safe calls don't get their own stack frame (yet).  Thanks for fixing this.

fatcerberus:
Hm... since you just wanted to copy the core loader from Duktape internals, it'd probably be best for you to do it.  I'm not familiar with how everything is wired up internally w.r.t. modules.

extras/module-node/Makefile
fatcerberus: Nevermind, I saw that you added an additional module to the loader. :)

svaarala: The test is trying to check that just assigning `exports.foo = <something>` works because none of the other tests do that: they all replace `module.exports` entirely.

svaarala: There's a lot of other gaps in coverage (same for other extras so it's not something being wrong here) but I wanted that basic `exports.foo = 123` to be covered because it's so central to modules.

fatcerberus:
I might revisit the Node.js loader at some point in the future, I'd like to see if I could implement `module.parent` and `module.children` to make this even more compatible with Node modules.  Well, at least the ones that don't need to call core Node.js APIs anyway.

fatcerberus:
The `nrets` on line 64 of `duk_module_node.c` is unreferenced now - you can delete it.

svaarala:
Added a few warning options to gcc:

```
duk_module_node.c: In function ‘duk__handle_require’:
duk_module_node.c:65:14: warning: variable ‘parent_id’ set but not used [-Wunused-but-set-variable]
  const char *parent_id;
              ^
duk_module_node.c:64:12: warning: unused variable ‘nrets’ [-Wunused-variable]
  duk_ret_t nrets;
            ^
```

svaarala:
`parent_id` is used via `duk_require_string()` to verify its existence but it can be cast to void.

extras/module-node/Makefile
fatcerberus: I like that you just return true or false for the test - could probably update the other tests to do the same and make them more compact.

svaarala: Oops, that's actually not enough because the 'make test' is now depending on exit code, which gets set to an error only if an error is thrown. That is, if one wants to run `make test` and assume that exit code 0 means everything is working. (None of the other extras do that, but that'd be a simple way to have at least some coverage.)

svaarala: I'll add a global `assert()` binding to allow the tests to actually throw when necessary. Not necessarily something to do in the long run (I still haven't figured out what's the next step for test coverage, as I had to scale down my expectations quite a bit for lack of time :-).

svaarala:
Ok, I added a global binding `assert(value, message)` so that the tests can actually throw on error:

- https://github.com/svaarala/duktape/blob/a5d3026f6ee01d4d8a4bd4544e529a7744e1365e/extras/module-node/Makefile

This should be enough for `make test` to fail with exit code if any of the tests fail. Assert-based unit tests are quite common so maybe this is a reasonable stop-gap for the extras.

fatcerberus:
Just thought of this: Might be worth adding a comment in the code that `__dirname` is not set because the loader can't make any assumptions about the underlying file system.

svaarala:
Right, could resolve/load return that because they are aware of the file system?

svaarala:
For example, could `load` set `module.dirname` if a directory name is applicable, and that would get passed to the module wrapper (and undefined if it isn't set)?

fatcerberus:
I assume `resolve` would be the better place to set it.

svaarala:
Why? It gets called every time a `require()` happens, even when the module turns out to be cached and we don't even create a module object.

Resolve callback could return both a resolved ID and a tentative dirname (in case a module gets created) but that'd mean returning an object from the resolve callback. That wouldn't necessarily be a bad idea, because it's not hard to foresee there might be interesting metadata to resolving than just getting the ID.

Another option would be to always create a module object when resolving, and throw it away in case it's already been loaded.

svaarala:
Ok, I'll open a pull for it today.

svaarala:
Hmm, so assuming `resolve` would be the callback where an application would most naturally figure out the `__dirname`, the alternatives would be:

- Change resolve() return value to be an object which provides both resolved ID, dirname, and maybe something else in the future: `{ id: "foo", "dirname": "modules/stuff" }`.
- Change resolve() to return a string but use some separator to allow a dirname to be included, e.g. `foo/bar.js|foo`. Seems too hacky to me at least :).
- Change the require() implementation to always create a tentative module object first, and give it to `require()`. If the module is loaded, the tentative module object gets used; otherwise it's discarded. This allows the resolve() callback to set e.g. `module.dirname = "modules/stuff";` directly without changing the plumbing, and also adds some extensibility. The downside is creating a throwaway object for each require() call.
- Something else? :-)

fatcerberus:
Hm, there's no `module.dirname` in Node (at least I don't think so), only the `_dirname` free variable which gets passed to the module wrapper as an argument.

Creating a fresh `module` for every `require()` might be overkill since in a large application it's possible for half (or more) of the `require()` calls to be already in the cache.  It wouldn't be very efficient to create a large number objects just to discard them immediately.

That said, `resolve` seems to be in the best position to come up with a canonical directory name, which is why it seemed awkward to me to give that responsibility to `load`--which might only care about getting source from the filename it was given without any concept of e.g. path format.

fatcerberus:
Following up on the throwaway module objects: It could be a concrete issue in the case of a Duktape build with refcounting and voluntary GC disabled.  Then the throwaways will only be collected in an emergency which is probably not ideal.

svaarala:
I proposed `module.dirname` just as a way of passing that information from `resolve()` to `load()` (which actually needs it) - not because it'd be the most logical place for it.

It would be cleanest IMO if resolve() returned an object containing the resolved ID and any other metadata; this intermediate result could then be consumed by load() as is so that that whatever metadata was looked up would be available without changing the compatibility related parts of the arrangement. This approach is not ideal when you just want the module ID and don't have any other metadata to pass; in this case the extra object is just extra work to deal with. OTOH you don't usually write dozens of resolve/load functions, so it's a one time thing usually and not as important as something that appears 100 times in a code base.

svaarala:
With automatic periodic mark-and-sweep disabled there are two situations where mark-and-sweep runs:

1. The user requests it.
2. An allocation attempt fails.

The second condition ensures that if we were to run out of memory, several rounds of mark-and-sweep are tried before actually erroring out on an out-of-memory error. This also means that even with voluntary GC disabled, no out-of-memory occurs *if* there is actually memory available.

svaarala:
My view is probably colored quite a bit by versioning concerns: almost every time you take a shortcut in API design, especially for interfaces quite likely to change, there's much more effort in dealing with versioning issues than with the actual functional changes.

Here, based on the experiences with the Duktape 1.x loader framework, it's inevitable there will be additions to the resolve/load process which one doesn't yet foresee. Some of them may be changes in the Node.js side, i.e. the public API, and others might be custom to the Duktape environment.

So the most version compatible (considering both C and Ecmascript source compatibility) would be something like:

```js

// args = all the context related to resolving a module
function resolve(args) {
    var requestedId = args.requestedId;

    // Come up with resolved ID
    var resolvedId = ...;

    return {
        resolvedId: resolvedId,
        // ... any other metadata
    };
}

// args = all arguments related to loading the module from the loader framework
// lookup = verbatim lookup result object from resolve() call
function load(args, lookup) {
    var id = args.id;  // resolved ID

    return {
        source: sourceCode,
        // whatever additional metadata is eventually needed
    };
}
```

Interaction with the module loader is via objects; lookup result passes as an object from resolution to loading, and load result is an object which is easy to extend (and it's quite easy to foresee source code is not the only possible result).

I'm not suggesting we concretely do this, but just wanted to illustrate what kind of API design I think works best regards versioning when the number of changes over time is high. Using objects establishes key/value containers which can be evolved without changing the formal signatures (C prototypes, value stack conventions, etc).

svaarala:
The above design approach is, by the way, one of the main reasons I'd very much like to offer good multi-property get/set APIs. It would make it easier to use object-based, versionable APIs both internally and externally.

fatcerberus:
I think I would be okay with it if `resolve` returned an object.  I can sympathize with the compatibility issues, as I've started trying to honor semantic versioning with minisphere too and it's not always easy.

svaarala:
So I suggest we finish this pull so that master has a fully functional version and use a separate pull to change the mechanics if that makes sense after sleeping on it.

fatcerberus:
I agree with that sentiment.  Especially since the version in master is currently broken. ;-)

svaarala:
Also might make sense to reconsider the current implementation in a week or two, if I find enough time to finally complete some multi-property API calls.

svaarala:
Ok, I'm merging this; `__dirname` issue will be handled separately.

PR #821, opened 2016-05-27
Move Duktape 1.x module loading framework to extras

svaarala:
- [x] Remove module loading support from core sources
- [x] Add Duktape 1.x compatible module loader as an extra
- [x] Dist/makefile changes
- [x] Add DUK_CMDLINE_MODULE_SUPPORT and integrate the extra into 'duk' (allows running existing testcases)
- [x] Documentation fixes
- [x] Wiki modules howto update - https://github.com/svaarala/duktape-wiki/pull/127
- [x] 2.0 migration notes
- [x] Releases entry


svaarala:
Footprint saved is around 1,7kB. If you add back module loading functionality via the extra, footprint increases by about 400 bytes.

extras/module-duktape/duk_module_duktape.c
```diff
@@ -0,0 +1,441 @@
...
+	/*
+	 *  Call user provided module search function and build the wrapped
+	 *  module source code (if necessary).  The module search function
+	 *  can be used to implement pure Ecmacsript, pure C, and mixed
+	 *  Ecmascript/C modules.
+	 *
+	 *  The module search function can operate on the exports table directly
+	 *  (e.g. DLL code can register values to it).  It can also return a
+	 *  string which is interpreted as module source code (if a non-string
+	 *  is returned the module is assumed to be a pure C one).  If a module
```
fatcerberus: > if a non-string is returned the module is assumed to be a pure C one

Should the Node.js loader use the same semantics?  Currently if the return value is anything other than string or undefined an error is thrown.

extras/module-duktape/duk_module_duktape.c
```diff
@@ -0,0 +1,441 @@
...
+	duk_push_string(ctx, "})");
+	duk_concat(ctx, 3);
+	if (!duk_get_prop_string(ctx, DUK__IDX_MODULE, "module")) {
+		/* module.filename for .fileName, default to resolved ID if
+		 * not present.
+		 */
+		duk_pop(ctx);
+		duk_dup(ctx, DUK__IDX_RESOLVED_ID);
+	}
+	duk_eval_raw(ctx, NULL, 0, DUK_COMPILE_EVAL);
```
fatcerberus: Hm, since this is external code now it probably shouldn't be calling "raw" variants of APIs.

svaarala: True, I haven't finished the pull yet, just pushed to get test results :)

extras/module-duktape/duk_module_duktape.c
svaarala: I think the Node.js loader behavior is better, but I don't want to change it here because this extra is trying to keep 1.0 behavior to keep migration as straightforward as possible. (I wouldn't want to add significant features to this loader either, but rather add a separate one.)

fatcerberus: Then again, I kind of like the stronger contract in the new loader.  Obviously we can't do that here because it would be a 1.x-incompatible change.

extras/module-duktape/duk_module_duktape.c
svaarala: The eval call should also be protected: the 1.x loader is more fragile than the Node.js loader is now because it tries to catch most errors manually instead of using a general safe call wrapper. I'll first port the existing semantics and then see if a similar safe call wrapper should be done here too.

svaarala:
Ok, should be a bit closer to finished now. I'll have to re-read this tomorrow (it's getting pretty late).

DukLuv test build fails which is interesting: DukLuv replaces the built-in framework but I think it's expecting `Duktape.modLoaded` to exist and be an empty object. I'll fix that in the DukLuv 2.x pull.

svaarala:
Documentation, migration notes are missing, so I'll add them tomorrow.

svaarala:
Name for the extra could be improved, for example `module-duktape-v1`. This is intended to replicate 1.x behavior and probably won't get much development beyond that, so it'd be good to identify the version in the name.

extras/minimal-printf/Makefile
```diff
@@ -1,5 +1,5 @@
 # Just for manual testing
 .PHONY: test
 test: duk_minimal_printf.c
-	gcc -fno-stack-protector -m32 -otest -Wall -Wextra -Os -fomit-frame-pointer duk_minimal_printf.c test.c
+	gcc -std=c99 -Wall -Wextra -fno-stack-protector -m32 -Os -fomit-frame-pointer -otest duk_minimal_printf.c test.c
```
fatcerberus: Is there a reason this one in particular is compiled with `-fno-stack-protector`?  It seems to be the only one.

fatcerberus: Actually now that I look at it closer there are a lot more compiler options used here than in the other tests.

svaarala: With the minimal printf I was trying to see what options matter re: the footprint it compiles down to.

These Makefiles are pretty placeholdery; I added the warnings options to catch obvious issues, but they're nowhere near good. For proper coverage there would also need to be a clang compilation, C++ compilation, etc.

fatcerberus:
Good thing I already moved to a custom module loader for minisphere, so I sidestepped this migration issue. :)

PR #822, opened 2016-05-28
Fix </p> tags instead of a </li> tag in some enumerations

ReneHollander:
Some enumerations in the API had broken list items.


svaarala:
Thanks!

PR #823, opened 2016-05-29
Rework Duktape.Logger init

svaarala:
Use eval helper approach, set property attributes to match 1.x. `Duktape.Logger` was enumerable before this fix.


fatcerberus:
I really like that string eval trick to get values on the stack.  I've started using it in minisphere in lieu of get_global_prop -> get_prop chains, it's quite convenient.

PR #824, opened 2016-05-29
Make math.random() provider configurable

svaarala:
Currently `math.random()` is provided using a very low entropy seed, and is not intended to be high quality. Because many platforms have a much better native provider, make the random provider configurable via `duk_config.h`.

- [x] Add `DUK_USE_GET_RANDOM_DOUBLE()` to allow replacing the internal `Math.random()` provider
- [x] Random numbers are also used for qsort, so that also needs a fix (it reads bits rather than IEEE doubles though)
- [x] Make current tinyrandom optional in compilation if a replacement is provided
- [x] Improve tinyrandom seeding
- [x] Documentation re: Math.random() quality and how to replace it
- [x] Documentation about random number quality, not cryptographic
- [x] Releases


fatcerberus:
This seems like a low priority.  It'd be good to have but a lot of JS programmers simply avoid `Math.random()` outright for exactly the reason that it's low-quality in the majority of engines.  In most cases it's just a basic linear congruential generator--not sure what the one in Duktape is.

svaarala:
What I'm trying to clean up in general is that platform or environment dependent stuff would be configurable so that Duktape would be maximally embeddable (the file I/O bindings needed to be removed for similar reasons). There are, for example, environments where having a deterministic random sequence is useful (e.g. for repeatable testing) which requires configurability.

One can also overwrite `Math.random()` of course (though that won't affect all contexts automatically). One motivation here is simply to reduce footprint when the platform already provides a reasonably good (most likely much better) PRNG or RNG. Making the provider configurable allows the internal provider to be dropped.

In any case the change is quite trivial.

fatcerberus:
> There are, for example, environments where having a deterministic random sequence is useful (e.g. for repeatable testing) which requires configurability.

Definitely agree there.  It for exactly this reason that I implemented an MT19937 generator in minisphere.  The ability to manually seed the RNG is often very useful in game development. :)

saghul:
Heh, I also opted for a MT19937 PRNG in sjs: https://github.com/saghul/sjs/commit/d889fe3a6675d306edbf00accc65f09ef9d99ef2

As a side note, when I tried to look up the algorithm Duktape's `Math.random` uses I ended up in a non-existing page :confused:

svaarala:
@saghul Mersenne Twister would be a logical choice for a reasonably high quality PRNG. But it's quite a bit larger than the current built-in one, and without good seed material it doesn't seem really worth the footprint. The current one is: http://www.woodmann.com/forum/archive/index.php/t-3100.html - which link did you try to follow?

saghul:
@svaarala sure, that's why I think this proposal is nice! 

Doh, I guess I pasted the link wrong or something... nevermind!

svaarala:
The current seeding is actually quite bad - the `duk_heap` pointer is cast to a 32-bit value and used as a seed. I'll add `DUK_USE_DATE_GET_NOW()` to the mix in this pull too.

fatcerberus:
Huh, that algorithm is quite elegant in its simplicity, quality notwithstanding.  It's a shame it only gives you one good random bit at a time, though.

fatcerberus:
@svaarala Regarding footprint of Mersenne Twister: How about TinyMT:
http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/TINYMT/#tiny

I doubt it's smaller than the current algorithm (it's hard to get any smaller than that, actually), but I wonder how it compares with a full MT19937 generator.

svaarala:
But what's the point of having slightly better but still not really good PRNG output? For me the logical alternatives are smallest possible (which makes sense for a hard-coded built-in generator considering very low memory targets), highest performance, or highest quality. Right now the built-in default is for smallest possible with reasonable output.

I guess one alternative would be to use e.g. Mersenne Twister as a default (figuring out some solution for getting a high quality seed), and recommend that low memory targets replace the default via `duk_config.h` to reduce footprint.

fatcerberus:
The seed quality, at least to me, seems to be the bigger portability issue here: Since one of Duktape's goals is small footprint for embedded platforms, and there are devices which don't have a (user-accessible) source of randomness or even a RTC, there's not really a good way to come up with a seed that works well for all platforms.  Deriving the seed from the pointer is low quality as you say but very portable.

Long story short: This proposal is a good idea, as long as the implementor gets to control seeding as well as generation.

svaarala:
The built-in seeding can be improved via `DUK_USE_GET_DATE_NOW()` (current time in milliseconds).

If one overrides the generator, control of the algorithm and the seed would naturally behind whatever macro implements the random number generator. I guess the macro should get the heap userdata in case that's useful in locating a per-heap seed state.

fatcerberus:
@saghul I see you had the same idea I just did to modify the MT19937 generator to be object-oriented.  Very useful.

@svaarala Apologies for hijacking the discussion. :)

svaarala:
Hehe :) Well it's not entirely off-topic :)

fatcerberus:
Re: Useful defaults: Mersenne Twister might be a bit heavy for inclusion in Duktape, even in light of this pull.  Most extant JS engines seem to have switched to xorshift128+ in recent times which looks similar in code size to the current generator but is apparently fairly high-quality.  Might be worth looking into.

Here's an article from the V8 blog about it:
http://v8project.blogspot.com/2015/12/theres-mathrandom-and-then-theres.html

svaarala:
The current generator is not statistically great (that's a pretty high bar), but it seems OK at least visually; with 0.5 threshold:

rnd-threshold-05

With no threshold (256 greyscale):

rnd-threshold-none

svaarala:
Threshold 0.1:

rnd-threshold-01

Threshold 0.9:

rnd-threshold-09

fatcerberus:
Not bad.  There do seem to be some hotspots but considering the simplicity of the algorithm I think it's more than acceptable.  Nobody should be using `Math.random()` for crypto anyway (which is how the bad V8 algorithm was discovered by the way... can't help shaking my head at that).

svaarala:
Cryptographic PRNG is a very different beast and quite definitely should be outside the engine :)

fatcerberus:
> random numbers are also used for qsort

Using random numbers in a sorting algorithm that's not bogosort?  This does not compute :P

svaarala:
It's used to choose the qsort pivot to avoid systematic pathological cases: https://en.wikipedia.org/wiki/Quicksort#Choice_of_pivot.

svaarala:
I could change the qsort to use a deterministic pivot strategy (like median-of-three). This would add footprint, but would remove the only dependency on random input other than `Math.random()`. Plain median-of-three won't work though because there's only a partial order for the elements.

fatcerberus:
Just noticed this was merged.  Maybe I'll take advantage of it and have Duktape use xoroshiro128+ like the rest of minisphere already does. :)

svaarala:
Hmm, I wonder if that generator would be a better default for Duktape too. It's small and presumably one could extract bytes (rather than bits) at a time with reasonable quality. The main drawback is reliance on 64-bit types which are not always available (or may be available but broken as is/was the case for VBCC for example).

fatcerberus:
One thing I like about using a 64-bit generator is that it's very easy to generate doubles in the [0,1[ range:
https://github.com/fatcerberus/minisphere/blob/master/src/shared/xoroshiro.c#L102-L107

You're right that platforms not supporting 64-bit types would need an alternative, though. :-(

svaarala:
Yep, it seems xoroshiro128+ has a reasonable distribution for all bits except the lowest so you can generate a double by taking the high 52 bits and constructing the double manually. I used that same approach for https://github.com/svaarala/duktape/blob/master/src-input/duk_util_tinyrandom.c#L55-L73 but it wasn't much better than the arithmetic version because the current tinyrandom only generates one bit at a time.

I could implement a 32-bit types only variant, it's just rotation so it's not difficult.

svaarala:
#970 does the switch (if merged).

PR #832, opened 2016-06-05
Make "uncaught error" fatal error more verbose

svaarala:
Stringify the uncaught error object safely into the fatal error message. This fatal error occurs quite commonly when protected calls are not used to wrap potentially unsafe operations, so it's useful if the fatal error message provides at least a clue of what went wrong.

- [x] Use verbose fatal error message when not size optimized
- [x] Refactor error-aware summarization to an internal helper
- [x] Test coverage for fatal error => skipped, tested manually
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1669,6 +1669,8 @@ Planned
 * Add a human readable summary of 'new MyConstructor()' constructor call
   target when the target is non-constructable (GH-757)
 
+* Add a safe summary for "uncaught error" fatal error (GH-832)
+
 * Remove duk_{get,put,has,del}_var() calls from API header; they were not
   fully implemented and not part of the documented public API (GH-762)
 
--- a/src/duk_api_internal.h
+++ b/src/duk_api_internal.h
@@ -96,9 +96,7 @@ DUK_INTERNAL_DECL duk_hstring *duk_to_hstring(duk_context *ctx, duk_idx_t idx);
 DUK_INTERNAL_DECL duk_hstring *duk_safe_to_hstring(duk_context *ctx, duk_idx_t idx);
 #endif
 DUK_INTERNAL_DECL void duk_to_object_class_string_top(duk_context *ctx);
-#if !defined(DUK_USE_PARANOID_ERRORS)
 DUK_INTERNAL_DECL void duk_push_hobject_class_string(duk_context *ctx, duk_hobject *h);
-#endif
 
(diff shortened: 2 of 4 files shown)


fatcerberus:
This has bugged me since I started using Duktape.  When I first started minisphere I was using the fatal error handler as a catch-all, but had to switch to duk_pcall() etc. to get the proper error message instead of a useless "uncaught error".  That's why up until recently minisphere handed off to `duk_fatal()`.

I know now that I should have been using the protected calls all along of course, but back then I didn't know that (I thought that's what the fatal handler was for) and was baffled by the nondescriptive errors.

svaarala:
Fatal errors are not very useful for most applications in the sense that there's no safe way to continue without potentially at least leaking something. But even so, uncaught errors are pretty typical; while the most useful fix is to use a protected call, it doesn't hurt if the message is descriptive. Then again, perhaps the message should be "fatal error, use a protected call" ;)

fatcerberus:
I still sometimes get uncaught errors during startup while setting up my API (before calling into any Ecmascript code), e.g. due to improper value stack handling with a `duk_def_prop()`.  When trying to debug that kind of situation "use a protected call" wouldn't be very good advice. :)

I guess one solution would be to wrap the startup code in a safe call but that's inconvenient.

svaarala:
That's really the only workable way in general because any API call can in theory throw due to out of memory etc. But I know it's a bit different running a desktop application/game where it's indeed rare to get initialization errors. That's one reason for this pull :)

Another change I was considering when reworking the fatal error arguments was that maybe the fatal error handler should be mandatory in heap creation; the default handler can't do I/O so it's going to hide the reason. Making the fatal error handler would force at least some kind of reasonable (and environment appropriate) error printing to be present.

fatcerberus:
As mentioned in other pulls in the past, my main issue with the safe call mechanism is that it requires writing the code to be protected in a separate function which doesn't allow you to directly use local variables etc.  So everything needs to be passed on the value stack which is sometimes a pain depending on the types of values you need to pass in/out.

If some mechanism could be devised to do an "inline safe call" a la try-catch that would be awesome.

svaarala:
Right, we discussed that earlier. Opened a separate issue for it, #833.

Now with safe calls having a userdata argument, you can pass a C struct pointer and use the struct for in/out variables. It's not as convenient as an inline call, but in many cases workable enough.

fatcerberus:
Yeah, the userdata was a welcome addition.  Of course I can't actually use it in minisphere yet because I still have to support Duktape 1.x at least until 2.0 is released. :)

PR #840, opened 2016-06-11
Better error message for strict mode undeclared variable write

### LINKED ISSUE #839: Show name in "identifier not defined" error
Currently an "identifier not defined" error caused by a strict mode reference to an undefined variable doesn't include the identifier that caused the error.  This sometimes makes debugging difficult.  Consider updating the error message to include it, to match the behavior for other, similar errors.

svaarala:
Fixes #839.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1670,7 +1670,7 @@ Planned
   fully implemented and not part of the documented public API (GH-762)
 
 * Minor changes to error messages for errors thrown by Duktape internals
-  (GH-827)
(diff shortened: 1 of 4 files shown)


fatcerberus:
Does it only throw on write?  I was under the impression that in strict mode a get would throw too.

svaarala:
Oops, answered the above to the issue instead of the pull: https://github.com/svaarala/duktape/issues/839#issuecomment-225399527

PR #841, opened 2016-06-12
Fix non-source module loading in Node loader

saghul:
I'm integrating the Nodejs-like loader in sjs and ran into this problem when loading binary addons.

### DIFF
--- a/extras/module-node/duk_module_node.c
+++ b/extras/module-node/duk_module_node.c
@@ -130,7 +130,7 @@ static duk_ret_t duk__handle_require(duk_context *ctx) {
 			duk_throw(ctx);  /* rethrow */
 		}
 	} else if (duk_is_undefined(ctx, -1)) {
-		/* nop */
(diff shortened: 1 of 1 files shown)


fatcerberus:
Good catch, I didn't notice this before the loader was merged.  How does it work? :)

saghul:
I think this wasn't noticed because the test only tests source modules, not binary addons which override module.exports.

So far it's working good for me! I hope to merge it into sjs soon. :-)

svaarala:
Yeah, there's no proper test coverage for extras yet. Hopefully soon :)

svaarala:
Oops, @saghul, could you add yourself to `AUTHORS.rst`? I remembered your were already there for some reason.

saghul:
@svaarala Thanks! Sure, here you go: https://github.com/svaarala/duktape/pull/842

svaarala:
Done :)

PR #843, opened 2016-06-13
Some fixes for the CommonJS extra

saghul:
I finished integrating the CommonJS extra into sjs. It works great! I made some adjustments, so I thought I'd share :-)

### DIFF
--- a/extras/module-node/duk_module_node.c
+++ b/extras/module-node/duk_module_node.c
@@ -202,7 +202,7 @@ static duk_int_t duk__eval_module_source(duk_context *ctx) {
 	 * way to implement CommonJS closure semantics and matches the behavior of
 	 * e.g. Node.js.
 	 */
-	duk_push_string(ctx, "(function main(exports,require,module,__filename,__dirname){");
+	duk_push_string(ctx, "(function(exports,require,module,__filename,__dirname){");
 	duk_dup(ctx, -2);  /* source */
 	duk_push_string(ctx, "})");
 	duk_concat(ctx, 3);
@@ -214,6 +214,11 @@ static duk_int_t duk__eval_module_source(duk_context *ctx) {
 	duk_call(ctx, 0);
 
 	/* [ ... module source func ] */
+
+	/* Set name for the wrapper function. */
+	duk_push_string(ctx, "name");
+	duk_push_string(ctx, "main");
+	duk_def_prop(ctx, -3, DUK_DEFPROP_HAVE_VALUE | DUK_DEFPROP_FORCE);
 
 	/* call the function wrapper */
 	(void) duk_get_prop_string(ctx, -3, "exports");   /* exports */
--- a/extras/module-node/duk_module_node.c
+++ b/extras/module-node/duk_module_node.c
@@ -91,7 +91,7 @@ static duk_ret_t duk__handle_require(duk_context *ctx) {
 		goto have_module;  /* use the cached module */
 	}
 
-	duk__push_module_object(ctx, id);
+	duk__push_module_object(ctx, id, 0 /*main*/);
 	duk__put_cached_module(ctx);  /* module remains on stack */
 
 	/*
@@ -158,10 +158,24 @@ static void duk__push_require_function(duk_context *ctx, const char *id) {
 	(void) duk_get_prop_string(ctx, -1, "\xff" "requireCache");
 	duk_put_prop_string(ctx, -3, "cache");
 	duk_pop(ctx);
+
+	/* require.main */
+	duk_push_global_stash(ctx);
+	(void) duk_get_prop_string(ctx, -1, "\xff" "mainModule");
+	duk_put_prop_string(ctx, -3, "main");
+	duk_pop(ctx);
 }
 
-static void duk__push_module_object(duk_context *ctx, const char *id) {
+static void duk__push_module_object(duk_context *ctx, const char *id, int main) {
 	duk_push_object(ctx);
+
+	/* Set this as the main module, if requested */
+	if (main) {
+		duk_push_global_stash(ctx);
+		duk_dup(ctx, -2);
+		duk_put_prop_string(ctx, -2, "\xff" "mainModule");
+		duk_pop(ctx);
+	}
 
 	/* Node.js uses the canonicalized filename of a module for both module.id
 	 * and module.filename.  We have no concept of a file system here, so just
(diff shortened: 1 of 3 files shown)


fatcerberus:
Is there a reason you made the wrapper anonymous?  I originally named it `main` to make it obvious in a stack trace that a module is being loaded.  Too many (anon) entries in a stack trace is annoying. :)

saghul:
Not really, but while writing a test I accidentally wrote `if (require.main === main)` instead of `if (require.main === module)` and I was going nuts :-P

I guess some more obvious name would also be fine.

fatcerberus:
It might be possible to fix the stack trace while leaving the function anonymous by setting its `.name` after it's peval'd.  That would solve the issue of global pollution.

fatcerberus:
Thanks for implementing `require.main` by the way - I wanted to do that, but wasn't sure of the best way to do it while still keeping things generic.

saghul:
I just realized there is one case not covered here: when you wanrt to eval some random code, for example coming from `sjs -e "code"`. That one should be wrapped in a CommonJS wrapper, but require.main should be undefined. I solved this in sjs by making main a parameter, but I'm not sure we want to have it here, since people are going to make small adjustments anyway (I made dirname work, for example).

fatcerberus:
Yeah, that's exactly the reason I didn't implement `require.main` in the extra - it's difficult to do in a generic way.  The semantics of what should be considered the "main" module will differ from application to application.

svaarala:
The original built-in module loader side-stepped that name problem by:

- First creating the wrapper function as an anonymous function. This creates a function which doesn't have an internal "I have a function name binding" flag.
- Then setting the `.name` property which doesn't affect that internal flag.

As a result, when the function is called, a scope with the function name doesn't get created so that the function name is not visible as a variable.

saghul:
@svaarala any hints on how I could add that to this code? Just setting the `.name` property of the compiled function before calling it?

svaarala:
That's basically it, here's how the Duktape loader does it: https://github.com/svaarala/duktape/blob/master/extras/module-duktape/duk_module_duktape.c#L367-L385

saghul:
I force-pushed with the suggested change. I'm at work so I couldn't test it out. It visually compiled fine :-)

fatcerberus:
Yeah, when I first implemented this I forgot that named function expressions bind the name in the function scope to support recursion.  Setting `.name` afterwards should prevent that and still make useful stack traces. :)

saghul:
@svaarala is there anything else I need to do to help get this landed? :-)

svaarala:
@saghul The change looks OK to me but maybe @fatcerberus could comment on this?

fatcerberus:
I'll review the diff in a bit.

extras/module-node/duk_module_node.c
```diff
@@ -91,7 +91,7 @@ static duk_ret_t duk__handle_require(duk_context *ctx) {
 		goto have_module;  /* use the cached module */
 	}
 
-	duk__push_module_object(ctx, id);
+	duk__push_module_object(ctx, id, 0  /* main */);
```
fatcerberus: Nit: One space between argument and comment, and no spaces inside comment, i.e.

```c
duk__push_module_object(ctx, id, 0 /*main*/);
```

extras/module-node/duk_module_node.c
```diff
@@ -160,7 +160,7 @@ static void duk__push_require_function(duk_context *ctx, const char *id) {
 	duk_pop(ctx);
 }
 
-static void duk__push_module_object(duk_context *ctx, const char *id) {
+static void duk__push_module_object(duk_context *ctx, const char *id, int main) {
```
fatcerberus: Should probably be `duk_bool_t` instead of `int`?

extras/module-node/duk_module_node.c
```diff
@@ -238,6 +249,20 @@ static duk_int_t duk__eval_module_source(duk_context *ctx) {
 	return 1;
 }
 
+duk_ret_t duk_module_node_peval_file(duk_context *ctx, const char* path) {
+	/*
+	 *  Stack: [ ... source ]
+	 */
+
+	duk__push_module_object(ctx, path, 1  /* main */);
```
fatcerberus: Nit: Comment style (see above)

extras/module-node/duk_module_node.c
```diff
@@ -182,6 +182,12 @@ static void duk__push_module_object(duk_context *ctx, const char *id) {
 
 	/* module.require */
 	duk__push_require_function(ctx, id);
+	if (main) {
+		duk_dup(ctx, -2);
+	} else {
+		duk_push_undefined(ctx);
```
fatcerberus: Wait, so doesn't this mean `require.main` is only set in the main module?  That doesn't seem right.

fatcerberus:
Mostly minor style nits in the diff, but there is one concrete issue: `require.main` only appears to be set within the main module itself, in all other modules it's set to `undefined`.  This differs from Node.js behavior and pretty much makes the property useless for its intended purpose.

saghul:
@fatcerberus AFAIK the whole point of require.main is to detect when a script was required()-d vs when it was executed. So you'd use duk_module_node_peval_file for files, and get the Node behavior.

https://nodejs.org/api/modules.html#modules_accessing_the_main_module

Node sets require.main to undefined when not running the file directly.

saghul:
I'll address the style issues throughout the day.

fatcerberus:
You misread that: Yes, `require.main` is set to a given file when that file is run directly, however it's *not* undefined for other modules--it points back to the original "main" file.  You can see this by require'ing another module which does `console.log(require.main)` in Node--it will print the module object of the main file.

Node even implies as much in the doc you linked:
> Because module provides a filename property (normally equivalent to __filename), the entry point of the current application can be obtained by checking require.main.filename.

i.e. It functions like GetModuleFileName() in Win32 and /proc/self/exe in Linux.

fatcerberus:
On a side note, this is the problem with the Node docs--you need to read between the lines because a lot is only implied without being stated explicitly.  For an API reference that's not a good thing.

saghul:
Oh, I see. I'll take a look.

saghul:
@fatcerberus Thanks for the help!

fatcerberus:
@saghul No problem.  I like to reverse engineer stuff, can you tell? :)

fatcerberus:
@svaarala Feel free to merge, I can make the adjustments for `require.main` next time I get a chance.

saghul:
Sorry for the delay! I found a way to do it, basically storing the main module in the global stash and then attaching that to require. I'll push a fix here later today.

fatcerberus:
It might be better to store it as an internal property on `require` itself (or make the `require.main` itself nonconfigurable, though this might clash with expected semantics), since `duk_eval_module()` pushes a new independent main module each time.

saghul:
@fatcerberus here is the code I'm using in sjs now: https://github.com/saghul/sjs/commit/4b1edd00caf777ce7474470d4632101af31bd9e4 So far it seems to do the job. Thoughts on it? It's what I was panning to add here.

fatcerberus:
That change is fine as long as there's only one "main" module active at a time (the usual case, definitely).  If a user calls `duk_peval_module()` more than once though, things will get weird.

That might be okay--as long as we document it.  Maybe make a note in the README.

saghul:
👍 ok, I'll add a note to the README then.

saghul:
Updated!

fatcerberus:
Looks good to me, thanks! :-)

svaarala:
So, I guess this pull is good to go? :)

saghul:
I'm done unless there is something else to be addressed, and @fatcerberus gave me his blessing :-)

PR #852, opened 2016-07-01
Minor fixes to module-node extra

svaarala:
- [x] Fix function call name in README
- [x] Fix forward declaration arguments for `duk__push_module_object()` (was missing "int main" argument)
- [x] Change "int main" to "duk_bool_t main"
- [x] Comment and C convention trivia
- [x] Consider renaming `duk_module_node_peval_file()`
- [x] Fix dependency on now removed error codes
- [x] Fix duk_safe_call() argument discrepancy in master

### DIFF
--- a/extras/module-node/duk_module_node.c
+++ b/extras/module-node/duk_module_node.c
@@ -12,7 +12,7 @@ static duk_int_t duk__eval_module_source(duk_context *ctx, void *udata);
 #else
 static duk_int_t duk__eval_module_source(duk_context *ctx);
 #endif
-static void duk__push_module_object(duk_context *ctx, const char *id);
+static void duk__push_module_object(duk_context *ctx, const char *id, duk_bool_t main);
 
 static duk_bool_t duk__get_cached_module(duk_context *ctx, const char *id) {
 	duk_push_global_stash(ctx);
@@ -133,7 +133,7 @@ static duk_ret_t duk__handle_require(duk_context *ctx) {
 		duk_pop(ctx);
 	} else {
 		duk__del_cached_module(ctx, id);
-		duk_error(ctx, DUK_ERR_API_ERROR, "invalid module load callback return value");
+		duk_error(ctx, DUK_ERR_TYPE_ERROR, "invalid module load callback return value");
 	}
 
 	/* fall through */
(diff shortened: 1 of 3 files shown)


svaarala:
@saghul Here are some trivia fixes to the module-node changes.

On closer inspection I find `duk_module_node_peval_file()` to be named rather confusingly: it doesn't load anything from a file but rather expects source code to be on the value stack and loads a module as the "main" module.

Would `duk_module_node_peval_main()` be more appropriate?

saghul:
👍

saghul:
Thanks for taking care of those @svaarala, I guess I couldn't see the forest because of the trees :-)

svaarala:
Ah, the module-node extra fails now because it depends on error code defines (DUK_ERR_API_ERROR) removed in master. I'll fix that here too.

svaarala:
Hmm, there's also a duk_safe_call() discrepancy.

saghul:
In my defense, I've only run this code on v1.5 😢 Sorry about that!

svaarala:
Right, the duk_safe_call() used the 1.x convention. Added the same DUK_VERSION guard as in the other calls so that the loader should now work on both 1.x and 2.x.

@saghul Could you test if this branch works for you too, on 1.x?

extras/module-node/duk_module_node.c
```diff
@@ -12,7 +12,7 @@ static duk_int_t duk__eval_module_source(duk_context *ctx, void *udata);
 #else
 static duk_int_t duk__eval_module_source(duk_context *ctx);
 #endif
-static void duk__push_module_object(duk_context *ctx, const char *id);
+static void duk__push_module_object(duk_context *ctx, const char *id, int main);
```
fatcerberus: Oops, this was never fixed.  `main` argument should be `duk_bool_t` for clarity right?

svaarala: Ok, fixed.

saghul:
> Could you test if this branch works for you too, on 1.x?

Sure, I'll give it a go.

svaarala:
@saghul Did you have time to test this?

saghul:
Not yet sorry! Thanks for the reminder, I'll test it tomorrow for sure!

saghul:
Tested. Works fine! 👍

svaarala:
Ok, thanks @saghul :)

PR #853, opened 2016-07-01
Minor cleanups to debugger detached_cb ctx argument change

svaarala:
- [x] Documentation fixes
- [x] 2.0 migration notes
- [x] Fix debug-trans-dvalue detached_cb ctx argument
- [x] Releases GH issue/pull references

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1654,8 +1654,8 @@ Planned
   DUK_ERR_API_ERROR and a plain Error for Ecmascript representation) into
   TypeErrors and RangeErrors to match common Ecmascript conventions (GH-827)
 
-* Incompatible change: add a ctx argument to detached_cb to allow easier
-  reattachment in detached callback
+* Incompatible change: add a "ctx" argument to the debugger detached_cb to
+  allow easier reattachment in detached callback (GH-758, GH-853)
 
 * Remove no longer needed platform wrappers in duk_config.h: DUK_ABORT(),
   DUK_EXIT(), DUK_PRINTF(), DUK_FPRINTF(), DUK_FOPEN(), DUK_FCLOSE(),
--- a/doc/release-notes-v2-0.rst
+++ b/doc/release-notes-v2-0.rst
@@ -336,6 +336,37 @@ To upgrade:
 
 * Update debug client code to support both versions 1 and 2, or version 2 only.
 
+Debugger detached callback has a duk_context pointer argument
+-------------------------------------------------------------
+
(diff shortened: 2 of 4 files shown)


fatcerberus:
@svaarala I pushed a fix to minisphere master to make it compatible with this change.  Not sure how the minisphere test passed here, your snapshot is from 5/16...

svaarala:
The test is pretty relaxed now and doesn't fail on warnings:

```
src/engine/debugger.c: In function ‘update_debugger’:
src/engine/debugger.c:129:5: warning: passing argument 8 of ‘duk_debugger_attach’ from incompatible pointer type [enabled by default]
     NULL);
     ^
In file included from src/engine/minisphere.h:29:0,
                 from src/engine/debugger.c:1:
src/shared/duktape.h:988:24: note: expected ‘duk_debug_detached_function’ but argument is of type ‘void (*)(void *)’
 DUK_EXTERNAL_DECL void duk_debugger_attach(duk_context *ctx,
                        ^
src/engine/debugger.c: In function ‘update_debugger’:
src/engine/debugger.c:129:5: warning: passing argument 8 of ‘duk_debugger_attach’ from incompatible pointer type [enabled by default]
     NULL);
     ^
In file included from src/engine/minisphere.h:29:0,
                 from src/engine/debugger.c:1:
src/shared/duktape.h:988:24: note: expected ‘duk_debug_detached_function’ but argument is of type ‘void (*)(void *)’
 DUK_EXTERNAL_DECL void duk_debugger_attach(duk_context *ctx,
```

svaarala:
So essentially what the minisphere test is doing is: (1) prepare the Duktape sources from the commit being tested for Minisphere, (2) run Minisphere `make`, and fail if that make command fails. But since Minisphere Makefile doesn't fail on warnings the test now succeeds.

I could maybe apply a diff to the Minisphere Makefile to enable warnings-as-errors for this particular test. Or maybe scan the compilation output and fail if anything gets printed.

fatcerberus:
Ah, it's just a warning, that makes sense.  I assumed that would be an error since the number of arguments doesn't match (potentially leading to an inconsistent stack).

svaarala:
It will in practice lead to issues but since the call is not direct but goes through a cast from the actual function to a function pointer, the compile doesn't actually detect an invalid call and only warns about the cast.

fatcerberus:
By the way the reason I didn't enable warnings-as-errors for the minisphere build is that clang, but not gcc, gripes about the code (a lot!), and I'm too lazy to fix it. :)

PR #856, opened 2016-07-12
Duk_harray fixes and cleanups

svaarala:
Issues from merge review:

- [x] Debugger changes and testing (especially inspecting array .length)
- [x] Debugger migration notes
- [x] Change duk_pack() and duk_unpack() visibility to internal
- [x] Add overflow check to duk_pack() `count * sizeof(duk_tval)`
- [x] Add helper for value stack range wipe for duk_pack() (may reduce footprint) => XXX note, doesn't seem worth changing now
- [x] Error message "array .length not writable" maybe too verbose, just "not writable"? => Actually "not configurable" is correct when writing .length to a smaller value and a non-configurable element prevents resize from completing
- [x] "property is virtual" error message inconsistent with other error messages ("not writable" etc) => juse use "not configurable" which is accurate in the sense that virtual properties are implicitly non-configurable
- [x] Add `DUK_HTHREAD_GET_TOP()` macro to replace `top = (duk_idx_t) (thr->valstack_top - thr->valstack_bottom)` => separate pull
- [x] `a->length = count` after creating array with predefined size unnecessary in duk_pack()
- [x] putprop comment "may realloc and compact properties and hence change e_idx" inaccurate
- [x] "new length is higher than old length, just update length, no deletions" log comment incorrect, "higher or same" would be correct
- [x] Error traceback C filename handling now reads `DUK_TVAL_GET_STRING()` etc; just copy value using `DUK_TVAL_SET_TVAL()`? => No footprint difference, leave as is
- [x] Mark-and-sweep and refcount `DUK_HOBJECT_IS_ARRAY()` check is unnecessary as there's nothing to mark/refzero process; just comment and remove check?
- [x] Array length write XXX note; calling convention for handling array length change (via value stack) => separate pull
- [x] "Creating an Array with a huge count creates a dense array but with no preallocated array part" comment in test case incorrect
- [x] Check `DUK_DPRINT()` calls in duk_hobject_props.c; should be very few in success paths
- [x] JX/JC handling of gappy arrays, now seems to use "null" for gaps but "undefined" might be more appropriate? => #859
- [x] Uint8Array etc typed array and Object.defineProperty() handling => #860
- [x] Lowmemory configuration: disable non-BMP sources and coroutines?
- [x] duk_error_augment.c cast warning


fatcerberus:
Regarding the error message, "not writable" might not be verbose enough.  If one were to, say:

```javascript
arr.length = foobar.count = 0;
```

...and threw an error, which property does "not writable" apply to?

svaarala:
Right, but the same applies to:

    foo.something = bar.somethingElse = 123;

fatcerberus:
Actually I thought this error was already changed a while back to "property 'foo' of [string representation] not writable"?

svaarala:
Those verbose error messages only apply to some common situations like the base value being invalid and such. In this particular case:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-323-g164fe31)
duk> a = { foo: 123 }
= [object Object]
duk> Object.defineProperty(a, 'foo', { writable: false })
= [object Object]
duk> function foo() { 'use strict'; a.foo = 123; }
= undefined
duk> foo()
TypeError: not writable
    at [anon] (duk_hobject_props.c:4086) internal
    at foo (input:1) strict
```

While this is something to improve on, I don't think Array `.length` should have an error message of its own but should just use the same error message as other properties (verbose or not).

fatcerberus:
Yes, definitely agree it shouldn't be special-cased (outwardly at least :).

fatcerberus:
#703 introduced a (most likely harmless) warning:
```
duk_error_augment.c(190): warning C4267: '=': conversion from 'size_t' to 'duk_int_t', possible loss of data (compiling source file ..\src\shared\duktape.c)
```

svaarala:
Thanks, fixed :)

svaarala:
I'll make a note of improving more of the property error messages when I next touch the property code (there's some pressure to rework it for better performance). BTW, there are >300 throw sites in Duktape internals and not every call site produces an awesome error message (yet) :-)

PR #861, opened 2016-07-13
Add internal DUK_GET_TOP() macro

svaarala:
- [ ] Add internal DUK_GET_TOP() macro
- [ ] Replace `(duk_idx_t) (thr->valstack_top - thr->valstack_bottom)` with it
- [ ] Check whether changing internal `duk_get_top(ctx)` call sites too makes a footprint/performance difference


svaarala:
The actual `duk_get_top()` call inlines as well as the macro so this doesn't seem useful, closing without merge.

PR #864, opened 2016-07-13
Change plain buffers to behave like subset ArrayBuffer instances

svaarala:
Change handling of plain buffers throughout so that they behave like ArrayBuffer instances in Ecmascript code. Because they lack an actual property table, properties are virtualized and as such plain buffers will be a memory efficient subset of full ArrayBuffer objects, similarly to how lightfuncs are a memory efficient subset of actual functions. Plain buffers behave in the C API as they did before (at least for now).

This change should simplify plain buffer behavior for Ecmascript code, and allows a later pull to remove `Duktape.Buffer`, perhaps simplify `duk_hbufobj` vs. `duk_hbuffer` distinction, etc.

Tasks:
- [x] Make changes throughout for plain buffers
- [x] Review diff, debug log levels
- [x] Fix any internal call sites which assume `duk_to_string()` for a buffer results in a string containing the buffer's data (need a `duk_buffer_to_string()` or something, at least internally)
- [x] For consistency, change `plainPointer instanceof Duktape.Pointer` to be true
- [x] Remove plain buffer index enumerability (keep the indices for now)
- [x] Testcases and testcase fixes
- [x] JSON/JX/JC fastpath and buffer types (separate handling in each)
- [x] Add a new testcase illustrating the updated behavior
- [x] Wiki issue / pull for buffer howto changes
- [x] Open issue: should `ToObject(plain)` return `plain` as is (as would happen for an actual ArrayBuffer) or return an actual ArrayBuffer which has the same underlying buffer (similarly to how a lightfunc coerces to an actual, full Function instance) => For now behaves like lightfuncs, object coerces to actual ArrayBuffer
- [x] Open issue: should lightfunc behavior also be changed to match plain buffer behavior in case ToObject() is a no-op? => No change for now
- [x] Open issue: remove virtual index properties from ArrayBuffer (they're now present but non-enumerable) => Keep for now.
- [x] Buffer behavior changes, release notes separate section or separate internal document? => Add to release notes, reference testcase.
- [x] Test debugger class numbers / class name handling
- [x] Test lightfunc built-ins manually
- [x] 2.0 migration notes for buffer changes
- [x] 2.0 migration notes for pointer changes
- [x] 2.0 migration notes for lightfunc changes
- [x] Releases entry

Notes for follow-ups:
- Coercing a plain object into an ArrayBuffer sharing the same storage will be problematic if an explicit underlying buffer reference is removed from ArrayBuffer objects. But because views will retain some indirect reference, maybe those ArrayBuffer instances could have an indirect reference too (in effect there would then be two kinds of ArrayBuffer backend storages).


fatcerberus:
For what it's worth, I'd say it'd be more in the spirit of the Ecmascript spec for `ToObject` to promote the plain buffer to a real ArrayBuffer, similarly to how strings are coerced to String objects, numbers to Number objects, etc.  Conceptually the plain buffer is still a primitive value, not an object--at least I _think_ this is the case, correct me if I'm wrong), its "value" just happens to be a memory address so it appears mutable.  With that in mind, `ToObject` doing boxing makes sense.

svaarala:
Then again, conceptually this change makes plain buffers behave like ArrayBuffers as closely as possible. As there's no plain buffer value in Ecmascript, we then have, for example:

- typeof plainString === 'string'
- typeof objectString === 'object'
- typeof plainBuffer === 'object'   (used to be 'buffer' in Duktape 1.x)
- typeof arrayBuffer === 'object'

In Duktape 1.x that string/String vs. plainBuffer/arrayBuffer analogy would be clear. But with no visible plain type that's no longer clear cut.

For objects it's always the case that `Object(anyObject) === anyObject`; if ToObject() coercion for plain buffers did an actual object conversion internally that wouldn't be true which would break the illusion of plain buffers behaving like ArrayBuffers.

The same issue exists for lightfuncs. With them I ended up making `ToObject(lightFunc)` to promote into an equivalent Function object. That was also a difficult choice because the other alternative also makes sense.

The upside of ToObject() doing an object conversion is that you can promote a plain value into an object variant in pure Ecmascript without custom calls (such as `Duktape.promote(lightFunc)` or something). On the other hand that's not very often necessary, and the coercion behavior may also be confusing and may happen by accident in some cases. Luckily there's no Ecmascript idiom that I'm aware of which might make this likely.

svaarala:
The reason why I want to make this change in behavior is so that the buffer typing could be conceptually simplified: every internal buffer variant (`duk_hbuffer_fixed`, `duk_hbuffer_dynamic`, `duk_hbuffer_external`, and `duk_hbufobj`) would behave the same as far as Ecmascript was concerned. There'd be natural limitations in plain buffers having no property table, for example, but it'd be up to C code what kind of buffer to create.

A related typing change might be to also eliminate this typing difference in the C API as far as possible. This is a bit more difficult; for example `duk_get_type()` currently maps 1:1 to the internal `duk_tval` tag which would then no longer be the case. So I'm not sure if this change is realistic for 2.x.

svaarala:
I'll implement the ToObject(plain) -> actual ArrayBuffer semantics first. That'll match how lightfuncs work (at present, at least) and poses fewest issues in call sites which expect `duk_to_object()` to always result in a concrete object. It also allows a first-cut implementation of accepting plain buffers in place of array buffers: coerce to ArrayBuffer first and then do whatever is done now. So behavior now is:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-339-g20fff2b-dirty)
duk> plain = Duktape.Buffer(4); obj = Object(plain);
= [object ArrayBuffer]
duk> plain
= [object ArrayBuffer]
duk> obj
= [object ArrayBuffer]
duk> plain === obj
= false
duk> plain[3] = 0x61
= 97
duk> obj[3]
= 97
duk> Duktape.info(plain)
= 7,0xe372e0,3,45
duk> Duktape.info(obj)
= 6,0xe37320,2,88,0,0,0,0,0
```

The ArrayBuffer object created uses the plain buffer as its underlying buffer so that the same memory region is shared which one would expect.

fatcerberus:
On an unrelated note, why does `Duktape.info()` return an array?  Wouldn't it be more intuitive (and readable) for it to return an object with named properties?

svaarala:
That'd be more readable but also a lot larger footprint for a very rarely accessed feature :) It might actually make more sense to drop the feature altogether or perhaps expose it to C code only. It's really just access to internals with very few non-debug use cases.

svaarala:
Opened #866 for figuring out what to do with Duktape.xxx() properties, now that it's possible to prune them.

svaarala:
One rather awkward result of this change is that ToString(buffer) no longer results in a string containing the exact bytes of the buffer - instead, it usually results in `[object ArrayBuffer]`. This is not a big deal in Ecmascript code because that's what happens to ArrayBuffers anyway.

However, this also happens for `duk_to_string()` whose semantics are Ecmascript `ToString()` directly. This breaks a lot of internals and a lot of testcases, and presumably some application code. The idiom used in the internals, for example, is to write out bytes into a dynamic buffer (e.g. for some conversion or escaping) and then converting that buffer to a string using `duk_to_string()`.

There's no obvious easy fix - changing `duk_to_string()` to deviate from Ecmascript `ToString()` is somewhat awkward too.

For now, I've added an internal `duk_buffer_to_string()` internal call to handle the internal call sites. Perhaps this could be added to the public API, along with a counterpart `duk_string_to_buffer()`. Maybe an encoding placeholder argument could also be included. In some sense it's also a good change that converting a buffer to a string is an explicit operation rather than just maybe an accidental coercion. Conceptually one should always think about character encoding in such a conversion (even if using the bytes as is it needs to be a conscious choice).

For Ecmascript code this does raise the question: how to convert buffers into strings conveniently? There's no nice standard answers; there are various approaches like `String.fromCharCode.apply()` and using a custom API like https://encoding.spec.whatwg.org/. In Duktape 1.x one could e.g. say `String(Duktape.Buffer(bufferObject))` which would work unless there were active slices involved.

svaarala:
One more option for buffer/string coercion: use the encode/decode calls for that. For example:

    var str = Duktape.dec('none', buf);
    var buf = Duktape.enc('none', str);

This would extend logically to different encodings:

    var str = Duktape.dec('iso-8859-1', buf);
    var buf = Duktape.enc('iso-8859-1', str);

For the C API adding `duk_encode()` and `duk_decode()` might make sense.

svaarala:
Current ways of converting a buffer 1:1 into a string (using the buffer bytes for the string internal representation which allows e.g. creation of internal strings):

```js
// Duktape 1.x, no longer works.
str = String(Duktape.Buffer(...));

// Node.js Buffer String() coercion in Duktape uses bytes as is. Node.js itself
// does UTF-8 parsing and replaces e.g. 0xff with U+FFFD.
str = String(new Buffer('abcd'));

// The Node.js .toString() call can also be used for other buffer types.
str = Buffer.prototype.toString.call(new Uint8Array([ 0xff, 0x61, 0x62, 0x63 ]));

// Duktape.Buffer .toString() also uses bytes as is, and works for other buffer types
// too (not just Duktape.Buffer); but this is going away later before 2.x release.
str = Duktape.Buffer.prototype.toString.call(new Uint8Array([ 0x61, 0x62, 0x63, 0x64 ]));
```

These may change before a 2.x release. I'll probably use the Node.js Buffer .toString() call to fix Ecmascript testcases for now, until the best idiom for such a conversion is clear.

svaarala:
There are a few functional issues left for this; other than that there's some internal refactoring to do for better footprint etc (lightfunc special casing should simultaneously handle plain buffers where possible), and a bunch of FIXMEs to resolve or handle in separate pulls. I'll be cleaning this pull for a couple of days and then merge.

PR #872, opened 2016-07-19
Add ROM built-in lightfunc conversion

svaarala:
For RAM built-ins there's an option to convert built-in functions into lightfuncs to save RAM. Add also support for ROM built-ins to do the same to reduce code footprint for environments where both RAM and ROM are tight.

Tasks:
- [x] Add metadata format for lightfuncs and check that their ROM conversion works (RAM conversion not needed at this point)
- [x] Add a metadata tag to indicate "no automatic lightfunc" conversion for functions that are otherwise eligible but shouldn't be lightfunc converted
- [x] Add genbuiltins.py option `--rom-auto-lightfunc` for automatic lightfunc conversion (specifying lightfuncs manually without this option should also be possible)
- [x] For testing convert all built-ins into lightfuncs, using the same rules as the RAM path
- [x] Check .length handling; metadata already normalized
- [x] Fix typing for lightfunc initializer
- [x] Testrunner coverage: test that ROM built-ins + lightfuncs build
- [x] Documentation, low-memory.rst for now (maybe migrate more concrete instructions to a Wiki page?)
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1715,6 +1715,10 @@ Planned
   and constructor call argument count from 511 to 255, and maximum Ecmascript
   function constant count from 262144 to 65536 (GH-903)
 
+* Add support for converting ROM function property values into lightfuncs in
+  genbuiltins.py to reduce code footprint for ROM-based built-ins and custom
+  bindings; footprint reduction is around 14-15kB on 32-bit targets (GH-872)
+
 * Allow ES6 Annex B unescaped right bracket (']') in regular expressions
   (non-standard before ES6 Annex B), left bracket ('[') not yet supported
   because it needs backtracking (GH-871)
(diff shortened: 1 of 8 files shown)


svaarala:
Potential for ROM footprint savings; ajduk with ROM builtins from master (much larger than a minimal build because there's debugger support, etc):

```
$ size ajduk
   text	   data	    bss	    dec	    hex	filename
 321053	    757	  66696	 388506	  5ed9a	ajduk
```

Ajduk with ROM built-ins converted to lightfuncs:

```
$ size ajduk
   text	   data	    bss	    dec	    hex	filename
 305461	    757	  66696	 372914	  5b0b2	ajduk
```

So, unless there's some bug (and I'm converting too many lightfuncs or similar), the potential footprint savings are around 15kB.

PR #875, opened 2016-07-21
Remove Duktape.Buffer support

### LINKED ISSUE #739: Remove Duktape.Buffer, use ArrayBuffer for coercion instead
The `Duktape.Buffer` object is unnecessary now that typed array support exists and `ArrayBuffer` can take its place.

Duktape.Buffer has two functions worth keeping:

- A plain buffer coerces to a Duktape.Buffer when coerced using ToObject() coercion. This could be changed to ArrayBuffer. (#864 makes plain buffers behave like (subset) ArrayBuffer instances which may invalidate this issue.)
- You can get the plain buffer from any buffer object using a plain function call to Duktape.Buffer, i.e. `var plain = Duktape.Buffer(bufobj)`. This can be provided using a separate function, e.g. `Duktape.plain(bufobj)` or `bufobj.toPlain()`. (#864 might make this unnecessary; there's also no way of coercing a full Function to a lightfunc.)

svaarala:
Remove Duktape.Buffer, ArrayBuffer is now used instead. Fixes #739.

- [x] Rebase once #864 has been merged
- [x] Idiom for getting plain buffer from a buffer object
- [x] Code changes
- [x] Testcase fixes
- [x] Update testcase idioms for string-to-buffer and buffer-to-string
- [x] ArrayBuffer.stringOf, String.fromBuffer, or other options
- [x] Change `duk_buffer_to_string()` to accept any buffer object as input, uses the active byte slice
- [x] Test `duk_push_buffer_object()` for ArrayBuffer with `DUK_USE_BUFFEROBJECT_SUPPORT` disabled
- [x] What to do when bufferobject support is disabled for now; ArrayBuffer is needed for coercion result and will be mandatory when Duktape.Buffer is removed? => duk_push_buffer_object() works for ArrayBuffers, but e.g. ArrayBuffer.prototype.slice() doesn't work; this is liable to change
- [x] Internal documentation updates
- [x] API documentation changes (at least `duk_to_object()`)
- [ ] Website changes (at least custom type behavior, introduction to buffers, etc)
- [ ] Guide section on plain buffer vs. ArrayBuffer behavior (table of operations, results on both, comments); some custom operations like Duktape.dec() result in plain buffers
- [ ] Guide section on lightfunc vs. Function behavior
- [x] Sandboxing document changes re: buffer-to-string coercion
- [x] Update test-dev-buffer-copy-example.js to match guide
- [x] Update dukweb example
- [x] Test debugger and DukLuv proxy
- [x] Test DukLuv basics
- [x] Add Duktape.Buffer polyfill (to the extent possible)
- [ ] Cleanup leftover from plain buffer pull: why does Proxy handler get lfunc promotion?
- [ ] Separate pull for string intern safety (see duk_buffer_to_string() comments)
- [x] 2.0 migration note
- [x] Releases entry


svaarala:
Comments on how to create plain buffers after this change: https://github.com/svaarala/duktape/issues/739#issuecomment-234810188.

svaarala:
How about:

- `ArrayBuffer.allocPlain()`
  + Allocate a plain buffer. Result is always a new buffer (see above).
  + Naming is similar to Node.js Buffer (which uses `Buffer.alloc()` for the newer APIs).
- `ArrayBuffer.plainOf()`
  + Get plain buffer underlying a buffer object.
  + Naming mimics e.g. `Object.prototype.valueOf()` which conceptually gets the "underlying" value of an object (e.g. plain string underlying a String object).

svaarala:
I'm still kind of on the fence whether these should be exposed to Ecmascript by default at all.

For example, creating a plain buffer would only create a fixed buffer (like `Duktape.Buffer`) but should there be support for:

- Creating a dynamic buffer: if so, there'd also need to be a resize primitive (`ArrayBuffer.resizePlain`?).
- Creating an external buffer: cannot really done be memory safely so this one probably only for C.

There's also no Ecmascript support for creating lightfuncs, so maybe it'd be consistent that plain buffer creation, configuration, etc would only be available for C code. The downside would be that every code base creating plain buffers from Ecmascript code would need their own binding and these wouldn't match which impacts code portability between Duktape environments.

fatcerberus:
Would it make more sense for low-memory situations just to have a compiler option to make all ArrayBuffers plain buffers by default, since they are indistinguishable to Ecmascript code now?   Would that even be possible?

svaarala:
The plain buffers are not indistinguishable by any means, they just behave like ArrayBuffers whenever possible. For example, they don't have a property table so `plainBuf.foo = 123` won't actually record any changes (same as `plainString.foo = 123`).

But adding an option that caused built-ins to create plain buffers by default (or perhaps downgrade any results to plain buffers whenever possible) might make sense.

svaarala:
`ArrayBuffer.allocPlain()` and `ArrayBuffer.plainOf()` are now in the pull. I've tried to route all testcase dependencies on these via `util-buffer.js` so that it'll be easy to change them if the bindings still change.

fatcerberus:
@svaarala Huh, you taught me something new.  I was always under the impression that `plainString.foo = 123;` automatically promoted `plainString` to a String object.  Turns out that's not actually the case.

svaarala:
That's what it conceptually does - the resulting object is then just discarded. But it doesn't update any storage locations referencing the plain buffer (or plain string for the string case) - there could be an arbitrary number of those.

svaarala:
The current size of allocPlain() and valueOf() is 321 bytes on x64 which is a bit much considering these bindings are not commonly used (heavy users are likely to do it in C anyway, especially because there's more flexibility in the C API). Couple of possibilities:

- Drop some argument type support from `allocPlain()`; it's now pretty flexible but might be enough to just accept a string and a length variant. The others can be achieved by creating a buffer of certain length and copying,
- Drop the `allocPlain()` binding entirely. You can then create a plain buffer e.g. as `ArrayBuffer.plainOf(new Uint8Array([ 1, 2, 3, 4]))`. A temporary buffer object is created but it'll get garbage collected right away.
- Try to refactor the buffer code so that more code is shared. There are unfortunately minor differences in coercion, negative index handling etc, which makes this a bit painful.

svaarala:
By sharing parts of the Node.js buffer constructor the allocPlain() + plainOf() size is now 153 bytes on x64. plainOf() is 106 bytes of that.

svaarala:
@fatcerberus Regarding that string behavior, this test:

```js
Object.defineProperty(String.prototype, 'foo', {
    set: function () { print('Setter called'); }
});

var plainString = 'foo';
plainString.foo = 123;
print(plainString.foo);
```

is required to print:

```
$ ./duk /tmp/foo.js
Setter called
undefined
```

The setter also sees the String object before it's destroyed. Not terribly useful in practice :-)

fatcerberus:
So now that plain buffers behave like ArrayBuffers, can you use them in any context where a real ArrayBuffer is expected, e.g. as backing for a typed array view?

svaarala:
Yes, they are accepted like ArrayBuffers wherever possible, so for example:

```js
duk> var plain = ArrayBuffer.allocPlain(8);
= undefined
duk> var u8 = new Uint8Array(plain);
= undefined
duk> plain[0] = 123;
= 123
duk> u8[0]
= 123
```

Btw, I'm rewriting the buffer howto for 2.x here: https://github.com/svaarala/duktape-wiki/pull/131 (work in progress). I'm doing it at the same time because it also serves as a view of what the complete picture is right now.

svaarala:
After changing semantics to match ArrayBuffer, `String(plainBuffer)` usually yields `[object ArrayBuffer]` so some new idiom is needed for 1:1 buffer-to-string conversion.

Node.js Buffer.prototype.toString() mostly serves this purpose now but it's rather obscure to depend on, especially when working with typed arrays only.

So I added one more method to ArrayBuffer:

```js
// Create a string whose internal bytes are copied 1:1 from the active buffer
// slice of any buffer type.
var str = ArrayBuffer.stringOf(anyBufferValue);
```

This is better than the previous `String(buf)` because it works for all types (such as Uint32Array); the `String()` coercion in 1.x only worked for plain buffers and Node.js buffers.

The naming is not that great, but I want to avoid `.toString()` as it's quite likely to clash at some point (and is a bit misleading because toString() conversions are often for readable outputs rather than exposing internal values directly).

There's still a separate issue to decide whether direct 1:1 string-to-buffer conversions should be left to C code only; but for the purposes of this pull that mechanism is needed to leave master in a functional intermediate state.

fatcerberus:
I think there's definitely a case to be made for providing buffer-to-string functionality in the base implementation.  This is the one weakness in the TypedArray spec (DataView has no methods for working with strings) which I feel makes the Node.js Buffer implementation superior.

svaarala:
Agree - the somewhat awkward alternative would be to recommend doing something like:

```
$ ./duk
((o) Duktape [linenoise] 1.99.99 (v1.5.0-371-g7269f8a)
duk> var u16 = new Uint16Array([0x6162, 0x6364, 0x6566]);
= undefined
duk> var str = Buffer.prototype.toString.call(u16);
= undefined
duk> str
= badcfe
```

This works in Duktape because the Node.js Buffer also accepts other buffer types and looks up their active byte slice (not a "view slice") - and because no encoding is applied (= bytes are copied 1:1 into the internal representation, which is an inherently a non-portable concept).

With no buffer type mixing one could look up the view `.buffer` and do something like:

```
var str = Buffer.prototype.toString.call(u16.buffer, null,
                                         u16.byteOffset, u16.byteOffset + u16.byteLength);
```

In other words, the active slice must be explicitly specified. This wouldn't work in Node.js though, because there's an encoding so bytes won't end up 1:1 in the internal representation (which is not a portable concept by any means anyway, but useful for *C coders* specifically).

In any case, neither is easy to explain for someone getting into using buffers :-) Some of the other alternatives (some previously discussed) are:

- Adding a 2nd parameter to `String` to allow signaling that a buffer-to-string conversion is to be done.
- Using `String.fromBuffer()` instead of placing the method in `ArrayBuffer`. For some reason I prefer it to be in `ArrayBuffer` (even though it applies to all buffer types which is definitely an argument against that).

fatcerberus:
From a purely aesthetic standpoint and all other things being equal, I tend to prefer `String.fromBuffer()` over `ArrayBuffer.stringOf()`.  The latter looks even more awkward if you add an additional parameter (for the encoding, say).

svaarala:
So for concreteness:

```js
// Alt 1
var str = ArrayBuffer.stringOf(plainBuf);
var str = ArrayBuffer.stringOf(plainBuf, 'utf8');

// Alt 2
var str = String.fromBuffer(plainBuf);
var str = String.fromBuffer(plainBuf, 'utf8');
```

One potentially confusing thing is that it's not necessarily clear whether a `String` built-in method results in a plain string or a String object - one could argue String.fromXXX() would produce a String, not a plain string. But that's of course the case for `ArrayBuffer.stringOf()` too in some sense, maybe not as much as for the String built-in. (This issue actually affects all Ecmascript built-ins because most of them have this dual role, e.g. String(arg) vs. new String(arg), etc.)

The branch now has `.stringOf()` but I'll change it if necessary. Let's see tomorrow :-)

PR #881, opened 2016-07-27
Console extra Proxy and JX improvements

svaarala:
- Tolerate Proxy setup errors in case Proxy throws (= disabled).
- Fix bug in JX encoding error handling.  Incorrect quote marks caused value to be number coerced (+v) rather than string coerced (''+v).


svaarala:
No releases entry because bug not published.

PR #884, opened 2016-07-30
Avoid side effects during string intern processing

svaarala:
When e.g. pushing a string from the contents of a buffer object there's an outside chance that the mechanics of string pushing (allocation) causes a mark-and-sweep, which triggers a finalizer that resizes or reconfigures the buffer object we're trying to string coerce. When the string interning code has allocated space for the string and is ready to copy the data over, the original address and length might no longer be valid.

There's an explicit workaround for this in `duk_push_buffer_object()` but the workaround is awkward: the code first pushes a fixed buffer and copies the buffer object data there to stabilize it (fixed buffers can't be reallocated/relocated) and then proceeds to string intern the fixed buffer contents.

A better solution would be to ensure string interning cannot trigger finalizers or other side effects. This would then make all call sites safe against this potential (but a bit contrived) issue, and would allow `duk_buffer_to_string()` to be simplified.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1793,6 +1793,11 @@ Planned
   .length is larger than the internal array part size (created e.g. when
   calling new Array(10)) without falling back to the slow path (GH-703)
 
+* Fix potential memory unsafe behavior when duk_push_(l)string() data pointer
+  is from a dynamic/external buffer (or any other relocatable data source)
+  and a finalizer side effect resizes/reconfigures the buffer, invalidating
+  the pointer before string table code has time to copy the data (GH-884)
+
 * Internal performance improvement: avoid one extra shift when computing
   reg/const pointers in the bytecode executor (GH-674)
 
--- a/src/duk_api_stack.c
+++ b/src/duk_api_stack.c
@@ -2343,38 +2343,16 @@ DUK_INTERNAL duk_hstring *duk_to_hstring(duk_context *ctx, duk_idx_t idx) {
  */
 DUK_EXTERNAL const char *duk_buffer_to_string(duk_context *ctx, duk_idx_t idx) {
 	void *ptr_src;
-	void *ptr_tmp;
 	duk_size_t len;
 	const char *res;
-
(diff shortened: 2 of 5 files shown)


fatcerberus:
I agree with this pull, string interning is somewhat of a mission-critical mechanism and should avoid side effects as much as possible.

fatcerberus:
By the way why are all the commits from #875 here?

svaarala:
This pull was a follow-up to the Duktape.Buffer removal branch; I rebased this now that the removal is in master. I quite often do that when a previous branch is finished except for some trivia, because it avoids an actual non-trivial rebase later.

svaarala:
I added this to the 1.5.1 fix list too. However, this issue would be very difficult to trigger in practice with the way typical finalizers work.

svaarala:
Added security tag because memory unsafe behavior would result if the issue is triggered in practice.

fatcerberus:
That being so, avoiding side effects is desirable anyway. :)

svaarala:
Well, not necessarily :) One of the benefits of GC side effects is that if we're truly low on memory, running the finalizers and freeing the related objects might allow the string push to succeed. With this change in place that won't be possible so we may run out of memory when there's technically collectable garbage in the heap. But because there's no easy and clean alternative that's an OK trade-off.

svaarala:
As a general goal the GC, especially emergency GC, tries to guarantee that:

- An allocation won't fail if there's collectable garbage in the heap which, if collected, would allow the allocation to succeed.

There are a few exceptions to this now. One is property table resize which prevents finalizers and other side effects to avoid a finalizer operating on the object whose property table we're resizing. This string intern handling is another exception now.

The JSON fast path also does this, but alloc failures there cause a fallback into the slow path so while the behavior is not ideal, it's still correct for the caller, as the slow path will be able to collect the collectable garbage if it's enough to succeed.

svaarala:
Ok, managed to write a test case which illustrates the problem concretely in master (segfault; the same should happen in 1.x) and shows it was fixed in this pull.

PR #886, opened 2016-07-30
Lightfunc and plain buffer as a Proxy handler

svaarala:
There's little reason to accept a lightfunc as a Proxy handler (which is normally an object containing traps). I'm not sure why I added this behavior originally; maybe to match how Function arguments are treated, e.g. Firefox:

```
x = new Proxy({}, function foo() {})
=> Object {  }
```

This makes little sense in practice but works, so maybe lightfuncs also should remain accepted. The only way that makes sense, though, is for the traps to be in Function.prototype which lightfuncs inherit from; this makes much less sense than using a Function which at least has an ordinary own property table.

If the current behavior remains, improve the code commenting a bit to explain why.

### DIFF
--- a/src/duk_bi_proxy.c
+++ b/src/duk_bi_proxy.c
@@ -24,11 +24,14 @@ DUK_INTERNAL duk_ret_t duk_bi_proxy_constructor(duk_context *ctx) {
 
 	/* Reject a proxy object as the handler because it would cause
 	 * potentially unbounded recursion.  (ES6 has no such restriction)
+	 *
+	 * There's little practical reason to use a lightfunc or a plain
+	 * buffer as the handler table: one could only provide traps via
+	 * their prototype objects (Function.prototype and ArrayBuffer.prototype).
+	 * Even so, as lightfuncs and plain buffers mimic their object
+	 * counterparts, they're promoted and accepted here.
 	 */
-	/* XXX: there's little reason to use a lightfunc as the handler
-	 * (which is usually an object with properties for traps).
-	 */
-	h_handler = duk_require_hobject_promote_lfunc(ctx, 1);
+	h_handler = duk_require_hobject_promote_mask(ctx, 1, DUK_TYPE_MASK_LIGHTFUNC | DUK_TYPE_MASK_BUFFER);
 	DUK_ASSERT(h_handler != NULL);
 	if (DUK_HOBJECT_HAS_EXOTIC_PROXYOBJ(h_handler)) {
 		return DUK_RET_TYPE_ERROR;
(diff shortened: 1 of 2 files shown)


fatcerberus:
What does the specification say about this?  SpiderMonkey generally sticks pretty closely to the spec, so it's interesting that a function is accepted there.  What would the function even do in this context?

fatcerberus:
Although now that I think about it functions are just objects in JS, so in that sense the construction makes perfect sense (grammatically).

For lightfuncs I guess the most reasonable behavior would be to coerce the lightfunc to a proper Function and use that as the Proxy handler.  In general I would expect lightfuncs to behave as much like real Function objects as possible, just like plain buffers now act like ArrayBuffers.

svaarala:
Right, that's the current behavior:
- The target must be an object. A lightfunc is promoted to an object. A plain buffer is now also promoted to an ArrayBuffer.
- The handler must be an object; a lightfunc is promoted to an object. A plain buffer is currently not promoted, but actually should.

For neither lightfunc nor a plain buffer the setup makes much sense, because you could only do something like:

```js
var target = { /* ... */ };
Function.prototype.get = function () { print('my get trap'); return 'dummy'; };
var proxy = new Proxy(target, myLightFunc);
print(proxy.foo);
```

This is actually tested for and demonstrated by the lightfunc test: https://github.com/svaarala/duktape/blob/master/tests/ecmascript/test-dev-lightfunc.js#L2817-L2833. The code comment is a bit unclear so I'll clarify that at least.

svaarala:
You should also be able to do the same inheritance trick for a plain buffer but currently can't:

```
duk> var target = {};
= undefined
duk> ArrayBuffer.prototype.get = function () { print('my get trap'); return 'dummy'; };
= function () {"ecmascript"}
duk> var proxy = new Proxy(target, ArrayBuffer.allocPlain(4));
TypeError: object required, found [buffer:4] (stack index 1)
    at [anon] (duk_api_stack.c:1636) internal
    at Proxy () native strict construct preventsyield
```

This should work like it does for ArrayBuffers:

```
duk> var target = { /* ... */ };
= undefined
duk> ArrayBuffer.prototype.get = function () { print('my get trap'); return 'dummy'; };
= function () {"ecmascript"}
duk> var proxy = new Proxy(target, new ArrayBuffer(4));
= undefined
duk> proxy.foo
my get trap
= dummy
```

Technically lightfuncs and plain buffers are of course not bound by Ecmascript behavior: they're custom types so they can behave in whatever way is most useful. In this case it might be more useful to reject them rather than coerce them because there's next to no use case for doing so.

I'll fix that plain buffer vs. lightfunc consistency issue; they should be consistent at least.

fatcerberus:
I agree that it's not very useful in practice to do this.  However from a standards compliance POV it's a bit of a gray area since, even though lightfuncs and buffers are not standard types, they pretend to be standard objects (Function and ArrayBuffer respectively).  So it's hard to say what the correct behavior is.

fatcerberus:
Then again, there's no property table which already has no precedent in the standard, so perhaps rejecting outright is okay.

svaarala:
I guess it doesn't matter much, so I'll just merge the current state where plain buffers behave the same as lightfuncs (i.e. are accepted). There's essentially no code size difference either because there are shared helpers for automatically promoting lightfuncs and plain buffers.

fatcerberus:
I agree, not a particularly pressing concern. :)

Issue #865, opened 2016-07-15
Duktape 2.x buffer behavior changes


svaarala:
This issue tracks the overall buffer behavior changes in Duktape 2.x:
- [x] Change plain buffers to behave like ArrayBuffers for Ecmascript code - #864
- [x] Remove Duktape.Buffer, use ArrayBuffer instead - #739
- [x] With Duktape.Buffer gone, how to get at plain buffer from Ecmascript (if at all; lightfuncs also can only be created from C code)? => ArrayBuffer.allocPlain()
- [x] `DUK_USE_BUFFEROBJECT_SUPPORT`; more ArrayBuffer functionality (like ArrayBuffer.slice?) when bufferobject support disabled? - #889 => For now all Ecmascript bindings are disabled when buffer object support is disabled
- [x] Should Duktape.dec() return a plain buffer or an ArrayBuffer object? Now returns a plain buffer.
- [x] Internal document or Wiki article detailing changes made (separate from buffer howto which should just describe current state) => Release notes

Minor issues / cleanups left over from main merges (to be fixed or promoted into issues):
- [x] Should `duk_buffer_to_string()` accept buffer objects too? => Yes, changed in #875
- [x] Idiom for creating a plain buffer from Ecmascript code; `Uint8Array.allocPlain(4)`
- [x] Idiom for creating a Node.js buffer sharing the same ArrayBuffer (now a copy is made); now follows Node.js Buffer behavior with ArrayBuffer having special (no copy) handling
- [x] Idiom for creating an ArrayBuffer from a Node.js buffer, sharing the same underlying buffer; now follows newer Node.js Buffer behavior so that you can just say `x.buffer`
- [x] Idiom for string-to-buffer in Ecmascript code => `ArrayBuffer.allocPlain('foo')` for plain buffers, no easy way for typed arrays, `new Buffer('foo')` for Node.js Buffers
- [x] Idiom for buffer-to-string in C code
- [x] Idiom for buffer-to-string in Ecmascript code => 1:1 coercion to be removed #1005, Node.js Buffer does an UTF-8 decoding step
- [x] Idiom for creating an internal key in Ecmascript code => not to be supported anymore #1005 by default
[...]

PR #889, opened 2016-07-31
Finalize buffer behavior when bufferobject support is disabled

svaarala:
Some issues when buffer object support is disabled:
- Should plain buffers inherit from Object.prototype or ArrayBuffer.prototype?
- Should `duk_push_buffer_object()` work for e.g. ArrayBuffer but not the other types? Or fail all buffer object types (allows more internals to be stripped)

Discussion: https://github.com/svaarala/duktape/issues/865#issuecomment-236402087

Tasks:

- [x] Only define `duk_hbufobj` when bufferobject support is enabled; fix fallout from that
- [x] Remove object coercion for plain buffers when bufferobject support disabled
- [x] Move "bufferobject supported" sections in duk_bi_buffer.c under same check
- [x] Allow ArrayBuffer.prototype.slice() to work with plain buffers only
- [x] Also allow ArrayBuffer.allocPlain() or maybe disable also ArrayBuffer.prototype.slice() for consistency? => Now also .slice() disabled
- [x] Config option metadata update
- [x] Update testcase documenting the behavior
- [x] Testrunner coverage for no bufferobjects (now covered by stripsize build, which also runs a hello world eval test)
- [x] 2.0 migration notes for disabled bufferobjects
- [x] Releases entry


fatcerberus:
IMO ArrayBuffer is a bit of a worthless type without the view objects, so probably better to disable it entirely with no bufferobject support.

svaarala:
I agree - the main reason I was comtemplating keeping it is that any operation that directly or indirectly object coerces a plain buffer will fail with a `TypeError` if that type doesn't exist. Another reason is that some methods are still useful, e.g. `buf.slice()`. That could be made to work via ArrayBuffer.prototype without enabling all the other buffer object support (IOW it would only accept and return plain buffers).

svaarala:
Some behavior examples:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-399-g2cacf4c-dirty)
duk> a = ArrayBuffer.allocPlain('foobarquux')
Error: unknown error (rc -1)
    at allocPlain () native strict preventsyield
duk> a = Duktape.dec('hex', 'deadbeef12345678')
= [object ArrayBuffer]
duk> Duktape.enc('jx', a)
= |deadbeef12345678|
duk> b = a.slice(3, 7)
= [object ArrayBuffer]
duk> Duktape.enc('jx', b)
= |ef123456|
duk> Object(b)
TypeError: not object coercible
    at [anon] (duk_api_stack.c:2539) internal
    at Object () native strict preventsyield
duk> b.valueOf()
TypeError: not object coercible
    at [anon] (duk_api_stack.c:2539) internal
    at valueOf () native strict preventsyield
```

If `ArrayBuffer.allocPlain()` doesn't work, then maybe it's not all that useful that `ArrayBuffer.prototype.slice()` works. It might be simpler if all Ecmascript bindings are non-functional. Or maybe also allow `ArrayBuffer.allocPlain()` - it's a bit weird you can create plain buffers using `Duktape.dec()` but there's no binding to create one to a specific size.

Re: footprint it's worth noting that the ArrayBuffer, Uint8Array, Buffer, etc objects are still in place and all the methods and other properties are defined. The methods just throw errors. It's possible to reduce footprint further by actually deleting the bindings. This can be done by using the YAML syntax for defining custom bindings (and editing built-in bindings) during `make_dist.py`. When that's done the stub methods will be dropped by a modern linker which will reduce footprint and RAM usage a bit more. I'll add examples of this as part of providing some documented low memory configurations.

svaarala:
Disabled `ArrayBuffer.prototype.slice()` when bufferobjects are disabled. This is maybe easiest to understand: all operations on standard buffer objects will fail. Custom built-ins working with plain buffers only (like Duktape.dec()) continue to work.

svaarala:
At the current state of this pull, plain buffers still inherit from ArrayBuffer.prototype. That can probably stay; it does have a slight impact because ArrayBuffer.prototype must be present in the `thr->builtins[]` array even with buffer objects disabled. There's no support for pruning that array in master at the moment anyway.

fatcerberus:
`String.fromBuffer()` throws with bufferobjects disabled?  Why?

svaarala:
It's not really a clear cut case what operations should throw.

If the basic idea is that buffer-related bindings throw and you only interact with buffers via the C API, then `String.fromBuffer()` should be unsupported in my view.

If it's not, and the idea is that bindings dealing with plain buffers do work, then `String.fromBuffer()` should work. But also `ArrayBuffer.allocPlain()`, `ArrayBuffer.plainOf()` (for a plain buffer argument), `ArrayBuffer.prototype.slice()` (for a plain buffer argument), etc should work, at least for a plain buffer argument.

A middle of the road would be to only support bindings dealing with plain buffers exclusively. That would include `ArrayBuffer.allocPlain()`, but exclude e.g. `String.fromBuffer()` because that also accepts buffer objects.

svaarala:
It's also possible to pick and choose the supported bindings arbitrarily based on some perceived general utility. What matters footprint-wise (which is really the only reason one would need to disable buffer object support) is what internal code those bindings need, i.e. how much stuff is pulled in.

svaarala:
To expand a bit on the above, the only real motivation of disabling buffer object support is for low memory targets where it's acceptable to drop core functionality to gain footprint or RAM space. If one doesn't need buffer objects but isn't working on a low memory target, there's no need to disable the support (it only gains a few kilobytes), one can simply just not use the buffer objects and maybe delete the bindings from the global object.

So for the low memory targets it's probably better if disabling buffer objects removes as much footprint as reasonably possible and one then adds back whatever bindings are needed. If one needs, say, `ArrayBuffer.allocPlain()` it's easy to add it back without introducing any Duktape diffs: the user can simply implement it themselves and replace the (throwing) built-in binding. The reverse means removing functionality which is not disabled by a config option - which in practice means floating a diff against Duktape and then going through to motions to keep it up-to-date. That said, there may be operations that are almost always needed *and* which have a low footprint, so including them might make sense.

So that's the primary concern I think. The secondary concern is that it should be intuitive what operations work and what don't. The easiest policy is that anything providing explicitly buffer-related functionality (all ArrayBuffer, typed array, Node.js Buffer methods, and String.fromBuffer) would be disabled. This is what the branch currently does; Duktape.dec() remains because it's not used to deal *only* with buffers.

Looking at the "pick and choose" approach:
- String.fromBuffer() is a useful binding even when dealing with plain buffers. The related code footprint is low.
- Having the ability to allocate them using ArrayBuffer.allocPlain() is also quite useful, because you can't create plain buffers conveniently from Ecmascript otherwise. The related code footprint is a bit larger than `String.fromBuffer()`.
- ArrayBuffer.plainOf() is quite pointless with only plain buffers in use.
- ArrayBuffer.slice() is quite useful, footprint is relatively low (around 100 bytes).
- Various typed array methods and Node.js `Buffer.prototype` also provide functionality for plain buffers. For example `Buffer.prototype.readUInt32BE.call(plainBuf, 6)` works for plain buffers and is quite useful. But providing these pulls in a lot of code so it seems counterproductive.

I'd be fine with being able to say "nothing works except these: X, X, X", as long as the footprint impact of them is low :-)

fatcerberus:
The main reason removing `String.fromBuffer()` surprises me is because it's still possible for C code to expose a plain buffer to Ecmascript code--which would have no way to stringify it.  I do understand the footprint motivation, though.

svaarala:
True, but equally C code might expose a function accepting a plain buffer with Ecmascript code having no convenient way of creating one?

svaarala:
I think the option name `DUK_USE_BUFFEROBJECT_SUPPORT` is actually a bit misleading. Based on the name only one would reasonably assume that:
- Plain buffers work normally.
- All buffer bindings work, as long as both arguments and results are plain buffers.
- Operations involving buffer objects would error with (probably) TypeError.

The problem with this is that the footprint benefit is marginal, and really the only motivation for disabling bufferobject support is footprint. Also, some operations for plain buffers are implemented internally by promoting them temporarily to an ArrayBuffer, and these would need separate implementations which doesn't seem worth the hassle.

So maybe I could rename the option so that it'd match the stripped semantics better. Not sure what that name would be, need to think about it a bit :)

svaarala:
@fatcerberus So if you consider what Ecmascript code might do with buffers provided by C code: typical operations include e.g.:
- Logging and printing, string conversion is useful there.
- Appending data to an accumulation buffer. Because buffers can't be resized from Ecmascript (and standard buffer objects can't be resized in a standard way at all) this involves creating temporary buffers, and copying data from one buffer to another. Getting a final result from an accumulation buffer involves a slice() operation.
- Operating on the buffer data inline or iterating the data and doing something with it (say printing out byte by byte, converting the data byte by byte to e.g. hex, etc). The basic index properties are enough for this.

If we try to provide a default set of bindings to work in common use cases, I think at least the following would be necessary:

- String.fromBuffer(): string to buffer conversion.
- ArrayBuffer.allocPlain(): allocating new buffers, e.g. for accumulation.
- ArrayBuffer.slice(): getting a slice of e.g. an accumulation buffer, to be passed back to native code.
- Uint8Array.prototype.set(): copying data from one buffer to another without resorting to slow looping. Duktape allows the `this` binding and argument both to be plain buffers, so this works even without views enabled.

ArrayBuffer.slice() could be emulated using ArrayBuffer.allocPlain() and Uint8Array.prototype.set() so it's arguably not necessary.

fatcerberus:
I guess what I find most odd is the functions which have been compiled out for footprint reasons do exist but are just stubs which throw.  I think it'd be less awkward just to have them not exist at all.

I assume there are technical reasons why this is the case though.

svaarala:
I agree wholeheartedly, and it's indeed a technical limitation now: the set of builtins is fixed at dist time. You can use config options to disable stuff but that won't remove the built-in objects or their properties so the functions must be present and to minimize their size, they'll just throw.

There's a way to remove the built-ins too: re-run dist and edit the built-in objects (there's a [YAML syntax](https://github.com/svaarala/duktape/blob/master/util/example_user_builtins1.yaml) for that) to remove the properties too. However, since dist and `duk_config.h` evaluation generally happen at different times it's not very convenient. Also the dist tools and source files are not shipped with the end user distributable now.

The only real solution is to unify the process so that when you create a `duk_config.h` you also simultaneously run the `make_dist.py` process. This is also my intent going forward: the distributable would provide a tool to configure Duktape which would generate `duk_config.h`, all the `make_dist.py` steps, and amalgamation. With this process the built-in object configuration can react to user requested options like `DUK_USE_BUFFEROBJECT_SUPPORT`, Unicode tables can be optimized based on options, etc.

svaarala:
The way disabled built-ins are now handled is a footprint issue but equally seriously it's contrary to the common idiom for detecting what the runtime environment supports. For example, this won't work to detect if ArrayBuffer.prototype.slice is implemented:

```js
if (typeof buf.slice === 'function') {
    console.log('slice works, yay!');
    return buf.slice(3, 7);
}
```

That will now throw which is quite unfortunate. But, it's a separate fix :-)

svaarala:
This is now in master with the behavior as described above (to get this pull out of the way): when `DUK_USE_BUFFEROBJECT_SUPPORT` is disabled the Ecmascript bindings don't work (ArrayBuffer, Node.js Buffer, typed array, String.fromBuffer()). Also didn't rename the option in this pull; these can still be changed for 2.0 via separate pulls.

fatcerberus:
Re: Compiler options - Are `DUK_OPT_*` going to be fully removed for 2.x?

svaarala:
That's the plan, yes.

Issue #834, opened 2016-06-06
Approaches to multi-threading


saghul:
I'm thinking about how to add multi-threading support to sjs and since everything I have come up with is kinda not ideal, I'm opening this issue so someone can elucidate me :-)

The way I see it there are 2 ways to do this: using "green" threads or coroutines (Duktape.Thread) or using native threads.

Duktape coroutines have severe limitations, the biggest one I see is the inability to yield while in C code. If this was a possibility one could write some network server and handle each request in a Duktape thread, and each blocking call would yield, and resume once the result is available. This would be similar to Python's Stackless sibling. The other problem with this is that we wouldn't be exploiting the multiple cores either, so we need to add some native threads into the mix, which brings us to point 2.

Using native threads looks like has another set of problems: the same heap cannot be used from 2 threads. So I guess I could create a new one for each thread, which basically would make these threads look like child processes, since no data sharing is possible. Some serialization mechanism would be necessary, which means code needs to be written with this in mind. Not necessarily bad, but multi-threading without the possibility of data sharing is... unexpected. The upside is that there are no possible race conditions! :-)

Right now I have the following idea in mind: use native threads and create a new heap for each. When creating the thread, I'd get a callable, which I compile to bytecode, and pass it around so I can execute it on the fresh heap. Then use a couple of pipes for communication. One important unsolved problem is that while this allows for threads A and B (assuming A spawned B) to communicate, it doesn't allow for threads B and C (assuming A spawned C) to do so, because they don't know about each other. I could use a global list to store thread identifiers and some associated data, but it sounds quite hackish.

Did I miss anything painfully obvious? Thoughts?

svaarala:
I don't think you missed anything obvious, that's pretty much the big picture.

Regarding coroutines, you can use coroutines for networking if the yielding code is Ecmascript. For example, for a socket read a native call can return "no data right now" and an Ecmascript read implementation can yield until more data is available. This same approach works whenever native code would block.

It'll be possible to relax the C yield/resume limitations a little bit, but it'll require native code to explicitly support a resume for a function. Having the resume work with no awareness of C code means, essentially, switching C stacks which is quite unportable (but would otherwise be nice).

That IPC issues sounds like a classic discovery problem, which often involves some kind of client registry, discovery broadcasts, or similar.

fatcerberus:
[...]

Issue #391, opened 2015-10-03
Relax multithreading requirements to allow multiple non-concurrent native threads


svaarala:
Current threading requirement is that only a single native thread can call into Duktape and another thread may enter only when the previous one exits.

When implementing blocking native calls it'd be useful to relax this so that:
- Thread A enters by calling into a function (`duk_call()` etc).
- Thread A calls a native function which, for example, sends a network request and then blocks waiting for a response (this may just be a blocking platform call or an explicit wait loop).
- At this point thread A has called into Duktape but is not actively executing.
- Thread B calls into Duktape.
- Thread B exits from Duktape.
- Thread A resumes.

Thread B could equally well also call into a blocking native function and allow thread C to execute while that happens. This policy requires that the user code manage and exclude multiple threads accurately, e.g. by using an "execution permission" lock. In the example above thread A would obtain the lock before executing, and release the lock just before blocking. Duktape API calls are only allowed when holding the lock. The policy also implies that the _Duktape threads_ involved in threads A and B (etc) never overlap.

Implementing this policy needs some internal state changes so that thread A can resume even when something has happened in the middle. For example, current code assumes `heap->curr_thread` which needs to be saved and restored or removed entirely.

Tasks to implement first version into Duktape 2.0:
- [x] Add `duk_suspend()` and `duk_resume()` API - #893
- [x] Go through a few alternatives to ensure API won't need changes in 2.x
- [ ] If `duktape.h` internal struct definitions are made visible to application code, remove the internal/external suspend state issue (external state is a char[128] while internal state is smaller)
- [x] API test case
- [x] Website update for http://duktape.org/guide.html#threading
- [x] Wiki howto on threading, with concrete example of suspend/resume - https://github.com/svaarala/duktape-wiki/pull/153
- [x] Releases entry

svaarala:
Since the user code will in most cases need to manage some kind of execution lock, it'd probably be quite OK to require user code to use some API calls to tell Duktape when the thread may change and when thread A resumes.

fatcerberus:
So in this scenario, when Thread A resumes (the blocking call is satisfied), it would have to request the mutex back before returning from the native function?

svaarala:
Yes, before making any Duktape API calls.

svaarala:
[...]

PR #893, opened 2016-08-03
Add suspend and resume API

oleavr:
This allows a native thread to suspend its Duktape thread while for
example calling out to an external blocking API, allowing other native
threads to cooperatively run code on the same heap in the meantime.

It is left up to the application to synchronize its native threads so
only a single native thread executes code at a time, each taking care to
use distinct Duktape threads while doing so.

Related to #834.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -34,6 +34,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Ondřej Jirman (https://github.com/megous)
 * Saúl Ibarra Corretgé <saghul@gmail.com>
 * Jeremy HU <huxingyi@msn.com>
+* Ole André Vadla Ravnås (https://github.com/oleavr)
 
 Other contributions
 ===================
@@ -71,7 +72,6 @@ bugs, provided ideas, etc; roughly in order of appearance):
 * Michael Drake (https://github.com/tlsa)
 * https://github.com/chris-y
 * Laurent Zubiaur (https://github.com/lzubiaur)
-* Ole André Vadla Ravnås (https://github.com/oleavr)
 
 If you are accidentally missing from this list, send me an e-mail
 (``sami.vaarala@iki.fi``) and I'll fix the omission.
--- a/src/duk_api_heap.c
+++ b/src/duk_api_heap.c
@@ -4,6 +4,15 @@
 
 #include "duk_internal.h"
 
+typedef struct duk_internal_thread_state duk_internal_thread_state;
+
+struct duk_internal_thread_state {
+	duk_ljstate lj;
+	duk_bool_t handling_error;
+	duk_hthread *curr_thread;
+	duk_int_t call_recursion_depth;
+};
+
 DUK_EXTERNAL
 duk_context *duk_create_heap(duk_alloc_function alloc_func,
                              duk_realloc_function realloc_func,
@@ -67,6 +76,55 @@ DUK_EXTERNAL void duk_destroy_heap(duk_context *ctx) {
 	DUK_ASSERT(heap != NULL);
 
 	duk_heap_free(heap);
+}
+
+DUK_EXTERNAL void duk_suspend(duk_context *ctx, duk_thread_state *state) {
+	duk_hthread *thr = (duk_hthread *) ctx;
+	duk_internal_thread_state *snapshot = (duk_internal_thread_state *) state;
+	duk_heap *heap;
+	duk_ljstate *lj;
+
+	DUK_ASSERT_CTX_VALID(ctx);
+	DUK_ASSERT(thr != NULL);
+	DUK_ASSERT(thr->heap != NULL);
+
+	heap = thr->heap;
+	lj = &heap->lj;
+
+	duk_push_tval(ctx, &lj->value1);
+	duk_push_tval(ctx, &lj->value2);
+
+	DUK_MEMMOVE((void *) &snapshot->lj, (const void *) lj, sizeof(duk_ljstate));
+	snapshot->handling_error = heap->handling_error;
+	snapshot->curr_thread = heap->curr_thread;
+	snapshot->call_recursion_depth = heap->call_recursion_depth;
+
+	lj->jmpbuf_ptr = NULL;
+	lj->type = DUK_LJ_TYPE_UNKNOWN;
+	DUK_TVAL_SET_UNDEFINED(&lj->value1);
+	DUK_TVAL_SET_UNDEFINED(&lj->value2);
+	heap->handling_error = 0;
+	heap->curr_thread = NULL;
+	heap->call_recursion_depth = 0;
+}
+
+DUK_EXTERNAL void duk_resume(duk_context *ctx, const duk_thread_state *state) {
+	duk_hthread *thr = (duk_hthread *) ctx;
+	const duk_internal_thread_state *snapshot = (const duk_internal_thread_state *) state;
+	duk_heap *heap;
+
+	DUK_ASSERT_CTX_VALID(ctx);
+	DUK_ASSERT(thr != NULL);
+	DUK_ASSERT(thr->heap != NULL);
+
+	heap = thr->heap;
+
+	DUK_MEMMOVE((void *) &heap->lj, (const void *) &snapshot->lj, sizeof(duk_ljstate));
+	heap->handling_error = snapshot->handling_error;
+	heap->curr_thread = snapshot->curr_thread;
+	heap->call_recursion_depth = snapshot->call_recursion_depth;
+
+	duk_pop_2(ctx);
 }
 
 /* XXX: better place for this */
--- a/src/duk_api_public.h.in
+++ b/src/duk_api_public.h.in
@@ -32,6 +32,7 @@ extern "C" {
  *  in Duktape web documentation.
  */
 
+struct duk_thread_state;
 struct duk_memory_functions;
 struct duk_function_list_entry;
 struct duk_number_list_entry;
@@ -40,6 +41,7 @@ struct duk_time_components;
 /* duk_context is now defined in duk_config.h because it may also be
  * referenced there by prototypes.
  */
+typedef struct duk_thread_state duk_thread_state;
 typedef struct duk_memory_functions duk_memory_functions;
 typedef struct duk_function_list_entry duk_function_list_entry;
 typedef struct duk_number_list_entry duk_number_list_entry;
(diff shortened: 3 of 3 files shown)


svaarala:
Good stuff, thanks :) Couple of comments and issues:

Use `DUK_MEMMOVE()` instead of `memmove()`; it's abstracted because it needs to be retargeted on some platforms. This causes the code policy check to fail now.

The external state is a 128-byte char array and not an exact match to the internal state. I can't think of a cleaner way right now unfortunately. I have a pending branch for 2.0 where the headers are rearranged so that all internal structures are visible also when the application does `#include "duktape.h"`. This will allow some optimizations (e.g. `duk_get_top()` can be a macro which peeks directly into the internal structure instead of making a call, value pushing can operate directly on the value stack for performance optimized builds) but will also allow that suspend state to be defined without this internal/external inconvenience. We can fix this later.

Please add yourself to AUTHORS.rst; you're already on the second list but not the code/documentation contributor list.

I can add API documentation entries etc as separate pulls. This probably need a small guide note on the threading section and a Wiki Howto for threading which would explain both the default limitations and what suspend/resume will allow (and what it won't).

svaarala:
So as a basic idiom, assuming a mutex for thread exclusion, one would do something like:

```c
duk_ret_t blocking_http_request(duk_context *ctx) {
    duk_thread_state suspend_state;

    /* ... figure out parameters etc ... */

    duk_suspend(ctx, &suspend_state);
    release_mutex();

    /* At this point all Duktape API calls are unsafe, but it's
     * still safe to reference string data pointers. Also buffer
     * data pointers can be OK -if- you know, for your own code,
     * that the buffers are not resized/reconfigured by other
     * threads.
     */

    do_blocking_http_request(uri, ...);

    acquire_mutex();
    duk_resume(ctx, &suspend_state);

    /* ... figure out return values ... */

    return 1;
}
```

oleavr:
Good stuff, thanks :) Couple of comments and issues:

> Use DUK_MEMMOVE() instead of memmove(); it's abstracted because it needs to be retargeted on some platforms. This causes the code policy check to fail now.

Oops! Fixed. I noticed that all existing call-sites perform explicit casts to `void *` and `const void *`, so I did that as well. I'm assuming this is for compilers that try to be smart and end up making bad assumptions about alignment?

> The external state is a 128-byte char array and not an exact match to the internal state. I can't think of a cleaner way right now unfortunately. I have a pending branch for 2.0 where the headers are rearranged so that all internal structures are visible also when the application does #include "duktape.h". This will allow some optimizations (e.g. duk_get_top() can be a macro which peeks directly into the internal structure instead of making a call, value pushing can operate directly on the value stack for performance optimized builds) but will also allow that suspend state to be defined without this internal/external inconvenience. We can fix this later.

Sounds good. I'm also not super-happy about this part, so it'll be great to get rid of this once the internal structures are externally visible.

> Please add yourself to AUTHORS.rst; you're already on the second list but not the code/documentation contributor list.

Done.

> I can add API documentation entries etc as separate pulls. This probably need a small guide note on the threading section and a Wiki Howto for threading which would explain both the default limitations and what suspend/resume will allow (and what it won't).

That would be great.

> Similarly an API testcase is needed, I can try to figure one out.

Awesome!

oleavr:
> So as a basic idiom, assuming a mutex for thread exclusion, one would do something like:
> (...snip...)

+1. Looking great!

svaarala:
The venerable #391 now tracks the changes for 2.0.

svaarala:
Since a new API is introduced any alternatives or variants to the public API syntax should also be considered. The internals don't matter that much.

So, for completeness: another alternative to the current approach where the caller provides a (usually stack allocated) storage would be for `duk_suspend()` to allocate a buffer for the suspend state and push it on the value stack for resume. The call site would then simply be:

    duk_suspend(ctx);

    ....

    duk_resume(ctx);

This would avoid adding a suspend state type and the call site would be simpler. Also if no suspend state is needed later it'd be cleaner to remove it. The downside would be that a suspend/resume would involve a temporary fixed buffer. It's not a big deal but using the stack for the suspend state is more efficient.

Another alternative would be to provide a define for the suspend state size and the caller would then allocate it from the stack like:

    char state[DUK_SUSPEND_STATE_SIZE];

This is worse from a typing point of view so I don't see upsides for this.

Finally, one alternative would be that there was no need for any suspend/resume API calls. That's not achievable on the short term and maybe not even the medium term (one issue being uncaught error handling) so for me this doesn't seem like a relevant option for 2.0.

svaarala:
@andoma If you have time, it'd be great to hear if you could test this out or see if it would work for you.

There's probably going to be a 1.6.0 release from `v1-maintenance` branch and this should be relatively simple to port to 1.x if that makes it easier for you to test.

andoma:
Looks great!

Just want to verify that I do the right thing here.

To do proper suspend/resume the calls need to happen on different duk threads too, right?
So basically when my code jumps into duktape it should look something like this:

```
udata *u = get_udata(superctx);
pthread_mutex_lock(&u->mutex);
duk_context *ctx = duk_get_context(superctx, duk_push_thread(superctx));
// push all the args on ctx
int rc = duk_pcall(ctx, ...);
if(rc) {
...
} 
duk_set_top(superctx, 0); // .. and the created thread is GC'd eventually
pthread_mutex_unlock(&u->mutex);
```

and then inside my blocking I/O call

```
duk_thread_state suspend_state;

udata *u = get_udata(ctx);
const char *url = duk_get_string(ctx);
duk_suspend(ctx, &suspend_state);
pthread_mutex_unlock(&u->mutex);

buffer *b = http_request(url, ...);
pthread_mutex_lock(&u->mutex);
duk_resume(ctx, &suspend_state);

// push return arg

buffer_free(b);
return 1;
```
I assume udata is shared between the thread-contexts using the same heap.

andoma:
Hm, clearly if two of these things execute `duk_set_top(superctx, 0);` will make both threads unreachable by GC so I suppose I need to store the thread-value in some object until duk_pcall() returns

svaarala:
This will need to be explained in the Wiki in some detail because it's not obvious what the caller should do. Couple of observations:

- The reachability of all threads must be guaranteed by the caller at all times.
- The C callstack of the suspended thread must not be unwound while suspended, otherwise one can't resume (and all bets would be off anyway).
- One obvious way to use this feature is when a native thread blocks and another native thread resumes. This implies some form of locking to control access to Duktape API. If the native threads are cooperating rather than pre-emptive, a simpler solution is possible.
- Another less obvious way is for the suspended thread to resume a different Duktape thread (using the same native thread) while the operation is pending. This runs the risk of unbounded C recursion if that thread also suspends, resumes another thread etc; this is ultimately up to the application to control.

andoma:
> @andoma If you have time, it'd be great to hear if you could test this out or see if it would work for you.
> 
> There's probably going to be a 1.6.0 release from v1-maintenance branch and this should be relatively simple to port to 1.x if that makes it easier for you to test.

Tested it (just applied the patch on 1.5.0) and it seems to work perfectly fine for my needs.

saghul:
@oleavr Thanks for making this! I'm the one who opened #834 some time ago. If you don't mind, I have a quick question: if my understanding is correct, with this we would have a model similar to that of CPython or the MRI, a GIL: only one thread can be executing JS code at a time, and syscalls would be some of the "yielding points", where one thread calls `duk_suspend` and waits until the syscall is done, to then call `duk_resume`. How does this model work when multiple threads execute JS code concurrently? CPtyhon, for instance, lets the GIL go every X bytecode instructions, so syscalls are not the only yielding point. How would we go about that?

Thanks!

svaarala:
@saghul There's is no locking here as far as Duktape is concerned - it's up to the application to control access to Duktape API. So there's no overhead or concurrency for interpretation. There are upsides and downsides to this, of course.

svaarala:
@andoma Good to hear :) So now we need to decide which API variant to adopt, and I'll then apply it as a patch to 1.6.0 release for those who don't want to jump into 2.0 just to get this feature.

saghul:
@svaarala I know, let me rephrase :-) In order to allow multiple threads to execute JS code on the same Duktape context one can use these APIs, for example around blocking C code, so another thread can continue with its business. But how would this work if there is no blocking C code around for one thread to suspend execution? Is there a way to register a callback which will be called every so often and allows for suspending execution of JS code wherever it might be? Thanks!

svaarala:
@andoma Good to hear :) So I guess the main open issue is which API variant to adopt, and of course documentation and such.

Regarding the API approaches, one thing that came to mind is that it's important that the C call stack of the suspending thread is not unwound, i.e. a `duk_resume()` must always happen from the same C call stack context as the corresponding `duk_suspend()`.

The explicit state pointer kind makes it harder to do that wrong because the most natural place for the state is a local variable, so it's also quite natural for both `duk_suspend()` and `duk_resume()` to be placed in the same function.

svaarala:
I'll merge this pull (not the alternate approach) pretty soon unless there are some voices to the contrary :) This will then be in master, at least as an experimental feature. API changes before 2.0 release would still be possible, if there are good reasons to do so.

oleavr:
@svaarala Awesome! Thanks a lot for looking into this. Totally agree that this stack-based approach makes it more intuitive to use the API correctly.

andoma:
> @svaarala Awesome! Thanks a lot for looking into this. Totally agree that this stack-based approach makes it more intuitive to use the API correctly.

+1

fatcerberus:
> stack-based approach

*Is* it stack-based?  `duk_suspend()` and `duk_resume()` take a pointer to a state structure.  It never touches the value stack from what I can tell.

oleavr:
> Is it stack-based? duk_suspend() and duk_resume() take a pointer to a state structure. It never touches the value stack from what I can tell.

I meant the *native* stack, sorry, confusing. :-) And yeah, it does touch the value stack because of `value1` / `value2`, but that's the case for both approaches.

fatcerberus:
Ah, that makes more sense then, thanks.

kattechou:
> @svaarala I know, let me rephrase :-) In order to allow multiple threads to execute JS code on the same Duktape context one can use these APIs, for example around blocking C code, so another thread can continue with its business. But how would this work if there is no blocking C code around for one thread to suspend execution? Is there a way to register a callback which will be called every so often and allows for suspending execution of JS code wherever it might be? Thanks!

@saghul Do you get an answer for that? Can you share it? Thanks. (To be clear, liblua has API suited for my need - lua_yield() and lua_resume())

saghul:
I don't rtemember, TBH... I'm no longer working on the project which prompted that question.

PR #894, opened 2016-08-03
Add suspend and resume API (alternative approach)

oleavr:
This explores a value stack-based approach to #893.

(Not yet tested, this is just a rough draft.)


svaarala:
Not merged, original approach chosen.

PR #895, opened 2016-08-04
Fix lightfunc constructable handling

svaarala:
Lightfuncs were rejected as constructor call targets. Fix so that lightfuncs are allowed both as direct and indirect (via bound function chain) constructor call targets. Also added to 1.5.1 release.


fatcerberus:
Is this a recent regression?  Seems like it'd have shown up sooner if it were a long-standing bug.

svaarala:
This is also present in 1.5.0. Lightfuncs are only used in low memory environments and rarely as constructors so it's gone unnoticed.

PR #896, opened 2016-08-04
Simplify opcodes for call handling a bit

svaarala:
Some of the opcodes support arbitrary source/target registers but in practice the compiler only uses them for "in place" initialization. Rework the opcodes a little bit for simplicity. There's a small footprint and performance improvement with the change.

Tasks:
- [x] CSREG improvements
- [x] CSPROP improvements
- [x] CSVAR improvements
- [x] NEW improvements
- [x] CALL improvements
- [x] Figure out what to do with new gaps in the direct dispatch space; perhaps move some extra opcodes there (maybe in a separate pull) => Separate pull, also check unused opcode field space (reduces shuffling)
- [x] Add some perf tests
- [x] Fix API testcase, dump/load output changes
- [x] Measure performance difference for call handling
- [x] Update bytecode metadata
- [x] Test bytecode forms in debugger bytecode dump
- [x] Note constructor and ordinary function call 255 argument limit (dropped from 511)
- [x] Add testcase for constructor call argument limit
- [x] Releases entry


svaarala:
Code footprint improvement varies quite a bit, for low memory x86 builds it is about 200-300 bytes.

svaarala:
Performance improvement for calls with a small amount of arguments is around 10%.

fatcerberus:
200-300 bytes footprint improvement is not bad at all for a change which is functionally transparent (?).

svaarala:
It's almost transparent: there are some limit changes like maximum number of constructor arguments dropping from 511 to 255. This is just a side effect of switching the `nargs` field from a 9-bit opcode field to an 8-bit one. It'd be relatively easy to maintain the old limit or extend it to 1023, but that would then compete with maximum number of registers in a value stack frame (which is now 262144).

svaarala:
Some preliminary perf tests:

```
test-call-basic-1.js                : duk.O2.call  8.49 duk.O2.master  9.42 duk.O2.150  9.58
test-call-basic-2.js                : duk.O2.call  8.55 duk.O2.master  9.56 duk.O2.150  9.50
test-call-basic-3.js                : duk.O2.call 14.06 duk.O2.master 15.03 duk.O2.150 15.17
test-call-basic-4.js                : duk.O2.call 35.25 duk.O2.master 36.54 duk.O2.150 38.45
test-call-native.js                 : duk.O2.call 13.85 duk.O2.master 14.24 duk.O2.150 14.06
test-call-prop.js                   : duk.O2.call  5.38 duk.O2.master  5.93 duk.O2.150  5.96
test-call-reg-new.js                : duk.O2.call  7.54 duk.O2.master  6.58 duk.O2.150  6.83
test-call-reg.js                    : duk.O2.call  3.19 duk.O2.master  3.62 duk.O2.150  3.70
test-call-var.js                    : duk.O2.call  8.47 duk.O2.master  8.50 duk.O2.150  8.69
```

The `test-call-reg-new` is slower with the changes in the branch right now, the test is for the NEW opcode. The change in the branch has smaller value stack use because it reuses the value stack more efficiently. But performance is lower for a constructor call with relatively few arguments (and higher when the constructor call has a lot of arguments). I'll probably revert that change or maybe use the smaller footprint version with `DUK_USE_PREFER_SIZE`. A similar trade-off exists for CALL which I haven't touched yet.

svaarala:
Final code footprint difference for minimal/stripped x86 build is around 500 bytes. I'll work a bit more on the CSVAR opcode and some other opcode improvements which have been pending.

PR #900, opened 2016-08-09
Improve assembly generated for interrupt counter in executor dispatch loop

svaarala:
Making the interrupt decision based on the updated value (rather than old value) seems to work better on at least x86/x64.

Tasks:
- [ ] Change interrupt counter mechanism in executor
- [ ] Fix counter init locations, hide the init behind a macro for further tweaks
- [ ] Test interrupt fixup handling, i.e. accurate counting
- [ ] Releases entry


svaarala:
No measurable impact, closing as unmerged.

PR #902, opened 2016-08-10
Dispatch DUK_OP_ADD separately

svaarala:
Just a small tweak to opcode dispatch: addition is handled using a different helper than other binary arithmetic so dispatch it separately to eliminate an awkward comparison.

- [x] Change dispatch
- [ ] Add perf tests (fastint or not? now mixed)
- [ ] Releases entry

### DIFF
--- a/src/duk_js_executor.c
+++ b/src/duk_js_executor.c
@@ -3053,7 +3053,17 @@ DUK_LOCAL DUK_NOINLINE void duk__js_execute_bytecode_inner(duk_hthread *entry_th
 			break;
 		}
 
-		case DUK_OP_ADD:
+		case DUK_OP_ADD: {
+			duk_small_uint_fast_t a = DUK_DEC_A(ins);
+			duk_tval *tv_b;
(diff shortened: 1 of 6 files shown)


fatcerberus:
Looking at the diff, it looks like it was done this way originally for footprint reasons?

svaarala:
Yes, the trade-off is not very good however: ~50 bytes but it affects all dispatched arithmetic operations. There's not yet a "prefer footprint" vs. "prefer speed" in the executor so this is the better default here I think.

svaarala:
The impact is quite small though - I've been prototyping a single level dispatch option (8 bit opcode) which may replace the current two level dispatch (6 bit main opcode field) so this pull may not matter if that gets merged later.

svaarala:
Copied the test from this pull into #903; that seems promising for a later merge so I'll probably close this without merging in a bit, as the change here is irrelevant if #903 is merged.

svaarala:
Closing as unmerged.

PR #907, opened 2016-08-16
JX formatting for duk result

### LINKED ISSUE #905: Print flattens arrays
evaluate (e.g. in the REPL): 

duk> [[1,2],[3,4,[5,6]]];
= 1,2,3,4,5,6

Why does duktape flatten arrays on output? This seems confusing.

svaarala:
Attempt JX formatting for `duk` interactive eval results which is more useful than string coercion for e.g. arrays and objects (see #905). When JX fails, e.g. due to JX support missing, a circular object, etc, `ToString()` coercion is used as before.

Example:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-434-g043d88a-dirty)
duk> 123
= 123
duk> [1,2,3,[4,5]]
= [1,2,3,[4,5]]
duk> {foo:123}
= 123
duk> var A = {}; var B = {}; A.ref = B; A
= {ref:{}}
duk> B.ref = A; B
= [object Object]
```

Fixes #905.

### DIFF
--- a/examples/cmdline/duk_cmdline.c
+++ b/examples/cmdline/duk_cmdline.c
@@ -327,6 +327,11 @@ static int wrapped_compile_execute(duk_context *ctx, void *udata) {
 		 *  safely using duk_safe_to_string().
 		 */
 
+		duk_push_global_stash(ctx);
+		duk_get_prop_string(ctx, -1, "dukFormat");
+		duk_dup(ctx, -3);
+		duk_call(ctx, 1);  /* -> [ ... res stash formatted ] */
+
 		fprintf(stdout, "= %s\n", duk_to_string(ctx, -1));
 		fflush(stdout);
 	} else {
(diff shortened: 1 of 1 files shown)


svaarala:
What would actually be useful for `duk` is some form of true inspection of the result. For example, the class of the result would be useful information for buffer values. The result for buffers will now be JX formatted, e.g.:

```
duk> new Uint8Array([1,2,3,4])
= |01020304|
```

instead of the previous:

```
duk> new Uint8Array([1,2,3,4])
= [object Uint8Array]
```

svaarala:
I'm going to leave this open for a bit in case there are better ideas. The change is not always an improvement, e.g. for functions:

```
duk> Math.cos
= {_func:true}
```

while `ToString()` coercion is:

```
duk> print(Math.cos)
function cos() {"native"}
```

The JX encoding of a function doesn't include it's name etc; JX is not an inspection format which is really what `duk` would need.

Because there are a bunch of different approaches to inspection the best longer term change would be to figure out a useful format for `duk`, and then ensure that the public API has the necessary inspection support to implement that using the public API. DukLuv, for example, implements inspection in Ecmascript and it uses e.g. `Duktape.info()` for some of the inspection stuff which is not always very clean.

In any case, within the scope of this pull the question is: is this change a net improvement over master or not.

fatcerberus:
One thing that might be useful is to look at the source for Node.js `util.inspect()` as I believe that's what's used in the REPL to generate output for expression results.

svaarala:
That's probably too "deluxe" to be included in the core. But it is useful to figure out what gaps there are in the API to implement a function like that yourself. And I wouldn't mind if Duktape distributable included a useful "deluxe" inspection function, as long as it's implemented on top of the public API.

Re: DukLuv, despite some of the difficulties in inspection, the result is pretty good I think. It displays buffer data (truncated), has coloring, etc. These features don't belong in the core functionality, but the ability to implement them via the public API does.

fatcerberus:
Agreed, robustness of the public API is paramount.  Low-level inspection and debugging type stuff is one area where Duktape is somewhat lacking I think.  Recent additions such as AppRequest and `duk_debugger_pause()` have gone a long way in this regard, but there's still a lot of room for improvement IMO.

svaarala:
I'm merging this because there were no objections and it seems like a (small) improvement over master. The proper improvement would be inspection support, #925.

PR #913, opened 2016-08-18
Avoid signed/unsigned conversion in bias handling

### LINKED ISSUE #911: Segfault in Duktape master (ec11270)
Here's the stack listing at the time of the crash.

```
 	spherun.exe!duk__js_execute_bytecode_inner(duk_hthread * entry_thread, unsigned __int64 entry_callstack_top) Line 2642	C
 	spherun.exe!duk_js_execute_bytecode(duk_hthread * exec_thr) Line 2349	C
 	spherun.exe!duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 1575	C
 	spherun.exe!duk_handle_call_protected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 1060	C
 	spherun.exe!duk_pcall(duk_hthread * ctx, int nargs) Line 134	C
>	spherun.exe!main(int argc, char * * argv) Line 253	C
```

This happens on minisphere startup.  The only thing I did leading up to the crash is to execute the main script, then call the global `game()` function.  During execution of that function, Duktape segfaults.

It looks like the segfault is on this line:
```c
		ins = *curr_pc++;
```

Hovering on `curr_pc`, Visual Studio lists the value after dereferencing as "???".  Expanding that shows `<unable to read memory>`, indicating `curr_pc` seemingly went WAY out of bounds (if it were only slightly past the end, it would have shown a guard value instead of "???").

svaarala:
This might be related to #911, I'll confirm that once the Windows build clears.

Fixes #911.

### DIFF
--- a/src/duk_js_executor.c
+++ b/src/duk_js_executor.c
@@ -2727,7 +2727,7 @@ DUK_LOCAL DUK_NOINLINE void duk__js_execute_bytecode_inner(duk_hthread *entry_th
 		case DUK_OP_LDINT: {
 			duk_int32_t val;
 
-			val = (duk_int32_t) (DUK_DEC_BC(ins) - DUK_BC_LDINT_BIAS);
+			val = (duk_int32_t) DUK_DEC_BC(ins) - (duk_int32_t) DUK_BC_LDINT_BIAS;
 			duk_push_int((duk_context *) thr, val);
 			DUK__REPLACE_TOP_A_BREAK();
 		}
@@ -2744,7 +2744,7 @@ DUK_LOCAL DUK_NOINLINE void duk__js_execute_bytecode_inner(duk_hthread *entry_th
 			duk_tval *tv1;
 			duk_int32_t val;
 
-			val = (duk_int32_t) (DUK_DEC_BC(ins) - DUK_BC_LDINT_BIAS);
+			val = (duk_int32_t) DUK_DEC_BC(ins) - (duk_int32_t) DUK_BC_LDINT_BIAS;
 			tv1 = DUK__REGP_A(ins);
 			DUK_TVAL_SET_I32_UPDREF(thr, tv1, val);  /* side effects */
 			break;
@@ -3795,7 +3795,11 @@ DUK_LOCAL DUK_NOINLINE void duk__js_execute_bytecode_inner(duk_hthread *entry_th
 		}
 
 		case DUK_OP_JUMP: {
-			curr_pc += DUK_DEC_ABC(ins) - DUK_BC_JUMP_BIAS;
+			/* Note: without explicit cast to signed, MSVC will
+			 * apparently generate a large positive jump when the
+			 * bias-corrected value would normally be negative.
+			 */
+			curr_pc += (duk_int_fast_t) DUK_DEC_ABC(ins) - (duk_int_fast_t) DUK_BC_JUMP_BIAS;
 			break;
 		}
 


svaarala:
All right, seems to fix the issue in my Wine test. Interesting issue - the issue was mixed signed/unsigned arithmetic on values which had a different bit size in MSVC.

svaarala:
@fatcerberus Could you confirm this fixes your Minisphere issue?

svaarala:
One of the many testcase coverage improvement issues is running a small test set for all variant builds. I don't want to run the full test set at least because it takes minutes per each configuration. I'd need a whole lot more VMs to do that :)

But, it'd be nice to have some subset of tests which exercised all basic features summarily and would be quick to run for all configuration. For example, for executor, a testcase which was designed to run every opcode in the dispatch table would be useful. In practice it might be enough to cherry pick a set of 10-20 tests from existing testcases to serve as a quick test with better coverage. The easiest short term alternative would be to write a single test which can be executed for every variant, and whose output can be checked to match byte for byte for success/failure determination.

svaarala:
So what I think happened here is that in the expression:

    DUK_DEC_ABC(ins) - DUK_BC_JUMP BIAS

- DUK_DEC_ABC(ins) yields a `duk_uint32_t` via `duk_instr_t`. So, result is 32-bit unsigned.
- DUK_BC_JUMP_BIAS is `(1L << 23)` which becomes a `long`, which would be the same as `duk_int32_t` on MSVC. Result is 32-bit signed.

There's mixed arithmetic on values of equal rank and they get both converted to unsigned for the computation. The result is an unsigned 32-bit value.

That unsigned 32-bit value is added to a 64-bit pointer which causes the problem. If pointers were also 32-bit, the unsigned nature of the addition wouldn't matter because the end result would be the same as the pointer would wrap.

On MSVC x86 there's again no problem because the 32-bit unsigned value is added to a 32-bit pointer so that the result is as intended as the pointer wraps.

This doesn't manifest on Linux because x64 has 64-bit integers so that `(1L << 23)` is 64 bits and the unsigned DUK_DEC_ABC(ins) value gets converted to a 64-bit *signed* value so that the result is 64-bit signed. Pointers are also 64-bit so even if the result were unsigned it wouldn't matter.

Anyway, the fix is to ensure the arithmetic is computed with signed values. I also fixed LDINT/LDINTX bias handling to ensure the arithmetic is explicitly signed there too.

svaarala:
Just out of curiosity, checked the MSVC x86 in master works via Wine, so the theory above seems correct :)

fatcerberus:
I'll check in a bit.  I kind of overslept :)

fatcerberus:
minisphere no longer crashes with a Duktape build against this branch.  Looks like you nailed it. :)

And yes, Windows is very odd.  It's LLP64, which differs from every other major platform.  Apparently MS did this to maintain source compatibility, because the Windows headers assume that a `long` is 32 bits.  This _despite_ the fact that said headers include typedefs for `INT`, `LONG`, etc. and already recommended that Win32 code use those typedefs instead of the built-in C types.

Source compatibility was a pipe dream in practice anyway - most Win32 code of the time still had to be rewritten for x64 because developers assumed pointers were 32 bits (and thus could be safely cast to `unsigned long`) :-\

svaarala:
So it no longer crashes - but does it work? :grin:

fatcerberus:
Well, there's no odd behavior that I could see and the game executes normally, so... I guess so? :)

svaarala:
Hehe, that's what I wanted to know :-) I'll merge this and let's see if anything else turns up.

Issue #914, opened 2016-08-18
Finalized Object being used after finalization


harold-b:
I added an anonymous function that does nothing to my test code when making sure my debug client was still working well with local-only scope (As I had been testing with the custom command for the better part of the week) and I started getting crashes right after.

I don't know what exactly introducing this function caused (said function does nothing at all but assign some local number variables to value passed in), but it started causing an finalizable object that I have be used even after it has been finalized.

The object in question is a wrapper to a native object, which allows me to draw debug text and line in my framework, so I use it every frame. For this particular object I don't keep any references to it stashed anywhere so it's eligible for garbage collection as per normal requirements in the scripting side.

The finalizer gets called just fine, and I then recycle the 'glue' object tat binds the native object to the scripting object. Right after, however, that very same object is used to draw some text!

If  I remove that dummy anonymous function (see `_game.update()` below), then it works normally as it did before. However, when I place it back in there, even when it's commented out, as shown below, it causes that error.

It's going to be a bit hard to show what is going on given that is deeply bound to my framework but I'll do my best to help.

This is the basic flow of the program *(typescript)*:
``` typescript

/// Called from the native side with the correct 'this' binding:
App.ts.update()

// Inside update, i do some debug drawing and updating a native renderable object.
// I call also call some dummy javascript functions that
// are simply there to help me test debugging:

public update( dt:number ):void
{
    this._time += dt;
    
    let time  = this._time;
    let math  = Math;

    let pt    = 24;
    let y     = 10-pt;
    let fps   = 1.0 / dt;
    let dd    = musa.DebugDraw.default;  // This is the object being collected then used
    let mouse = this._mouse;

    // This does nothing
    let func = ( aLocal:boolean ) => {

        let mouse = this._mouse;
        let keys = this._keys;
    }

    // Here, normally, I have no errors.
    // However, after adding that dummy function within _game.update(),
    // when it gets to this next call, the finalizer has already been called.
    dd.drawText( "FPS       " + math.round(fps), 10, y+=pt, pt );
    dd.drawText( "Time      " + time, 10, y+=pt, pt );
    dd.drawText( "DeltaTime " + dt,   10, y+=pt, pt );
    dd.drawText( `Mouse     ${mouse.x} ${mouse.y}`, 10, y+=pt, pt  );

    func(true); // Call dummy function

    // _game.update is another dummy function,
    // inside of which is the culprit, the anonymous function
    // which is causing all the trouble.
    // Please see the next code block to see it's internals.
[...]

PR #917, opened 2016-08-19
Add a pointer-not-pending-finalization assert to duk_push_heapptr()

svaarala:
Add a `duk_push_heapptr()` to catch cases where the application supplied pointer references an object already moved to the "pending finalization" mark-and-sweep list.

### DIFF
--- a/src/duk_api_stack.c
+++ b/src/duk_api_stack.c
@@ -4179,6 +4179,30 @@ DUK_EXTERNAL duk_idx_t duk_push_heapptr(duk_context *ctx, void *ptr) {
 
 	DUK_ASSERT_CTX_VALID(ctx);
 
+	/* Reviving an object using a heap pointer is a dangerous API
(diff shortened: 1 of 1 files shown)


svaarala:
@harold-b If you have a chance to test that this would have caught the issue in #914, it'd be nice to know :) I wrote this against v1.5.0 so it'd be easy for you to check. I'll then rebase this to master before merge.

harold-b:
OK, will test it out in a little bit.

harold-b:
Must DUK_DEBUG be set? (Just tested without it, no assertion was launched before the unreachable pointer was used)

fatcerberus:
You need to enable `DUK_OPT_ASSERTIONS`.

harold-b:
@fatcerberus Thank you, that did it :+1: 

The assert was indeed triggered.

svaarala:
Good to know - I'll merge this to master to help other users :)

PR #919, opened 2016-08-20
Implement Ecmascript 6 Math functions

### LINKED ISSUE #838: Support new ES6 Math methods
Specifically I'm looking for `Math.hypot()`, however it seems like it would be low-hanging fruit to add all the new `Math` functions introduced in ES6.

fatcerberus:
Fixes #838.  Work in progress.
- [ ] Implement new `Math` functions added in ES6
- [ ] Investigate the best way to implement `Math.hypot`, perhaps look at how other engines do it?
- [ ] Add new ES6 `Number` functions: `Number.isNaN()`, etc. => Seperate pull
- [ ] Testcases for ES6 Math functions

### DIFF
--- a/src-input/builtins.yaml
+++ b/src-input/builtins.yaml
@@ -2266,6 +2266,12 @@ objects:
           magic:
             type: math_onearg
             funcname: "floor"  # BI_MATH_FLOOR_IDX
+      - key: "hypot"
+        value:
+          type: function
+          native: duk_bi_math_object_hypot
+          length: 2
+          varargs: true
       - key: "log"
         value:
           type: function
--- a/src-input/duk_bi_math.c
+++ b/src-input/duk_bi_math.c
@@ -18,6 +18,47 @@
 typedef double (*duk__one_arg_func)(double);
 typedef double (*duk__two_arg_func)(double, double);
 
+DUK_LOCAL duk_ret_t duk__math_hypot(duk_context *ctx) {
+	duk_idx_t n = duk_get_top(ctx);
+	duk_idx_t i;
+	duk_double_t res = 0;
+	duk_double_t t;
+
+	/*  ES6 20.2.2.18 Math.hypot
+	 *
+	 *  - If no arguments are passed, the result is +0.
+	 *  - If any argument is +inf, the result is +inf.
+	 *  - If any argument is -inf, the result is +inf.
+	 *  - If no argument is +inf or -inf, and any argument is NaN, the result is
+	 *    NaN.
+	 *  - If all arguments are either +0 or -0, the result is +0.
+	 */
+
+	/* FIXME: naive implementation */
+
+	/*
+	 *  Note: every input value must be coerced with ToNumber(), even
+	 *  if we know the result will be a NaN anyway: ToNumber() may have
+	 *  side effects for which even order of evaluation matters.
+	 */
+
+	for (i = 0; i < n; i++) {
+		t = duk_to_number(ctx, i);
+		if (DUK_FPCLASSIFY(t) == DUK_FP_INFINITE || DUK_FPCLASSIFY(res) == DUK_FP_INFINITE) {
+			res = (duk_double_t) DUK_DOUBLE_INFINITY;
+		} else if (DUK_FPCLASSIFY(t) == DUK_FP_NAN || DUK_FPCLASSIFY(res) == DUK_FP_NAN) {
+			/* Note: not normalized, but duk_push_number() will normalize */
+			res = (duk_double_t) DUK_DOUBLE_NAN;
+		} else {
+			res += (duk_double_t) (t * t);
+		}
+	}
+
+	duk_push_number(ctx, (duk_double_t) DUK_SQRT(res));
+
+	return 1;
+}
+
 DUK_LOCAL duk_ret_t duk__math_minmax(duk_context *ctx, duk_double_t initial, duk__two_arg_func min_max) {
 	duk_idx_t n = duk_get_top(ctx);
 	duk_idx_t i;
@@ -243,6 +284,10 @@ DUK_INTERNAL duk_ret_t duk_bi_math_object_twoarg_shared(duk_context *ctx) {
 	fun = duk__two_arg_funcs[fun_idx];
 	duk_push_number(ctx, (duk_double_t) fun((double) arg1, (double) arg2));
 	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_math_object_hypot(duk_context *ctx) {
+	return duk__math_hypot(ctx);
 }
 
 DUK_INTERNAL duk_ret_t duk_bi_math_object_max(duk_context *ctx) {
--- a/tests/ecmascript/test-bi-math.js
+++ b/tests/ecmascript/test-bi-math.js
@@ -354,6 +354,34 @@ prE(Math.floor(0.5)); print(zeroSign(Math.floor(0.5)));
 prE(Math.floor(1));
 
 /*===
+hypot
+0
+Infinity
+Infinity
+NaN
+NaN
+0
+5
+7.0710678118654755
+7.0710678118654755
+812
+===*/
+
+print('hypot');
+
+prE(Math.hypot());
+prE(Math.hypot(Infinity, NaN));
+prE(Math.hypot(NaN, -Infinity));
+prE(Math.hypot(1, 2, NaN));
+prE(Math.hypot(1, 2, 'pig'));
+prE(Math.hypot(-0));
+
+prE(Math.hypot(3, 4));
+prE(Math.hypot(3, 4, 5));
+prE(Math.hypot(3, '4', 5));
+prE(Math.hypot(-812));
+
+/*===
 log
 4812184
 NaN


fatcerberus:
List of Math functions added to ES6:
* `Math.acosh`
* `Math.asinh`
* `Math.atanh`
* `Math.cbrt` (cubic root)
* `Math.clz32` (count leading zeroes [in 32-bit binary representation])
* `Math.cosh`
* `Math.expm1`
* `Math.fround`
* `Math.hypot` (square root of sum of squares)
* `Math.imul`
* `Math.log1p`
* `Math.log2`
* `Math.log10`
* `Math.sinh`
* `Math.tanh`
* `Math.trunc`

Looks like I have my work cut out for me.

svaarala:
By the way, regarding ES6 vs. ES5.1, ES6 is probably too large for some embedded targets. So I think the big goal forwards would be to adopt ES6 features one by one but keep the ability to build a smaller, clean cut ES5.1 version too, with the idea that ES6 features would very minimally contribute to ES5.1 feature footprint.

Right now there isn't good infrastructure in place for optional built-ins; as we discussed before, the properties will remain in place even if the implementation is stubbed out so it's pretty clumsy. My current plan for that is to make the dist process part of the distributable and merge it with genconfig, so that preparing a configuration involves both a configuration file and creating the prepared sources. This setup would then make it easy to drop built-ins cleanly. For example, you could just say "no ES6 Math" or "no Math.exp()".

The concrete issue for this pull is that:

- ES6 Math functions should maybe be behind `DUK_USE_ES6_MATH` (enabled by default).
- Low memory build should say `DUK_USE_ES6_MATH: false` as it already does for e.g. ES6 Proxy.
- When ES6 Math is disabled, the Math functions unfortunately need to be present, but with a stubbed out `return DUK_RET_ERROR;` implementation; to be improved later.

fatcerberus:
Granted, the compiler still needs to be rewritten in any case, which will likely impact footprint.  That said, I agree on having the ability to compile out ES6 features to reduce footprint, as ES5.1 support is often more than enough for many applications.  My head spins every time I look at the amount of stuff that was added to the standard in ES6.

fatcerberus:
One interesting tidbit from the ES6 spec about `Math.hypot()` specifically:
> Implementations should take care to avoid the loss of precision from overflows and underflows that are prone to occur in naive implementations when this function is called with two or more arguments.

This appears to be more of a guideline than a strict requirement, but I'm not sure how to satisfy it.  Maybe do the intermediate calculations as `long double`?  Other than that I can't see how you'd prevent an overflow.

fatcerberus:
On the plus side (emphasis mine):
> `Math.hypot` returns an **implementation-dependent** approximation of the square root of the sum of squares of its arguments.

That makes things a little easier for me. :)  The spec only prescribes specific behaviors for Infinity, NaN and -0.  Everything else is left up to the implementation.

fatcerberus:
ES6 also adds:

* Number.isNaN()
* Number.isFinite()
* Number.isInteger()
* Number.isSafeInteger()
* Number.parseFloat()
* Number.parseInt()
* Number.EPSILON
* Number.MAX_SAFE_INTEGER
* Number.MIN_SAFE_INTEGER

As these are somewhat related, it might be worth investigating to see if they can be added to this pull.

svaarala:
> Granted, the compiler still needs to be rewritten in any case, which will likely impact footprint.

True that - but I don't think I could justify a merge that added, say, 30kB to the minimum footprint. So the challenge for me will be to find a way to implement an ES6-capable compiler that also compiles to a reasonably small footprint when used in ES5.1 (if the ES6 version doesn't).

svaarala:
> >  Implementations should take care to avoid the loss of precision from overflows and underflows that are prone to occur in naive implementations when this function is called with two or more arguments.
> This appears to be more of a guideline than a strict requirement, but I'm not sure how to satisfy it. Maybe do the intermediate calculations as long double? Other than that I can't see how you'd prevent an overflow.

I think it's referring to the basic numerical computation problems re: precision. For example, if you're adding one large positive number and a bazillion small positive numbers, you should add up the small numbers first before adding in the large number - otherwise you'll lose precision.

For `hypot` the naive approach is add the squares in sequence which will work very badly if the first argument is large and the subsequent arguments are small - but large enough to contribute to the result if added in a proper order.

A less naive approach is to sort the numbers (absolute value here) and add the squares in increasing order. I don't recall exactly but probably a better approach is to keep adding numbers that have magnitude (exponent) as close to each other as possible, in sequence, so that you'll always minimize the precision loss.

In any case it's not a trivial thing which is why any functions not provided by the standard library are painful to implement. If you look into implementing trigonometric functions from base functions, for example, it's a lot of work to implement that correctly.

fatcerberus:
`hypot` is part of the C standard library, but unfortunately it's not a variadic function, and (unlike min/max) can't be composed into one.  So unfortunately we have to do it manually. :(

svaarala:
Yeah. I'm not saying this should be used for the implementation but a numerically OK approach would probably be something like:
- Replace all values in the value stack with their square; special case N=0
- While duk_get_top(ctx) > 1:
  + Find two smallest (absolute, automatically) values from the value stack
  + Remove them and push the result of adding them together to the value stack
- In the end there's only one value, return sqrt(value)

I'm not sure if that'd be the best way to do it - but this should be a pretty basic numeric computation problem so it should be easy to find the "right way".

It's then of course a separate question how good the implementation needs to be :)

svaarala:
For large argument count something like that would be too naive and would need an efficient data structure, etc. Again, not really sure how good this needs to be for embedded scripting - in other words, what's the expectation from the application writer?

svaarala:
If there are multiple algorithms that work with variadic arguments and end up adding a bunch of numbers (here the squares), the "add N values minimizing precision loss" should probably be a shared helper.

svaarala:
If you're interested in the theory part, this is one early paper (which at least used to be suggested reading when I was studying numerical computation): http://perso.ens-lyon.fr/jean-michel.muller/goldberg.pdf.

Looks like pairwise summation would be pretty good: https://en.wikipedia.org/wiki/Pairwise_summation, and there are better ones which get more complicated, like https://en.wikipedia.org/wiki/Kahan_summation_algorithm. Not really an area I know well however :)

fatcerberus:
It's been my experience that hypot() is generally used for its trigonometric purpose ("given side lengths A and B, calculate hypotenuse C) - passing more than two arguments to this function feels odd.  In that light, the naive implementation seems fine to me.

I guess we could use a simple implementation for the initial merge and if it becomes an issue in practice, it can always be changed later. :)

svaarala:
That works for me :)

svaarala:
For one- and two-argument cases it'd be good if the shared one-arg/two-arg helper could be used because it minimizes footprint impact when the functions are intentionally disabled.

fatcerberus:
I noticed the one- and two-arg helpers, but it seemed they weren't actually used for vararg functions?  e.g. min/max use a separate helper.

svaarala:
Re: ES5.1 / ES6, the builtins YAML metadata has some tags for flagging ES6 objects and properties. That information doesn't yet have a direct impact, but the intent is to allow e.g. "--no-es6" to automatically drop properties flagged with ES6. So it would be nice to add those flags here too (see existing usage) - but unfortunately nothing will detect if a flag is missing yet.

svaarala:
> I noticed the one- and two-arg helpers, but it seemed they weren't actually used for vararg functions? e.g. min/max use a separate helper.

Yes, I meant any other ES6 functions which are one/two argument kind.

svaarala:
Also if you think it makes sense to implement these in multiple pulls, that'd be fine too. I don't mind small pulls, as long as they are otherwise clean (i.e. squash fixups and such :-)

fatcerberus:
Yes, I will probably implement the new `Number` functions in a separate pull at least.  They are related, but only tangentially (no pun intended!).

fatcerberus:
Just checked Travis, looks like all `Math.hypot()` tests pass.  I wasn't sure if I got all the corner cases right, but it looks like it. The function has the someone odd requirement that if any argument is +/- infinity, the result is +Infinity, *even if a NaN is present*.

fatcerberus:
By the way, I noticed min/max (and possibly others) include a clause to call ToNumber on all arguments, even when bailing out on a NaN.  Is this a requirement in the specification?

svaarala:
Yes: http://www.ecma-international.org/ecma-262/5.1/#sec-15.8.2.12.

fatcerberus:
Thanks for the reference.  I scrolled up a bit from there too, and it's actually required for all `Math` functions:

> Each of the following `Math` object functions applies the ToNumber abstract operator to each of its arguments (in left-to-right order if there is more than one) and then performs a computation on the resulting Number value(s).

So it seems I did the right thing by doing that in `Math.hypot()` too :-)

svaarala:
Usually that doesn't matter but the specification guarantees side effects like:

```
duk> a={valueOf:function() {print('A'); return 123;}}
= [object Object]
duk> b={valueOf:function() {print('B'); return 234;}}
= [object Object]
duk> Math.max(a,b,b,a)
A
B
B
A
= 234
```

This probably doesn't have much real world importance though.

svaarala:
From a quick glance current test cases don't cover the Math ToNumber() coercion side effects.

fatcerberus:
Hm, I could write those tests and submit a pull for that, should be easy enough. :)

fatcerberus:
Oh, I still have this pull open too.  I forgot about this one :)

fatcerberus:
@svaarala I'm going to close this pull and open a laundry-list issue for ES6 Math.  There are enough functions to implement that it's probably better to do it in stages so that the implementation doesn't end up getting rushed.

I will open a pull for `Math.hypot()` though, I need that one :)

svaarala:
Sounds good to me :)

PR #922, opened 2016-08-22
Implement an API to safely summarize any value as a string

### LINKED ISSUE #776: API call to summarize any value into a string without side effects
The property access error messages use an internal mechanism to "summarize" any tagged value safely without side effects. For example, strings are automatically quoted and sanitized (and truncated if too long), etc.

This internal API call could be cleaned up and made available to user code.

fatcerberus:
Fixes #776.

This adds a function, `duk_to_summary()` to the public API which exposes the `duk_tval` summarization mechanism used internally for verbose errors for use by applications.
- [x] Initial draft implementation => `duk_to_summary()`
- [ ] API structure - Multiple calls, one call with flags, something else?
- [ ] Semantics - Should we coerce the value to a summary, replacing it; or push the summary to the stack without modifying the original value?
- [ ] API documentation for new call(s)
- [ ] API testcases for new call(s)

### DIFF
--- a/tests/api/test-safe-push-summary.c
+++ b/tests/api/test-safe-push-summary.c
@@ -1,0 +1,64 @@
+/*===
+duk_is_string()=1
+'pig'
+'the big fat pig ate everyone and...'
+duk_is_string()=1
+812
+9000.1
+false
+true
+null
+undefined
+Error: the pig ate it
+===*/
+
+void test(duk_context *ctx) {
+	duk_push_string(ctx, "pig");
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("duk_is_string()=%d\n", duk_is_string(ctx, -1));
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_string(ctx, "the big fat pig ate everyone and now it's too fat");
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_int(ctx, 812);
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("duk_is_string()=%d\n", duk_is_string(ctx, -1));
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_number(ctx, 9000.1);
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_false(ctx);
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_true(ctx);
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_null(ctx);
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_undefined(ctx);
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_error(ctx, DUK_ERR_ERROR, "the pig ate it");
+	duk_safe_push_summary(ctx, -1, 0x0);
+	duk_safe_push_summary(ctx, -2, DUK_SUMMARY_ERROR_AWARE);
+	printf("%s\n", duk_get_string(ctx, -2));
+	//printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_3(ctx);
+}
--- a/src-input/duk_api_public.h.in
+++ b/src-input/duk_api_public.h.in
@@ -175,6 +175,9 @@ struct duk_time_components {
 #define DUK_HINT_STRING                   1    /* prefer string */
 #define DUK_HINT_NUMBER                   2    /* prefer number */
 
+/* Summarization flags */
+#define DUK_SUMMARY_ERROR_AWARE           (1 << 0)
+
 /* Enumeration flags for duk_enum() */
 #define DUK_ENUM_INCLUDE_NONENUMERABLE    (1 << 0)    /* enumerate non-numerable properties in addition to enumerable */
 #define DUK_ENUM_INCLUDE_INTERNAL         (1 << 1)    /* enumerate internal properties (regardless of enumerability) */
@@ -628,6 +631,7 @@ DUK_EXTERNAL_DECL void *duk_to_pointer(duk_context *ctx, duk_idx_t idx);
 DUK_EXTERNAL_DECL void duk_to_object(duk_context *ctx, duk_idx_t idx);
 DUK_EXTERNAL_DECL void duk_to_defaultvalue(duk_context *ctx, duk_idx_t idx, duk_int_t hint);
 DUK_EXTERNAL_DECL void duk_to_primitive(duk_context *ctx, duk_idx_t idx, duk_int_t hint);
+DUK_EXTERNAL_DECL void duk_safe_push_summary(duk_context *ctx, duk_idx_t idx, duk_uint_t flags);
 
 #define DUK_BUF_MODE_FIXED      0   /* internal: request fixed buffer result */
 #define DUK_BUF_MODE_DYNAMIC    1   /* internal: request dynamic buffer result */
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -17,6 +17,7 @@
  */
 
 DUK_LOCAL_DECL duk_idx_t duk__push_c_function_raw(duk_context *ctx, duk_c_function func, duk_idx_t nargs, duk_uint_t flags);
+DUK_LOCAL const char *duk__push_string_tval_readable(duk_context *ctx, duk_tval *tv, duk_bool_t error_aware);
 
 /*
  *  Global state for working around missing variadic macros
@@ -2337,6 +2338,22 @@ DUK_EXTERNAL const char *duk_to_string(duk_context *ctx, duk_idx_t idx) {
  skip_replace:
 	DUK_ASSERT(duk_is_string(ctx, idx));
 	return duk_require_string(ctx, idx);
+}
+
+DUK_EXTERNAL void duk_safe_push_summary(duk_context *ctx, duk_idx_t idx, duk_uint_t flags) {
+	duk_tval *tv;
+	duk_bool_t error_aware = 0;
+
+	DUK_ASSERT_CTX_VALID(ctx);
+
+	tv = duk_get_tval_or_unused(ctx, idx);
+	DUK_ASSERT(tv != NULL);
+
+	if (flags & DUK_SUMMARY_ERROR_AWARE) {
+		error_aware = 1;
+	}
+
+	duk__push_string_tval_readable(ctx, tv, error_aware);
 }
 
 DUK_INTERNAL duk_hstring *duk_to_hstring(duk_context *ctx, duk_idx_t idx) {


fatcerberus:
@svaarala Thoughts on the current naming and semantics?

src/duk_api_stack.c
```diff
@@ -2339,6 +2340,19 @@ DUK_EXTERNAL const char *duk_to_string(duk_context *ctx, duk_idx_t idx) {
...
 }
 
+DUK_EXTERNAL const char *duk_to_summary(duk_context *ctx, duk_idx_t idx) {
+	duk_tval *tv;
+
+	DUK_ASSERT_CTX_VALID(ctx);
+
+	idx = duk_require_normalize_index(ctx, idx);
+	tv = DUK_GET_TVAL_POSIDX(ctx, idx);
+	DUK_ASSERT(tv != NULL);
```
svaarala: Here index normalization is not necessary because the index is not reused when the value stack has changed (e.g. new values pushed). So a shorter alternative is: `tv = duk_require_tval(ctx, idx);`.

website/api/duk_to_summary.yaml
```diff
@@ -0,0 +1,26 @@
...
+proto: |
+  const char *duk_to_summary(duk_context *ctx, duk_idx_t idx);
+
+stack: |
+  [ ... val! ... ] -> [ ... summary! ... ]
+
+summary: |
+  <p>Replace the value at <code>idx</code> with a readable summary of the value
+  in string form.  If <code>idx</code> is invalid, throws an error.</p>
+
```
svaarala: Maybe one or two sentences about the intent and specific guarantees:

- Summary is a short string with a maximum length, strings and buffer values are truncated.
- Coercing a value to a summary does not cause side effects such as calling `.toString()` methods or such. Garbage collection side effects (like running a finalizer) are possible.

A table of example summary values for API types, and perhaps a few common objects, would be quite useful when browsing the API docs.

src/duk_api_stack.c
```diff
@@ -2339,6 +2340,19 @@ DUK_EXTERNAL const char *duk_to_string(duk_context *ctx, duk_idx_t idx) {
...
+
+	duk__push_string_tval_readable(ctx, tv, 0 /*error_aware*/);
+	duk_replace(ctx, idx);
```
fatcerberus: Actually the index *is* reused here.  `duk__push_string_tval_readable()` modifies the stack, then I use `duk_replace()` to overwrite the original value.

svaarala: Ah true, my bad. For some reason I had the idea the pushed value would be left on the stack when glancing the change.

svaarala:
For value stack semantics the two basic alternatives would be:

- The internal helper semantics: push a value of the value stack rather than replacing an existing value.
- Coercing and replacing an existing value.

Both are useful in some ways but the formed, currently in the pull, is probably more in line with existing APIs?

fatcerberus:
That's what I figured, coercion seems to be the usual idiom in the public Duktape API.  I was on the fence with this one because it's a bit of a gray area (summary being useful for human debugging but not much else), but decided it would be best to err on the side of consistency.

website/api/duk_to_summary.yaml
fatcerberus: I agree, this was just a quick draft.  It's mostly a placeholder so I don't forget to write the API doc later :)

svaarala:
Re: naming, I think the current name is pretty good, the only thing it might be missing is "string" to indicate the result is always a string. So I guess the logical alternative would be `duk_to_summary_string()` or `duk_to_string_summary()`?

svaarala:
The error awareness thing is also a somewhat related issue. Similarly, for errors one might often want to extract the stack trace safely.

So either there would be several API calls or a few flags?

fatcerberus:
I actually didn't know what the `error_aware` argument was for; from reading the code it seemed to only be used in actual error paths (which this is not).

svaarala:
When error_aware is set, the internal helper will check if the value is an object that inherits from `Error.prototype`; if so, extracts the error message being careful not to invoke side effects such as a getter.

svaarala:
Anyway, I was commenting more on the generic needs for application code, which are quite various:

- Safe `ToString()` coercion. Exists as `duk_safe_to_string()` at present.
- Safe generic value summary.
- Safe generic value summary which is aware of specific object types like Errors, maybe Buffers, etc. Application code might want to use a variant which also supports application specific types, but this is hard to support in the API beyond just providing the generic fallback summarization call.
- Safe error dumping, including stack trace. This is useful in e.g. protected call idioms to write out the error conveniently and safely.

So from API design standpoint these can all be separate calls, or if there's some overlap, perhaps some of them could share a more generic safe string coercion call with a few behavioral flags.

For example, `duk_safe_to_string()` could get an additional flag which could provide the summarization, stack trace dumping, etc via flags.

In any case some consistent "end goal" for these coercion needs should be in sight. It's of course possible to just approach it incrementally and then merge calls later, but it's also nice to avoid that if possible.

fatcerberus:
I guess we have two options for that:

* Add additional APIs like `duk_to_error_summary()`.  This needs a bit more thought to avoid bloating the public API.
* Add a flag argument to the base call which specifies the type of summary, e.g. `DUK_SUMMARY_ERROR` would get an error-aware summary.  This cuts down on API bloat but may be less intuitive unless well-documented.

svaarala:
One useful approach is to think what the application call site would look like. For example, summary printing might happen in some debug paths:

```c
if (DEBUG) {
    duk_dup(ctx, 0);
    printf("First argument: %s\n", duk_to_summary(ctx, -1));
    duk_pop(ctx);
}
```

For this case the `duk_dup()` is needed because the original value cannot be disturbed. For this specific case the "push summary" variant works more cleanly:

```
if (DEBUG) {
    printf("First argument: %s\n", duk_to_summary(ctx, 0));
    duk_pop(ctx);
}
```

It would be nice if that `duk_pop()` could also be eliminated, but there's no easy to way to return a string which is safe to print without also leaving it somewhere on the value stack :(

Error handling is a use case of its own and I'd like to do something like:

```c
if (duk_pcall(...) != 0) {
    printf("Call failed: %s\n", duk_to_error_summary(ctx, -1));
}
```

and that would be (1) safe, (2) side effect free, and (3) always print the stack trace when available.

fatcerberus:
Folding the functionality into `duk_safe_to_string()` is an interesting idea, I missed that bit.

Maybe that's not too intuitive though: I'd expect developers to assume the call was usually the same as `duk_to_string()`, just without the ability to throw. :)

svaarala:
`duk_safe_to_string()` actually isn't equal to `duk_to_string()`: if an error happens during the string coercion, it proceeds to summarize the error. So the result may not be related to the input value. If that error-to-string coercion fails it just produces the fixed string `"Error"`.

svaarala:
One naming issue is also that `duk_to_summary()` doesn't convey it's safe nature (compare to `duk_safe_to_string()`). API calls being safe (= no throw) or protected (= catch, return success/fail) is something that'd be nice to be readily apparent - so `duk_safe_to_summary()`?

svaarala:
Incidentally, these API design questions are the reason it's not trivial just to add new calls. The core change here is trivial, but ensuring that the API as a whole is as minimal and consistent as possible is the difficult issue. It often needs prototyping (well, more like simulating) application call sites for various alternatives. This is actually why I like that pulls provide documentation and testcases too, because both of these provide some incentive of trying to see the application point of view.

fatcerberus:
Re: `duk_safe_to_string()` behavior differences: That's why I said "usually the same" - i.e. outside of an error, it's a typical string coercion. :)

Agree on conveying the no-throw nature of the call, that's useful information to have at the call site.

svaarala:
The current convention for those call safety differences is roughly:

- All calls can by default throw.
- `duk_pXXX` are protected calls.
- `duk_safe_XXX` are no-throw calls which is different from protected because they don't return a success/error indicator.

As a whole that's not maybe very consistent :-)

fatcerberus:
Just for the sake of argument, `duk_safe_call()` returns a success/error indicator AND leaves the error on the stack (in place of the first return value).  So that's more of a "protected" call than a "safe" one, by the convention above.

svaarala:
Agree, `duk_safe_call()` is actually named incorrectly. It's a protected call so should be something like `duk_pnativecall()`.

fatcerberus:
By the way, I don't mind if it takes a bit of thought (and time) to get this merge ready.  I just figured it would be more conducive to progress if there was some concrete change to discuss.

fatcerberus:
A related question: Should this mechanism be exposed to Ecmascript code?

svaarala:
One more comment regarding API design: the most important first concern is IMO how the API looks like to the application and how intuitive it is to use. For example, having several different string coercion variants (safe coercion, summary, error trace, etc) is fine if that is a logical and easy-to-understand API design. The secondary concern is the most logical internal implementation (which then changes over time). If the most natural implementation is a single helper with flags, those API calls can then be implemented as macros achieving two goals at the same time.

In fact, API call variants are maybe slightly preferable if the set of variants is reasonably change and not liable to multiply over time: in such a case the calls be implemented either individually or via macros, whichever works best internally.

However, if an API has a lot of variability and is likely to grow more variable over time, e.g. `duk_def_prop()`, then using an explicit flags field makes more sense and leads to better API stability.

If an API uses an explicit flags field and flags defines, that's more awkward to convert into separate internal calls if that happened to be the most logical internal implementation. It's not impossible but relatively awkward; as a fictional example:

```c
#define duk_compile(ctx,flags) \
    ((flags) & DUK_COMPILE_EVAL ? duk_compile_eval_internal : duk_compile_other_internal)(ctx)
```

The problem with this is that it's quite quickly necessary to evaluate `flags` multiple times which violates the API guarantees. On the other hand, assigning it temporarily somewhere is not possible in standard C because you can't declare new variable inside an expression.

svaarala:
As a sidenote, maybe there should be a small "API design guidelines" internal document discussing some of the trade-offs and their current solutions :)

svaarala:
Ok, #924 is an initial draft of that.

fatcerberus:
If we go with the solution of pushing the summary to the stack without touching the original value (which is probably more in line with how an application would use it), I guess the function(s) would need to be renamed.

Candidates:
* `duk_push_xxx_summary` (where 'xxx' is value, error, etc.)
* `duk_summarize_xxx`
* `duk_push_value_summary` (with flags, maybe)
* Something else?

svaarala:
A few more points of comparison:
- `duk_safe_to_string()` works in-place for an index; it's a similar function in some sense. But also different because it's an extended version of `ToString()` rather than a totally new conversion.
- `duk_json_encode()` and other codec functions also work in-place with an index argument.
- Context dump helper, `duk_push_context_dump()` pushes a value. But it doesn't have a logical argument so couldn't work in place.

One reason I left this primitive internal for the time being was because I was debating the best API for it.

W.r.t. naming, I think a logical name would be: `duk_safe_to_summary_string()`. That's pretty long though.

Btw, on second thought, the current implementation is not a "safe" function right now. It may throw on out-of-memory, which safe/protected calls shouldn't generally do. They may throw on API call errors, which are within the developer's control, but out-of-memory errors should be caught. It's a small distinction and doesn't always matter - but it does matter for e.g. sandboxing.

svaarala:
Regarding push/in-place, there's actually a good argument for making `duk_safe_to_string()` to be `duk_push_to_string_safe()` (or something) instead. I think I made that in-place previously because it was quite often used to print errors, where the value is already on the value stack and you often don't need to preserve the original when coercing (assuming you just print out the error).

So maybe keep this in-place for the purposes of this pull, and change it later if there's some insight?

fatcerberus:
That's fine by me.  If this makes 2.0.0 release though, we will have to make a final decision before the release due to semantic versioning guarantees.

fatcerberus:
Thinking about this pull a bit more, it seems that the typical use cases for summarization would be:

1. Logging
2. Debugging, e.g. communicating internal state to a debug client
3. Diagnostics - related to 2, printing stuff on-screen

In all of these cases, the program generally still wants the original value, so a `duk_dup()` would almost always be required with in-place coercion.  So I think ultimately, push behavior is most intuitive for this call.

fatcerberus:
Related, it might be good to also have a summarization mechanism in the debug protocol.  The debug client could then send a dvalue to the target and get a summary string back.

svaarala:
> So I think ultimately, push behavior is most intuitive for this call.

Agree.

> Related, it might be good to also have a summarization mechanism in the debug protocol. The debug client could then send a dvalue to the target and get a summary string back.

I think the priority there should be that the debugger protocol provides all the necessary low level primitives (it probably already does?) so that one can implement a summarization mechanism in the debug client. If that's generally useful, it might be nice to document an example algorithm for that, e.g. what debug requests to use to do that.

Debug clients have quite different user interfaces and the most appropriate way to present summarized data varies a great deal. For example, if there are hover hints that changes the situation a little bit, hex data can maybe be clicked to get a full hex dump, the amount of visible data depends on how much space there is in the layout (which may depend on the UI context), etc.

Having said that it'd still probably be useful to have a debug request which provided the same result as the API call, which is also intuitive for a programmer when e.g. comparing a debug log to the debugger view. But it's maybe not as generally useful as the C API call.

fatcerberus:
When this is merged, it will of course be possible to implement debug summarization in the target using an AppRequest.  I'm just wondering if the functionality is useful enough to be a built-in command.

Then again, maybe it's better going forward just to provide a more robust debugging API in Duktape itself and have AppRequest/Notify be the basic primitive in the protocol.  Doing that would have the added benefit that the application could "debug itself" to some extent, even if it's just for printing diagnostic info on-screen.

fatcerberus:
Currently, APIs provided in this pull are:
* `duk_summarize(ctx, idx)` - Summarizes a value.
* `duk_summarize_error(ctx, idx)` - Summarizes a value (error aware)

Based on the current state of things, this makes the most sense.  Having a single function with a flags argument would be more intuitive if there's a possibility of implementing more types of summarization in the future, however.

fatcerberus:
The alternative API (which I think I like better) is a single function:

```c
duk_safe_summarize(ctx, idx, hint);
```

Where `hint` is one of:
* `DUK_SUMMARY_NORMAL`
* `DUK_SUMMARY_ERROR`

Another option is for `hint` to be documented instead as `flags`, which is more extensible.  If in the future we have multiple options for summarization and it makes sense to combine them, a flags argument would allow that, e.g.:

```
DUK_SUMMARY_ERROR_AWARE | DUK_SUMMARY_FOO_AWARE
```

fatcerberus:
@svaarala Do you agree with the naming of ‘duk_safe_summarize()‘?

svaarala:
Most pushers are named `duk_push_xxx()` but there are no safe pushers yet - but maybe `duk_safe_push_summary()`?

This call is still a bit awkward in that it *may* throw (e.g. due to an out-of-memory) so it's not really a safe call now. It's a *side effect free* call yes. I think the best fix for that is to make the call actually safe - internals also except it to work that way. When it fails, it then needs to push an already interned string (which causes no memory allocations), e.g. `Error`. I can fix that part separately.

fatcerberus:
`duk_safe_push_summary()` would work and would still be concise enough to be readable.  Another option would be `duk_safe_push_summary_string()` which makes it clearer that the output is a string, but that might be too verbose.   We can just make sure to document that the summary is always a string.

Agreed that the call shouldn't be able to throw since its main use case is debugging/diagnostics where throwing anything at all might lead to a double error. :)

fatcerberus:
Usage is now:
```c
duk_push_string(ctx, "pig");
duk_safe_push_summary(ctx, -1, 0x0);
// -> "'pig'"

duk_push_error(ctx, DUK_ERR_ERROR, "the pig ate it");
duk_safe_push_summary(ctx, -1, DUK_SUMMARY_ERROR_AWARE);
// -> "Error: the pig ate it" + stack trace
```

Awaiting code review.

tests/api/test-safe-push-summary.c
```diff
@@ -0,0 +1,64 @@
...
+	duk_push_undefined(ctx);
+	duk_safe_push_summary(ctx, -1, 0x0);
+	printf("%s\n", duk_get_string(ctx, -1));
+	duk_pop_2(ctx);
+
+	duk_push_error(ctx, DUK_ERR_ERROR, "the pig ate it");
+	duk_safe_push_summary(ctx, -1, 0x0);
+	duk_safe_push_summary(ctx, -2, DUK_SUMMARY_ERROR_AWARE);
+	printf("%s\n", duk_get_string(ctx, -2));
+	//printf("%s\n", duk_get_string(ctx, -1));
```
fatcerberus: I wasn't sure what the error-aware summary would look like, that's why this is commented out.  I assume it's "Error: message" followed by a stack trace starting on the next line.  However, I have no idea what the stack will look like here since we're running under control of the test harness.

fatcerberus:
Okay, I'm utterly confused now:
```
test-safe-push-summary.c: fail; 15 diff lines
```

There aren't even 15 lines of output in the test?

svaarala:
The 15 lines refers to the line length of the diff, which includes context.

svaarala:
The summary of errors doesn't include a stack trace because it's intended to be a short "one liner" like other summary values. The code implementing is here: https://github.com/svaarala/duktape/blob/master/src-input/duk_api_stack.c#L4779-L4797.

In short, the result is `.message` of the error if it can be safely read without side effects. So for `new TypeError('foo')` the summary would be just `foo`.

svaarala:
By the way, I added that error aware behavior specifically for getting better fatal error messages. It's not necessarily generally useful behavior - which is one reason I didn't want to add this to the API as is. What works internally may not be good for a public API commitment.

fatcerberus:
I thought I remembered you saying that error_aware included a stack trace in the summary?  If that were the case I'd consider the behavior useful to have in the API.

Anyway I can remove the `DUK_SUMMARY_ERROR_AWARE` flag if you think that's better.  I wonder if we should leave the `flags` argument in as "reserved for future use" though.

Re: Error summary format: I would expect the error's `.name` to be included also (i.e. the same way string coercion works for errors).  Otherwise it's not immediately obvious from the summary string that it's an error object.

svaarala:
> I thought I remembered you saying that error_aware included a stack trace in the summary? If that were the case I'd consider the behavior useful to have in the API.

But then the result would be a multi-line string and different from how string values are summarized (newlines are escaped)? To me that would be odd. In any case the current behavior should be pretty apparent from the code (it's not long, just a 100 lines or so).

> Anyway I can remove the DUK_SUMMARY_ERROR_AWARE flag if you think that's better. I wonder if we should leave the flags argument in as "reserved for future use" though.

I don't know if it's better :) Just wanted to mention that the main reason I didn't make this API public originally is because there are a lot of alternatives and use cases so coming up with something that's worth versioning guarantees needs some work.

> Re: Error summary format: I would expect the error's .name to be included also (i.e. the same way string coercion works for errors). Otherwise it's not immediately obvious from the summary string that it's an error object.

That sounds reasonable, as long as the `.name` is also read safely without side effects.

svaarala:
Regarding the stack trace: what I said above was related to potential use cases in actual application code. String coercing an arbitrary value into a string safely, and dumping an error stack trace is a useful use case. However there's no primitive for that. There are also a bunch of other use cases, and it's a bit awkward introducing primitives for safe coercion piecemeal because the result might then be larger than necessary. That's why I wanted to raise that application use case question.

fatcerberus:
I'll read the code later.  The mobile GitHub site is buggy and doesn't want to highlight or scroll to the lines you pointed me to.  And doesn't list the line numbers either.  :/

fatcerberus:
Okay, I see now: Errors summarize as their `.message` if error_aware is set, or as generic objects otherwise.  If that's the case, it doesn't seem to make sense for the public API to ever not be error aware?

Conceptually the idea is that the summary string should be as unambiguous as possible - it identifies the type as well as the value.  Errors are an important class of object just like functions are, so summarization should always be error aware IMO.

svaarala:
>  If that's the case, it doesn't seem to make sense for the public API to ever not be error aware?

I think it depends on the application use case. I might want to see `[object Error]` for errors (like for all other object types)? Otherwise that typing info is lost, regardless of what string is substituted for errors.

Then again, if I'm specifically expecting to deal with an error that behavior makes sense, and it's still useful to summarize other types in case the object turned out to be something else (like a boolean).

It all comes down to practical application call sites and what works for them. That's why I'd like to understand the use cases a bit more before making API commitments, and is the reason I made this API internal because I hadn't had time to do that :/

svaarala:
> Conceptually the idea is that the summary string should be as unambiguous as possible - it identifies the type as well as the value.

This specifically isn't the case right now:
- `new TypeError('foo')` summarizes to `'foo'`
- The string "foo" summarizes to `'foo'`

The situation would be the same even if an error summarized to `'TypeError: foo'` because it could still be confused for a string.

A better alternative would maybe be something like:
- `[object TypeError]` when not error aware
- `[object TypeError foo]` when error aware

If this were the case, then error awareness could be the default (provided the summarization has truncating, etc.

svaarala:
Generally it'd be quite nice if there didn't need to be many variants of a summarization / friendly printing mechanism. Error awareness can probably be worked in like that.

However, in some use cases the application code would want to print a value but if the value turned out to be an error, print its stack trace. For example:

```c
rc = duk_pcall(...);
printf("pcall result %d: %s\n", rc, duk_XXX(ctx, -1));
duk_pop(ctx);
```

This is "lazy code" but avoids an unnecessary if-else for informatively printing out the result. That `duk_XXX()` would ideally be some kind of useful string coercion but have the ability to return stack traces for errors.

svaarala:
... Forgot to say:

A single primitive can't provide a "oneliner" and a stack trace at the same time so there is a need for two or more mechanisms or flags.

svaarala:
Maybe there should be a pre-step pull to improve the summarization a little bit so that we wouldn't need variants? These come to mind right away (they've been bugging me but the internal variant didn't need them):
- For errors, something like `[object name message]`, e.g. `[object TypeError invalid argument]`.
- For functions, current is `[object Function]` but it'd be nice to know at least the function name if it exists (safely of course), e.g. `[object Function cos]` for `Math.cos`.
- For buffer/view objects indicate some basic parameters; current is `[object Uint16Array]`, maybe summarize some initial values like `[object Uint16Array 1 8 3123]` or bytes like `[object Uint16Array 01000800330c]` (which is more useful for C programmers usually and also used by JX).

The remaining question would then be how this interacts with error dumping and some of the other use cases.

svaarala:
Ok, #969 is a pull where we can improve the summarization first before exposing it through the API.

fatcerberus:
Agreed, the implementation in master right now is a bit too rudimentary to expose through the public API as-is.  The idea is sound, but it needs further polishing first.  Priorities. :)

fatcerberus:
Hm, the more I think about this the less I think it has a place in the public API.  Different applications may have different needs WRT summary formatting and so forth, so it'd probably be better to focus on having a robust set of inspection primitives to allow the application to implement its own summary mechanism on top.

That said, Duktape does have need of string summaries internally, so it's still worth it to pursue #969.  I'm just not sure whether "summarize to string" needs to be a publically accessible primitive.

svaarala:
Right, that was my point too: as discussed earlier, there are quite many use cases and they are in some cases quite different. For example, in some situations you might want to display some properties (maybe just own properties, or also inherited properties but limited to a certain maximum length or property count) and not just a top level summary.

For error messages to be friendly, Duktape will need a mechanism to summarize a single value usefully. It's debatable if `[Object]` is useful or not though, but atleast the basic type can be described (null, true, string "foo", arraybuffers, etc).

That can be exposed as an API, but it will just fit a certain set of use cases - those aligning with error messages, debug messages, and the like.

fatcerberus:
Closing this pull too since Duktape 2.0.0 is nigh and it's not clear whether this specific approach is viable yet.

PR #923, opened 2016-08-22
Add testcase for Math ToNumber argument coercion

fatcerberus:
ES5 requires that all `Math` function arguments are ToNumber coerced from left to right, even when the function would otherwise return early (on a NaN, e.g.).  This adds a test to the Ecmascript test suite to verify this behavior.

Work in progress.  See https://github.com/svaarala/duktape/pull/919#issuecomment-241313769

ES5 reference:
http://www.ecma-international.org/ecma-262/5.1/#sec-15.8.2

### DIFF
--- a/tests/ecmascript/test-bi-math-tonumber.js
+++ b/tests/ecmascript/test-bi-math-tonumber.js
@@ -1,0 +1,81 @@
+/*
+ *  Math object ToNumber side effects (ES5 15.8.2)
+ *
+ *  Ecmascript 5.1 requires that all Math object methods coerce all arguments
+ *  with ToNumber from left to right.  This ensures that all side effects
+ *  caused by ToNumber coercion have pig chance to be evaluated.
+ */
+
+var pig = { valueOf: function() { print("pig"); return 812; } };
+var cow = { valueOf: function() { print("cow"); return 1208; } };
+var ape = { valueOf: function() { print("ape"); return -128; } };
+var goku = { valueOf: function() { print("goku"); return 9001.5; } };
+var badger = { valueOf: function() { print("badger"); return 8; } };
+var munch = { valueOf: function() { print("*munch*"); return NaN; } };
+
+/*===
+pig
+===*/
+Math.abs(pig);
+
+/*===
+pig
+cow
+ape
+pig
+cow
+ape
+goku
+badger
+===*/
+Math.cos(pig);
+Math.sin(cow);
+Math.tan(ape);
+Math.acos(pig);
+Math.asin(cow);
+Math.atan(ape);
+Math.atan2(goku, badger);
+
+/*===
+badger
+*munch*
+pig
+goku
+ape
+badger
+*munch*
+===*/
+Math.exp(badger);
+Math.exp(munch);
+Math.log(pig);
+Math.pow(goku, ape);
+Math.sqrt(badger);
+Math.sqrt(munch);
+
+/*===
+pig
+cow
+ape
+===*/
+Math.floor(pig);
+Math.ceil(cow);
+Math.round(ape);
+
+/*===
+pig
+cow
+ape
+goku
+goku
+ape
+cow
+pig
+*munch*
+pig
+*munch*
+cow
+===*/
+Math.min(pig, cow, ape, goku);
+Math.max(goku, ape, cow, pig);
+Math.min(munch, pig);
+Math.max(munch, cow);


fatcerberus:
Incidentally, is there some sort of "index" I have to add new testcases to, or will the test harness pick them up automatically?

svaarala:
The test harness will pick up all test files automatically from `tests/ecmascript` and `tests/api`.

fatcerberus:
From ES5.1 15.8.2:
> Each of the following `Math` object functions applies the ToNumber abstract operator to each of its arguments (in left-to-right order if there is more than one) and then performs a computation on the resulting Number value(s).

What does this mean for functions like `Math.abs()`, which nominally only takes one argument but due to the nature of JavaScript will allow you to pass more than one?  The spec is a bit ambiguous here: Do we have to coerce *all* arguments, even those which aren't used?

svaarala:
I agree it's a bit ambiguous - I'd assume the coercion is required for "intended arguments", i.e. for `Math.abs()` only for the single argument.

fatcerberus:
That's what made the most sense to me too.  Like you said it probably doesn't have much real-world relevance, but if I'm writing a test anyway it should try to be compliant :)

fatcerberus:
The above behavior for `Math.abs()` (only coerce intended arguments) is consistent with at least V8:

```javascript
var a = { valueOf: () => (console.log("pig"), 812) };
var b = { valueOf: () => (console.log("cow"), 1208) };
var c = { valueOf: () => (console.log("ape"), -128) };
var d = { valueOf: () => (console.log("goku"), 9001.5) };
var e = { valueOf: () => (console.log("*MUNCH*"), NaN) };
console.log(Math.max(a, b, c, d, e));
console.log("");
console.log(Math.max(e, d, c, b, a));
console.log("");
console.log(Math.abs(c, a, b, d, e));
```

Output:
```
pig
cow
ape
goku
*MUNCH*
NaN

*MUNCH*
goku
ape
cow
pig
NaN

ape
128
```

tests/ecmascript/test-bi-math-tonumber.js
```diff
@@ -0,0 +1,62 @@
...
+print("FLOOR");
+Math.floor(d);
+
+print("MIN/MAX");
+Math.min(d, c, b, a);
+Math.max(a, b, c, d);
+
+print("POW");
+Math.pow(e, e);
+
```
svaarala: Not sure if this pull is ready to merge, but just in case: it'd be better to use different arguments for Math.pow() so that the order was confirmed by the test.

fatcerberus: Good idea, thanks.  And no, it's not merge ready yet. :)

fatcerberus:
I'll look at this tonight and get it polished off and ready for merge.  Then we'll have an additional Ecmascript compliance test in master. :)

Do new testcases get a releases entry?

svaarala:
Sounds good :)

> Do new testcases get a releases entry?

No, they're not visible in the distributable so they're not included in RELEASES.rst.

fatcerberus:
Okay, finished writing tests, waiting on Travis result.  Anything else you want me to do for this pull?

fatcerberus:
@svaarala Travis says the new test is failing, but I'm not sure if the bug is in the test or in Duktape, as it doesn't show the output, only "19 diff lines".  Could you review the diff for this pull and verify?

svaarala:
I just ran the test manually and compared the output to expected and I think `atan2()` arguments are ToNumber() coerced but not in a left-to-right order. The call site has multiple coercion calls in one expression which doesn't have a guaranteed evaluation order:

- https://github.com/svaarala/duktape/blob/master/src/duk_bi_math.c#L302

So this is actually incorrect now in master. I'll write a pull to fix it.

Btw, the expect output would be easier to parse if coercion of `a` printed `a` and not `pig`;)

svaarala:
Ok, let's see if #943 fixes the issue.

svaarala:
There's a lot to improve in the test harness (even simple stuff like logging at least small diffs explicitly), and also to split execution of a single run into multiple hosts to make the runs complete faster. Unfortunately no time right now, but probably there's some time before 2.0 release to improve that :)

svaarala:
I've also avoided touching the test stuff too much for a while, because the whole dist/prepare process is undergoing change and it makes more sense to work on the test harness after that.

fatcerberus:
Hehe.  In an early draft I had `print("a")` etc. but found that the single letters were hard to differentiate in the expect output.  So I switched to my usual method of using different animals, like in the Node.js module extra tests.

Maybe it would be easier to follow if I renamed the variables, though.

svaarala:
Ok, if try rebasing against (updated) master and let's see if Travis test passes.

fatcerberus:
Looks like it passed this time. :)

svaarala:
Ok, shall I merge? :)

tests/ecmascript/test-bi-math-tonumber.js
```diff
@@ -0,0 +1,81 @@
+/*
+ *  Math object ToNumber side effects (ES5 15.8.2)
+ *
+ *  Ecmascript 5.1 requires that all Math object methods coerce all arguments
+ *  with ToNumber from left to right.  This ensures that all side effects
+ *  caused by ToNumber coercion have pig chance to be evaluated.
+ */
```
svaarala: "pig change" probably caused by overeager s/a/pig/ :)

fatcerberus:
Yeah, I think I covered all the functions in ES 5.1, should be good to go.

tests/ecmascript/test-bi-math-tonumber.js
fatcerberus: Oops. :)

fatcerberus:
I already left for work so I can't fix the typo :(

svaarala:
Ok, I can fix it in master.

PR #924, opened 2016-08-22
Add initial draft of API design guidelines document

svaarala:
Just a quick draft of some trade-offs involves in API design and how they've been solved at the moment. The intent is to make it easier to recall current conventions when adding new calls.

### DIFF
--- a/doc/api-design-guidelines.rst
+++ b/doc/api-design-guidelines.rst
@@ -1,0 +1,169 @@
+=====================
+API design guidelines
+=====================
+
+Introduction
+============
+
+A consistent API design makes it easier to write applications: it should be
+easy to remember API call naming, argument order, and other conventions, to
+make it as easy as possible to write error-free code.
+
+API design decisions involve several important concerns:
+
+* Consistency: easy to remember API calls and arguments.
+
+* Versioning: easy to extend the API, avoid dead ends in design.
+
+* Footprint: minimize code footprint of Duktape internals and application
+  call site.
+
+* Performance: some API designs allow better performance than others.
+
+This document provides some notes on common API design issues and trade-offs.
+The document is not intended to be complete, but just cover practical issues
+as they come up.
+
+Naming
+======
+
+API calls
+---------
+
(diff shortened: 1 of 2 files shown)


svaarala:
@fatcerberus This is not intended to be complete by any means. Within the scope of this pull, I'd like to at least add some placeholder sections for issues that may be relevant to document however, so suggestions welcome!

svaarala:
Couple of issues that come to mind:

- Argument naming
- Argument typing
- Varargs API calls; when provided

fatcerberus:
Regarding the flags vs. multiple calls debate: For me this decision usually comes down to - What would a combined API call (with flags) look like internally?

If you have function with a large switch statement or if/else tower switching on the flags with little to nothing in common between the cases, that almost certainly would do well to be split up; each function should only do one basic thing.

Where flags make more sense is in the case of calls that *already* do only one basic thing (e.g. `duk_pcompile()`) and have only subtle variations in the code paths given different flags.  For instance you can compile source code as either strict or non-strict, but the compiler does mostly the same thing in either case.

svaarala:
> This seems easy enough to avoid: Simply wrap the macro body in a do { ... } while (0); and assign flags to a local variable. C has block scope so there should be no issue there.

Unfortunately that doesn't work for expressions which must evaluate to a value.

svaarala:
> Regarding the flags vs. multiple calls debate: For me this decision usually comes down to - What would a combined API call (with flags) look like internally?

I view this quite differently: often a public API outlives the internal implementation, sometimes several times. So what's crucial is that the public API is conceptually good, and that it can be mapped to different internal implementations over time.

fatcerberus:
Sorry, lapsed into "developer mode" there - despite speaking in concrete terms, I really meant that from a conceptual standpoint: if a call can do 3 totally different things depending on what flags you pass to it, maybe you shouldn't be using flags in that call. :)

svaarala:
Sure, agree with that. I don't think that has been a concrete issue though :)

PR #926, opened 2016-08-23
Add improved ES6 Annex B regexp identity escape

### LINKED ISSUE #643: Escaped unicode identifier parts in duktape regexp
Found something accepted by browsers but not by duktape:
```javascript
var re=/[\Æ]/;
```
Browser ignores the identity escape, but duktape reports "invalid regexp escape".

A quick search didn't find anything standard-offending in that expression. Since it's rather unlikely for someone to mistakenly write a bogus escape with a unicode character, I think it wouldn't hurt to relax those `duk_unicode_is_identifier_part()` checks.

svaarala:
Allow almost all source characters as identity escapes when `DUK_USE_ES6_REGEXP_SYNTAX` is enabled. However, the pull doesn't add support for backtracking out of e.g. an invalid hex escape and treat it literally.

Fixes #643.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1715,9 +1715,14 @@ Planned
   and constructor call argument count from 511 to 255, and maximum Ecmascript
   function constant count from 262144 to 65536 (GH-903)
 
-* Allow ES6 unescaped right bracket (']') in regular expressions (non-standard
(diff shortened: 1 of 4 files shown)


svaarala:
Still need to implement the character class case.

svaarala:
Ok, seems to fix #643.

Issue #927, opened 2016-08-24
Rework tooling into a combined config-and-prepare-sources model


svaarala:
At the moment there is a `make_dist.py` tool which creates an end user distributable package. The dist tool creates a directory structure and copies over files, config metadata, etc. It also autogenerates some code (like Unicode tables) and creates the separate and combined end user distributable sources, as well as a few `duk_config.h` examples. The end user distributable doesn't include the ability to combine the sources from scratch.

This model doesn't work well for a few cases:

- ROM built-ins involve automatic generation of object initializers, driven by YAML metadata which may also be supplied by the application writer (e.g. to embed custom bindings into ROM).
- Some config options should affect the dist processing. For example, if buffer object support is disabled, the corresponding buffer built-in bindings should be dropped (rather than left in but implemented as error throwing stubs like at present).

Those two cases need a combined step which both prepares a `duk_config.h` header and simultaneously prepares sources for compilation. The inputs to this process include everything genconfig currently accepts (`DUK_USE_xxx` options) but also custom built-in YAML metadata and other miscellaneous options (like whether to use `#line` directives or not). The combined tool could provide shortcut options like `--no-es6` which can be implemented based on YAML metadata tags in `builtins.yaml`.

Tasks:

- [x] Move scripts that should be in the end user distributable into `tools/` and keep scripts not part of the distributable in `util/`.
- [x] Fix `make_dist.py` and other basic build steps for the relocation. At this point builds and disting works as before.
- [x] Split `make_dist.py` into two parts. One part creates the end user distributable directory hierarchy and copies files. Another part is a separate utility for generating Unicode tables, combining sources, etc. This second utility ("prepare source") goes into the distributable, and make_dist.py would run the utility to generate prepared sources with default configuration for convenience (as is the case in the distributable now).
- [x] Merge genconfig with the prepare source utility to form an initial config-and-prepare utility which doesn't yet take advantage of the combined nature of the process.
- [x] Implement initial support for dropping optional bindings entirely rather than having them present with functions that always throw.
- [x] Fix testrunner jobs which run genconfig to use config-and-prepare instead
- [x] Fix example scripts (like util/example_rombuild.sh) also
- [x] Tool naming; `prepare_sources.py` is not an ideal name for "the frontend"
- [x] Fix prepare_sources.py to create either separate or combined sources into the target directory (with no further hierarchy which is confusing); Unicode handling will be slow, figure out if there's any way to speed it up
- [x] Remove dist/config/genconfig.py copy
[...]

PR #928, opened 2016-08-25
Reorganize tooling for a split dist / config-and-prepare process

svaarala:
Main changes:

- [x] Reorganize Python scripts into `tools/` (goes into distributable) and `util/` (development time).
- [x] Create a `tools/prepare_sources.py` tool which combines the functions of `genconfig.py`, `genbuiltins.py` and combining sources for compilation.
- [x] Change make_dist.py copy both `tools/` and the main sources directory (which gets the name `src-input/` in the dist for now) into the distributable.
- [x] Change make_dist.py to only copy files over and then run `tools/prepare_sources.py` (from the dist directory) to create the default source and config files expected by applications (`src/`, `src-noline/`, and `src-separate`).
- [x] Add Python PEP8 indent test (no tabs) to code policy check and fix Python scripts to comply.
- [x] Remove obsolete genconfig.py barebones-header and autodetect-header commands (deprecated in 1.x and `duk-config-header` command replaces them)

This implements parts of #927. The end result is usable but not very pretty so a separate pull will improve the code structure a bit. I would like to use Python utilities via relative imports but that's unfortunately not entirely intuitive.


svaarala:
Tests will fail for a bit because they rely on previous file organization. I'll fix that before merging.

svaarala:
Some tests will fail until this change is in master so merging with some test errors.

There's still work to do, but I've confirmed that the prepared sources in the distributable (`src/`, `src-noline`, and `src-combined/`) come out as before. Also dist `config/genconfig.py` is still present (it will be moved to `tools/genconfig.py`) to avoid upsetting the situation in master yet.

PR #929, opened 2016-08-26
Cleanups for dist and prepare-and-config split

svaarala:
Implement parts of #927:
- [x] Remove make_dist.py obsolete comments, explicit src-input copy
- [x] Remove dist `config/genconfig.py` copy
- [x] Git grep for references to moved tools (e.g. in comments) and fix
- [x] Reduce tool output (combine_src.py)
- [x] Add option for custom UnicodeData.txt and SpecialCasing.txt for prepare-and-config, useful for controlling footprint
- [x] Add pre-stripped 8-bit limited UnicodeData.txt and SpecialCasing.txt into dist
- [x] Low memory note for Unicode tables (can save about 3kB of footprint)
- [x] Use stripped Unicode tables for stripsize build
- [x] Underscore test shim fixes for newer Underscore
- [x] PEP8 indent fix for testrunner code


svaarala:
This pull adds stripped Unicode tables to the "stripsize" continuous integration build - which now compiles a REPL to less than 115kB on Linux x86. That stripped profile still includes Array, String, Date, etc built-ins (but doesn't include regexp support or bufferobject support). I'll refine the profile later into a fully stripped build so that it can be more easily compared to "stripped subset" builds of other engines; it'll probably compile down to ~100kB or so on x86. That's a nice figure because it'll be possible (although just barely) to run Duktape on 128kB flash with a barebones OS, and 192kB flash will be relatively straightforward.

PR #932, opened 2016-08-27
Use Math replacements by default on AmigaOS 3

svaarala:
Use math replacements even when not compiling using VBCC: even when when cross compiling with GCC some math asserts fail.

- [x] Figure out best check: all AmigaOS regardless of version, or is there some version after which replacements are not necessary?
- [x] Releases entry


svaarala:
Best fix so far: enable math replacements with AmigaOS and M68K, regardless of AmigaOS version or compiler.

PR #933, opened 2016-08-27
More cleanups for dist and prepare-and-config split

svaarala:
- [x] Rename `prepare_sources.py` tool; naming matters because it'll become an important external tool.
- [x] Refactor prepare_sources.py so that it only generates separate or combined sources (with or without #line directives) as requested per execution, rather than creating all variants. Rework make_dist.py to run the tool multiple times.
- [x] Metadata files for build and prepared sources: `duk_dist_meta.json` at top of distributable, `duk_source_meta.json` in a prepared source directory
- [x] Copy config metadata directly into dist package (rather than tar.gz'ing it)
- [x] Remove pre-built `duk_config.h` examples, they're pretty pointless
- [x] Check dist on Windows
- [x] Unicode speedups? => Small improvements made, separate pull for more
- [x] Short form options? Limit to a small set because they make any call sites less obvious => Nothing apparent, keep current set
- [x] Wiki draft for 2.x source preparation
- [x] Compare dist package to the one in master and 1.5.0, check any extra files are intentional
- [x] Fix testrunner jobs that use prepare_sources.py directly (path changes)
- [x] Update Duktape repo snapshot for testrunners

Implements parts of #927.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1067,7 +1067,7 @@ cloc:	dist cloc-1.60.pl
 # XXX: make prints a harmless warning related to the sub-make.
 dist:
 	@make codepolicycheck
-	$(PYTHON) util/make_dist.py --create-spdx
+	$(PYTHON) util/dist.py --create-spdx
 
 .PHONY:	dist-src
 dist-src:	dist
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Automatically generated bleeding edge snapshots from master are available at
 [duktape.org/snapshots](http://duktape.org/snapshots).
 
 You can also clone this repository, make modifications, and build a source
-distributable on Linux, OSX, and Windows using `python util/make_dist.py`.
+distributable on Linux, OSX, and Windows using `python util/dist.py`.
 
 Getting started: modifying and rebuilding the distributable
 -----------------------------------------------------------
@@ -74,17 +74,17 @@ distributable in Linux, OSX, or Windows:
 
     # Linux; can often install from packages or using 'pip'
     $ sudo apt-get install python python-yaml
-    $ python util/make_dist.py
+    $ python util/dist.py
 
     # OSX
     # Install Python 2.7.x
     $ pip install PyYAML
-    $ python util/make_dist.py
+    $ python util/dist.py
 
     # Windows
     ; Install Python 2.7.x from python.org, and add it to PATH
     > pip install PyYAML
-    > python util\make_dist.py
+    > python util\dist.py
 
 The source distributable directory will be in `dist/`.
 
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1761,8 +1761,8 @@ Planned
 * Incompatible change: genconfig.py has been relocated to tools/genconfig.py
   in the end user distributable (GH-929)
 
-* Incompatible change: make_dist.py no longer supports ROM built-ins, use
-  tools/prepare_sources.py instead (GH-929)
+* Incompatible change: util/dist.py no longer supports ROM built-ins, use
+  tools/configure.py instead (GH-929)
 
 * Include raw input sources and a prepare-and-config tool in the distributable,
   which allow user code to regenerate a config file and source code files for
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,7 +25,7 @@ build_script:
   # Make dist.
 
   - cmd: cd C:\projects\duktape
-  - cmd: python util\make_dist.py
+  - cmd: python util\dist.py
 
   # --- Visual Studio 2015 ---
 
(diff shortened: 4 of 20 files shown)


svaarala:
@fatcerberus Any name suggestions for the "prepare config and sources" tool? Current name: `tools/prepare_sources.py`.

Some possibilities:

- `tools/prepare.py`
- `tools/configure.py`
- `tools/config.py`
- `tools/setup.py`

Nothing sounds immediately right. But the name matters because it'll be "the tool" for setting up a build later on (and will subsume genconfig.py).

svaarala:
Unicode table preparation is the most important time consumer in prepare_sources.py: out of ~11.5 seconds for one run 8.5 seconds are spent on preparing Unicode data. Now that make_dist.py runs prepare_sources.py three times (to create each of the prepared source variants which is cleaner) this adds up to a significant amount of make_dist.py time.

svaarala:
Example of current state: to create combined (amalgamated) sources using Unicode data limited to 8-bit codepoints, create a temporary directory `/tmp/out` and:

```
$ python2 tools/prepare_sources.py \
      --config-metadata config \
      --source-directory src \
      --output-directory /tmp/out \
      --line-directives \
      --unicode-data src/UnicodeData-8bit.txt \
      --special-casing src/SpecialCasing-8bit.txt 
$ ls -l /tmp/out
total 3356
-rw-rw-r-- 1 duktape duktape   31419 Aug 29 04:26 duk_build_meta.json
-rw-rw-r-- 1 duktape duktape   89062 Aug 29 04:26 duk_config.h
-rw-rw-r-- 1 duktape duktape 3223220 Aug 29 04:26 duktape.c
-rw-rw-r-- 1 duktape duktape   68201 Aug 29 04:26 duktape.h
-rw-rw-r-- 1 duktape duktape   17201 Aug 29 04:26 metadata.json
```

So the tool now creates a flat output directory with just the sources and some metadata files. The metadata files will be changed, they're a bit inconsistent now.

fatcerberus:
> Any name suggestions for the "prepare config and sources" tool?

Hm.  I would agree `tools/prepare_sources.py` is too verbose.  Honestly the ideal situation for me would be if I could just do:

```
python configure.py
```

Even with the current setup, it always annoys me that I have to tack on `tools/` to the beginning of the script every time I want to make a dist in Windows (yeah, I could make a batch file, but where's the fun in that? :-)

If it's not possible to put the script in the root though, I'd be okay with `tools/configure.py`.  "configure" has the nice benefit that it's close to what, e.g. Unix developers would typically expect.  I'll also be glad to see that underscore disappear from the filename, whatever gets picked.  That underscore in `make_dist` is even more annoying to type than the `tools/` prefix!

svaarala:
`tools/configure.py` would conjure an association with autoconf `./configure` which is not too much from the truth: ./configure also allows generation of source files and/or header files based on options.

I could rename `util/make_dist.py` to simply `util/dist.py`?

svaarala:
Moving the script to the top level would also be possible, but it would also then need to be moved in master - otherwise the same commands wouldn't work in master as in the dist. I'll be moving a few files around to ensure that ideally the same relative paths would work from master. The main issue there is the `src/` directory which is the "raw" input sources in master and the single-file, #line-directive inclusive combined source in dist.

svaarala:
By the way, you'll still want to use a script file for preparing sources: there are quite a few options because the tool combines both `duk_config.h` and source preparation aspects. The options are intentionally long form (though I'll probably add short forms for the most common ones) because the call sites for the tool should be as close to self documenting as possible.

fatcerberus:
I assumed calling the script with no options would just use defaults for everything, i.e. it would use the same options as in an official dist.

svaarala:
But you still need to define a source directory, an output directory, etc.

svaarala:
The minimum command, which creates amalgamated sources with default configuration, is:

```
$ python tools/configure.py \
      --output-directory /tmp/out \
      --source-directory src \
      --config-metadata config
```

svaarala:
Wiki documentation draft for configuring 2.x for build: https://github.com/svaarala/duktape-wiki/blob/02b4d59f982a5a4affb6d63044f666d4b68fd424/Configuring2x.md.

fatcerberus:
Currently in master, `python util\make_dist.py` still works as before.  Should I expect that to stop working at some point in the near future?

svaarala:
It will be renamed but otherwise work as before, with the exception that it won't support ROM built-ins options (`--rom-support`, `--rom-auto-lightfunc`) or user-provided built-in object metadata (`--user-builtin-metadata`).

svaarala:
I'll merge this pull soon so master will have the new tools and new naming. I'll use at least one more pull to tweak option names and such, and maybe try to optimize Unicode preparation a bit.

PR #942, opened 2016-08-28
Add __attribute__ ((unused)) for internal functions for gcc/clang

### LINKED ISSUE #916: Compile fails with -Werror=unused-function
I imported duktape into PacRunner to handle proxy PAC scripts.

The build fails because it uses -Werror=unused-function:
```
duktape/duktape.c:9796:19: error: ‘duk_err_unsupported_defmsg’ defined but not used [-Werror=unused-function]
DUK_INTERNAL void duk_err_unsupported_defmsg(duk_hthread *thr, const char *filename, duk_int_t linenumber) {
```

For the GCC/DUK_SINGLE_FILE build, this can be avoided by adding `__attribute__((unused))` to the definition of `DUK_INTERNAL`, thus:
`#define DUK_INTERNAL static __attribute__((unused))`

svaarala:
Fixes #916.

Tasks:
- [x] Implement fix in default configuration files
- [x] Check fix works in practice
- [x] Check both combined and separate source compilation
- [x] Releases entry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -272,7 +272,7 @@ CCOPTS_SHARED =
 CCOPTS_SHARED += -D_POSIX_C_SOURCE=200809L  # to avoid linenoise strdup() warnings
 CCOPTS_SHARED += -pedantic -ansi -std=c99 -fstrict-aliasing
 # -Wextra is very picky but catches e.g. signed/unsigned comparisons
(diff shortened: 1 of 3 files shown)


svaarala:
Gcc/clang manuals seem to indicate the attribute should be applied to the function declaration but that doesn't seem to work in practice. Adding the attribute to both declaration and definition seems to work.

PR #946, opened 2016-08-30
Add -lstring variants for property access API calls

jchlapinski:
I am using Duktape trough FFI in my project written in Rust. In Rust, strings are not null-terminated and have specified length. Many of the duktape API functions already have "lstring" variants that allow me to pass Rust strings directly, without the need to copy and null-terminating them, which is great. However a few rather useful variants for property access are missing. In this PR I would like to add the following functions to the public API:
```
DUK_EXTERNAL_DECL duk_bool_t duk_get_prop_lstring(duk_context *ctx, duk_idx_t obj_idx, const char *key, duk_size_t key_len);
DUK_EXTERNAL_DECL duk_bool_t duk_put_prop_lstring(duk_context *ctx, duk_idx_t obj_idx, const char *key, duk_size_t key_len);
DUK_EXTERNAL_DECL duk_bool_t duk_del_prop_lstring(duk_context *ctx, duk_idx_t obj_idx, const char *key, duk_size_t key_len);
DUK_EXTERNAL_DECL duk_bool_t duk_has_prop_lstring(duk_context *ctx, duk_idx_t obj_idx, const char *key, duk_size_t key_len);
DUK_EXTERNAL_DECL duk_bool_t duk_get_global_lstring(duk_context *ctx, const char *key, duk_size_t key_len);
DUK_EXTERNAL_DECL duk_bool_t duk_put_global_lstring(duk_context *ctx, const char *key, duk_size_t key_len);
```

Those additional implementations are not-so trivial to implement outside of Duktape API, and probably adding them to the API will not enlarge library size by much. This change will make much easier to integrate Duktape into languages with non null-terminated strings.

Regards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.strip
 /_*
 /duk.*
+/duk
 /libduktape.*
 /build/
 /dist/
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -37,6 +37,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Ole André Vadla Ravnås (https://github.com/oleavr)
 * Harold Brenes (https://github.com/harold-b)
 * Oliver Crow (https://github.com/ocrow)
+* Jakub Chłapiński (https://github.com/jchlapinski)
 
 Other contributions
 ===================
--- a/src/duk_api_object.c
+++ b/src/duk_api_object.c
@@ -44,6 +44,15 @@ DUK_EXTERNAL duk_bool_t duk_get_prop_string(duk_context *ctx, duk_idx_t obj_idx,
 	return duk_get_prop(ctx, obj_idx);
 }
 
+DUK_EXTERNAL duk_bool_t duk_get_prop_lstring(duk_context *ctx, duk_idx_t obj_idx, const char *key, duk_size_t key_len) {
+	DUK_ASSERT_CTX_VALID(ctx);
+	DUK_ASSERT(key != NULL);
+
+	obj_idx = duk_require_normalize_index(ctx, obj_idx);
+	duk_push_lstring(ctx, key, key_len);
+	return duk_get_prop(ctx, obj_idx);
+}
+
 DUK_EXTERNAL duk_bool_t duk_get_prop_index(duk_context *ctx, duk_idx_t obj_idx, duk_uarridx_t arr_idx) {
 	DUK_ASSERT_CTX_VALID(ctx);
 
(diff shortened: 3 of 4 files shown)


svaarala:
Sounds good to me. These API call variants are logical extensions of existing calls and I hadn't considered your use case (I mostly work with plain C strings). I'll add some testcases, documentation, etc, in a follow-up.

Could you add yourself to AUTHORS.rst (in a separate commit for clarity)?

fatcerberus:
@svaarala What was the original motivation for the lstring variant calls that _do_ currently exist?

svaarala:
Completeness: you may want to push strings from non-terminated buffers (e.g. when decoding binary serialized data structures), or maybe a string might contain a NUL character; Duktape strings are byte pure so these cases happen occasionally. But in normal C code you don't encounter them very often, so they don't need to be very convenient.

Btw, Lua API actually has a third variant, "literal", which is used when the argument is a C literal so that `strlen()` can be done at compile time (it's a macro) but you don't need to specify the length explicitly. This results in faster code but the call sites need to know which strings are literals (not always obvious if an argument is, say, a macro).

jchlapinski:
I have added myself to AUTHORS.rst, sorry for the mess in commits, I have very little experience in syncing forks yet...

svaarala:
No worries - the commits seem to be a bit broken right now, but should be fixable by doing an interactive rebase (git rebase -i) against the current master, and squashing the commits into two (main change and AUTHORS.rst).

jchlapinski:
That was really painful, I nearly cried, but hopefully those commits now are acceptable?

svaarala:
Unfortunately the commits seem to be a bit broken :/

Maybe you could do the following:

- git checkout master; git pull
- git checkout <this branch>
- git rebase -i master

In the interactive commit list, use "pick" for the first one and "squash" for the others. Edit the commit message to be something useful. The end result would be a single commit against master with hopefully only the intended changes.  Then:

- git diff master

To check that the changes are what you intended.

jchlapinski:
hmm, according to https://github.com/svaarala/duktape/blob/master/CONTRIBUTING.md I have created a branch in my fork named api-property-access-lstring-variants, which I then used to make this PR. Should I just make my changes in master in my fork?

jchlapinski:
I think I might have gotten this right this time :)

svaarala:
@jchlapinski Looks good now - thanks :) I'll merge tomorrow and open a follow-up for tests etc.

PR #948, opened 2016-08-31
More cleanups for dist/configure tools

svaarala:
- [x] Option naming, e.g. --user-builtin-metadata is long and not very accurate => --user-builtin-metadata renamed --builtin-file to be consistent with genconfig.py --option-file
- [x] Add explicit options for dist input/output
- [x] Unicode performance => Future work (XXX note)
- [x] Unicode table generation could now be simplified: only generate the tables indicated by options or add specific --ascii-only, --bmp-only, options (XXX note)
- [x] Dist directory final changes, e.g. naming of directories, src-input in particular => Future work
- [x] Remove src-separate and/or src-noline from dist package? => Keep for now, no meaningful size difference
- [x] Don't autodelete a directory in `dist.py` (dangerous if directory argument is incorrect) but create one
- [x] Consistent --output-directory behavior for `dist.py` and `configure.py`
- [x] --quiet option
- [x] --verbose option
- [x] Use Python logging and info/debug separation
- [x] Review info log level
- [x] Review debug log level (comment out excessive spam)


svaarala:
Ideally I'd like to remove the pre-configured `src`, `src-noline`, and `src-separate` directories from the dist package and instead include a Makefile or a Python script to create them.

However, the default configuration is useful for quite many users and they may not want to use the Python tools to use the distributable.

fatcerberus:
I would agree with keeping at least the pre-configured `src`.  I've been known myself to just pull the premade sources from an official dist when upgrading Duktape for minisphere.  `src-noline` and maybe `src-separate` have more limited application and I feel like it's okay to require custom configuration to arrive at them, as long as everything needed to do so is available directly in the dist.

svaarala:
The only difference between running `tools/configure.py` from the repo root vs. dist root is that the input source directory name is different: `src/` in repo root, `src-input` in dist root.

I could change the dist to use `src/` for the raw input sources and `src-combined` for the combined source, but I'm pretty sure people would then try to compile the source files in `src/` (which need preparation before use).

So, maybe I'll rename the repo root source directory to something which matches the dist package, e.g. `src-input` (as it is now) or `src-raw` or something.

jchlapinski:
As a side note please also consider adding *.pyc to .gitignore

svaarala:
@jchlapinski Good idea, I'll do that.

svaarala:
Apart from dist time, there's not that much benefit in removing `src-separate` and `src-noline`: the resulting tar.xz is practically the same size than before (compression works really well for the mostly duplicated sources). Unpacked size doesn't usually matter all that much.

So, I'm leaving the variants in for now.

fatcerberus:
The main reason I suggested removing them was for purity more than anything.  Otherwise it potentially becomes a slippery slope - "Well, you include preconfigured sources for (insert scenario here), so why not (insert other scenario here)."  Then you end up with a bloated dist. :)

That said, `src-seperate` does justify its existence in this regard I think--it's a pretty common need to have individual sources.

svaarala:
Right, but why include *any* prepared sources in that case? That's the beginning of the slippery slope ... :)

svaarala:
In particular, if there were no prepared sources by default, users would get immediately familiar with how to prepare a config header and sources for build.

PR #949, opened 2016-08-31
Remove DUK_OPT_xxx feature option support

### LINKED ISSUE #236: Remove support for DUK_OPT_xxx flags
Remove support for `DUK_OPT_xxx` from `duk_config.h` and `genconfig`. Update documentation to match.

svaarala:
With the new configuration tooling in place, it's finally time to remove DUK_OPT_xxx support:

- [x] Remove DUK_OPT_xxx support from genconfig
- [x] Change dist to include a legacy feature option check (#error if present) for prepared default sources
- [x] Makefile fixes
- [x] Testrunner fixes
- [x] Fix dist-files
- [x] Fix config metadata
- [x] Matrix compile fixes - separate pull
- [x] Website fixes
- [x] Internal documentation fixes
- [x] Git grep DUK_OPT_ and fix any remaining call sites and documentation
- [x] 2.0 migration notes
- [x] Releases entry
- [x] Configure.py optional `--source-directory` and `--config-metadata`, default from script path
- [x] Cleaner output for configure.py defaulting, log final directories
- [x] Assert fix for pushing readable tval string (null dereference if error-aware and error .message not a plain string; not in 1.5.0 so no bug issue needed)

Fixes #236.


svaarala:
It'll take a while to fix all the internal call sites like the Makefile, so tests will be failing for some time.

svaarala:
@fatcerberus This pull will be merged in a little while - as it removes DUK_OPT_xxx it breaks Minisphere build which depends on `DUK_OPT_HAVE_CUSTOM_H`.

One approach to fixing this is to add the following pre-step to your build (committing the prepared sources in `src/shared/` if that's preferable):

```
# Executed from distributable (only difference if running from Duktape repo
# is the 'src-input' path which is 'src' in repo).

$ python tools/configure.py \
      --output-directory /tmp/output \
      --source-directory src-input \
      --config-metadata config \
      --fixup-file src/engine/duk_custom.h
$ cp /tmp/output/duktape.c /tmp/output/duktape.h /tmp/output/duk_config.h src/shared/
```

This integrates `duk_custom.h` into the `duk_config.h` statically. If you'd rather keep the file as separate and make `duk_config.h` include it dynamically (so that it can be more easily edited):

```
$ python tools/configure.py \
      --output-directory /tmp/output \
      --source-directory src-input \
      --config-metadata config \
      --fixup-line '#include "duk_custom.h"'
$ cp /tmp/output/duktape.c /tmp/output/duktape.h /tmp/output/duk_config.h src/shared/
```

The above embeds the `#include "duk_custom.h"` directive into `duk_config.h` so that it gets included dynamically which is perhaps what you'd prefer.

fatcerberus:
Thanks for the heads up.  I've been meaning to improve that situation for a while, but was too lazy to fix what wasn't broken... until now. :-)

svaarala:
If you want to keep Minisphere compatible with 1.x too, you can leave in the `DUK_OPT_HAVE_CUSTOM_H` option. By default configure.py won't emit a sanity check for 1.x feature options (though the sources in the default distributable do have that check enabled).

fatcerberus:
Yeah, I'd like to maintain 1.x compatibility at least until 2.0.0 goes gold.  There are too many untested changes going into Duktape master right now for me to depend solely on the latest builds. :)

fatcerberus:
This got merged, so looks like I'd better get to work on fixing minisphere. :)

svaarala:
The fix the testcase is using is simply: https://github.com/svaarala/duktape/blob/master/testrunner/client-simple-node/run_commit_test.py#L827-L834.

In other words, the fixup file is integrated into `duk_config.h`. You could do the same - and you can leave in the `DUK_OPT_xxx` feature options because the generated header will ignore them (as long as you don't explicitly request for feature option sanity checks for the header, they won't be present).

fatcerberus:
The main reasons I like using a fixup file instead of genconfig:
* Easy to turn on and off options when troubleshooting, e.g. assertions, no need to re-run genconfig
* When adding an option, no need to search through the large auto-generated `duk_config.h` to find the option(s) I need, and I can see at a quick glance which non-default options are in use.

svaarala:
> Easy to turn on and off options when troubleshooting, e.g. assertions, no need to re-run genconfig

Hmm, but what's the difference to keeping the options in a YAML file and running genconfig automatically for every build? The end result should be the same: you edit some file to enable/disable assertions and rebuild?

> When adding an option, no need to search through the large auto-generated duk_config.h to find the option(s) I need, and I can see at a quick glance which non-default options are in use.

Same question here :-) If you keep your options in a YAML file, that YAML file (given to configure.py) lists the options that you want to be different from defaults.

In other words, you don't need to read or edit `duk_config.h` in normal use.

svaarala:
In any case, you can use a fixup file with Duktape 2.x too: just run configure.py once with `--fixup-line '#include "duk_custom.h"` and the resulting duk_config.h will dynamically include your fixup file and you can edit it as before.

fatcerberus:
Mainly just that I don't have to put `genconfig.py` in my repo and avoid wrestling with MSVC's build system.  I do most of my compiling in Windows in the MSVC IDE, and MSBuild doesn't always cooperate when using post-build steps.  For example MSBuild doesn't re-run post-build steps if it thinks the rest of the project is up to date.  Since the fixup *is* part of the project and MSVC recognizes it as such, modifying it triggers a recompilation and everything works smoothly.

svaarala:
> Since the fixup is part of the project and MSVC recognizes it as such, modifying it triggers a recompilation and everything works smoothly.

Hmm, so I guess that your YAML options file should be part of the project as far as dependencies are concerned. But since I don't use MSVC IDE I'm not sure how that works in practice.

Anyway, you can use a fixup header if you add that `--fixup-line` option and run configure.py once, and then store the resulting headers and sources just as you would do with a pre-built distributable.

fatcerberus:
Yeah, I got that, thanks :-)

svaarala:
Let me know if that works for you in practice :) I wrote that as one of the upgrade approaches by the way:

> If you're using a duk_custom.h header there are three simple approaches:
>
>   - To embed your custom header into duk_config.h statically, use --fixup-file duk_custom.h in tools/configure.py options.
>   - To include your custom header at compilation time, similarly to DUK_OPT_HAVE_CUSTOM_H, use --fixup-line '#include "duk_custom.h"' in tools/configure.py options.
>   - Finally, you can remove your custom header and add the equivalent options to tools/configure.py when possible.

fatcerberus:
Just took my first dive into the new configure.py method.  Much faster than making a dist every time, however I have a complaint: the output directory must be removed each time I build.  configure.py won't build in a directory that already exists, which is annoying.  This wasn't an issue with make_dist.py: If the `dist` directory already existed, it would just be reused.

As for `--fixup-line`, it works for minisphere, however it prevents Cell from compiling since it doesn't include a `duk_custom.h`.  I now remember this was the reason I liked `DUK_OPT_HAVE_CUSTOM_H`, because I could share a single `duk_config.h` between multiple projects.  So it looks like I will have to change my Duktape workflow now. :)

svaarala:
> This wasn't an issue with make_dist.py: If the dist directory already existed, it would just be reused.

It was actually deleted and recreated - but I think it's rather dangerous if the directory is autodeleted. Consider what would happen if you said `--output-directory /` or if output directory (at some later date) defaulted to some unintuitive location and got nuked.

> I now remember this was the reason I liked DUK_OPT_HAVE_CUSTOM_H, because I could share a single duk_config.h between multiple projects. So it looks like I will have to change my Duktape workflow now. :)

You can still get the previous behavior if you want. Just use `--fixup-file XXX` and provide the following contents:

```
#if defined(DUK_OPT_HAVE_CUSTOM_H)
#include "duk_custom.h"
#endif
```

The generated sources will then behave as before.

svaarala:
> > This wasn't an issue with make_dist.py: If the dist directory already existed, it would just be reused.

> [...] Consider what would happen if you said --output-directory / or if output directory (at some later date) defaulted to some unintuitive location and got nuked.

One safer but more convenient option would be for `configure.py` to:

- Create the directory if it's missing.
- If the directory exists, check that there are only known files there (`duktape.c`, `duktape.h`, `duk_config.h`, `duk_source_meta.json`) and if so, delete the known files only. If there are any other files there, error and delete nothing.

fatcerberus:
Good point, I'll try that.

As for the autodeletion behavior, I understand why it works this way now and generally agree, it doesn't make it any less annoying when I forget to delete the directory first though. :)

However I'll have to make a batch file/script to run configure.py anyway since typing all the options all the time is tedious, so I can just add `rmdir /q /s src` before running Python.  So it's probably a non-issue in practice for most users.

svaarala:
#957 adds that reuse behavior - it's conservative so it never reuses a directory if there are any other files there (other than known output files) and never deletes anything but those known output file names specifically.

fatcerberus:
> You can still get the previous behavior if you want. Just use --fixup-file XXX and provide the following contents

Thanks, just tried this and it worked like a charm.  I broke down and started automating the process, I added an `ms` directory to my local copy of the Duktape repo where I can just run "make" to get Duktape sources for minisphere.  I should probably move it elsewhere soon so I don't accidentally wipe it with an errant `git clean -xdf` though. :-)

Sadly I will no longer be able to use virgin sources from the dist, but I'll deal with it.  Small price to pay for all the benefits the new system brings.

PR #959, opened 2016-09-08
Fix duk_put_prop_string() bug when target and value are the same slot

svaarala:
When both the target object and property value are the same value stack slot for duk_put_prop_string() the result is not as expected.  No memory unsafe behavior happens though.

There's no similar issue in duk_put_prop(). The issue is caused by how duk_put_prop_string() adds the key to the value stack before calling duk_put_prop(); the normalized object index remains the same but is actually incorrect if the target was also the value:

```c
	obj_idx = duk_require_normalize_index(ctx, obj_idx);
	duk_push_string(ctx, key);
	duk_swap_top(ctx, -2);  /* [val key] -> [key val] */
	return duk_put_prop(ctx, obj_idx);
```

The same issue does affect duk_put_prop_lstring(), duk_put_prop_index(), and duk_put_prop_stridx().

Two possible fixes:
- Reject the situation as an API error as confusing.
- Support the situation. => This is implemented by the pull.

Tasks:
- [x] Figure out preferred fix
- [x] Add bug testcases for current issues
- [x] Releases entry
- [x] Backport to v1.5-maintenance and v1-maintenance => tracked via 1.5.2 fix list

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1877,6 +1877,11 @@ Planned
   for footprint optimized builds (ensure DUK_USE_ARRAY_PROP_FASTPATH is
   disabled for low memory builds) (GH-934)
 
+* Fix incorrect value stack handling in duk_put_prop_(l)string() and
+  duk_put_prop_index() when the target object and the property value
+  are in the same value stack slot (which is unusual but conceptually
+  clear) (GH-959)
+
 * Fix buffer object (duk_hbufobj) JSON serialization (bug present in 1.5.0):
   buffer objects were omitted from serialization when they should be
   serialized as normal objects instead (GH-867)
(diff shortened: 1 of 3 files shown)


fatcerberus:
What is the bug exactly?  In minisphere I set a global variable `global` to be the global object (like Node.js) and that works fine.

In any case I don't think it should be an API error since then the scenario described above is impossible to implement.

svaarala:
The concrete issue is described in the test case, but to repeat it here:

```c
duk_push_object(ctx);

/* Here the object is both the property set target -and- the input value. */
duk_dup_top(ctx);
duk_put_prop_string(ctx, -1, "example");
```

fatcerberus:
Ah, I see, so it's just value stack behavior.  The way you described it I thought it was triggered any time `target === value`.

svaarala:
I tried to avoid that interpretation by using the word "slot" in the title. But yeah, it's a bit of a corner case and only an issue in the value stack handling.

PR #968, opened 2016-09-16
Fixes to debugger document inspection text

### LINKED ISSUE #744: Ambiguity/error in debugger.rst
Under `GetObjPropDesc`, it says this:
> Properties stored in the internal "array part" are indexed using numeric string keys, e.g. "3", not integers, to avoid unnecessary string churn on the target.

This is already confusing (how does using string keys avoid string churn?), but then a few paragraphs down under "Key" it says:
> Number for array index properties of "dense arrays", i.e. arrays which have an internal array part present

Only one of these can be correct.  Which one is it? ;-)

svaarala:
Fixes #744.

### DIFF
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -2016,8 +2016,7 @@ result is either:
 * A "not found" error if the property doesn't exist.
 
 Properties stored in the internal "array part" are indexed using numeric
-string keys, e.g. ``"3"``, not integers, to avoid unnecessary string churn
-on the target.
+string keys, e.g. ``"3"``, not integers.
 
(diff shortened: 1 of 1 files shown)


fatcerberus:
I see you clarified the semantics for the "unused" values, that's good.  It's completely okay for the client to just ignore them, right?  i.e. Duktape will never send an "unused" value for a valid property so the client can drop them.

svaarala:
Right, they're just used for the dense array part when it has gaps. Eliminating them from the virtual index space would be straightforward but tedious which is why they're left in.

PR #969, opened 2016-09-17
Improve value summarization, remove explicit error aware variant

svaarala:
- [ ] Remove explicit error aware summarization variants; be error aware always
- [ ] Improve summarization of errors
- [ ] Improve summarization of functions
- [ ] Improve summarization of buffer objects / views
- [ ] Improve summarization of plain buffers


svaarala:
Starting with errors, a few possibilities and notes.

First alternative that comes to mind is to extend the class-based summary `[object TypeError]` to:
```
[object .name .message], e.g. [object TypeError foo]
```

Fall back to `[object TypeError]` if the .name or .message cannot be safely read of course.

A slight issue here is that if an error has .name `Uint8Array` (rare but possible) it can be confused with a buffer object. This could maybe be avoided by making the successful error aware serialization:

```
[error .name .message], e.g. [error TypeError foo]
```

The `[error...` prefix indicates safely that we're specifically serializing an error.

One small question is what to do if e.g. `.name` cannot be safely summarized but `.message` can (or vice versa). One can fall back to the generic variant of course but in some cases better behavior is possible. For example, if .name is not safe to serialize but .message is, one can print out `[object <classString> message]`.

Another question is whether the class name should still be kept (or just .name). For built-in errors the results are almost always the same, e.g. class name for a TypeError is `TypeError` like its `.name`. The `.name` can of course be modified so they *can* differ, they just almost never do.

For this my initial reaction is to use the `.name` if safe: if it has been overridden there's usually a good reason so displaying that makes more sense than the class-based named. On the other hand hiding the class-based name means that the error summarization is not a superset of the generic object summarization. But maybe that's small enough a technicality to ignore; showing both would look a bit silly, e.g. `[error TypeError TypeError foo]`.

svaarala:
For concreteness on the last example:
```
[error TypeError TypeError foo]     for new TypeError('foo')
[error TypeError MyErrorName foo]   for err = new TypeError('foo'); err.name = 'MyErrorName';
[error Object MyErrorName foo]      for a user error which is a subclass of TypeError
```

Using the .name only:

```
[error TypeError foo]     for new TypeError('foo')
[error MyErrorName foo]   for err = new TypeError('foo'); err.name = 'MyErrorName';
[error MyErrorName foo]   for a user error which is a subclass of TypeError
```

Using class only:

```
[error TypeError foo]   for new TypeError('foo')
[error TypeError foo]   for err = new TypeError('foo'); err.name = 'MyErrorName';
[error Object foo]      for a user error which is a subclass of TypeError
```

For me the .name only variant is most intuitive.

svaarala:
This will need a bit of groundwork: there's no internal primitive yet for looking up a property with no side effects, except for own properties. But `.name` is almost always inherited.

svaarala:
Quoting for .name and .message is a potential issue but more of a technicality than a real issue:
- If .name has spaces the name and the message will not be cleanly delineated, e.g. if .name is `TypeError foo` and .message is `bar` the result is `[error TypeError foo bar]` which is indistinguishable from .name `TypeError` and .message `foo bar`. However, error names don't by convention have spaces.
- Message may contain e.g. a closing bracket which may be confusing. But still readable.

fatcerberus:
Agreed on `.name` only being most intuitive, preferably with an optional fallback to the class name if that isn't safe.

svaarala:
Added a `summary(x)` binding to "duk" for testing:

```
duk> err = new TypeError('asdasdasdasdasdashdasdijdsaoidsaoijdsaoijdsaijdsaoijdsaoijdsadsa')
= {}
duk> err.name='TypeError'  // needed for now
= "TypeError"
duk> summary(err)
= "[error TypeError asdasdasdasdasdashdasdijdsaoidsa...]"
```

It will have to be removed before merge because the API is not yet public.

svaarala:
Changed .name handling so that if it's not safe to summarize, the class string is used instead. As a side effect the class string is used now when the property is not found in the object itself (no inheritance check yet) so the result is now:

```
duk> summary(new TypeError('aiee\u0000foo'));
= "[error Error aiee\\x00foo]"
```

I remembered the internal class numbering incorrectly: built-in errors have just one class so that e.g. `TypeError` is just an `Error` as far as its class is concerned. Which means .name is clearly more useful.

svaarala:
Even in its current state (not having Function or bufferobject support) the pull adds slightly less than 1kB to the minsize build so I'll probably need to work in some footprint controls.

fatcerberus:
Ouch, yeah, 1kB is pretty high especially for a feature that's mostly only useful for debugging.  Maybe there should be a compiler option for "verbose" summarization, and use basic summaries otherwise.

fatcerberus:
Since errors are being summarized as `[error ...]` here, maybe functions should be e.g. `[function ...]` (instead of `[object Function]`).

svaarala:
For functions interesting information includes:
- Function type: native, Ecmascript, bound. Though one could indicate "bound" and also the type of the final function in the bound function chain (e.g. bound and native, bound and Ecmascript).
- Function .name
- Function .fileName
- Function argument count and/or external .length

No idea what's in practice useful. The Function.prototype.toString() stringification includes function .name and its type, categorized to "native", "ecmascript", or "bound". Summarization matching that would be e.g.:

```
[function native cos]
```

or:

```
[function cos native]
```

svaarala:
> Maybe there should be a compiler option for "verbose" summarization, and use basic summaries otherwise.

Agree, it's somewhat similar to verbose/non-verbose errors.

fatcerberus:
I believe Node.js REPL summarizes functions like `[Function cos]`, might be misremembering that though.

svaarala:
I think it's `[Function: cos]`.

svaarala:
For errors:

```
> new TypeError('aiee')
[TypeError: aiee]
> err = new TypeError('aiee'); err.name = 'FooError'; err
{ [FooError: aiee] name: 'FooError' }
```

svaarala:
Although that's a bit different because it's also summarizing property values etc. Which gets us back to expected application use cases: should properties also be summarized safely, should that be done only for the top level or N levels, etc. There are a lot of use cases unfortunately to make good API design choices, so I still think one important goal is to provide enough inspection capabilities to allow applications or libraries do to inspection by themselves.

svaarala:
Agree with using `[Function ...` over `[object Function`. Similarly:
- Plain buffer can be just `[buffer ...`; the lowercasing indicates it's a primitive value. Alternative would be maybe `[PlainBuffer ...`.
- Buffer objects don't need a `[buffer ...` prefix because the class names are known so they could be directly `[Uint16Array...` etc.

The error summary requires that the object inherits from `Error.prototype`, has a string `.name` and a string `.message`. Current summarization is `[error <name> <message>]` but maybe we can overlook the potential confusion and use `[<name> <message>]` directly. For example: `[MyError invalid argument]`.

The potential for confusion is relatively low because this only affects actual Error instances (values inheriting from Error prototype) and `.name` has been explicitly set so it's presumably something useful. Sure, it can be `ArrayBuffer` but it's not likely to be a concrete concern.

To summarize, these changes would result in:

```
[buffer deadbeef 4 bytes]
[PlainBuffer deadbeef 4 bytes]   // alternative for plain buffer
[Uint16Array 0100020003000400 8 bytes]
[MyError invalid argument]
[Function cos]
```

svaarala:
Removing the "object" from generic object summary would also be possible, e.g. `[Array]` instead of `[object Array]`; it doesn't really add any information. It does have the merit of being recognizable as the Object.prototype.toString() coercion but that's not very important I think.

svaarala:
@fatcerberus Added a testcase for manual summary tests: https://github.com/svaarala/duktape/pull/969/commits/3a36e5825b056df5e36fdd5ddda1290311dee5b4.

This is the current printout:

```
error .name getter!
errd.name: FakeName
undefined
null
true
false
123
Infinity
-Infinity
NaN
''
'foo'
'foo쫾\x00bar'
'longstring1234567890abcdefghijkl...'
'x\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00...'
[buffer 666f6fecabbe626172 9 bytes]
[buffer 31323334353637383930313233343536 16 bytes]
[buffer 31323334353637383930313233343536... 30 bytes]
[buffer 00000000000000000000000000000000... 100 bytes]
[ArrayBuffer 0000000000 5 bytes]
[ArrayBuffer 00000000000000000000000000000000 16 bytes]
[ArrayBuffer 00000000000000000000000000000000... 50 bytes]
[Uint16Array 0100020003000400feca0dd0 12 bytes]
[Uint16Array 0100020003000400feca0dd034122143 16 bytes]
[Uint16Array 01000200030004000500060007000800... 32 bytes]
[DataView 0100020003000400feca0dd0 12 bytes]
[DataView 0100020003000400feca0dd034122143 16 bytes]
[DataView 01000200030004000500060007000800... 32 bytes]
[URIError failed to convert URI]
[Error]
[TypeError invalid argument]
[Error]
[MyError aiee]
[Function cos]
[Function myFunction]
[Function myFunction2]
[Function cos]
[String]
[Boolean]
[Number]
[Date]
```

In the pull the `[error TypeError ...` prefix is now just `[TypeError ...`. Functions only have the function name (no info on arguments or type). The `[object XXX]` generic output has been shortened to `[XXX]`.

There's not yet support for boxed types String, Boolean, Number, which would make sense. Similarly summarizing Date objects by either showing the time value or ISO 8601 UTC formatting of the Date would be useful.

svaarala:
Random notes:
- Maybe indicate truncated string character length? Including it for all strings would be spammy.
- Maybe omit byte length for non-truncated buffers (would then match string behavior). "... 32 bytes" could be shortened to "... (32)" or "... 32b".
- Buffer bytes are now joined without spacing, maybe `de ad be ef` is more readable than `deadbeef`.
- For views the segments would ideally match view byte size, e.g. Uint16Array `0100 0200 0300 0400`.
- Anonymous function (name is empty string) now comes out as `[Function ]` while `[Function]` would be better.
- String, Boolean, Number should maybe be unboxed, e.g. `[String foo\x00bar]`.
- Date support, e.g. `[Date 1970-01-15T06:56:07.890Z]`.
- Thread support, maybe indicate state?
- RegExp support, e.g. `[RegExp /myRegexp/g]`.

svaarala:
The more I work on this the more I have nagging doubts about the different available mechanisms. For example, summaries don't include property values now (not even for the first level) so one may still need to use JX. But JX doesn't provide the level of introspection summaries now provide.

Contrast to the Node command line:

```
> x = [1,2,3]; x.foo = 'bar'; x;
[ 1, 2, 3, foo: 'bar' ]
> y = new Date(1234567890); y.myProperty = 123; y;
{ Thu, 15 Jan 1970 06:56:07 GMT myProperty: 123 }
> z = {}; Object.defineProperties(z, { foo: { get: function myGetter() {}, enumerable: true }, bar: { value: 123, enumerable: true } }); z;
{ foo: [Getter], bar: 123 }
> w = new TypeError('invalid argument'); w;
[TypeError: invalid argument]
```

This kind of introspection both shows useful type-specific information (like the formatted date, similar to summary strings in this pull) but still shows further properties in the objects (like JX). Getters/setters are indicated as such and are not invoked (not supported in JX).

Maybe I should be working first on the missing introspection features to allow Node-style introspection to be implemented from the application side. With that in place it might be easier to figure out the best roles for JX and summaries.

I can complete this pull without answering that by sticking to formatting the object specifics without going into property keys at all. It'd then be an improvement over master anyway (provided that the low memory variant didn't increase footprint significantly) so it can be merged in any case. But this whole API area is not very clean cut right now :-/

fatcerberus:
w.r.t. introspection: It might be worth looking into the `Reflect` API introduced in ES6.  It looks like it might align with some of the goals here.

svaarala:
I don't think Reflect brings much more capability than what's already present e.g. via Object bindings? What bindings did you have in mind?

Somewhat related: it's useful to have stable introspection API calls that align with how object behave in Ecmascript. It's also useful to have less stable introspection APIs to expose things like "is this array internally dense?" or "is this buffer object backed by a fixed, dynamic, or external buffer?" or "is this value represented internally by a fastint?". I would really like "duk" to display this information because it'd make it easier to understand how e.g. fastints work.

fatcerberus:
For some reason when I wrote that, I thought `Reflect` had more comprehensive introspection stuff than it does.  I found it otherwise when I actually implemented it :)

Honestly though JavaScript doesn't really need that much reflection stuff, as a programming platform it's naturally pretty transparent already.

svaarala:
Agree, it's quite possible to implement a reasonably useful colorized, pretty printed value summary like DukLuv does. It uses just Ecmascript code and `Duktape.info()` to get at a few details under the hood. 

But the problem with that plain Ecmascript approach is that it's difficult to do safely without triggering any side effects: any properties can be getters/setters and any value can be a Proxy. While you can work around getters/setters using Object.getOwnPropertyDescriptor(), Proxies are a bit more difficult because traps can be invoked even if you just try to read a property descriptor. And a trap may then busy loop for example.

So what I'm mostly missing is a side effect free way to inspect values "deeply" so that you could get access to internals too (even if they won't be version stable). For example, a more versioning friendly take on `Duktape.info()` which could return an object like:

```js
{
    "type": 3,
    "internalType": 65531,
    "classNumber": 30,
    // ..
}
```

For versioning friendliness it could accept an object argument to provide control flags and parameters if they become necessary. An object argument is, again, quite version friendly and robust against parameters being added and removed.

This could provide the basis for doing various console dumps and such without introducing a lot of API primitives. The same binding could also be exposed to C code, through an API which just pushed an object on the value stack (a C struct would be pretty hostile versioning-wise).

svaarala:
@fatcerberus Regarding Reflect, I don't really understand why it duplicates so much of what Object already does, but just with small variations. I could understand the variations if Object was left untouched, but also Object semantics (e.g. for coercion) have also been changed slightly. Maybe there's a longer term plan.

PR #970, opened 2016-09-18
Switch to xoroshiro128+ for built-in PRNG

svaarala:
- [x] Add manual test to generate doubles using the original xoroshiro128+ implementation
- [x] Test generated double sequence manually against manual test
- [x] Add 64-bit implementation, all endianness variants
- [x] Add 32-bit implementation, all endianness variants => Skipped, use previous algorithm (Shamir's three-op algorithm) for targets without 64-bit types
- [x] Documentation and/or license changes
- [x] Releases

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1886,6 +1886,10 @@ Planned
   to override the built-in random number generator (which is very simple
   and low footprint optimized) with something faster or better (GH-824)
 
+* Change default built-in PRNG algorithm to xoroshiro128+ with SplitMix64
+  seed mixing; previous algorithm (Shamir's three-op algorithm) is still
+  used for low memory targets and targets without 64-bit types (GH-970)
+
 * Fix incorrect value stack handling in duk_put_prop_(l)string() and
   duk_put_prop_index() when the target object and the property value
   are in the same value stack slot (which is unusual but conceptually
--- a/config/config-options/DUK_USE_GET_RANDOM_DOUBLE.yaml
+++ b/config/config-options/DUK_USE_GET_RANDOM_DOUBLE.yaml
@@ -8,9 +8,9 @@ description: >
   Override the default internal random number generator which is used for
   Math.random() and some other internal call sites (currently, for example,
   Array.prototype.sort()).  The default random number generator has a very
-  low footprint, but has low performance and is not statistically very good.
-  It is also not suitable at all for cryptographic applications.  Overriding
-  the random number generator may thus be useful in some environments.
+  low footprint but is not suitable for serious statistics algorithms or
+  cryptography.  Overriding the random number generator may thus be useful
+  in some environments.
 
   The macro gets a heap userdata argument and must provide an IEEE double
   in the range [0,1[.
--- a/dist-files/README.rst
+++ b/dist-files/README.rst
@@ -120,9 +120,11 @@ commit @GIT_COMMIT@ (@GIT_DESCRIBE@).
 Duktape is copyrighted by its authors (see ``AUTHORS.rst``) and licensed
 under the MIT license (see ``LICENSE.txt``).  String hashing algorithms are
 based on the algorithm from Lua (MIT license), djb2 hash, and Murmurhash2
-(MIT license).  Duktape module loader is based on the CommonJS module
-loading specification (without sharing any code), CommonJS is under the
-MIT license.
+(MIT license).  Pseudorandom number generator algorithms are based on
+Adi Shamir's three-op algorithm, xoroshiro128+ (public domain), and SplitMix64
+(public domain).  Duktape module loader is based on the CommonJS module
+loading specification (without sharing any code), CommonJS is under the MIT
+license.
 
 Have fun!
 
--- a/licenses/splitmix64.txt
+++ b/licenses/splitmix64.txt
@@ -1,0 +1,7 @@
+/*  Written in 2015 by Sebastiano Vigna (vigna@acm.org)
+
+To the extent possible under law, the author has dedicated all copyright
+and related and neighboring rights to this software to the public domain
+worldwide. This software is distributed without any warranty.
+
+See <http://creativecommons.org/publicdomain/zero/1.0/>. */
--- a/licenses/xoroshiro128plus.txt
+++ b/licenses/xoroshiro128plus.txt
@@ -1,0 +1,7 @@
+/*  Written in 2016 by David Blackman and Sebastiano Vigna (vigna@acm.org)
+
+To the extent possible under law, the author has dedicated all copyright
+and related and neighboring rights to this software to the public domain
+worldwide. This software is distributed without any warranty.
+
+See <http://creativecommons.org/publicdomain/zero/1.0/>. */
--- a/misc/splitmix64.c
+++ b/misc/splitmix64.c
@@ -1,0 +1,28 @@
+/*  Written in 2015 by Sebastiano Vigna (vigna@acm.org)
+
+To the extent possible under law, the author has dedicated all copyright
+and related and neighboring rights to this software to the public domain
+worldwide. This software is distributed without any warranty.
+
+See <http://creativecommons.org/publicdomain/zero/1.0/>. */
+
+#include <stdint.h>
+
+/* This is a fixed-increment version of Java 8's SplittableRandom generator
+   See http://dx.doi.org/10.1145/2714064.2660195 and
+   http://docs.oracle.com/javase/8/docs/api/java/util/SplittableRandom.html
+
+   It is a very fast generator passing BigCrush, and it can be useful if
+   for some reason you absolutely want 64 bits of state; otherwise, we
+   rather suggest to use a xoroshiro128+ (for moderately parallel
+   computations) or xorshift1024* (for massively parallel computations)
+   generator. */
+
+uint64_t x; /* The state can be seeded with any value. */
+
+uint64_t next() {
+	uint64_t z = (x += UINT64_C(0x9E3779B97F4A7C15));
+	z = (z ^ (z >> 30)) * UINT64_C(0xBF58476D1CE4E5B9);
+	z = (z ^ (z >> 27)) * UINT64_C(0x94D049BB133111EB);
+	return z ^ (z >> 31);
+}
--- a/misc/splitmix64_test.c
+++ b/misc/splitmix64_test.c
@@ -1,0 +1,27 @@
+/*
+ *  Generate a few random doubles to compare Duktape internal implementation
+ *  against the original splitmix64 source.
+ *
+ *  $ gcc -o /tmp/test -std=c99 splitmix64.c splitmix64_test.c
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+
+extern uint64_t x;
+extern uint64_t next(void);
+
+int main(int argc, char *argv[]) {
+	union {
+		uint64_t x;
+		double d;
+	} u;
+	int i;
+
+	x = 0xdeadbeef12345678ULL;
+
+	for (i = 0; i < 10; i++) {
+		printf("%08llx\n", (unsigned long long) next());
+	}
+}
--- a/misc/xoroshiro128plus.c
+++ b/misc/xoroshiro128plus.c
@@ -1,0 +1,69 @@
+/*  Written in 2016 by David Blackman and Sebastiano Vigna (vigna@acm.org)
+
+To the extent possible under law, the author has dedicated all copyright
+and related and neighboring rights to this software to the public domain
+worldwide. This software is distributed without any warranty.
+
+See <http://creativecommons.org/publicdomain/zero/1.0/>. */
+
+#include <stdint.h>
+
+/* This is the successor to xorshift128+. It is the fastest full-period
+   generator passing BigCrush without systematic failures, but due to the
+   relatively short period it is acceptable only for applications with a
+   mild amount of parallelism; otherwise, use a xorshift1024* generator.
+
+   Beside passing BigCrush, this generator passes the PractRand test suite
+   up to (and included) 16TB, with the exception of binary rank tests,
+   which fail due to the lowest bit being an LFSR; all other bits pass all
+   tests. We suggest to use a sign test to extract a random Boolean value.
+
+   Note that the generator uses a simulated rotate operation, which most C
+   compilers will turn into a single instruction. In Java, you can use
+   Long.rotateLeft(). In languages that do not make low-level rotation
+   instructions accessible xorshift128+ could be faster.
+
+   The state must be seeded so that it is not everywhere zero. If you have
+   a 64-bit seed, we suggest to seed a splitmix64 generator and use its
+   output to fill s. */
+
+uint64_t s[2];
+
+static inline uint64_t rotl(const uint64_t x, int k) {
+	return (x << k) | (x >> (64 - k));
+}
+
+uint64_t next(void) {
+	const uint64_t s0 = s[0];
+	uint64_t s1 = s[1];
+	const uint64_t result = s0 + s1;
+
+	s1 ^= s0;
+	s[0] = rotl(s0, 55) ^ s1 ^ (s1 << 14); // a, b
+	s[1] = rotl(s1, 36); // c
+
+	return result;
+}
+
+
+/* This is the jump function for the generator. It is equivalent
+   to 2^64 calls to next(); it can be used to generate 2^64
+   non-overlapping subsequences for parallel computations. */
+
+void jump(void) {
+	static const uint64_t JUMP[] = { 0xbeac0467eba5facb, 0xd86b048b86aa9922 };
+
+	uint64_t s0 = 0;
+	uint64_t s1 = 0;
+	for(int i = 0; i < sizeof JUMP / sizeof *JUMP; i++)
+		for(int b = 0; b < 64; b++) {
+			if (JUMP[i] & 1ULL << b) {
+				s0 ^= s[0];
+				s1 ^= s[1];
+			}
+			next();
+		}
+
+	s[0] = s0;
+	s[1] = s1;
+}
--- a/misc/xoroshiro128plus_test.c
+++ b/misc/xoroshiro128plus_test.c
@@ -1,0 +1,34 @@
+/*
+ *  Generate a few random doubles to compare Duktape internal implementation
+ *  against the original xoroshiro128+ source.
+ *
+ *  Doesn't work on middle endian hosts.
+ *
+ *  $ gcc -o /tmp/test -std=c99 xoroshiro128plus.c xoroshiro128plus_test.c
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+
+extern uint64_t s[2];
+extern uint64_t next(void);
+
+int main(int argc, char *argv[]) {
+	union {
+		uint64_t x;
+		double d;
+	} u;
+	int i;
+
+	s[0] = 0xdeadbeef12345678ULL;
+	s[1] = 0xcafed00d12345678ULL;
+
+	for (i = 0; i < 100; i++) {
+		if (i == 10) {
+			printf("--- duk_heap_alloc init mixing ends here\n");
+		}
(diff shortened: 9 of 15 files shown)


svaarala:
Xoroshiro128+ is a little bit larger than the previous algorithm. I'll probably leave the original algorithm in for low memory targets - and it can also be used for targets without 64-bit operations. It'll save me some bit fiddling ;)

fatcerberus:
What's the footprint impact on x64?

fatcerberus:
One thing that bugs me looking over the changes is the seeding method.  For xoroshiro128+ it's recommended to use a different PRNG to seed the generator because otherwise seeds which are close in value will lead to undesirable correlation.  Since you're using the current time as the seed, this is a real concern.

minisphere uses a SplitMix64 generator for seeding, but maybe Duktape could use Shamir's algorithm for that purpose?

fatcerberus:
This was the relevant text, right from the xoroshiro128+ page:

> Nonetheless, SplitMix64 is very useful to initialize the state of other generators starting from a 64-bit seed, as research has shown that initialization must be performed with a generator radically different in nature from the one initialized to avoid correlation on similar seeds.

svaarala:
+144 bytes for 'make duk' x64, so pretty meaningless for desktop targets.

svaarala:
Yeah - that's why I added 10 rounds of reading from the PRNG to mix up the state. They'll still be correlated of course. But what's the practical alternative?

svaarala:
I could enable the Shamir three-op algorithm and use that, but whether it reduces the correlation or not is not a trivial thing to be convinced of. Running some correlation tests for nearby seeds would be useful to see that it has at least superficial impact.

svaarala:
Also just to put things into perspective: user code should always replace the built-in PRNG or use specific high quality PRNG bindings for statistics or cryptography applications. What remains is casual use of randomness, e.g. to generate images or sound, timer back-offs, spawn enemies, and such. Seed correlation might still be undesirable but is it a concrete issue?

For example, here are some Math.random() numbers generated by launching new Duktape heaps in a busy loop and printing one random number from each:

```
1474233638921 0.027313505862523124
1474233638922 0.14244310363856405
1474233638924 0.12584209587910733
1474233638925 0.7967707704011076
1474233638927 0.18310068630166954
1474233638928 0.3000993220985475
1474233638930 0.11968672972505012
1474233638932 0.7344654106195236
1474233638933 0.8443269813107293
1474233638935 0.970020924724039
1474233638937 0.49071643113174157
1474233638939 0.6690293678136026
1474233638940 0.8507609238999883
1474233638942 0.7270452028908565
1474233638944 0.1884892122406039
1474233638945 0.16815491247949899
1474233638947 0.22058734304156502
```

First number is `Date.now()` (there's usually 1-3 milliseconds between launches) and second is one `Math.random()` evaluated in that Duktape heap. It's not at least obviously correlated even for very close millisecond counts. That doesn't mean there isn't a detectable correlation (most likely there is), but whether that matters for practical non-serious applications or not, is another thing.

I *did* notice that if I didn't do that heap initialization loop where PRNG data is consumed, the first few Math.random() values would come out as clearly correlated, so I added that loop.

svaarala:
For the built-in generator one shortcoming is that it can't rely on almost any viable seed data other than what comes out of `duk_config.h`. Without that limitation it's usually easy to get good enough seed data, e.g. on Linux one could just read 16 bytes off `/dev/urandom` and that'd be good enough for most applications.

Note that without some improved seed data source, two Duktape heaps launched on the exactly same millisecond will have the same PRNG stream, no matter what seed mixing strategy is used.

One alternative would be to require an application to provide seed material one way or another. For example, providing it via a `duk_config.h` macro or adding an explicit API to (re)seed the PRNG.

My gut feeling, though, is that this is an overkill for the default generator. In other words, when this matters an application is quite likely to be using a replacement anyway. On the other hand, when using Duktape as a library that's not possible so in that case a (re)seed API might be workable.

svaarala:
I'll leave this unmerged for now, in case a simple practical solution is found :)

fatcerberus:
> user code should always replace the built-in PRNG or use specific high quality PRNG bindings for statistics or cryptography applications

I agree 100%, however do note that what people _should_ do vs. what they _actually_ do in practice are two very different things. ;)

Ultimately I agree that the current seeding method is probably fine.  xoroshiro128+ will never be suitable for cryptography regardless of the seeding method, and the period is too short for anything too heavy duty anyway, so I guess leave it as-is and we can revisit it later if it becomes a concrete issue. :)

svaarala:
Also as noted above, any kind of seed mixing or preprocessing is not enough as such: for two Duktape heaps created at exactly the same millisecond the PRNG sequences will be identical.

To actually solve the problem better seed input is needed in one way or another. Previous code XORed in the `duk_heap` pointer (for the Duktape heap being allocated) but that may even be static in some environments with an allocation pool and no address space randomization. So I don't see a good solution other than providing more (platform dependent) seed input.

svaarala:
Well, another sort-of solution would be to mix in more data into the seed at runtime. As a trivial example, XOR in the current timestamp every time a function is entered which presumably varies so that any correlated PRNG streams would diverge. Or maybe XOR in the string hash of any source string compiled (but this might be the same for every run). No easy solutions :-/

fatcerberus:
One option might be to initialize `s[1]` using the heap pointer and leave `s[0]` based on `Date.now()` like it is now.  This would at least ensure that different heaps within the same process would get a different seed, even if they are created at the same time.

fatcerberus:
To clarify, my main concern with the current implementation is that half of the state is initialized using a constant.  Essentially seeding 128 bits of state using only 64 bits of "entropy" (quoted because `Date.now()` is hardly entropic :)

Of course, that issue isn't solved at all by seed mixing, regardless of the algorithm used.  So if there could be found a suitable variable to use to initialize `s[1]`, that would be enough for me.

svaarala:
Note however that seed entropy should ideally not affect the PRNG *sequence* quality. If the generator is good, creating sequences from seeds S, S+1, S+2 should generate good individual sequences - although those sequences themselves may correlate. Within a *single* sequence a seed chosen from a timestamp shouldn't create a bad sequence unless the PRNG has good and bad seeds (all zeroes is a bad seed for example) to be avoided.

So unless you're working with cryptography, having a predictable seed is not really an issue. An attacker can guess the seed and regenerate the sequence, but so what if that's not a relevant attack scenario?

svaarala:
Also note that there is far, far less entropy than 64 bits when initializing a 64-bit value from `Date.now()`. For an outsider which can guess the time to within 1 second there are 1000 possible values and log2(1000) ~= 10 bits of entropy in the initializer.

But having good seed entropy is not really the point for a non-cryptographic PRNG. The sequence just needs to be high quality and unlikely to be reused/correlated where it matters.

svaarala:
As a toy example, suppose there were a global (as in truly global, for all hosts on Earth) counter to initialize successive seed sequences. This would be a good solution for non-cryptographic applications if the sequences didn't correlate (or that correlation didn't matter): every host would get a separate sequence without seeds being reused anywhere.

It would be totally insecure for security applications because anyone could guess your seed to within a few bits of entropy. For non-cryptographic applications that doesn't usually matter. For cryptographic work one should always use a cryptographic PRNG anyway: a statistically good PRNG with a perfectly unguessable seed is very often highly insecure for cryptography.

The above is not correct if there are "bad seeds" which create deformed sequences. However the only "bad seed" I can see documented for xoroshiro128+ is zero:

> The state must be seeded so that it is not everywhere zero.

So presumable other seeds generate acceptable sequences, at least after a few initial values. If not, it'd be important to know the criteria for a good seed (cf. "weak keys" in cryptography), they're not in the source file at least.

svaarala:
Ideally I'd prefer using a cryptographic PRNG for the default PRNG and allow user code to control the seed. That way the built-in random provider would be suitable for cryptographic applications too as long as the user provided a good seed. But cryptographic PRNGs are much, much slower which people seem to mind. They also have much larger footprint. And if I wrote a cryptographic application, I'd be very hesitant to rely on an in-engine PRNG for cryptography: maybe its algorithm changes in some upgrade silently for example.

The next logical alternative (to me) is for the built-in generator to be very small, reasonably fast, and provide a reasonably good sequence; as in not generating the same sequence always or otherwise surprising the user who doesn't care enough to use a specific provider and seeding mechanism.

The main question from that perspective is: does it surprise the user if two PRNG output streams from different Duktape heaps correlate? I'm not sure what the answer is, but I find it hard to imagine an application that doesn't really care about PRNG quality (at least enough to use a specific provider with known seed input) but would still manage to stumble on this. Real world often surprises me, so maybe there's a case where this happens :)

svaarala:
Coming back to concrete improvement possibilities:

- Use Shamir's three-op algorithm to mix the seed. Shamir's three-op algorithm would be simple to enable for this purpose with no additional new code, but I'd need to check if it actually fixes any correlation issues.
- Use SplitMix64 for mixing the seed which is the recommendation. SplitMix64 is quite simple so maybe this is the best option. It won't impact low memory targets anyway because they use Shamir's three-op algorithm.
- Adding in the heap pointer would eliminate the chance of two Duktape heaps created on the same millisecond generating identical sequences on the same host. It does introduce a compilation warning for some platforms though :/
- Maybe adding a "add this to the PRNG seed" API would be useful (similar to `srand()` which is known practice). That would also need to be plumbed into an external replacement PRNG so that those seeding API calls were not lost when using an external provider.
- Adding a platform provider into `duk_config.h` for proper seed entropy would work, but would be an extra step in porting. If handled poorly, the PRNG seeding would then be bad, which would be a silent surprise for applications on that platform.

svaarala:
So if I get over having to implement PRNG to seed the new PRNG ;) maybe the simplest practical fix is: add SplitMix64 and include heap pointer into the seed state. This follows xoroshiro128+ recommendation and also eliminates identical PRNG sequences within the same host (even if created in the same millisecond). (Grumble... ;)

svaarala:
ShiftMix64 might also be good enough alone as a built-in PRNG by the way. However it's slower than xoroshiro128+ and Shamir's three-op algorithm so maybe I'll keep xoroshiro128+ as the main algorithm.

svaarala:
Ok, let's see what I broke :)

fatcerberus:
SplitMix64 has a period of only 2^64 (since the state is 64 bits), which as the page notes, is probably not suitable for many applications:

> For 64 bits, we suggest to use a SplitMix64 generator, but 64 bits of state are not enough for any serious purpose. Nonetheless, SplitMix64 is very useful to initialize the state of other generators starting from a 64-bit seed, as research has shown that initialization must be performed with a generator radically different in nature from the one initialized to avoid correlation on similar seeds.

Regarding `Date.now()` not having 64 bits of entropy: Well, that's why I quoted "entropy" and stated the caveat that it's not really entropic at all. :)  Also, using a CSPRNG as the default provider is probably a bad idea - these tend to be too slow for a lot of application use cases (games, e.g.).

By the way the recommendation for the xoroshiro128+ seed to be "not everywhere zero" is confusing to me.  I never really figured out what it means.  If it's just that the seed can't be all zero bits (i.e. `s == 0`), then that should be stated directly.  I always interpreted it to mean that the initial seed should have, if possible, a roughly 50/50 distribution of 0 and 1 bits.  Maybe I'm way off base with that, but in any case it's not well documented. :(

As for applications not caring about quality of non-CS random numbers, well, let me just point this out (this article is also how I ended up discovering xoroshiro128+ in the first place):

* http://v8project.blogspot.com/2015/12/theres-mathrandom-and-then-theres.html

Sometimes they care very much, even outside of cryptographic applications. :)

fatcerberus:
Regarding correlation between runs: Sometimes this is indeed significant.  In a game, for example, if the sequences between two runs correlate, this can make save scumming (abuse of the save/load system) a lot easier.

fatcerberus:
One benefit to using SplitMix64 instead of direct seeding is that the SplitMix64 generator can be seeded with any value.  Your input seed can be zero and you still get a usable "not everywhere zero" value out.

fatcerberus:
@svaarala What is the period of Shamir's algorithm, by the way?

fatcerberus:
Anyway sorry for making this more difficult. :)

svaarala:
> SplitMix64 has a period of only 2^64 ...

Ahem, ever tried to run a 2^64 counter in a busy loop? It's a pretty long sequence. I'll wait while you do it ;-D

> Sometimes they care very much, even outside of cryptographic applications. :)

Sure, like I said, cryptography *and* statistics applications. Both of those should definitely use a high quality provider. But lacking proper seed material, the built-in generator cannot currently serve as a high quality provider regardless of the algorithm. It could only be fixed by improving the seeding process - but that has portability impact so it's maybe just easier to point out that "if you care, use an external provider".

There's of course the "doesn't pass visual inspection" test which matters for even casual applications, and this is really what was the issue with V8 as far as I can tell. But that's not really the issue here, because both Shamir's three-op algorithm and xoroshiro128+ (even with a low quality seed) pass that simple test.

> Well, you did break the AppVeyor build :)

I'll fix that :)

svaarala:
> Ahem, ever tried to run a 2^64 counter in a busy loop? It's a pretty long sequence. I'll wait while you do it ;-D

So to clarify, the actual issue with a "short" sequence like 2^64 is that the seed just chooses where you start in the sequence, and the sequence can be relatively easily reused. This makes it not very suitable for serious applications - but again, for randomizing game elements, timer back-offs, and the like, it's not really an issue. (Again, the V8 link you pointed to is a whole different issue because the randomization didn't even pass visual inspection, which is a totally different issue.)

svaarala:
Alright, so final footprint impact over master (with SplitMix64 + xoroshiro128+) is +240 bytes for x64.

svaarala:
> Anyway sorry for making this more difficult. :)

Not at all, seed material was/is indeed an issue. And it's definitely worth figuring out what makes sense for the built-in provider. It'd be much easier if there was high quality seed material available because then the goal of being "statistically good" out-of-the-box would be feasible. It'd be a clear and achievable goal, and intuitive for applications. But now that's not really possible due to lack of proper seed material, so the question becomes: what's a good goal for the built-in generator?

fatcerberus:
Did you see my note above on correlation between independent sessions?  That sometimes matters in game development scenarios even if the sequence is otherwise of high-quality.

svaarala:
> Did you see my note above on correlation between independent sessions? That sometimes matters in game development scenarios even if the sequence is otherwise of high-quality.

I did. But is that a concrete issue in this particular case? Does xoroshiro128+ with nearby seeds correlate enough to matter? It's rather difficult to assess (or fix) without any more detail than that.

Moreover, note that there are a lot of theoretical attacks. For example, non-cryptographic PRNGs can be generally reverse engineered from their output. So unless you're careful it may possible for a player to deduce the PRNG state and take advantage of it. If this is a concrete issue, then the only real solution is to use a cryptographic PRNG with a high quality secure seed.

With *that* as the measuring stick, Duktape should provide a cryptographic PRNG because anything else is potentially exploitable one way or another. But "playing it safe" also means a higher footprint, much worse performance, and platform dependencies to come up with good seed material. And even so it wouldn't necessarily fit all applications (for example, some environments may mandate a specific algorithm).

One important thing to note is that there's no single "high quality" generator; it may mean high performance, high statistical quality, cryptographic quality, good portability, etc. Because a single algorithm cannot address all those simultaneously (they're to some extent mutually exclusive), it makes sense to me to:
- Include a simple built-in algorithm which doesn't try to provide too many guarantees of quality.
- Allow the generator to be easily swapped out with a high quality generator for the task at hand. Depending on what you're doing the ideal generator may vary.

Just to be clear: I don't dispute at all that high quality randomness is needed in some situations. But this pull is about the *default* generator which doesn't (at least currently) have access to high quality seed material, and can be swapped out to use a custom generator. A built-in generator clearly cannot address all "high quality" concerns at once because some of the concerns conflict.

svaarala:
Translating the specific PRNG concerns of games, embedded targets, cryptographic applications, etc, into concrete requirements is not necessarily very easy.

There would be two easy-to-understand guarantees IMO:
- High performance statistically good PRNG with good seed material. Not suitable for cryptography but suitable for statistical applications. Also not suitable for advanced players who will figure out your PRNG internal state and exploit it by predicting the sequence ;-)
- Cryptographic PRNG with good seed material. Probably needs an interface to provide secure material from an external source.

The built-in generator unfortunately cannot achieve either one now due to lack of good seed material. So failing those two, what's a clear and easy-to-understand goal for the PRNG guarantees? The current answer, in this pull, is essentially:
- High performance PRNG, seeded from `Date.now()`. Individual sequences are statistically good quality, but sequences may correlate due to low quality seed. Use with caution, replace if PRNG really matters.

The "good statistical PRNG" goal would be achievable by improving the seeding. Doing so has API and portability impacts. Also user code will need to provide the seed somehow (otherwise the guarantee is not achieved), and it's unlike all applications would actually provide a good seed. So overall this seems like an overkill, especially because the built-in generator can be replaced entirely with a few lines of `duk_config.h` defines.

But I'm open to improving the seeding or somehow improving the guarantees (or their understandability). It's just important to avoid over-engineering the default generator because it will need to be replaced in many environments regardless of how it's implemented.

svaarala:
Re: this pull, I'd be ready to merge it. It should improve the state of the PRNG in master (the low quality seed issues were already present there) and if there's a good concrete idea for next steps they can be handled in separate pulls.

svaarala:
Added #971 to discuss next steps.

fatcerberus:
Sure, I'm okay with the current state of this pull, I'd definitely consider it an improvement over master.

By the way if my players want to run the math and figure out my PRNG's sequence then more power to them, they deserve every prize they get (outside of multiplayer at least - For that the developer should probably use a CSPRNG). :)  I was more concerned with _legitimate_ play causing a sequence to repeat between sessions; however based on my experience with xoroshiro128+ this doesn't actually happen in practice.

svaarala:
>  I was more concerned with legitimate play causing a sequence to repeat between sessions; however based on my experience with xoroshiro128+ this doesn't actually happen in practice.

Right, even if the seed has 1 bit of difference the sequences are still (at least to plain inspection) quite different.

svaarala:
> By the way if my players want to run the math and figure out my PRNG's sequence then more power to them, they deserve every prize they get (outside of multiplayer at least - For that the developer should probably use a CSPRNG).

Yeah, I used that example just to illustrate there are a lot of different threats and that every application writer views them differently. So it's difficult to answer them all without actually playing it safe. That said there are two clean guarantees (discussed above): good statistical PRNG, good cryptographic PRNG.

fatcerberus:
I want to note that it's exactly because of the compact state space of xoroshiro128+ that I made the switch in minisphere.  I originally used Mersenne Twister, but felt that its large state space was unwieldy.  Only having 128 bits of state is nice because it can be serialized as a small 32-digit hex string, while still having a large enough period even for very long-running games such as RPGs that can span many, many sessions.

svaarala:
> Only having 128 bits of state is nice because it can be serialized as a small 32-digit hex string, while still having a large enough period even for very long-running games such as RPGs that can span many, many sessions.

Being able to serialize state conveniently is definitely handy.

Regarding the period, a 2^64 period means that if you generate a billion random numbers per second you'll repeat the sequence in around 580 years. So the sequence length by itself is not really much of a problem. The real problem (which you probably also refer to) is that since it's just a single sequence and you just vary the starting point (via the seed), the sequence is easily reused when you start over from a new seed. With a larger state space it's easier to guarantee (at least probabilistically) that you won't reuse PRNG streams. A 2^128 state is much easier to work with in this respect.

fatcerberus:
> Regarding the period, a 2^64 period means that if you generate a billion random numbers per second you'll repeat the sequence in around 580 years.

...wow.  I knew 2^64 was large, but didn't realize it was _that_ much.  Thanks for running the math on that, I remembered reading on MSDN back when I used to do win32 coding that`QueryPerformanceCounter()` which runs at around 3MHz and uses a 64-bit counter would roll over after just 55 days or so.  Someone done goofed. :)

svaarala:
Maybe it was GetTickCount()? http://stackoverflow.com/questions/727918/what-happens-when-gettickcount-wraps, that seems to wrap around 49.7 days. Although it seems QueryPerformanceCounter() also had some wrapping issues (not 64-bit related though). Interesting, though a bit of a side track :-D

PR #974, opened 2016-09-21
Minor optimizations to RegExp compilation and execution

svaarala:
Some minor improvements to RegExp compilation and execution. For example, take advantage of the fact that regexp opcodes are small (well within 7 bits) so they can be emitted as is without UTF-8 encoding.


svaarala:
```
test-regexp-compile.js              : duk.O2  2.66 duk.O2.master  2.82 duk.O2.150  2.82
test-regexp-execute.js              : duk.O2  1.62 duk.O2.master  1.75 duk.O2.150  1.96
test-regexp-string-parse.js         : duk.O2  6.42 duk.O2.master  9.89 duk.O2.150 10.33
```

PR #975, opened 2016-09-21
Implement WHATWG Encoding API support

### LINKED ISSUE #961: Implement TextDecoder/Encoder objects
This allows writing and reading strings to/from ArrayBuffers and is implemented by most browsers:
https://encoding.spec.whatwg.org/

Currently manipulating strings embedded in a buffer requires using the Node.js Buffer object or `String.fromBuffer()` (Duktape 2.0+), both of which are nonstandard and not likely to exist outside of Duktape (and Node.js of course).  TextDecoder seems to be the standard way to do it with ArrayBuffers, since `ArrayBuffer#toString()` just gives "[object ArrayBuffer]".

It would be nice to have this standard implemented in Duktape, even if it only provides UTF-8 support without all the other encodings.

fatcerberus:
Implements #961.
 
Add `TextEncoder` and `TextDecoder` built-ins as described by the WHATWG Encoding specification.  For this pull, only UTF-8/CESU-8 support is implemented.  UTF-16 and all legacy encodings are left as future work.

Specification:
https://encoding.spec.whatwg.org/#api

Checklist:
- [x] Implement the WHATWG Encoding API: `TextEncoder` and `TextDecoder`
- [x] Config option - `DUK_USE_ENCODING_BUILTINS`
- [x] Support Encoding API with bufferobjects disabled?  Bindings could then work with plain buffers instead of Uint8Arrays (out of spec).
- [x] Node.js Buffer `.toString()` can share code with `TextDecoder`? => Save for later
- [x] Test cases for both encode and decode
- [x] Website note (Encoding API listed in features)
- [x] RELEASES entry

### DIFF
--- a/src-input/duk_bi_encoding.c
+++ b/src-input/duk_bi_encoding.c
@@ -1,0 +1,372 @@
+/*
+ *  WHATWG Encoding API built-ins
+ *
+ *  API specification: https://encoding.spec.whatwg.org/#api
+ *  Web IDL: https://www.w3.org/TR/WebIDL/
+ */
+
+#include "duk_internal.h"
+
+#if defined(DUK_USE_ENCODING_BUILTINS)
+typedef struct {
+	duk_uint8_t *out;      /* where to write next byte(s) */
+	duk_codepoint_t lead;  /* lead surrogate */
+} duk__encode_context;
+
+typedef struct {
+	duk_bool_t bom_seen;
+	duk_bool_t fatal;
+	duk_bool_t ignore_bom;
+
+	/* UTF-8 decoding state */
+	duk_codepoint_t codepoint;  /* built up incrementally */
+	duk_int_t needed;           /* how many more bytes we need */
+	duk_uint8_t upper;          /* max value of next byte (decode error otherwise) */
+	duk_uint8_t lower;          /* min value of next byte (ditto) */
+} duk__decode_context;
+
+#define DUK__U8_CONTINUE   0  /* continue to next byte */
+#define DUK__U8_CODEPOINT  1  /* have codepoint */
+#define DUK__U8_ERROR      2  /* decoding error */
+#define DUK__U8_RETRY      3  /* retry last byte, constraint: RETRY > ERROR */
+
+DUK_LOCAL void duk__utf8_decode_init(duk__decode_context *dec_ctx) {
+	dec_ctx->codepoint = 0x0000L;
+	dec_ctx->needed = 0;
+	dec_ctx->upper = 0xbf;
+	dec_ctx->lower = 0x80;
+	dec_ctx->bom_seen = 0;
+}
+
+DUK_LOCAL int duk__utf8_decode_next(duk__decode_context *dec_ctx, duk_uint8_t byte, duk_codepoint_t *cp) {
+	/*
+	 *  UTF-8 algorithm based on the Encoding specification:
+	 *  https://encoding.spec.whatwg.org/#utf-8-decoder
+	 */
+
+	if (dec_ctx->needed == 0) {
+		/* process initial byte */
+		if (byte <= 0x7f) {
+			/* U+0000-U+007F, 1 byte (ASCII) */
+			*cp = (duk_codepoint_t) byte;
+			return DUK__U8_CODEPOINT;
+		} else if (byte >= 0xc2 && byte <= 0xdf) {
+			/* U+0080-U+07FF, 2 bytes */
+			dec_ctx->needed = 1;
+			dec_ctx->codepoint = byte & 0x1f;
+			return DUK__U8_CONTINUE;
+		} else if (byte >= 0xe0 && byte <= 0xef) {
+			/* U+0800-U+FFFF, 3 bytes */
+			if (byte == 0xe0) {
+				dec_ctx->lower = 0xa0;
+			} else if (byte == 0xed) {
+				dec_ctx->upper = 0x9f;
+			}
+			dec_ctx->needed = 2;
+			dec_ctx->codepoint = byte & 0xf;
+			return DUK__U8_CONTINUE;
+		} else if (byte >= 0xf0 && byte <= 0xf4) {
+			/* U+010000-U+10FFFF, 4 bytes */
+			if (byte == 0xf0) {
+				dec_ctx->lower = 0x90;
+			} else if (byte == 0xf4) {
+				dec_ctx->upper = 0x8f;
+			}
+			dec_ctx->needed = 3;
+			dec_ctx->codepoint = byte & 0x7;
+			return DUK__U8_CONTINUE;
+		} else {
+			/* not a legal initial byte */
+			return DUK__U8_ERROR;
+		}
+	} else {
+		/* process continuation byte */
+		if (byte >= dec_ctx->lower && byte <= dec_ctx->upper) {
+			dec_ctx->lower = 0x80;
+			dec_ctx->upper = 0xbf;
+			dec_ctx->codepoint = (dec_ctx->codepoint << 6) | (byte & 0x3f);
+			if (--dec_ctx->needed > 0) {
+				/* need more bytes */
+				return DUK__U8_CONTINUE;
+			} else {
+				/* got a codepoint */
+				*cp = dec_ctx->codepoint;
+				dec_ctx->codepoint = 0x0000L;
+				dec_ctx->needed = 0;
+				return DUK__U8_CODEPOINT;
+			}
+		} else {
+			/* We just encountered an illegal UTF-8 continuation byte.  This might
+			 * be the initial byte of the next character; if we return a plain
+			 * error status and the decoder is in replacement mode, the character
+			 * will be masked.  We still need to alert the caller to the error
+			 * though.
+			 */
+			dec_ctx->codepoint = 0x0000L;
+			dec_ctx->needed = 0;
+			dec_ctx->lower = 0x80;
+			dec_ctx->upper = 0xbf;
+			return DUK__U8_RETRY;
+		}
+	}
+}
+
+DUK_LOCAL void duk__utf8_encode_char(void *udata, duk_codepoint_t codepoint) {
+	duk__encode_context *enc_ctx;
+
+	enc_ctx = (duk__encode_context *) udata;
+	if (codepoint > 0x10ffffL) {
+		/* cannot legally encode in UTF-8 */
+		enc_ctx->out += duk_unicode_encode_xutf8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, enc_ctx->out);
+	} else if (codepoint >= 0xd800L && codepoint <= 0xdbffL) {
+		/* high surrogate */
+		if (enc_ctx->lead != 0x0000L) {
+			/* consecutive high surrogates, consider first one unpaired */
+			enc_ctx->out += duk_unicode_encode_xutf8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, enc_ctx->out);
+		}
+		enc_ctx->lead = codepoint;
+	} else if (codepoint >= 0xdc00L && codepoint <= 0xdfffL) {
+		/* low surrogate */
+		if (enc_ctx->lead != 0x0000L) {
+			codepoint = 0x010000L | ((enc_ctx->lead - 0xd800L) << 10) | (codepoint - 0xdc00L);
+			enc_ctx->out += duk_unicode_encode_xutf8(codepoint, enc_ctx->out);
+			enc_ctx->lead = 0x0000L;
+		} else {
+			/* unpaired low surrogate */
+			enc_ctx->out += duk_unicode_encode_xutf8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, enc_ctx->out);
+		}
+	} else {
+		if (enc_ctx->lead != 0x0000L) {
+			/* unpaired high surrogate */
+			enc_ctx->out += duk_unicode_encode_xutf8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, enc_ctx->out);
+			enc_ctx->lead = 0x0000L;
+		}
+		enc_ctx->out += duk_unicode_encode_xutf8(codepoint, enc_ctx->out);
+	}
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textencoder_constructor(duk_context *ctx) {
+	/*
+	 *  TextEncoder currently requires no persistent state, so the constructor
+	 *  does nothing on purpose.
+	 */
+
+	if (!duk_is_constructor_call(ctx)) {
+		duk_error(ctx, DUK_ERR_TYPE_ERROR, "constructor must use 'new'");
+	}
+	return 0;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textencoder_prototype_encoding_getter(duk_context *ctx) {
+	duk_push_string(ctx, "utf-8");
+	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textencoder_prototype_encode(duk_context *ctx) {
+	duk__encode_context enc_ctx;
+	duk_size_t len;
+	duk_uint8_t *output;
+
+	if (!duk_is_undefined(ctx, 0)) {
+		duk_to_string(ctx, 0);
+	} else {
+		duk_push_string(ctx, "");
+		duk_replace(ctx, 0);
+	}
+
+	len = duk_get_length(ctx, 0);
+	if (len >= DUK_HBUFFER_MAX_BYTELEN / 3) {
+		duk_error(ctx, DUK_ERR_TYPE_ERROR, "input too large");
+	}
+
+	/* Allowance is 3*len because all bytes can potentially be replaced with
+	 * U+FFFD--which rather inconveniently encodes to 3 bytes in UTF-8.
+	 *
+	 * XXX: The buffer allocation strategy used here is rather inefficient.
+	 * Maybe switch to a chunk-based strategy, or preprocess the string to
+	 * figure out the space needed ahead of time?
+	 */
+	output = (duk_uint8_t *) duk_push_dynamic_buffer(ctx, 3 * len);
+
+	/* XXX: duk_decode_string() is used to process the input string.  For strings
+	 * originating in C, this means we will see the string the same way Ecmascript
+	 * code would.  But maybe we need something stricter in this case?
+	 */
+	enc_ctx.lead = 0x0000L;
+	enc_ctx.out = output;
+	duk_decode_string(ctx, 0, duk__utf8_encode_char, (void *) &enc_ctx);
+	if (enc_ctx.lead != 0x0000L) {
+		/* unpaired high surrogate at end of string */
+		enc_ctx.out += duk_unicode_encode_xutf8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, enc_ctx.out);
+		DUK_ASSERT(enc_ctx.out <= output + (3 * len));
+	}
+
+#if defined(DUK_USE_BUFFEROBJECT_SUPPORT)
+	duk_push_buffer_object(ctx, -1, 0, (duk_size_t) (enc_ctx.out - output), DUK_BUFOBJ_UINT8ARRAY);
+#else
+	/* shrink buffer to include only the encoded text */
+	duk_resize_buffer(ctx, -1, (duk_size_t) (enc_ctx.out - output));
+#endif
+	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_constructor(duk_context *ctx) {
+	duk__decode_context *dec_ctx;
+	duk_bool_t fatal = 0;
+	duk_bool_t ignore_bom = 0;
+
+	if (!duk_is_constructor_call(ctx)) {
+		duk_error(ctx, DUK_ERR_TYPE_ERROR, "constructor must use 'new'");
+	}
+	if (!duk_is_undefined(ctx, 0)) {
+		/* XXX: for now, 'label' argument is ignored */
+		duk_to_string(ctx, 0);
+	}
+	if (!duk_is_null_or_undefined(ctx, 1)) {
+		if (duk_get_prop_string(ctx, 1, "fatal")) {
+			fatal = duk_to_boolean(ctx, -1);
+		}
+		if (duk_get_prop_string(ctx, 1, "ignoreBOM")) {
+			ignore_bom = duk_to_boolean(ctx, -1);
+		}
+	}
+
+	dec_ctx = (duk__decode_context *) duk_push_fixed_buffer(ctx, sizeof(duk__decode_context));
+	dec_ctx->fatal = fatal;
+	dec_ctx->ignore_bom = ignore_bom;
+	duk__utf8_decode_init(dec_ctx);
+
+	duk_push_this(ctx);
+	duk_dup(ctx, -2);
+	duk_put_prop_string(ctx, -2, "\xff" "Context");
+	return 0;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_encoding_getter(duk_context *ctx) {
+	duk__decode_context *dec_ctx;
+
+	duk_push_this(ctx);
+	duk_get_prop_string(ctx, -1, "\xff" "Context");
+	dec_ctx = (duk__decode_context *) duk_require_buffer(ctx, -1, NULL);
+	DUK_ASSERT(dec_ctx != NULL);
+
+	/* XXX: change to look up from dec_ctx once other encodings are supported */
+	DUK_UNREF(dec_ctx);
+	duk_push_string(ctx, "utf-8");
+	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_fatal_getter(duk_context *ctx) {
+	duk__decode_context *dec_ctx;
+
+	duk_push_this(ctx);
+	duk_get_prop_string(ctx, -1, "\xff" "Context");
+	dec_ctx = (duk__decode_context *) duk_require_buffer(ctx, -1, NULL);
+	DUK_ASSERT(dec_ctx != NULL);
+
+	duk_push_boolean(ctx, dec_ctx->fatal);
+	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_ignorebom_getter(duk_context *ctx) {
+	duk__decode_context *dec_ctx;
+
+	duk_push_this(ctx);
+	duk_get_prop_string(ctx, -1, "\xff" "Context");
+	dec_ctx = (duk__decode_context *) duk_require_buffer(ctx, -1, NULL);
+
+	duk_push_boolean(ctx, dec_ctx->ignore_bom);
+	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_decode(duk_context *ctx) {
+	const duk_uint8_t *input = (const duk_uint8_t *) "";
+	duk__decode_context *dec_ctx;
+	duk_size_t len = 0;
+	duk_bool_t stream = 0;
+	duk_codepoint_t codepoint;
+	int ret;
+	duk_uint8_t *output;
+	const duk_uint8_t *in;
+	duk_uint8_t *out;
+
+	if (!duk_is_undefined(ctx, 0)) {
+		input = (const duk_uint8_t *) duk_require_buffer_data(ctx, 0, &len);
+	}
+	if (!duk_is_null_or_undefined(ctx, 1)) {
+		if (duk_get_prop_string(ctx, 1, "stream")) {
+			stream = duk_to_boolean(ctx, -1);
+		}
+	}
+	duk_push_this(ctx);
+	duk_get_prop_string(ctx, -1, "\xff" "Context");
+	dec_ctx = (duk__decode_context *) duk_require_buffer(ctx, -1, NULL);
+	DUK_ASSERT(dec_ctx != NULL);
+
+	if (len >= (DUK_HBUFFER_MAX_BYTELEN / 3) - 3) {
+		duk_error(ctx, DUK_ERR_TYPE_ERROR, "input too large");
+	}
+
+	/* Allowance is 3*len in the general case because all bytes may potentially
+	 * become U+FFFD.  If the first byte completes a non-BMP codepoint it will
+	 * decode to a CESU-8 surrogate pair (6 bytes) so we allow 3 extra bytes to
+	 * compensate: (1*3)+3 = 6.  Non-BMP codepoints are safe otherwise because
+	 * the 4->6 expansion is well under the 3x allowance.
+	 *
+	 * XXX: As with TextEncoder, need a better buffer allocation strategy here.
+	 */
+	output = (duk_uint8_t *) duk_push_fixed_buffer(ctx, 3 + (3 * len));
+
+	in = input;
+	out = output;
+	DUK_ASSERT(DUK__U8_RETRY > DUK__U8_ERROR);
+	while (in < input + len) {
+		ret = duk__utf8_decode_next(dec_ctx, *in++, &codepoint);
+		if (ret == DUK__U8_CONTINUE) {
+			continue;
+		}
+		if (ret >= DUK__U8_ERROR) {
+			/* decoding error */
+			DUK_ASSERT(ret == DUK__U8_ERROR || ret == DUK__U8_RETRY);
+			if (dec_ctx->fatal) {
+				/* fatal mode: throw a TypeError */
+				goto fatal;
+			} else {
+				/* replacement mode: replace with U+FFFD */
+				codepoint = DUK_UNICODE_CP_REPLACEMENT_CHARACTER;
+			}
+		}
+		if (ret == DUK__U8_RETRY) {
+			--in;  /* retry last byte */
+		}
+		if (!dec_ctx->bom_seen) {
+			dec_ctx->bom_seen = 1;
+			if (codepoint == 0xfeffL && !dec_ctx->ignore_bom) {
+				continue;
+			}
+		}
+		out += duk_unicode_encode_cesu8(codepoint, out);
+		DUK_ASSERT(out <= output + (3 + (3 * len)));
+	}
+
+	if (!stream) {
+		if (dec_ctx->needed != 0) {
+			/* truncated sequence at end of buffer */
+			if (dec_ctx->fatal) {
+				goto fatal;
+			} else {
+				out += duk_unicode_encode_cesu8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, out);
+				DUK_ASSERT(out <= output + (3 + (3 * len)));
+			}
+		}
+		duk__utf8_decode_init(dec_ctx);
+	}
+
+	duk_push_lstring(ctx, (const char *) output, (duk_size_t) (out - output));
+	return 1;
+
+fatal:
+	duk_error(ctx, DUK_ERR_TYPE_ERROR, "cannot decode as utf-8");
+	DUK_UNREACHABLE();
+}
+#endif  /* DUK_USE_ENCODING_BUILTINS */
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1818,6 +1818,10 @@ Planned
   e.g. "2 ** 10" evaluates to 1024, avoiding the cost of an Ecmascript call to
   Math.pow() while also being more readable (GH-987, GH-997)
 
+* Add TextEncoder and TextDecoder built-ins (the Encoding API) which allow
+  Ecmascript code to read and write text stored in an ArrayBuffer or Duktape
+  buffer (GH-975)
+
 * Remove no longer needed platform wrappers in duk_config.h: DUK_ABORT(),
   DUK_EXIT(), DUK_PRINTF(), DUK_FPRINTF(), DUK_FOPEN(), DUK_FCLOSE(),
   DUK_FREAD(), DUK_FWRITE(), DUK_FSEEK(), DUK_FTELL(), DUK_FFLUSH(),
--- a/config/config-options/DUK_USE_ENCODING_BUILTINS.yaml
+++ b/config/config-options/DUK_USE_ENCODING_BUILTINS.yaml
@@ -1,0 +1,13 @@
+define: DUK_USE_ENCODING_BUILTINS
+introduced: 2.0.0
+default: true
+tags:
+  - encoding-api
+description: >
+  Provide TextEncoder and TextDecoder built-ins (the Encoding API) which allow
+  Ecmascript code to encode and decode text stored in a buffer.  Only UTF-8 is
+  currently supported for decoding.
+
+  DUK_USE_BUFFEROBJECT_SUPPORT is recommended but not required: If it is
+  disabled, TextEncoder will encode to a plain buffer instead of a Uint8Array.
+  TextDecoder always accept plain buffers as input.
--- a/src-input/builtins.yaml
+++ b/src-input/builtins.yaml
@@ -359,6 +359,20 @@ objects:
         es6: true
         present_if: DUK_USE_BUFFEROBJECT_SUPPORT
 
+      # Encoding API
+      - key: "TextEncoder"
+        value:
+          type: object
+          id: bi_textencoder_constructor
+        encoding_api: true
+        present_if: DUK_USE_ENCODING_BUILTINS
+      - key: "TextDecoder"
+        value:
+          type: object
+          id: bi_textdecoder_constructor
+        encoding_api: true
+        present_if: DUK_USE_ENCODING_BUILTINS
+
       # Functions
       - key: "eval"
         value:
@@ -4159,3 +4173,139 @@ objects:
           length: 4
           varargs: false
         nodejs_buffer: true
+
+  #
+  # Encoding API
+  #
+
+  - id: bi_textencoder_constructor
+    class: Function
+    internal_prototype: bi_function_prototype
+    nargs: 0
+    native: duk_bi_textencoder_constructor
+    callable: true
+    constructable: true
+    bidx: true
+    encoding_api: true
+    present_if: DUK_USE_ENCODING_BUILTINS
+
+    properties:
+      - key: "length"
+        value: 0
+        attributes: ""
+        encoding_api: true
+      - key: "prototype"
+        value:
+          type: object
+          id: bi_textencoder_prototype
+        attributes: ""
+        encoding_api: true
+      - key: "name"
+        value: "TextEncoder"
+        attributes: ""
+        encoding_api: true
+
+  - id: bi_textencoder_prototype
+    internal_prototype: bi_object_prototype
+    class: Object
+    bidx: true
+    encoding_api: true
+    present_if: DUK_USE_ENCODING_BUILTINS
+
+    properties:
+      - key: "constructor"
+        value:
+          type: object
+          id: bi_textencoder_constructor
+        attributes: "wc"
+        encoding_api: true
+      - key: "encoding"
+        value:
+          type: accessor
+          getter: duk_bi_textencoder_prototype_encoding_getter
+          getter_nargs: 0
+          getter_magic: 0
+        attributes: "ec"
+        encoding_api: true
+      - key: "encode"
+        value:
+          type: function
+          native: duk_bi_textencoder_prototype_encode
+          length: 0
+          nargs: 1
+        attributes: "wec"
+        encoding_api: true
+
+  - id: bi_textdecoder_constructor
+    class: Function
+    internal_prototype: bi_function_prototype
+    nargs: 2
+    native: duk_bi_textdecoder_constructor
+    callable: true
+    constructable: true
+    bidx: true
+    encoding_api: true
+    present_if: DUK_USE_ENCODING_BUILTINS
+
+    properties:
+      - key: "length"
+        value: 0
+        attributes: ""
+        encoding_api: true
+      - key: "prototype"
+        value:
+          type: object
+          id: bi_textdecoder_prototype
+        attributes: ""
+        encoding_api: true
+      - key: "name"
+        value: "TextDecoder"
+        attributes: ""
+        encoding_api: true
+
+  - id: bi_textdecoder_prototype
+    internal_prototype: bi_object_prototype
+    class: Object
+    bidx: true
+    encoding_api: true
+    present_if: DUK_USE_ENCODING_BUILTINS
+
+    properties:
+      - key: "constructor"
+        value:
+          type: object
+          id: bi_textdecoder_constructor
+        attributes: "wc"
+        encoding_api: true
+      - key: "encoding"
+        value:
+          type: accessor
+          getter: duk_bi_textdecoder_prototype_encoding_getter
+          getter_nargs: 0
+          getter_magic: 0
+        attributes: "ec"
+        encoding_api: true
+      - key: "fatal"
+        value:
+          type: accessor
+          getter: duk_bi_textdecoder_prototype_fatal_getter
+          getter_nargs: 0
+          getter_magic: 0
+        attributes: "ec"
+        encoding_api: true
+      - key: "ignoreBOM"
+        value:
+          type: accessor
+          getter: duk_bi_textdecoder_prototype_ignorebom_getter
+          getter_nargs: 0
+          getter_magic: 0
+        attributes: "ec"
+        encoding_api: true
+      - key: "decode"
+        value:
+          type: function
+          native: duk_bi_textdecoder_prototype_decode
+          length: 0
+          nargs: 2
+        attributes: "wec"
+        encoding_api: true
--- a/tests/ecmascript/test-bi-textdecoder.js
+++ b/tests/ecmascript/test-bi-textdecoder.js
@@ -1,0 +1,177 @@
+/*
+ *  TextDecoder
+ */
+
+/*===
+basic decode
+utf-8
+false
+false
+2
+55357 56375
+true
+5
+65533 65 66 67 68
+true
+0
+true
+===*/
+
+function decodeTest() {
+    var decoder;
+    var ref, input, output;
+
+    decoder = new TextDecoder();  // defaults to UTF-8
+    print(decoder.encoding);
+    print(decoder.fatal);
+    print(decoder.ignoreBOM);
+
+    // non-BMP codepoint should become a surrogate pair
+    input = new Uint8Array([ 0xf0, 0x9f, 0x90, 0xb7 ]);
+    ref = "\ud83d\udc37";
+    output = decoder.decode(input);
+    print(output.length);  // should be 2
+    print(output.charCodeAt(0), output.charCodeAt(1));
+    print(output === ref);
+
+    // incomplete codepoint should not mask characters
+    input = new Uint8Array([ 0xf0, 0x41, 0x42, 0x43, 0x44 ]);
+    ref = "\ufffdABCD";
+    output = decoder.decode(input);
+    print(output.length);
+    print(output.charCodeAt(0), output.charCodeAt(1), output.charCodeAt(2), output.charCodeAt(3), output.charCodeAt(4));
+    print(output === ref);
+
+    // no argument is the same as an empty buffer
+    ref = "";
+    output = decoder.decode();
+    print(output.length);
+    print(output === ref);
+
+}
+
+try {
+    print("basic decode");
+    decodeTest();
+} catch (e) {
+    print(e.stack || e);
+}
+
+/*===
+fatal mode
+true
+false
+foo
+true
+TypeError
+===*/
+
+function fatalTest() {
+    var fatal, nonFatal;
+    var ref, input, output;
+
+    fatal = new TextDecoder('utf-8', { fatal: true });
+    print(fatal.fatal);
+    nonFatal = new TextDecoder('utf-8', { fatal: false });
+    print(nonFatal.fatal);
+
+    // well-formed input, should not throw
+    input = new Uint8Array([ 0x66, 0x6f, 0x6f ]);
+    ref = "foo";
+    output = fatal.decode(input);
+    print(output);
+
+    // buffer contains invalid UTF-8 code unit 0xFF
+    input = new Uint8Array([ 0x62, 0x61, 0x72, 0xff ]);
+    ref = "bar\ufffd";
+    print(nonFatal.decode(input) === ref);
+    try {
+        print(fatal.decode(input));
+    } catch (e) {
+        print(e.name);
+    }
+}
+
+try {
+    print("fatal mode");
+    fatalTest();
+} catch (e) {
+    print(e.stack || e);
+}
+
+/*===
+ignore BOM
+false
+true
+true
+true
+===*/
+
+function ignoreBOMTest() {
+    var stripBOM, keepBOM;
+    var ref, input, output;
+
+    var stripBOM = new TextDecoder('utf-8', { ignoreBOM: false });
+    print(stripBOM.ignoreBOM);
+    var keepBOM = new TextDecoder('utf-8', { ignoreBOM: true });
+    print(keepBOM.ignoreBOM);
+
+    input = new Uint8Array([ 0xef, 0xbb, 0xbf, 0x42, 0x4f, 0x4d ]);
+    ref = "\ufeffBOM";
+    output = keepBOM.decode(input);
+    print(output === ref);
+    ref = "BOM";
+    output = stripBOM.decode(input);
+    print(output === ref);
+}
+
+try {
+    print("ignore BOM");
+    ignoreBOMTest();
+} catch (e) {
+    print(e.stack || e);
+}
+
+/*===
+streaming
+true
+true
+===*/
+
+function streamTest()
+{
+    var decoder;
+    var ref, input, output;
+
+    // include a multi-byte character to test streaming correctness
+    decoder = new TextDecoder();
+    ref = "pig \ud83d\udc37";
+    input = new Uint8Array([ 0x70, 0x69, 0x67, 0x20, 0xf0, 0x9f, 0x90, 0xb7 ]);
+    output = "";
+    for (var i = 0; i < input.length; ++i) {
+        var byte = input.subarray(i, i + 1);
+        output += decoder.decode(byte, { stream: true });
+    }
+    output += decoder.decode();
+    print(output === ref);
+
+    // incomplete codepoint at end of string is caught in final decode().
+    // note that it's safe to reuse the decoder instance for another go as long
+    // as the last call to decode() was with {stream:false} (the default).
+    ref = "pig \ud83d\udc37\ufffd";
+    input = new Uint8Array([ 0x70, 0x69, 0x67, 0x20, 0xf0, 0x9f, 0x90, 0xb7, 0xf0 ]);
+    output = "";
+    for (var i = 0; i < input.length; ++i) {
+        var byte = input.subarray(i, i + 1);
+        output += decoder.decode(byte, { stream: true });
+    }
+    output += decoder.decode();
+    print(output === ref);
+}
+
+try {
+    print("streaming");
+    streamTest();
+} catch (e) {
+    print(e.stack || e);
+}
--- a/tests/ecmascript/test-bi-textencoder.js
+++ b/tests/ecmascript/test-bi-textencoder.js
@@ -1,0 +1,77 @@
+/*
+ *  TextEncoder
+ */
+
+// determine if two arrays contain the same values, used here to compare buffers
+function equiv(arr1, arr2) {
+    if (arr1.prototype !== arr2.prototype) {
+        return false;
+    }
+    if (arr1.length !== arr2.length) {
+        return false;
+    }
+    for (i = 0; i < arr1.length; ++i) {
+        if (arr1[i] !== arr2[i]) {
+            return false;
+        }
+    }
+    return true;
+}
+
+/*===
+basic encode
+utf-8
+true
+4
+true
+10
+true
+6
+true
+0
+true
+===*/
+
+function encodeTest() {
+    var encoder;
+    var ref, input, output;
+
+    encoder = new TextEncoder();
+    print(encoder.encoding);  // always "utf-8"
+
+    // non-BMP surrogate pair should become a single codepoint
+    input = "\ud83d\udc37";
+    ref = new Uint8Array([ 0xf0, 0x9f, 0x90, 0xb7 ]);
+    output = encoder.encode(input);
+    print(output instanceof Uint8Array);
+    print(output.length);  // should be 4
+    print(equiv(output, ref));
+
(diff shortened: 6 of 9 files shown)


svaarala:
Somewhat related: current Node.js Buffer binding does a 1:1 buffer-to-string coercion (this example assumes parts of #980 but the behavior is the same in 1.x and master):

```
duk> x = new Buffer('xfoo'); x[0] = 0xff; print(typeof String(x));
symbol
```

The resulting string is an internal string which cannot normally be created from Ecmascript code.

The actual Node.js binding replaces the 0xFF byte with U+FFFD:

```
> x = new Buffer('xfoo'); x[0] = 0xff; String(x).charCodeAt(0)
65533
```

Maybe the Buffer.toString() coercion could share a helper with this pull so that the Buffer binding would be in line with Node.js.

fatcerberus:
Hm, agree on that change of behavior.  Node.js `Buffer#toString()` ostensibly defaults to (canonical) UTF-8 so Duktape doing a 1:1 coercion is actually incompatible with Node.

svaarala:
Yes, it's certainly not correct now. It's an intentional placeholder because Duktape hasn't had any encoding/decodign support built in. But now with buffer objects optional and a scalable mechanism for config options in place, adding proper encoding/decoding support at least for UTF-8 would make a lot of sense.

fatcerberus:
Here's my just-implemented helper to decode Duktape strings:
https://github.com/fatcerberus/minisphere/blob/master/src/shared/lstring.c#L369-L452

...along with the comparatively much simpler UTF-8 codepoint encoder:
https://github.com/fatcerberus/minisphere/blob/master/src/shared/unicode.c#L44-L79

Just, nobody say the word "surrogate" to me... If they do, I might have to murder them. :)

fatcerberus:
One issue right now is that TextDecoder supports a "stream" parameter which allows decoding to be done in stages.  This means it's not enough for the helper to be a dumb function like mine is now - it needs to maintain state between invocations.

svaarala:
The most robust approach would be to implement a function that processes a single byte and updates state as it goes, so that all different stream chunk boundaries will be invisible to the decoding process. If that byte processing primitive is inside a loop handling N bytes (with a switch based on current state) it can be reasonable in practice and will still be oblivious to chunk boundaries because state is updated for every byte.

A more efficient approach is of course to handle full codepoints inline with no intermediate state updates, and just have the ability to resume them depending on the stream chunks. This is quite error prone to implement and needs a good regression test harness which exercises all the different cases.

fatcerberus:
That dovetails nicely with the WHATWG specification too, which describes the encoding and decoding algorithms in terms of byte streams.

This style of byte streaming is used elsewhere in Duktape too, e.g. the debug protocol, so it would slot in well I think.

fatcerberus:
So I hit an issue with Duktape allowing extended UTF-8 in strings pushed by C:
```
var str1 = "🐽 *munch*";
var buf = new TextEncoder().encode(str1);
var str2 = new TextDecoder().decode(buf);
```

After running this code, `str1 === str2` evaluates to `false` even though they represent the exact same text.  This turns out to be because TextDecoder encodes the pignose emoji atomically, while the original string used a surrogate pair for it, per Ecmascript specification.  As far as Duktape is concerned, they are not equivalent.

The only way for me to fix it that I can see is to encode the pignose using a surrogate pair, i.e. CESU-8.  But this is inconvenient since the rest of minisphere uses UTF-8.  Is there anything I can do to coax Duktape into "normalizing" the string?

svaarala:
Could you elaborate a bit? In Duktape strings are treated as byte sequences and their identity comes directly from that. So a CESU-8 surrogate pair and a direct UTF-8 encoding in string data is conceptually a different string.

It's also different for Ecmascript code, e.g. their `.length` will be different. So they're not equivalent in most senses.

fatcerberus:
Right, I understand why it happens, I'm just trying to figure out how to work around it.  The question is whether I can get Duktape to "normalize" the non-compliant string without directly encoding in CESU-8 myself.

svaarala:
Ultimately an application writer needs to decide whether:
- They want to use plain UTF-8 in C code - this means you'll be technically operating outside Ecmascript string space, but because Duktape allows that it works out fine in most cases.
- The want to use surrogate pairs i.e. CESU-8 in C code. This means dealing with surrogates, but the Ecmascript counterpart sees standard strings where surrogates are distinct codepoints.

In this case, wouldn't a TextDecoder() need to always emit CESU-8 for the internal representation, as far as codepoints [U+10000,U+10FFFF] are concerned, so that the result is a standard string? If it emitted UTF-8 instead, the resulting string would be non-standard because even Ecmascript code would see codepoints above BMP.

svaarala:
... That said, I'd find it quite annoying myself *not* being able to work in pure UTF-8, even if it's beyond Ecmascript standard. It's already allowed by e.g. String.fromCharCode() which accepts 32-bit values in Duktape to allow creation of pure UTF-8 strings. This is useful for C code convenience - and is ultimately up to the application writer to decide.

fatcerberus:
Yes, that's my goal - to return a standard string from `.decode()`.  Currently that means encoding in CESU-8 before pushing the string, which is inconvenient and also depends on me knowing Duktape's internal string representation when I should really be treating as a black box.

Anyway, to state it directly: What I'm looking for is some primitive or algorithm that will take my pushed UTF-8 string (which Duktape treats as valid but is out of spec) and normalize it to be Ecmascript-compliant.

fatcerberus:
To be clear I'm asking on behalf of application code, a TextEncoder/TextDecoder which is built into Duktape should of course just use the correct representation out of the gate :)

svaarala:
Ah OK, I thought `str1` was a surrogate pair encoded string because all UTF-8 codepoints are normalized to CESU-8 by the lexer automatically. But that was just a conceptual idea - `str1` comes in UTF-8 from C code.

There's nothing to normalize strings now, but adding API support to normalize a pushed UTF-8 string into CESU-8 would probably be useful. Having added, it'd be logical to request a reverse operation too.

svaarala:
Going on a bit of a tangent:

Just to be clear, all byte sequences are valid and supported strings as far as Duktape is concerned. This is convenient and important (at least to me) so that C code can freely alternate between buffers and strings. When you're working outside CESU-8 things mostly work fine. Some built-ins won't work well with strings outside of CESU-8 (or even outside of extended UTF-8) but the guarantee is that they won't cause memory unsafe behavior.

Many application writers prefer to work with UTF-8 throughout, and wouldn't want to convert to CESU-8 for scripting, even if it means they'll occasionally work with non-compliant strings. Some prefer to work with CESU-8 strings only so that their Ecmascript code remains fully compliant, even if it means translating between the two (for example a platform UI API wants UTF-8 but scripting uses CESU-8).

Because of this it's important that compliant strings work in a compliant fashion. But it's also important that non-standard strings work reasonably well. That's the reason why, for example, `String.fromCharCode()` is able to create non-CESU-8 strings:

```
duk> String.fromCharCode(0xdeadbeef).length
= 1
duk> String.fromCharCode(0xdeadbeef).charCodeAt(0)
= 3735928559
```

This is non-standard of course: the standard result is one codepoint U+BEEF (ToUint16 coercion). But it's convenient.

Now coming back to TextDecoder... The same users preferring UTF-8 throughout would probably prefer getting straight UTF-8 (rather than CESU-8) back from their TextDecoder. That would be non-compliant but that's not a problem because they've already chosen to work with straight UTF-8 for convenience.

svaarala:
Regarding API calls, maybe something like:

```c
duk_utf8_to_cesu8(ctx, -5);
duk_cesu8_to_utf8(ctx, -5);
```

Ideally I'd like to allow extended UTF-8 (beyond U+10FFFF) but that's not possible with surrogate pairs so the calls would deal with plain UTF-8 only.

svaarala:
For convenience it might be useful to be able to push a string and have it normalized at the same time (this would also reduce memory churn):

```c
duk_push_(l)string_normalize(ctx, "my utf-8 string");
```

The same isn't possible for reading back because `duk_get_string()` pointers point directly to the internal representation so it's not possible to return a C string pointer for a normalized-back-to-UTF8 string without actually modifying something on the value stack.

svaarala:
By the way, one reason I've tried to steer clear of adding codec, normalization, etc, support is that there are several use cases and places in the API where they might be useful, and it's not a trivial design issue to come up with a good and consistent minimal set for that. It's also a quite prone to feature creep (like codec registrations etc) so it's not a small piece of functionality in the end w.r.t. maintenance and code/API footprint. Making wrong decisions here is easy and reworking/obsoleting the wrong decisions takes a lot of cumulative development time.

But... This is no longer a workable position I think; with buffer support improving, at least UTF-8 support will be eventually needed for the Node.js Buffer interface. But could Duktape core functionality be restricted to pure UTF-8 and CESU-8? If so, it should provide good internal and external support for them, but not necessarily provide a generic codec interface (with extensible codec registration etc).

The concrete situation is anyway that Duktape internals already need to deal with extended UTF-8 and CESU-8 in many places. That's handled by a single format which covers all codepoints up to 32 bits (beyond strict UTF-8) and also allows arbitrary 16-bit codepoints (invalid surrogates included) with no special cases. The internals *don't* need to deal with strict non-extended UTF-8, surrogate pair conversions etc yet, though. But supporting these operations for UTF-8 and CESU-8 only would still be a reasonably compact target as opposed to generic codec support.

In any case it's good to step back and figure out what the different codec related use cases actually are. For example, should the C API for codecs be conceptually a mirror of TextEncoder/TextDecoder in the Ecmascript API? The fewer "patchwork" APIs, the fewer concepts, the easier the API is to remember and use.

svaarala:
> To be clear I'm asking on behalf of application code, a TextEncoder/TextDecoder which is built into Duktape should of course just use the correct representation out of the gate :)

Note however that there are many "correct" representations: for applications preferring to work with UTF-8 exclusively, it'd be very convenient if TextDecoder was able to produce UTF-8 (rather than CESU-8) strings. For example, the decoded string may be passed to C code which requires it to be UTF-8 so it's quite useful to be able to work with non-standard strings in Ecmascript code too.

From a pure compliance perspective there's of course a "correct" representation. But when an application chooses to work with an extended string concept for practical convenience, it's annoying if the scripting environment is "judgmental" and works against you (especially when pure UTF-8 strings otherwise work quite transparently) :-)

The concrete impact for TextEncoder/TextDecoder is that it may be useful to allow a user to request the encoder/decoder to work with (extended) UTF-8 rather than CESU-8 so that it plays nice with these extended strings. If that's easy to model in the Ecmascript API so that it's visible in the Ecmascript code call sites, that'd be nice. If not, a config option for compliant vs. practical behavior may be the way to go (and is the approach used for String.fromCharCode()'s 32-bit codepoint support).

fatcerberus:
I meant "correct" in the sense of "round tripping this string will compare equal at both ends".  Since the original string is a literal in the JS, the only way that can be the case is if the output is CESU-8 encoded.  No other representation will work. :)

svaarala:
> I meant "correct" in the sense of "round tripping this string will compare equal at both ends". Since the original string is a literal in the JS, the only way that can be the case is if the output is CESU-8 encoded. No other representation will work. :)

Right, that's the compliant case.

But if the input string is UTF-8 from C code, it should maybe round trip back to UTF-8 if the user wants it to. If there's a way to request that behavior in the encoder/decoder API, that would probably be useful.

fatcerberus:
GitHub really needs to move that Close & Comment button somewhere else. ;-)

svaarala:
I think I linked this earlier already but: http://wiki.duktape.org/HowtoNonBmpCharacters.html. I wrote this because there were questions about the possible approaches and it turns out some application writers prefer CESU-8 and compliance and others (very strongly) prefer working with UTF-8 because it's practical and Duktape supports it pretty well.

svaarala:
> GitHub really needs to move that Close & Comment button somewhere else. ;-)

Agree :+1: My laptop has a poor touchpad which leads to a lot of error clicks... :/

fatcerberus:
By the way it seems not worth the trouble for Duktape to go out of its way supporting the full 32-bit range for UTF-8 when according to ISO 10646 and Unicode themselves there will never, ever be a character assigned beyond U+10FFFF:

> Q: Will UTF-16 ever be extended to more than a million characters?
>
> A: No. Both Unicode and ISO 10646 have policies in place that formally limit future code assignment to the integer range that can be expressed with current UTF-16 (0 to 1,114,111). Even if other encoding forms (i.e. other UTFs) can represent larger intergers _[sic]_, these policies mean that all encoding forms will always represent the same set of characters. Over a million possible codes is far more than enough for the goal of Unicode of encoding characters, not glyphs. Unicode is not designed to encode arbitrary data. If you wanted, for example, to give each “instance of a character on paper throughout history” its own code, you might need trillions or quadrillions of such codes; noble as this effort might be, you would not use Unicode for such an encoding. [AF]
```

By extension this means the 1-4-byte forms are all that will ever be needed for UTF-8.

**Comment by svaarala** [comment]
Duktape uses the 32-bit codepoint space internally for regexp bytecode.

**Comment by svaarala** [comment]
But yes, I agree external APIs don't necessarily need to (and probably shouldn't) support extended UTF-8. Duktape does need to do that internally at least for RegExp bytecode where arbitrary jump offsets are necessary and those offsets need to go beyond Unicode range. That would be fixable by using some other encoding for RegExps: extended UTF-8 is just used as a variable integer encoding there to avoid introducing yet another format.

But other than that, Duktape internals wouldn't necessarily need to support codepoints beyond UTF-8 range. For example `String.fromCharCode()` could accept codepoints up to U+10FFFF (and even that's custom behavior) but above that a TypeError could be thrown. Unless the RegExp encoding is changed there's very little practical source code or footprint benefit because the encoding/decoding code will need to remain.

In general, when providing an UTF-8 API (and not some "relaxed form" like in Duktape internals) it should be pretty strict: UTF-8 mandates checking for invalid forms like surrogate pairs encoded in the CESU-8 style, non-minimum length encodings, etc. These should lead to a TypeError or replacement characters rather than be accepted as "close enough" which would be surprising behavior for an API explicitly described as supporting standard UTF-8.

**Comment by svaarala** [comment]
Somewhat related:

Apart from actual encoding/decoding scenarios user code is allowed to push arbitrary byte strings into the Duktape value stack and work with them. The current guarantee is that regardless of the byte sequence, operations won't croak but will do something consistent.

For example, codepoint lookups and substring operations will decode extended UTF-8 to figure out where the character boundaries are. When faced with invalid bytes, continuation bytes (0x80 to 0xbf) are generally ignored and other bytes are imagined to be leading bytes. This means that invalid bytes like 0xFF are considered individual characters and any continuation bytes after it are skipped when looking for the next leading byte.

I chose this approach because I like the guarantee that you can work with arbitrary byte sequences which is quite useful from a C code persective. While arbitrary byte strings don't need to work perfectly, they should work reasonably consistently. Throwing errors, for example, when doing a substring operation would be highly annoying.

Coming back to strict UTF-8 vs. extended UTF-8, this same approach could be used if extended UTF-8 interpretation was removed.

That would just mean that any extended UTF-8 sequences previously allowed would go through that "deal with garbage" path. For example, when scanning character boundaries, 0xFE would be considered a leading byte of an unknown byte sequence; it wouldn't be decoded as UTF-8 but would stand as is (for example a codepoint lookup would then return the integer 254 rather than decoding the sequence and returning the extended codepoint).

So all in all, we could prototype removing the extended UTF-8 support and see what the overall effect is. I'd need to rework the RegExp encoding to do that, but that's not difficult. Supporting only UTF-8 and CESU-8 would also remove the need to explain extended UTF-8 sequences to users. But unfortunately CESU-8 still needs to be mentioned which is often a source of confusion :/

**Comment by svaarala** [comment]
Ok, #983 is for prototyping/discussing that potential change.

**Comment by fatcerberus** [comment]
Agree, that's why a standard API like TextDecoder is good to have - it offers predictable behavior for strings regardless of how the engine treats text internally.  In most cases Duktape's internal string handling is (intentionally) too lenient for this purpose.

By the way, if this pull is merged, should `String.fromBuffer()` be removed?

**Comment by svaarala** [comment]
Re: removing `String.fromBuffer()`, it's still an open question, at least for me.

On the one hand, Duktape allows working with the internal string representation (if you choose to do so) so being able to create a string 1:1 from a buffer is a useful operation. With #979 it would also no longer pose sandboxing issues.

But on the other hand, maybe the Ecmascript environment shouldn't encourage working with the internal string representation. Removing the binding would make that clear, and you'd do it from custom C bindings instead.

**Comment by svaarala** [comment]
One practical reason I needed to keep String.fromBuffer() is that a lot of testcases rely on it to create internal strings. But an easy workaround for that is to add a few custom bindings to `duk` instead, so that if you run with say `duk --testing` it would register those additional bindings needed by the testcases.

**Comment by fatcerberus** [comment]
Progress update: I have written a decoder matching the requirements laid out in the standard.
* https://github.com/fatcerberus/minisphere/blob/master/src/shared/unicode.c#L100-L183

I'm not sure of the performance of this yet compared to Björn Höhrmann's lookup table based algorithm. which is what minisphere was using before.

**Comment by fatcerberus** [comment]
@svaarala Following up on our email conversation, w.r.t. `TextDecoder` in particular, there are a few complications:

* A compliant `TextDecoder` must support streaming, potentially a single byte at a time.  For multi-byte characters, buffering (and thus state tracking) is required, and you can't look ahead in the stream.
* If the string turns out to be not well-formed, replacement characters must be emitted.  Performing this validation requires some pre-decoding.
* Surrogate pairs: If you see a high surrogate, you MUST buffer it - and if the _next_ codepoint is not a low surrogate, you then need to emit two codepoints: U+FFFD followed by the codepoint you just decoded.
* Security: If you encounter an invalid continuation byte, you have to backtrack to avoid masking a character (see https://encoding.spec.whatwg.org/#security-background).

The last two points especially mean that it's not enough to have a helper that just returns a codepoint or status - you need some sort of actual streaming system in place.  That being said, it's certainly possible (if tedious) to do all this in a single pass, but I do wonder about the footprint and performance impact in any case.

**Comment by fatcerberus** [comment]
I guess what I'll do to start is work on `TextEncoder` first, as that always works with whole strings and requires no buffering.  Validation is still needed because C code may push malformed strings, but it should be overall easier to write a transcoder for compared to `TextDecoder`.

**Comment by svaarala** [comment]
Sounds like a good approach. For master it'd be easiest to first merge "as small as possible" algorithms and then look at performance variants later. But table driven first is also OK, whatever is nicest to implement first.

Doing a table driven direct transcoding (UTF8 <-> CESU8) would be interesting because many bytes can pass through as is (with validation for multibyte sequences). It makes most sense if (1) it's clearly faster and (2) it's unlikely new encodings will be plugged in later. If the second isn't true, a separate algorithm which stream decodes codepoints and encodes them whenever they pop up will be needed anyway. Also practical complications like "look behind" for surrogate pair validation might make this approach too cumbersome in practice.

So maybe it's easiest to use a table driven UTF-8 / CESU-8 decoder, and just encode completed codepoints as they pop up from the decoder. We can look at a table driven transcoder separately if it makes sense.

**Comment by fatcerberus** [comment]
Re: Other encodings - At some point these will need to be supported (only for decoding though) **IF** we aim for full compliance with the specification.  But UTF-8 only is a good first approximation, I think.

**Comment by fatcerberus** [comment]
As for CESU-8/UTF-8, I feel the approach that makes the most sense is to support both directly.  By that I mean you'd be able to construct a `TextDecoder` with a label of either "utf-8" or "cesu-8".  Same goes for `TextEncoder`.

I think it might also be feasible to use the same helper(s) for both encodings.  In the end they're not that different - they only differ in how they encode non-BMP characters.

**Comment by fatcerberus** [comment]
I think it might not be possible to use Bjoern's algorithm: When it rejects a byte, it doesn't differentiate between initial bytes and continuation bytes, whereas the WHATWG standard requires the decoder to backtrack *only* when encountering an invalid continuation byte to avoid masking the following character.  For obvious reasons you can't backtrack an on invalid IB (= infinite loop).

I will probably implement an algorithm similar to the one currently used in minisphere:
https://github.com/fatcerberus/minisphere/blob/v4.2.3/src/shared/unicode.c#L72-L168

**Comment by fatcerberus** [comment]
I kind of wish GitHub would order posts by newest first, it's annoying to have to scroll to the bottom all the time especially on mobile (no End key there). :(

Anyway, the build is horribly, horriby broken right now but I mocked up a quick draft of a UTF-8 decoder that meets the requirements of the specification, see https://github.com/svaarala/duktape/pull/975/commits/3b559194c82637ac35aea6d2f71a19d2ae49528b

**Comment by fatcerberus** [comment]
Ugh, I assume this is a PyYAML issue but the error messages for YAML files are useless.  I'm getting errors on builtins.yaml file after rebasing and I have no idea what I broke.  All I get out of the configure process is this:

```
KeyError: 'bi_textencoder_constructor'
```

No line number or anything, just a Python stack trace which doesn't help me in this situation.  I know the YAML worked before the rebase, so this makes no sense.

Here's the full stack trace anyway:
```
  File "D:\src\duktape\tools\genbuiltins.py", line 3058, in <module>
    main()
  File "D:\src\duktape\tools\genbuiltins.py", line 2922, in main
    ramobj_data_le = gen_ramobj_initdata_bitpacked(ram_meta, ram_native_funcs, ram_natfunc_name_to_natidx, 'little')
  File "D:\src\duktape\tools\genbuiltins.py", line 1952, in gen_ramobj_initdata_bitpacked
    count_obj_normal, count_obj_func = gen_ramobj_initdata_for_props(meta, be, o, string_index, natfunc_name_to_natidx, objid_to_idx, double_byte_order)
  File "D:\src\duktape\tools\genbuiltins.py", line 1722, in gen_ramobj_initdata_for_props
    _bidx(prop_constr['value']['id'])
  File "D:\src\duktape\tools\genbuiltins.py", line 1681, in _bidx
    be.bits(objid_to_bidx[bi_id], BIDX_BITS)
KeyError: 'bi_textencoder_constructor'
  File "..\tools\configure.py", line 966, in <module>
    main()
  File "..\tools\configure.py", line 663, in main
    exec_print_stdout(cmd)
  File "..\tools\configure.py", line 60, in exec_print_stdout
    ret = exec_get_stdout(cmd, input=input, print_stdout=True)
  File "..\tools\configure.py", line 51, in exec_get_stdout
    raise Exception('command failed, return code %d: %r' % (proc.returncode, cmd))
Exception: command failed, return code 1: ['C:\\Python27\\python.exe', 'D:\\src\\duktape\\tools\\genbuiltins.py', '--git-commit', '3b559194c82637ac35aea6d2f71a19d2ae49528b', '--git-branch', 'whatwg-encoding', '--git-describe', 'v1.5.0-716-g3b55919-dirty', '--duk-version', '19999', '--used-stridx-metadata', 'c:\\users\\fatce\\appdata\\local\\temp\\tmp-duk-prepare-nf9kuz\\duk_used_stridx_bidx_defs.json.tmp', '--strings-metadata', 'D:\\src\\duktape\\src-input\\strings.yaml', '--objects-metadata', 'D:\\src\\duktape\\src-input\\builtins.yaml', '--active-options', 'c:\\users\\fatce\\appdata\\local\\temp\\tmp-duk-prepare-nf9kuz\\duk_config_active_options.json', '--out-header', 'c:\\users\\fatce\\appdata\\local\\temp\\tmp-duk-prepare-nf9kuz\\src\\duk_builtins.h', '--out-source', 'c:\\users\\fatce\\appdata\\local\\temp\\tmp-duk-prepare-nf9kuz\\src\\duk_builtins.c', '--out-metadata-json', 'c:\\users\\fatce\\appdata\\local\\temp\\tmp-duk-prepare-nf9kuz\\genbuiltins_metadata.json', '--ram-support']
```

**Comment by fatcerberus** [comment]
So it turns out I was missing `bidx: true`.  Of course I had to study the existing metadata pretty closely to figure that out - no way I could have divined that from "KeyError: 'bi_textencoder_constructor'" :-/

**Comment by fatcerberus** [comment]
@svaarala In general, when a standard calls for an accessor property with only a getter (that always returns the same value), is it safe to substitute for a static property?  For example `TextEncoder.encoding` is always "utf-8".

**Comment by svaarala** [comment]
> So it turns out I was missing bidx: true. Of course I had to study the existing metadata pretty closely to figure that out - no way I could have divined that from "KeyError: 'bi_textencoder_constructor'" :-/

Sorry, it's unfortunately quite a lot of work for error messages to be very specific when talking about disposable tooling. Even more so if the error messages should tell exactly what's to be fixed :o

**Comment by svaarala** [comment]
> In general, when a standard calls for an accessor property with only a getter (that always returns the same value), is it safe to substitute for a static property? For example TextEncoder.encoding is always "utf-8".

There's no clear cut answer to that I'm afraid. It's safe in the sense that the majority of practical code won't care about the difference. But it's not safe in the sense that the difference is visible e.g. via `Object.getOwnPropertyDescriptor()` so it's easy to write could code that would act on the difference. It's obviously also not technically compliant and writing code to demonstrate that is easy. Even so it's quite typical for engines to use a getter when that makes internal sense or vice versa.

In this case it seems `.encoding` is inherited from TextDecoder prototype and is presumably a getter because the result is looked up from the specific instance (and might theoretically vary between instances) without burderning each instance with a concrete, guaranteed property?

**Comment by fatcerberus** [comment]
The spec is actually a bit vague in this instance: It describes `.encoding` as an "attribute" that is always "utf-8". It does later mention "getter", though.

If Duktape wasn't footprint sensitive there wouldn't even be a question - I'd just automatically implement it as an accessor.  I was just thinking of safe ways to reduce overhead. :)

In the end it's probably best to implement a getter regardless - if we ultimately support CESU-8 in addition to UTF-8 then the value of `.encoding` is no longer static.

**Comment by svaarala** [comment]
I appreciate the footprint awareness :-)

In general the inherited getter approach can improve footprint if data fields recorded for the instances can be reduced (and "internal") and be decoded by inherited accessors. For example, errors record a `_Tracedata` property and the inherited accessors like `.stack` and `.fileName` return their values based on that raw data. This effect is not very relevant if only a few instances are in play simultaneously (which might be the case for encoding instances).

For encoders/decoders it would, for example, be possible to record only a single integer per instance, containing:
- Encoding from an enumeration (lazy approach: internal stridx)
- Fatal flag
- IgnoreBOM flag

These could then be decoded by the accessors to provide actual string/boolean values. Each instance would record only a single integer property which only takes 13 bytes on 32-bit packed duk_tval targets.

But for the current discussion, a getter providing a fixed string would be around 20 bytes:

    duk_push_hstring_stridx(ctx, DUK_STRIDX_UTF8);
    return 1;

So that's not bad, for now.

**Comment by fatcerberus** [comment]
Proof of concept:
```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-719-g49d4ce8)
duk> var te = new TextEncoder();
= undefined
duk> var td = new TextDecoder('utf-8')
= undefined
duk> var buf = te.encode("pig\uD83D\uDC37")
= undefined
duk> buf
= |706967f09f90b7|
duk> buf.length
= 7
duk> td.decode(buf)
= "pig\ud83d\udc37"
duk> 
```

**Comment by fatcerberus** [comment]
Current implementation of `TextEncoder.encode()` uses `duk_decode_string()` to traverse the input string.  Having now looked at the code for that function, it is not adequate: C code may push ill-formed strings (in particular, internal strings/symbols) which will cause it to throw.  Instead replacement characters should be emitted, the same way it currently handles unpaired surrogates.

**Comment by fatcerberus** [comment]
@svaarala This branch is still very much a diamond in the rough, but it's a working implementation regardless.  If you get some free time and want to play around with it, feel free :-)

**Comment by fatcerberus** [comment]
@svaarala I looked through the Duktape repo as well as the comment block at the top of `builtins.yaml` but can't see any way to define an accessor property with only a getter.  If I don't include a `setter` field for an accessor in `builtins.yaml` then genbuiltins.py fails at build time.  Is this currently possible to do?

**Comment by svaarala** [comment]
It's not supported at the moment - it's primarily a limitation of the RAM built-in data which genbuiltins.py enforces. I can fix that in master prior to merging this if it makes things easier.

**Comment by fatcerberus** [comment]
I could include a dummy setter, but it's probably not footprint friendly to have functions that don't do anything. :)  Plus whether a non-writable property throws is dependent on caller strictness.  Strict mode = throw, not-strict mode = ignore.

The only other feasible option I can see would be to define the `readonly` attributes as plain properties (non-writable) in the constructor, but that doesn't seem very clean.

**Comment by svaarala** [comment]
An initial Encoding API pull could be "broken" in that way (having a dummy setter) and be fixed later on if that were easier. There would be a baseline to work against.

But I can add support for the missing getter/setter also in master, it should be very straightforward so I can probably get it done tonight so we can avoid an ugly intermediate solution?

**Comment by fatcerberus** [comment]
It's up to you, I don't want to put you out.  I can finish the pull either way :)

**Comment by fatcerberus** [comment]
Do I have to add the method and property names to strings.yaml?  It seems to work fine without them being present.

**Comment by svaarala** [comment]
That shouldn't be necessary. Strings.yaml contains the strings that need an internal `DUK_STRIDX_xxx` because code references them directly. Property names that don't need direct internal references don't need a string index.

**Comment by fatcerberus** [comment]
Alright, that's what I thought, thanks :)

**Comment by fatcerberus** [comment]
There seems to be some memory unsafe behavior leading to a segfault in `TextDecoder` when using streaming mode.  I'll run it through Valgrind tonight, no more time right now.

**Comment by svaarala** [comment]
Ok, optional setter/getter for built-ins is now in master, hopefully fixing the issue here :)

File: src-input/duk_bi_encoding.c [hunk 1]
```diff
@@ -0,0 +1,333 @@
...
+			return DUK__U8_RETRY;
+		}
+	}
+}
+
+DUK_LOCAL void duk__utf8_encode_char(void *udata, duk_codepoint_t codepoint) {
+	duk__encode_context *enc_ctx;
+
+	enc_ctx = (duk__encode_context *) udata;
+	if (codepoint > 0x10ffff) {
```
svaarala: This constant and any other signed constants not in 16-bit range should be `0x10fffffL`. It's a bit theoretical but currently Duktape does compile with a compiler using 16-bit ints in which case integer constants are also 16-bit unless they have `L` suffix.

fatcerberus: Good to know in the future, thanks.  I assumed the compiler would automatically promote constants that were too large for an int.

svaarala: Here's an example that I encountered a while back which still gives me the creeps:

```c
#include <stdio.h>
#include <stdint.h>
int main(int argc, char *argv[]) {
	(void) argc; (void) argv;
	int64_t test = -0x80000000;
	printf("%lld\n", (long long) test);
	return 0;
}
```

This prints:

```
2147483648
```

A positive rather than negative value. The situation remains the same using the constant `-0x80000000L` but is fixed by using `-0x80000000LL`.

The reason? As far as I can tell, the literal is parsed as an unary minus followed by a non-negative constant. `0x80000000` parses as an unsigned integer because it doesn't fit into a signed integer (but fits into 32 bits), so unary minus of that is still a positive (32-bit unsigned) value. *That* then gets coerced into signed 64 bits.

This was very much fun figuring out ;)

Anyway, here too the issue is avoided if the constant matches the width *and* sign of the intended result instead of relying on automatic conversion.

svaarala: So considering that example above with a compiler using 16-bit `int`s, a constant `-0x8000` would be interpreted as an unsigned int `0x8000` which would matter if that was then coerced to a 32 bit type like `duk_int_t`. Adding an `L` suffix would fix the issue.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,333 @@
...
+		if (duk_get_prop_string(ctx, 1, "stream")) {
+			stream = duk_to_boolean(ctx, -1);
+		}
+	}
+	duk_push_this(ctx);
+	duk_get_prop_string(ctx, -1, "\xff" "Context");
+	dec_ctx = (duk__decode_context *) duk_require_pointer(ctx, -1);
+
+	/* FIXME: 100% USDA Choice Grade A memory pork */
+	output = (duk_uint8_t *) duk_push_fixed_buffer(ctx, length * 3);
```
fatcerberus: I think this is the cause of the segfault.  For whole strings 3x is enough because the worst that will happen is all bytes will become U+FFFD (3 byte sequence).  In streaming mode however we're sending partial codepoints so that up to 6 bytes may be emitted in any given call.

This will need more intelligent buffer management, methinks.  Unconditionally allocating 6x the number of bytes being decoded would be a bit excessive.

svaarala: I agree - maybe the streamed input could be processed in chunks, e.g. max 256 bytes at a time so that the overallocation would be 6x256?

svaarala: For a `DUK_USE_PREFER_SIZE` that chunk size could maybe be a lot smaller, say 16 (for a spare allocation of 6x16 = 96 bytes which is no longer outrageous ;)

fatcerberus: Had some time to reflect on this and it should be good enough for now to keep the allocation as `3 * len` in the general case, but set a lower bound on the allocation of 6 bytes.  In the worst case you get 6 bytes of output for 1 byte of input, but that's a corner case--in general you can never get more than three times the number of bytes you put in because codepoints are emitted as soon as possible.  The only time it will expand by 6x is when the caller sends the final byte of a 4-byte sequence--and *only* that byte.

We can optimize for space later, but this simple fix should be good enough for master for now. :)

fatcerberus: It's rather annoying that Unicode assigned the replacement character to a codepoint which encodes to 3 bytes in UTF-8.  They could have used, say, ASCII 26 instead and then the overallocation wouldn't be needed at all.

svaarala: Hmm, so if the first byte is a final byte of a 4-byte sequence and the next byte is an invalid initial byte (say 0xFF), wouldn't that create 7 bytes out output for 2 bytes of input -- which is larger than 2*3?

This reservation bit needs a short and snappy comment on what the worst case is :-)

svaarala: Sorry, meant 6 + 3 = 9 bytes of output for 2 bytes of input.

svaarala: Anyway, even a 3x expansion is pretty large for a low memory target if the input is, say, 1kB of data and a temporary buffer of 3kB is used. Especially if the data is mostly ASCII.

But like you said, we can take care of that in a separate pull. A simple chunking approach should work well, especially because the streaming approach makes the segmentation arbitrary so that the chunking process doesn't need awareness of codepoint boundaries.

svaarala: Btw, in theory that chunking will be necessary for inputs that are in the gigabyte range: otherwise `3 * len` may be too large (or worse, wrap the C type). It wouldn't be technically correct to reject such inputs because it *may* be that the data only expands to `1 * len` because it's all ASCII.

As an initial implementation something like this just to detect the problem (without actually doing anything more sensible):

    if (len >= DUK_HBUFFER_MAX_BYTELEN / 3) {
        /* Would be too large. */
    }

would be enough. Because that limit exists, a wrap check is not required. If it were necessary, it's important to do wrap checks (e.g. `if (3 * len < len)`) using unsigned types, because signed wrapping is compiler dependent behavior.

Chunking makes this case work pretty well; you won't quite reach the buffer maximum size but pretty close if the chunk size is, say 1kB.

fatcerberus:
@svaarala Pull is merge ready, pending review.

I notice some of the tests are failing, but it doesn't seem to be my fault.  e.g. "linux-x64-hello-ram" errored out during genbuiltins with "KeyError: setter_id" so I guess it's not building against the latest master for some reason?

svaarala:
It's probably building against master, but master doesn't have any missing setter/getter cases for built-ins so it's something I overlooked. I can fix that in master.

fatcerberus:
Most importantly, the TextEncoder/TextDecoder tests passed :)

svaarala:
Ok, the (possible) fix is now in master - try rebasing :)

svaarala:
Since this is close to merge ready, I'll do a read through tonight. The primary thing is to ensure memory safety and that the testcase stresses the relevant corner cases (I can add more tests separately).

svaarala:
Footprint impact on x86 seems to be about +2,2kB. Pretty much for the lowest memory builds (+3% for stripped build) so stripped build should disable this by default (as it's not an ES5 binding).

The config options are maybe now a bit inconsistent - the builtins.yaml refers to `DUK_USE_ENCODING_BUILTINS` but I don't think that's added to config-options?

There are separate config options for TextEncoder and TextDecoder. I'm fine with that, but it'd also be fine to enable/disable them both I think?

fatcerberus:
Whoops, I originally had them combined into one config option, but decided to split them.  Maybe an embedded target only needs decoding, for example.  I forgot to update builtins.yaml it seems.

The main reason for the split though was to be consistent with the other config options for built-ins, there seems to be one option for each built-in constructor.  Then again, typed array support only has one option to enable ArrayBuffer and all views, so I guess this could be a combined setting too.

svaarala:
Some built-ins have separate config options for the binding and functionality when that matters.

For example, it's possible to disable JSON bindings *but* keep the JSON C functionality intact. This requires two config options (`DUK_USE_JSON_SUPPORT` for JSON in general, and `DUK_USE_JSON_BUILTIN` for the bindings).

But if there's no distinction like this here (at least now - since there's no C API) it's not that important yet.

fatcerberus:
One thing I notice is that there's an unnecessary explicit range check for surrogate codepoints in the UTF-8 decoder.  That's actually redundant because the byte ranges are already designed to exclude them.

Overall I think the UTF-8 decoding can be optimized.  The implementation here is basically a verbatim transcription from the WHATWG spec; it should be possible to reduce the footprint impact by reorganizing some things.  Something for future work. :)

fatcerberus:
Note that the functions are vararg on purpose - WebIDL makes a distinction between "argument is not present" (use default value) vs. "arg value is undefined" (coerce).

svaarala:
Good to know, there are a couple of such built-in functions in core Ecmascript too.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+			dec_ctx->codepoint = byte & 0x1f;
+			return DUK__U8_CONTINUE;
+		} else if (byte >= 0xe0 && byte <= 0xef) {
+			/* U+0800 - U+FFFF, 3 bytes */
+			if (byte == 0xe0) {
+				dec_ctx->lower = 0xa0;
+			} else if (byte == 0xed) {
+				dec_ctx->upper = 0x9f;
+			}
+			dec_ctx->needed = 2;
```
svaarala: dec_ctx->needed is set to 2 for both 2- and 3-byte cases - is this correct?

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+
+DUK_INTERNAL duk_ret_t duk_bi_textencoder_constructor(duk_context *ctx) {
+	/*
+	 *  TextEncoder currently requires no persistent state, so this constructor
+	 *  does nothing on purpose.
+	 */
+
+	if (!duk_is_constructor_call(ctx)) {
+        duk_error(ctx, DUK_ERR_TYPE_ERROR, "constructor must use 'new'");
+    }
```
svaarala: Indent broken here.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textencoder_prototype_encode(duk_context *ctx) {
+	duk__encode_context enc_ctx;
+	duk_size_t len;
+	duk_uint8_t *output;
+
+	if (duk_get_top(ctx) >= 1) {
+		duk_dup(ctx, 0);
```
svaarala: Maybe `duk_set_top(ctx, 0)` here instead?

svaarala: Firefox doesn't seem to distinguish between no argument and undefined btw:

```
new TextEncoder().encode()
Uint8Array [  ]
new TextEncoder().encode(undefined)
Uint8Array [  ]
```

If this is against the WHATWG specification, maybe a one sentence comment why the varargs approach is correct. Current behavior (as you of course know :) differs from Firefox:

```
duk> new TextEncoder().encode()
= ||
duk> new TextEncoder().encode(undefined)
= |756e646566696e6564|
```

For `null` behavior matches.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+	duk_decode_string(ctx, -2, duk__utf8_encode_char, (void *) &enc_ctx);
+	if (enc_ctx.lead != 0x0000L) {
+		/* unpaired high surrogate at end of string */
+		enc_ctx.out += duk_unicode_encode_xutf8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, enc_ctx.out);
+	}
+#if defined(DUK_USE_BUFFEROBJECT_SUPPORT)
+	duk_push_buffer_object(ctx, -1, 0, (duk_size_t) (enc_ctx.out - output), DUK_BUFOBJ_UINT8ARRAY);
+#else
+	/* shrink buffer to include only the encoded text */
+	duk_resize_buffer(ctx, -1, (duk_size_t) (enc_ctx.out - output));
```
svaarala: Typically internals create fixed buffers but I can see why a dynamic buffer might make sense here.

Another issue is memory safety: the code assumes `enc_ctx.out` is a stable pointer throughout the encoding process. This is generally not a safe assumption for dynamic buffers if anything, anywhere is able to cause a resize. But in this case the buffer isn't reachable to anyone else so the assumption is safe. Maybe just a single sentence comment that dynamic buffer pointer is assumed to be stable and why that's the case?

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_constructor(duk_context *ctx) {
+	duk__decode_context *dec_ctx;
+	duk_bool_t fatal = 0;
+	duk_bool_t ignore_bom = 0;
+	duk_idx_t nargs;
+
+	nargs = duk_get_top(ctx);
+	if (!duk_is_constructor_call(ctx)) {
+        duk_error(ctx, DUK_ERR_TYPE_ERROR, "constructor must use 'new'");
+    }
```
svaarala: Indent broken above.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	/* XXX: The buffer allocation strategy here (3x input length) is quite
+	 * inefficient.  Maybe switch to a chunk-based strategy?
+	 */
+
+	output = (duk_uint8_t *) duk_push_dynamic_buffer(ctx, len * 3);
+	enc_ctx.lead = 0x0000L;
+	enc_ctx.out = output;
+	duk_decode_string(ctx, -2, duk__utf8_encode_char, (void *) &enc_ctx);
```
fatcerberus: I assumed it would just decode the string the same way Duktape itself does.  Which made the most sense to me: `TextEncoder` sees the noncompliant string the same way other Ecmascript code does.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+		/* XXX: for now, label is ignored */
+		duk_to_string(ctx, 0);
+	}
+	if (nargs >= 2 && !duk_is_null_or_undefined(ctx, 1)) {
+		if (duk_get_prop_string(ctx, 1, "fatal")) {
+			fatal = duk_to_boolean(ctx, -1);
+		}
+		if (duk_get_prop_string(ctx, 1, "ignoreBOM")) {
+			ignore_bom = duk_to_boolean(ctx, -1);
+		}
```
svaarala: These reminded me again that an internal helper for `duk_get_prop_string_toboolean()` would be useful and probably save footprint. I'll maybe prototype that separately :)

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	dec_ctx = (duk__decode_context *) duk_alloc(ctx, sizeof(duk__decode_context));
```
svaarala: Using `duk_alloc()` is a serious leak risk: if an error causes an unwind, it won't be freed as the allocations are not part of the Duktape heap. It's generally safer to use `duk_push_fixed_buffer()` which is very low overhead but still automatically garbage collected.

src-input/duk_bi_encoding.c
fatcerberus: No, that'd be a bug.  Good catch, I need to improve the tests it seems. :)

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+
+	duk_push_this(ctx);
+	duk_get_prop_string(ctx, -1, "\xff" "Context");
+	dec_ctx = (duk__decode_context *) duk_require_pointer(ctx, -1);
+	duk_push_boolean(ctx, dec_ctx->ignore_bom);
+	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_decode(duk_context *ctx) {
+	duk_uint8_t *data = (duk_uint8_t *) "";
```
svaarala: This should probably be `const duk_uint8_t *data`?

src-input/duk_bi_encoding.c
fatcerberus: WebIDL is quite clear (IIRC) that undefined becomes the string "undefined" when coerced to a DOMString.  Firefox must just do the obvious thing and strict-compare against undefined.

I'll check Chrome behavior too though.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+	duk_bool_t fatal = 0;
+	duk_bool_t ignore_bom = 0;
+	duk_idx_t nargs;
+
+	nargs = duk_get_top(ctx);
+	if (!duk_is_constructor_call(ctx)) {
+        duk_error(ctx, DUK_ERR_TYPE_ERROR, "constructor must use 'new'");
+    }
+	if (nargs >= 1) {
+		/* XXX: for now, label is ignored */
```
fatcerberus: The "label" argument for TextDecoder specifying the encoding to use.  I'll clarify the comment.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+	 */
+	output = (duk_uint8_t *) duk_push_fixed_buffer(ctx, 3 + (3 * len));
+	in = data;
+	out = output;
+	while (in < data + len) {
+		ret = duk__utf8_decode_next(dec_ctx, *in++, &codepoint);
+		if (ret == DUK__U8_CONTINUE) {
+			continue;
+		}
+		if (ret >= DUK__U8_ERROR) {
```
svaarala: Maybe comment that we're handling both `DUK__U8_ERROR` and `DUK__U8_RETRY` here, or maybe even assert:

    DUK_ASSERT(DUK__U8_RETRY > DUK__U8_ERROR);
    DUK_ASSERT(ret == DUK__U8_ERROR || ret == DUK__U8_RETRY);

src-input/duk_bi_encoding.c
svaarala: Hmm. Elsewhere the intent is to be as loose as possible but would it make more sense here to replace e.g. invalid CESU-8 sequences with replacement characters?

This is of course not a compliance question but rather what behavior would be most in line with user expectations for non-standard strings. Maybe an XXX note here?

svaarala:
Awesome work overall! This will be great in master :)

Added some review comments. I have some small style etc issues I can address fastest in a small cleanup pull.

Here are some issues not mentioned in review comments which might need changes:
- [ ] It'd be nice to have some short comments for the struct members, e.g. needed, seen, upper, lower. They're mostly apparent by reading the code but still.
- [ ] The `DUK__U8_RETRY > DUK__U8_ERROR` constraint should probably be asserted in a relevant code section (review comment about this too). It's better to assert for constraints rather than just document them (which is better than nothing of course :).
- [ ] The property attributes for TextEncoder, TextEncoder.prototype, TextDecoder, and TextDecoder.prototype differ from at least Firefox:
  - TextEncoder.length is configurable
  - TextEncoder.name is configurable
  - TextEncoder.prototype.encoding is enumerable and configurable
  - TextEncoder.prototype.encode is writable, enumerable, and configurable
  - TextDecoder.length is configurable
  - TextDecoder.name is configurable
  - TextDecoder.prototype.encoding is enumerable and configurable
  - TextDecoder.fatal is enumerable and configurable
  - TextDecoder.ignoreBOM is missing (but I assume it should match the other accessors, i.e. enumerable and configurable)
  - TextDecoder.decode is writable, enumerable, and configurable
  - NOTE: these attributes are a tiny issue - we can address it separately if the specification is not clear on what attributes are correct.

Other related comments; separate pulls probably best:
- Improving test case coverage to cover all code paths exhaustively would be nice. For example a test case decoding all 3-byte sequences, random byte sequences, etc. Mainly useful for valgrind memory safety coverage.
- Test coverage for maximum expansion cases with comments would be nice.
- Performance tests for pure ASCII and mixed content would be nice.

svaarala:
The `duk_alloc()` call causes an active memory leak now by the way:

```
for (i = 0; i < 100000; i++) {
    td = new TextDecoder();
    td.decode(new Uint8Array(10));
}
```

Results in:

```
    MB
3.943^                                                                       #
     |                                                                     @@#
     |                                                                 @@@@@ #
     |                                                              @@@@@ @@ #
     |                                                          @@@@@@ @@ @@ #
     |                                                      @@@@@@@ @@ @@ @@ #
     |                                                  @@@@@@@ @@@ @@ @@ @@ #
     |                                               :@:@@ @@@@ @@@ @@ @@ @@ #
     |                                           @@:::@:@@ @@@@ @@@ @@ @@ @@ #
     |                                        @@:@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |                                     @@@@ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |                               ::::::@@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |                            @@@: ::  @@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |                        ::@@@ @: ::  @@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |                    :@@@: @@@ @: ::  @@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |                 @@@:@ @: @@@ @: ::  @@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |             @@@@@ @:@ @: @@@ @: ::  @@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |         :@@@@ @ @ @:@ @: @@@ @: ::  @@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |      ::@:@ @@ @ @ @:@ @: @@@ @: ::  @@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
     |  @:::: @:@ @@ @ @ @:@ @: @@@ @: ::  @@ @ :@ : :@:@@ @@@@ @@@ @@ @@ @@ #
   0 +----------------------------------------------------------------------->Gi
     0                                                                   1.844
```

and:

```
==12905== HEAP SUMMARY:
==12905==     in use at exit: 2,800,000 bytes in 100,000 blocks
==12905==   total heap usage: 1,404,083 allocs, 1,304,083 frees, 78,046,590 bytes allocated
==12905== 
==12905== LEAK SUMMARY:
==12905==    definitely lost: 2,800,000 bytes in 100,000 blocks
==12905==    indirectly lost: 0 bytes in 0 blocks
==12905==      possibly lost: 0 bytes in 0 blocks
==12905==    still reachable: 0 bytes in 0 blocks
==12905==         suppressed: 0 bytes in 0 blocks
==12905== Rerun with --leak-check=full to see details of leaked memory
==12905== 
==12905== For counts of detected and suppressed errors, rerun with: -v
==12905== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

The `duk_alloc()` API documentation tries to describe that the memory is not automatically garbage collected but there should maybe be a bigger warning: calling the allocation primitives directly makes sense in some specific situations where you want to use the same allocator as the Duktape heap (perhaps the platform doesn't have any allocator by default) but want Duktape to be entirely unaware of it. It's like calling `malloc()` when `malloc()` is the allocation provider.

svaarala:
It's too bad the automatic valgrind testcase is too slow to be of much use (it still hasn't executed for the latest commit because there's still enough commit backlog).

The TextEncoder test case passes without valgrind issues:

```
sva-p@keppihevonen:/tmp/duktape$ node runtests/runtests.js --valgrind --log /tmp/log --verbose --run-duk --cmd-duk ./duk --util-include-path tests/ecmascript/ --prep-test-path util/prep_test.py --minify-uglifyjs2 UglifyJS2/bin/uglifyjs tests/ecmascript/test-bi-textencoder.js 
Executing 1 testcase(s) with 1 engine(s) using 4 thread(s), total 1 task(s), valgrind enabled (for duk)
  duk/vg: test-bi-textencoder
Testcase execution done, running diffs

----------------------------------------------------------------------------


SUMMARY: 1 pass, 0 fail

----------------------------------------------------------------------------

All done.
```

The TextDecoder leak is caught by the valgrind test:

```
sva-p@keppihevonen:/tmp/duktape$ node runtests/runtests.js --valgrind --log /tmp/log --verbose --run-duk --cmd-duk ./duk --util-include-path tests/ecmascript/ --prep-test-path util/prep_test.py --minify-uglifyjs2 UglifyJS2/bin/uglifyjs tests/ecmascript/test-bi-textdecoder.js 
Executing 1 testcase(s) with 1 engine(s) using 4 thread(s), total 1 task(s), valgrind enabled (for duk)
  duk/vg: test-bi-textdecoder
Testcase execution done, running diffs

----------------------------------------------------------------------------

                               test-bi-textdecoder: pass; valgrind {"Leak_DefinitelyLost":1}

SUMMARY: 1 pass, 0 fail

----------------------------------------------------------------------------

All done.
```

src-input/duk_bi_encoding.c
svaarala: Right, the practical alternative would be to push a second fixed buffer for the final result. Not very nice either which is why a dynamic buffer might be the best solution here.

In any case the resulting buffer object will be different than a buffer object created using e.g. `new ArrayBuffer()` or `new Uint8Array()` which are always backed by fixed buffers now. Just thought this is noteworthy if you weren't aware of it, as it's a change to the current behavior for Ecmascript-binding-created buffer objects.

There may be some documentation that needs to be updated to reflect this; it shouldn't have a large impact in any case because C code can always create buffers with fixed, dynamic, or external backing.

Re: memory safety: Duktape won't do that, but it's critical to prove there can be no access from finalizers etc. I think that should be a safe assumption here but it's easy to violate in the general case.

src-input/duk_bi_encoding.c
fatcerberus: That would coerce `undefined` to `false` right?  That'd work here since both `fatal` and `ignoreBOM` are false by default, but what if you want "value-not-present = true"?

svaarala: Maybe a default value argument for missing property?

svaarala: In any case, my point was simply that if there are ~5 call sites, an internal helper will reduce code footprint and also make the call sites shorter (though not necessarily more easy to read unless the helper is named well). For internal helpers long term viability or consistency are not critical because they can be reworked at will.

src-input/duk_bi_encoding.c
svaarala: A small addition here: there are already some internal properties which have type or value assumptions which are simply `DUK_ASSERT()`ed for. In other words, if they are maliciously modified weird (potentially memory unsafe) things may happen, which is why access to internal properties is indeed an important sandboxing issue (and discussed in sandboxing.rst in detail).

So my comment is not really that there is any unique problem here. But rather, it may be worth noting with a short comment that the internal property is memory safety critical.

svaarala: Also small addition re: `duk_alloc()`. It's useful to allocate memory beyond Duktape control, using the same allocator as the Duktape heap, with the extra benefit that a GC will be performed to satisfy the request if necessary. If you just used say `malloc()`, it might fail even when a little GC would have freed up enough memory to satisfy the request. This is pretty annoying on low memory targets (been there, done that ;).

But in any case, `duk_alloc()` and friends should only really be used in these narrow situations.

svaarala:
I'm off to sleep for a bit - let me know when this would be ready to merge. The only merge blockers were really the "needed" bug for 2-byte encodings and the use of `duk_alloc()`. Everything else is something that can be addressed in separate pulls if you'd prefer to get the bulk of the work first into master.

src-input/duk_bi_encoding.c
```diff
@@ -0,0 +1,348 @@
...
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_encoding_getter(duk_context *ctx) {
+	duk_push_string(ctx, "utf-8");
+	return 1;
+}
+
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_fatal_getter(duk_context *ctx) {
+	duk__decode_context *dec_ctx;
+
+	duk_push_this(ctx);
+	duk_get_prop_string(ctx, -1, "\xff" "Context");
```
fatcerberus: I would like a function like that in the _public_ API with the number of times I have to call `duk_push_this()` ;-)

fatcerberus:
Thanks for the tip on the property descriptors, I meant to verify those and then forgot about it.  Chrome agrees with the values you list, so they are probably correct.  I'll look over the relevent section(s) of Web IDL too, but the descriptors should be easy to change before merging, just some quick edits to builtins.yaml.

fatcerberus:
I'll address the `duk_alloc()` issue here so it doesn't get lost in the sea of diffs: I was aware it was unmanaged. :)  That was actually a quick draft I mocked up until I had verified the way tracks data for other built-in objects.  I ended up forgetting to change it because I didn't add a FIXME note.  Silly me!

I admit it didn't occur to me to use a fixed buffer for that purpose, that's actually a lot cleaner than what I'm currently doing.

fatcerberus:
Regarding treatment of `undefined`, Chrome behavior agrees with Firefox:
```
> new TextEncoder().encode()
[]
> new TextEncoder().encode(undefined)
[]
```

Web IDL specification for `TextEncoder`:
```webidl
[Constructor,
 Exposed=(Window,Worker)]
interface TextEncoder {
  readonly attribute DOMString encoding;
  [NewObject] Uint8Array encode(optional USVString input = "");
};
```

Both browsers _seem_ to violate the specification.  Without any annotations, `DOMString` coercion basically boils down to ToString():
https://www.w3.org/TR/WebIDL/#es-DOMString

HOWEVER: the latest Web IDL draft prescribes no special treatment for `undefined`:
https://heycam.github.io/webidl/#es-DOMString

There are other passages in the latter document that imply `undefined` is to be treated the same as "value not present", e.g.:
> It is strongly suggested not to use default value of true for boolean-typed arguments, as this can be confusing for authors who might otherwise expect the default conversion of undefined to be used (i.e., false).

So I think the browsers are actually doing the right thing here, they are just following a more recent version of Web IDL than the one published by W3C.  Ultimately it's probably best if Duktape follows suit to avoid surprises.  This is also how I use `undefined` in my own code, so no complaint from me. ;)

fatcerberus:
One downside to replacing `undefined` with a (sensible) default value is that it can mask bugs.  `undefined` normally string coerces to "undefined" (and not, say, an empty string) so that when you start seeing "undefined" as output, it's pretty obvious you have a bug.  Substituting meaningful behavior can hide that bug, or at least make it more difficult to track down.

fatcerberus:
Hm, genbuiltins apparently doesn't allow `length` for functions to be configurable, it errors out with "non-default length attribute for unexpected object".  This seems to be the cause:
https://github.com/svaarala/duktape/blob/master/tools/genbuiltins.py#L1623-L1634

So I can't fix *all* the property attributes just yet (most of them are fixed though). :)

svaarala:
Heh, yes that's true, genbuiltins.py limits it now: `.length` has consistent property attributes throughout ES5 and the built-in format just assumes that to be the case for all `.length` attributes:

    > In every case, the length property of a built-in Function object described in this clause has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }. Every other property described in this clause has the attributes { [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true } unless otherwise specified.

In fact I'd personally prefer `.length` and `.name` to be non-writable and non-configurable so it was a bit surprising that they're not so for the Encoding API in Firefox. But then, `String.length` is also configurable (but not writable) in Firefox, so maybe the above quote is no longer true for ES6 (haven't checked) and property attributes will need an update once again... :)

In any case genbuiltins.py does have a limitation for this now, so maybe add an XXX note on preferred attributes but keep the genbuiltins.py enforced values for now?

svaarala:
Also one thing that bothered me about Firefox behavior is that the `.encode()` and `.decode()` methods are enumerable. Virtually no built-in methods are enumerable and for good reason IMO so it looks like an oversight.

But then, Chrome does the same: :face_with_head_bandage:

```
> Object.getOwnPropertyDescriptor(TextEncoder.prototype, 'encode')
Object {writable: true, enumerable: true, configurable: true}
```

If it's specifically required that the property be enumerable then, grudgingly, OK. But if it's uncertain (and possible that Firefox and Chrome are "wrong") then it'd probably be better not to have it enumerable and figure it out later. Your call, you have a better idea of the WHATWG requirements :)

svaarala:
Looks good - could you squash the fixes so the commits in master would be clean?

fatcerberus:
Sure.  I'll also give Web IDL a quick look and see what it says w.r.t. method/property enumerability.  It seems awkward to me too.

fatcerberus:
From Web IDL, mapping is as follows:
https://www.w3.org/TR/WebIDL/#es-interfaces

* **interface => constructor**
`{ [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }`

* **attribute => property (on prototype)**
` { [[Get]]: G, [[Set]]: S, [[Enumerable]]: true, [[Configurable]]: configurable }` (this also confirms they must be accessors :)

* **operation => method (on prototype)**
` { [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: true }`

So Firefox and Chrome are right here, unfortunately. 😞

fatcerberus:
Alright, this is ready to go now, feel free to merge whenever. :)

fatcerberus:
Found this:
https://esdiscuss.org/topic/classes-and-enumerability

Prototypes for DOM objects (what Web IDL covers) by default have enumerable properties/methods, in contrast with core Ecmascript where they usually don't.  ES6 classes were considered to have enumerable members by default, but it seems they ultimately went with non-enumerable, like core objects.

svaarala:
Interesting find - I wasn't aware of this apparently long term rift :)

svaarala:
I'll check out the final changes and merge, and then open an Encoding API improvements issue for a bunch of small stuff that can be improved upon incrementally.

fatcerberus:
One possible improvement: TextDecoder constructor could throw if the label doesn't map to UTF-8.  Real-world code may expect it to throw for an unsupported encoding.  Because the label is currently ignored (other to string coerce it), this is not the case right now.

src-input/duk_bi_encoding.c
svaarala: Sure, adding it to the public API would probably a good idea too. But IMO it's best to experiment with it internally first to see if it works out.

PR #978, opened 2016-09-24
Fix regexp executor zero buffer assumption

svaarala:
RegExp executor had an assumption that buffer data allocated using `duk_push_fixed_buffer()` was zeroed which is not the case when `DUK_USE_ZERO_BUFFER_DATA` is not set (default is set).

Labeled security: if `DUK_USE_ZERO_BUFFER_DATA` is explicitly set to false (non-default value), the initial save pointer array may contain garbage that might get used later.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1890,6 +1890,9 @@ Planned
   seed mixing; previous algorithm (Shamir's three-op algorithm) is still
   used for low memory targets and targets without 64-bit types (GH-970)
 
+* Fix incorrect buffer zeroing assumption in regexp executor, triggered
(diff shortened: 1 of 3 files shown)


svaarala:
To reproduce: compile Duktape 1.5.1 with `DUK_USE_ZERO_BUFFER_DATA` unset and run the following:

```js
var re = /x|(y)|(z)/;
var res = re.exec('x');
print(res);
```

When the regexp executes, the capture group save pointers are uninitialized. The regexp matches without writing to the capture groups so that the uninitialized values get used and cause a segfault.

svaarala:
With the fix in place the above works.

PR #980, opened 2016-09-29
Add documentation for ES6 Symbol implementation

svaarala:
Some drafting of a possible Symbol implementation which would be memory efficient and reframe the existing internal keys as specific kinds of Symbols.

Note: I'm not sure this approach is actually viable yet, I'll do some prototyping and specification reading to figure out if this works.

### DIFF
--- a/doc/symbols.rst
+++ b/doc/symbols.rst
@@ -1,0 +1,230 @@
+==========================
+ES6 Symbols in Duktape 2.x
+==========================
+
+Overview
+========
+
+Duktape 2.x adds ES6 Symbol support.  Duktape 1.x internal keys are unified
+with the Symbol concept, and are considered a custom "hidden symbol" type
+which is not normally visible to Ecmascript code.  C code can access hidden
+symbols, however.
+
+The internal implementation is similar to existing internal keys.  Symbols
+are represented as ``duk_hstring`` heap objects, with the string data
+containing a byte prefix which is invalid (extended) UTF-8 so that it can
+never occur for normal Ecmascript strings, or even strings with non-BMP
+codepoints.  Object coerced strings have a special object class and the
+underlying symbol is stored in ``_Value`` similarly to e.g. Number object.
+
+Representation basics:
+
+* Symbols have an external type ``DUK_TYPE_STRING``.
+
+* Symbols have internal type tag ``DUK_TAG_STRING``.
+
+* Symbols can be distinguished internally from ordinary strings by looking
+  up the ``DUK_HSTRING_FLAG_SYMBOL`` flag.  Hidden symbols also have the
+  ``DUK_HSTRING_FLAG_HIDDEN`` set.
+
+Behavior basics:
+
+* Symbols are visible to Ecmascript code as required by ES6 and later.
+  Hidden symbols are not visible through e.g.
+  ``Object.getOwnPropertySymbols()``.  They can only be accessed if a
+  reference to the hidden symbol string is somehow available, e.g. via a
+  C binding.
+
+* Symbols are visible to the public C API as strings: ``duk_is_string()``
+  is true, ``duk_get_string()`` returns a pointer to the symbol internal
+  string representation, etc.  C code can create symbols simply by pushing
+  C strings with a specific format, see below.
+
+* While symbols are strings in the C API, coercion semantics are based on
+  the Ecmascript behavior.  For example, ``duk_to_string()`` applied to a
+  symbol throws a ``TypeError``.
+
+See:
+
+* https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Symbol
+
+* http://www.2ality.com/2014/12/es6-symbols.html
+
+Internal key formats
+====================
+
+Duktape custom hidden Symbols have an initial 0xFF byte prefix, which matches
+the existing convention for Duktape 1.x internal keys.  While all bytes in the
+range [0xC0,0xFE] are valid initial bytes for Duktape's extended UTF-8 flavor,
+the continuation bytes [0x80,0xBF] are never a valid first byte so they are used
+for ES6 symbols (and reserved for other future uses) in Duktape 2.x.
+
++-----------------------------------------------+-----------------------------------------------------------------+
+| Internal string format                        | Description                                                     |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <ff> SomeUpperCaseValue                       | Hidden symbol (Duktape specific) used by Duktape internals.     |
+|                                               | Previously called internal properties.  First byte is 0xFF,     |
+|                                               | second is from [A-Z].                                           |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <ff> anyOtherValue                            | Hidden symbol (Duktape specific) used by application code.      |
+|                                               | First byte is 0xFF, second is ASCII (0x00-0x7f) but not         |
+|                                               | from [A-Z].                                                     |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <ff> <ff> anyOtherValue                       | Hidden symbol (Duktape specific) used by application code.      |
+|                                               | First and second bytes are 0xFF, remaining bytes arbitrary.     |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <80> symbolDescription                        | Global symbol with description 'symbolDescription' created      |
+|                                               | using Symbol.for().                                             |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <81> symbolDescription <ff> uniqueSuffix      | Local symbol with description 'symbolDescription'.  Trailing    |
+|                                               | unique string makes the symbol unique.  The unique suffix is    |
+|                                               | opaque and chosen arbitrarily by Duktape.  It's unique within a |
+|                                               | Duktape heap (across all global environments).                  |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <81> <ff> uniqueSuffix                        | Local symbol with an empty description.  Unique suffix makes    |
+|                                               | each such symbol unique.  The unique suffix is arbitrary but    |
+|                                               | must not contain the 0xFF byte.                                 |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <81> <ff> uniqueSuffix <ff>                   | Local symbol with undefined description.  ES6 differentiates    |
+|                                               | internally between symbols with an empty string description vs. |
+|                                               | symbols with an undefined description.                          |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <81> symbolDescription <ff>                   | Well known symbol with description 'symbolDescription'.  Well   |
+|                                               | known symbols (like Symbol.iterator) are local symbols which    |
+|                                               | are still shared across "code realms".  Any fixed suffix never  |
+|                                               | colliding with runtime generated unique local symbols works,    |
+|                                               | currently an empty suffix is used.                              |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <82 to bf>                                    | Initial bytes 0x82 to 0xBF are reserved for future use.         |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <00 to 7f>                                    | Valid ASCII initial byte.                                       |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <c0 to f7>                                    | Valid standard UTF-8 (or CESU-8) initial byte.                  |
++-----------------------------------------------+-----------------------------------------------------------------+
+| <f8 to fe>                                    | Valid extended UTF-8 initial byte.                              |
++-----------------------------------------------+-----------------------------------------------------------------+
+
+Useful comparisons (``p`` is pointer to string data) for internal use only:
+
+* ``p[0] == 0xff || (p[0] & 0xc0) == 0x80``: some kind of Symbol, either Duktape
+  hidden Symbol or an ES6 Symbol.
+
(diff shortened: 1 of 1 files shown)


svaarala:
Regarding existing internal keys there are at least two logical approaches:

* Treat internal keys like global symbols, e.g. `<0xff>Value` would be the same as `Symbol.for('Value')`. This would be compatible with Duktape 1.x behavior because in 1.x internal keys can be used as property keys and they're not fully sandboxed. For example, you can create an internal key via a buffer object. Having done that, add a custom "hidden Symbol" type for truly hiding properties from Ecmascript code (in the sense of #979).
* Treat internal keys as "hidden Symbols", i.e. hide them from Ecmascript code as in #979. This is incompatible with Duktape 1.x for any code which expects to be able to access internal keys from Ecmascript code (and not just C code). Then add ES6 Symbols which are visible to Ecmascript code too.

Both approaches have their merits. In some cases application writers would have preferred internal keys to be hidden from Ecmascript code to begin with, so that alternative is more of a bug fix or upgrade for them. Then again any code expecting internal keys to be visible to Ecmascript code will break and need to be upgraded to use ES6 Symbols (which are visible to user code).

svaarala:
With #979 buffer-to-string coercion would no longer allow access to internal properties (which might become a custom "hidden symbol" type). However, it would still be possible to use buffer-to-string coercion to come up with a string matching an ES6 Symbol - potentially even a symbol created as `Symbol()` if one can guess how they unique identifier for such a symbol is created. This is something that must be documented.

fatcerberus:
> However, it would still be possible to use buffer-to-string coercion to come up with a string matching an ES6 Symbol - potentially even a symbol created as `Symbol()` if one can guess how they unique identifier for such a symbol is created.

Having read the diff for #979 and assuming Symbols are implemented internally as `<0xff>UniqueID` like proposed, wouldn't such an attempt to subvert the sandbox be caught during handling of [[GetProperty]]?  I'd assume that property access through a legitimate Symbol would use a different code path from normal property lookup.

svaarala:
> Having read the diff for #979 and assuming Symbols are implemented internally as <0xff>UniqueID like proposed, wouldn't such an attempt to subvert the sandbox be caught during handling of [[GetProperty]]? I'd assume that property access through a legitimate Symbol would use a different code path from normal property lookup.

ES6 Symbols would need to be internally distinguishable from "hidden symbols" (a custom feature).

So what I meant above is that if the internal representation for `Symbol.for('foo')` were the string `<0xA0>foo`, one could simply:

```js
var buffer = Duktape.dec('hex', 'a0666f6f');
var string = String.fromBuffer(buffer);
```

And `string` would now be identical to the return value of `Symbol.for('foo')` which could then be used to access properties etc.

It's not really a big deal in the usual case because ES6 Symbols can be "peeked" via `Object.getOwnPropertySymbols()` so they can't be used as is for secure hiding of properties. But there's still a small difference in needing to find the property from an existing object vs. creating Symbols via their internal representation.

With #979 any Duktape specific "hidden symbols" would not be accessible from Ecmascript code. You could still create such strings, and C code could pass them to you, but they'd still be invisible to even `Object.getOwnPropertySymbols()` and, most importantly, attempts to read/write the property. The same treatment cannot be applied to ES6 Symbols because they must be readable from Ecmascript code.

fatcerberus:
I thought `Symbol.for()` returned a Symbol object?  `String.fromBuffer()` gives you a string back.  So the handling for `obj[symbol]` vs `obj[string]` would be necessarily different, and you could implement the sandboxing cleanly that way.

svaarala:
> I thought Symbol.for() returned a Symbol object? String.fromBuffer() gives you a string back. So the handling for obj[symbol] vs obj[string] would be necessarily different, and you could implement the sandboxing cleanly that way.

Right, but the draft design is to use `duk_hstring` for symbols, just like it is used for internal keys now. So a string with a certain internal representation (not possible to create via standard string operations) would be interpreted as a symbol.

svaarala:
And just to be clear `Symbol.for()` returns a "plain" symbol. It's different from `Object(Symbol.for())`.

fatcerberus:
Ah, I see now.  That makes sense then, thanks for clarifying. :)

svaarala:
The main reason I'll try that approach first is that it's compact and works well with the existing internal key mechanism. A concrete example of where it'd be convenient is that property tables are now string keyed, and changing them to support actual two separate key types would be much more hassle than dealing with symbols as a class of non-standard strings (not overlapping with standard strings).

That said the other approach of adding an explicit symbol type would also be logical. I have a hunch it'd lead to larger code, but on the plus side it'd more cleanly separate the types.

svaarala:
Also for C code it's useful that symbols and string keys are just `const char *` (with the caveat of potential NUL types in both).

fatcerberus:
...wait, I take that back, it still doesn't make sense.  Since we're only intending to sandbox Ecmascript code, the handling for the GetProp opcode (NOT the internal GetProperty handler) can easily differentiate between Symbol and string keys and reject only the invalid strings.  Symbols would be allowed to propagate to the actual lookup stage which could then still use the "invalid UTF-8" form for the symbols/internal properties.

In other words, I'm proposing to put the check directly at the Ecmascript boundary (the bytecode executor).

svaarala:
> ... can easily differentiate between Symbol and string keys and reject only the invalid strings.

What I'm talking about specifically is looking up a property using a string which has the *same* internal byte representation as a corresponding Symbol - not an arbitrary invalid string. At that property lookup point they're indistinguishable for the property code.

svaarala:
> In other words, I'm proposing to put the check directly at the Ecmascript boundary (the bytecode executor).

In #979 the check is there, indeed. But it *must* pass through ES6 Symbols. And if they can be created via buffers (as I described above), the created-via-buffer symbols cannot be distinguished at the property lookup stage.

fatcerberus:
But you can't create a symbol from a buffer, you can only create a string... I must be missing something here. :confused:

I think I get what you're saying, though.  Since Symbols are implemented as internal properties (at least initially), that means you can just do `Symbol.for("InternalPropName")` to get at it and this couldn't be prevented.  With #979 in place you *couldn't* do `obj[String.fromBuffer(internalNameBuffer)]` because the check would catch that.

svaarala:
> But you can't create a symbol from a buffer, you can only create a string... I must be missing something here. :confused:

Again, what I'm saying that:

```
var sym = Symbol.for('foo');
var str = String.fromBuffer(Duktape.dec('hex', 'a0666f6f'));
```

would not just be similar but would be the exactly same `duk_hstring` object, interned to the exactly samer native pointer. So there would be no type tag or anything else to separate strings and symbols, just that symbols are in a separate space due to their invalid byte prefix. As a consequence a buffer-to-string coercion can create *both* strings and symbols.

svaarala:
Another way of putting it: consider the current Duktape 1.x arrangement. Internal strings are in a separate conceptual space because you can't create them with any standard Ecmascript code. But they don't have a separate type, it's just a convention in the C API and the internal representation.

What I'm doing is just interpreting the convention as "Symbols". There's no internal typing change at all.

fatcerberus:
Okay, NOW I understand, you intend that `Symbol.for()` returns a plain string that just happens to be prefixed with an `FF` byte and thus distinguishable at the runtime level as an ES6 Symbol.

At that point I don't think there's anything you can do except to document that any string of the form `"\xffXXX"` is exactly equivalent to `Symbol.for(`XXX`)`, and leave it at that.  It's not _technically_ a sandboxing violation if it's documented "by-design" behavior. :-)

svaarala:
Well, the format I was proposing is not 0xFF but 0xA0 so that ES6 Symbols are distinguished from Duktape specific "hidden symbols" which are useful for sandboxing and they can get a different treatment (like #979).

> At that point I don't think there's anything you can do except to document that any string of the form "\xffXXX" is exactly equivalent to Symbol.for(XXX), and leave it at that. It's not technically a sandboxing violation if it's documented "by-design" behavior. :-)

Right, that's why I said it needs to be documented :)

But ES6 Symbols are not really a sandboxing feature anyway, because you can always just say `Object.getOwnPropertySymbols()` and then just lookup the properties. So it's not necessarily a big deal you can use non-standard code to create Symbols. Sandboxing is a whole different thing.

fatcerberus:
In that case, I don't think there's an issue - if a Symbol is prefixed by 0xA0 while Duktape internals are 0xFF, then it's easy to reject the latter while passing the former.

Anyway, I was under the impression above that Symbols would get their own internal typing, and only the actual property lookup would be string-keyed.  Hence all the confusion. :)  I *am* curious what the ES6 specification has to say about this design, would it be compliant?

svaarala:
Again just to be sure the Ecmascript expression:

    "\xffXXX"

would not be a symbol because internally that \xFF is represented in CESU-8. The interpretation only applies when you create strings outside of the normal Ecmascript string operations. In other words:

```
duk> x = Duktape.dec('hex', 'ff666f6f')
= |ff666f6f|
duk> y = String.fromBuffer(x)
= "\xfffoo"
duk> typeof y
= "symbol"
duk> typeof '\xfffoo'
= "string"
```

svaarala:
> Anyway, I was under the impression above that Symbols would get their own internal typing, and only the actual property lookup would be string-keyed.

This should be explained in the draft document in the pull... :)

fatcerberus:
I admit I didn't read the document.  I usually don't pay much attention to brand-new diffs because a lot of the time they begin life as a pile of FIXMEs.  Sorry about that, at least we're on the same page now. :-)

svaarala:
> I am curious what the ES6 specification has to say about this design, would it be compliant?

Yes, because there's no standard code that can create internal strings / symbol strings.

fatcerberus:
Good point, even `TextDecoder` (if implemented) can't be used to create such a string because it's required to decode to a well-formed string using replacement characters if necessary, or else throw.

svaarala:
Right, the downside of the internal strings approach is that for sandboxing you need to remove the ability to do a 1:1 buffer-to-string coercion (e.g. String.fromBuffer) and ensure that all C bindings are safe in that they don't do the equivalent. (These are described in the sandboxing document in some detail.)

This is rather error prone for actual sandboxing, which is why #979 would make sense to me: even if you have such a string, you can't use it to access hidden internal properties.

For ES6 Symbols an implementation actually separating strings and symbols would provide better separation between buffer and strings (for that non-standard 1:1 buffer-to-string coercion or any C equivalent). I'm not sure if it's worth the price of adding another tagged type with all the internal changes (and footprint) that entails. But it's of course a legitimate option otherwise.

svaarala:
Btw, this is why it's a good question whether a 1:1 buffer-to-string function should exist in the Ecmascript API by default. If it didn't exist, it'd always be up to user C code to do so (or not, if that matters). But because writing all C code with that in mind is error prone, #979 makes sense to me regardless.

fatcerberus:
Another question not mentioned in the draft is how symbols should be exposed in the debug protocol.  Should they be provided through GetObjPropDesc[Range] (with a flag to distinguish symbols), or a new command?

svaarala:
Sure, there are a lot of issues like that. I'm working on a prototype draft to see how many places are affected by this possible implementation approach. It's then easier to figure out if a different approach would be better.

Regarding debug protocol, I'd be happy with just exposing symbols in their internal representation, i.e. as specially formed strings. The internal string format needs to be understood by the debug client anyway.

svaarala:
So far this approach seems doable. But one issue has more impact than I originally thought:
- Because Symbol is a separate primitive type in the specification, any place that says "is X a string?" must not accept a symbol.
- This means all places using `duk_is_string()`, checking for a tagged type using a type mask, or handling a duk_tval `DUK_TAG_STRING` must explicitly handle symbols as opposed to ordinary strings.

This is also doable but it does bring the implementation impact closer to adding a tagged type; it'd be possible to add a symbol tagged type *but* have it backed by `duk_hstring` for example.

In any case it's easiest to say after some experimentation.

fatcerberus:
> This is also doable but it does bring the implementation impact closer to adding a tagged type; it'd be possible to add a symbol tagged type _but_ have it backed by `duk_hstring` for example.

This is what I was getting at yesterday when I said there would be a good amount of special casing involved with the symbol-as-string approach so in the end the footprint impact may not differ as much as you think between the two approaches.

svaarala:
Also, if an explicit symbol tag is accompanied by an explicit heap object type for symbols, *that* will bloat internals also considerably. The main issue is e.g. property tables whose keys are now just an array of `duk_hstring` pointers but would now need to handle both kinds of pointers.

One workaround is to use a tagged type backed by `duk_hstring`. But that means that any places in code dealing with a `duk_hstring` outside of a tagged value will *still* need the symbol special casing.

All in all, there are a lot of small issues and my hunch is not very strong - I need to prototype it a bit to see what works.

fatcerberus:
Interestingly, SpiderMonkey seems to distinguish between undefined description vs. empty string in a way that's externally visible:
```
fatce@pig-envy /c/src/mozilla-central/js/src/build_OPT.OBJ
$ dist/bin/js --version
JavaScript-C52.0a1
```

```
js> var sym1 = Symbol()
js> var sym2 = Symbol("")
js> sym1
Symbol()
js> sym2
Symbol("")
js> 
```

Mozilla's REPL seems to be able to pull the Symbol's [[Description]] directly.  Both coerce the same of course:
```
js> sym1.toString()
"Symbol()"
js> sym2.toString()
"Symbol()"
```

svaarala:
It would be safest to implement the internals so that the distinction is made because it seems deliberate in ES6 and it's just one binding away from being exposed in some way.

svaarala:
Merging the description now. It's been updated for the distinction between an undefined description and an empty description and should match the current implementation.

PR #982, opened 2016-09-30
Add initial ES6 symbol support

svaarala:
This is a prototype for an implementation approach where Symbols are `duk_hstring`s with a specific byte prefix which cannot occur for valid (extended) UTF-8 strings. The existing internal string concept is essentially taken a step further to support ES6 Symbols. There's no new internal type for Symbols as such, and the C API sees Symbols as just strings.

- [x] Code change for symbol support
- [x] Fix compiler warnings
- [x] Fix Reflect.keys()
- [x] Testcase support
- [x] Fix perftest coercion issues
- [x] Check key coercion for duk_def_prop() and other property calls
- [x] Wiki pull for symbols howto (should cover the C API aspects especially) - https://github.com/svaarala/duktape-wiki/pull/152
- [x] 2.x migration notes
- [x] Releases entry
- [x] Read through pull comments and add cleanup issues to #1120


tests/ecmascript/test-bi-symbol-basic.js
```diff
@@ -0,0 +1,247 @@
...
+    o1 = Object(s1);
+    print(s1 == o1, s1 === o1);
+    o2 = Object(o1);
+    print(s1 == o2, s1 === o2);
+    print(o1 === o1);
+
+    // ToString coercion is a TypeError.
+    try {
+        s1 = Symbol('123');
+        t = String(s1);
```
fatcerberus: If I understand this test correctly, then V8 (at least) is not compliant:

```
> String(Symbol("pig"))
'Symbol(pig)'
```

However:
```
> Symbol(Symbol("pig"))
TypeError: Cannot convert a Symbol value to a string
    at <snip>
```

svaarala: I noticed the same thing, but ES6 behavior seems pretty clear: http://www.ecma-international.org/ecma-262/6.0/#sec-tostring:

> Symbol: Throw a TypeError exception.

svaarala: The requirement is actually a bit awkward for Duktape because the internals directly or indirectly assume that `ToString()` is fine for internal keys and such. So if they're made Symbols, and that coercion is not allowed, all the related call sites need to be converted to use some internal "duk_to_string_accept_symbol()"-like helper. For now I've rejected ES6 Symbol string conversion but to allow string conversion (which is a nop) for the custom internal keys (0xFF prefix).

svaarala:
Looking at the footprint change which is roughly 0.5kB now, the final footprint impact would probably be around 1-2kB. It may be large enough to need a config option to turn off; symbols would still work if you constructed them from C code, but the constructor, `Symbol.for()` etc bindings would be removed (similarly to disabling buffer objects, which still leaves behind the basic buffer types etc).

tests/ecmascript/test-bi-symbol-basic.js
svaarala: Ah, my bad: in ES6 the String() constructor call has a special case for Symbols so that it doesn't end up invoking the built-in ToString() operation: http://www.ecma-international.org/ecma-262/6.0/#sec-string-constructor-string-value.

But anything that causes `ToString()` to happen should TypeError. So V8 is probably right here, Symbol() uses ToString().

svaarala:
Scanned through ES6 specification for:
- All "symbol" references
- All "string" references
- All "@@" references (for well known symbols)
- For ToPropertyKey and IsPropertyKey references

Scanned code base for:
- duk_is_string() (duk_to_string() and variants are not an issue: they reject symbols already)
- duk_get_string() and variants
- duk_require_string() and variants
- DUK_TAG_STRING
- DUK_TYPE_STRING and DUK_TYPE_MASK_STRING
- DUK_TVAL_IS_STRING

Relevant places in Dukape code are either fixed or marked FIXME. This coverage is probably not 100% but close to it.

svaarala:
So assuming I understood well-known symbols correctly this time, there are essentially three kinds of Symbols in ES6:
- Global symbols, Symbol.keyFor(sym) returns its description. Knowing its description you can create/access the symbol using Symbol.key(). For Duktape there's no global registry, but rather an identical symbol is created whenever the same description is requested which satisfies the conceptual registry requirements as far as I understand.
- Local symbols, each call to Symbol() creates a unique new instance. There's no way to recreate the same Symbol using standard APIs. For Duktape C code has that ability if it happens to know the underlying Symbol representation, and 1:1 buffer-to-string conversions allow that. May or may not have a description.
- Well-known symbol, essentially a local symbol but shared across code realms so it can't be "unique". As a practical matter well-known symbols must be created with a fixed internal suffix which won't overlap with the internal suffix used for dynamically created local Symbols.

This raises a few C API issues. The simplest approach for now is to provide no API for Symbol creation: one can simply create them as byte strings following the documented internal string format. But if API calls (or macros) are added:
- There needs to be an API call to create a global symbol with a given description.
- There needs to be an API call to create a local symbol with a given description (or NULL), and an automatic unique suffix guaranteed to be unique across at least that Duktape heap (preferably others).
- There also needs to be an API call to create a local symbol with a given description (or NULL) and a forced unique suffix, so that user code can also create its own "well known" symbols with a known internal suffix. Otherwise the user would need to pass the symbols across Duktape threads manually rather than creating them on the spot.

All in all I'll probably leave the C API out of the initial merge - there's no C API for internal symbols either. But at least C string initializer macros would be very nice to hide the representations in a more version compatible manner.

svaarala:
Now the branch disables the Symbol built-in bindings for stripped builds (also minimal because minimal configuration is ES5 only). There's still a +500-600 byte impact from symbol support.

Note that disabling symbol built-ins doesn't disable symbol handling as such, and one can still create symbols from C code directly into their byte representation and they behave like symbols should, to the extent this pull supports them.

Most likely there needs to be two separate defines: one for symbol support in general (that can be dropped entirely) and another for the presence of the built-in symbol bindings. Let's see if that's possible in a reasonable manner.

svaarala:
I'll focus on this for the next few days and choose one implementation option or the other - there will then be a symbol baseline (with no explicit C API) in master to work against. Changing the implementation approach (from duk_tval tag or not) can be done in master, as the difference is not visible to Ecmascript code (except in very contrived situations).

fatcerberus:
One part of ES6 I'm having a hard time wrapping my head around is "code realms".  Best I can tell the closest analogue in Duktape would be to create multiple contexts - since each realm gets a unique global object, etc.

svaarala:
That should be pretty close I think. Duktape doesn't actually track "code realms" as such: threads created into the same global space just happen to have the same built-in object pointers so they behave as if they live in the same space. But there's no internal mechanism preventing a thread that mostly shares the same pointers but with one pointer (say Number built-in) replaced with something else. There's no API to do that but as far as internal state goes, that's the case now.

It may be necessary to formalize that arrangement somehow, e.g. so that a `duk_hthread` explicitly points to some "code realm" object which would "own" the built-in objects. Any changes to the built-ins e.g. during sandboxing setup would then propagate to any threads that already exist within that realm.

svaarala:
That indirection will make some call sites worse (specifically those that deal with built-in objects) but it will also make `duk_hthread` smaller because the built-in table is not copied to every thread in the same global space.

fatcerberus:
Apparently objects can be passed between realms, otherwise there would be no reason to explicitly specify that well-known symbols are shared.  So yeah, the closest thing we have right now seems to be creating multiple threads, each with its own global object.

svaarala:
That symbol sharing works itself out quite naturally in this pull (or any approach where symbols are internal strings): a globally registered symbol (GlobalSymbolRegistry conceptually) has a fixed representation which is just a prefix and the symbol name. So, if the same symbol is registered by name (Symbol.for()) anywhere, they'll get the same `duk_hstring` for their internal representation.

svaarala:
A slight correction: Symbol.for() globally registered symbols are distinct from well-known symbols:

```
> Symbol.iterator
Symbol(Symbol.iterator)
> x = Symbol.for('Symbol.iterator')
Symbol(Symbol.iterator)
> Symbol.iterator == x
false
```

But they can still be handled the same way: if they have a fixed representation, they'll map to the same `duk_hstring` anywhere in the same heap.

fatcerberus:
I assume the `Symbol.for()` registry is also shared between realms?

svaarala:
Yes, http://www.ecma-international.org/ecma-262/6.0/#sec-symbol.for:

> The GlobalSymbolRegistry is a List that is globally available. It is shared by all Code Realms.

fatcerberus:
Thanks, that makes sense why `Symbol.for()` exists then.  I couldn't figure out what purpose it served since within a single thread/realm you can easily pass a regular symbol around to share it.  But if you want to access a symbol from a different realm it's indeed more convenient to just do Symbol.for("name").

svaarala:
It's also convenient in some cases to create a symbol with a known name using `Symbol.for(name)` rather than pass the symbol around.

svaarala:
Been working off and on in fixing issues in the pull. There are a bunch of FIXMEs still left but I hope to get this into mergeable order early next week. That doesn't mean it will be merged though, but it's then time to figure out if this merge is good enough as a first implementation for 2.x. It's then possible to rework it in an API compatible manner within 2.x but more than likely reworking the whole approach would also impact the API and would be most natural in 3.x.

svaarala:
Getting close: not many actual fixes needed for this to be mergeable as a MVP... :)

fatcerberus:
Good to hear.  Looking forward to having Symbol support in master, then I'll have to find all sorts of new and wonderful ways to abuse it. :)

fatcerberus:
@svaarala What is the convention for well-known symbols?  Is it easy to reference them from C code as, e.g. `"\xa0" "toStringTag"`?

svaarala:
Yes, C code will be able to create local symbols, global symbols, well-known symbols, and "hidden" symbols (= previous internal keys). Hidden symbols are \xFF prefixed as a carryover from 1.x, and will be different in that they're actually hidden from even `Object.getOwnPropertySymbols()`. They can be accessed through C code and are intended for hidden state.

But I thought I'd add some API macros to create those forms without needing to use direct hex escapes, e.g.:

```
#define DUK_SYMBOL_HIDDEN(desc) ("\xff" (desc))
#define DUK_SYMBOL_GLOBAL(desc) ("\x80" (desc))
/* ... */
```

so that one could then:

```
duk_push_string(ctx, DUK_SYMBOL_GLOBAL("scoreCounter"));
```

For unique symbols it may be useful to have a C API to create one, because there's a unique part that might not be best filled by a macro. One simple way to do that in a macro is to use a `__FILE__ ":" STRINFIGY(__LINE__)` string as the unique suffix, which should work in most cases reliably. But the resulting symbol strings would be rather long so an API call is probably preferable.

svaarala:
A bit unrelated but regarding stashes, once symbols are implemented, and #979 merged, there's not much need for the stash objects if one is using hidden symbols. Hidden symbols would be reliably isolated from Ecmascript code, so that:
- Global stash could be hidden symbols on the global object.
- Thread stash could be hidden symbols on the thread object.
- Heap stash still needs its own object.

There are also downsides to this, e.g. the need to use a hidden symbol where previously a plain string key on the stash object would be isolated from C. So, this is something to consider after 2.x.

svaarala:
Alright, no more FIXMEs in the code or tooling. That doesn't mean the code is finished or correct, but no obvious things anymore. I'll work on testcases next - that'll probably reveal something that needs fixing but other than that this is slowly coming to completion.

fatcerberus:
I could check out this branch and do some manual testing too - there are a few use cases I have in mind involving private properties that are currently awkwardly handled using closures.

svaarala:
That'd be great - I'll be improving the testcases for maybe a day or two so there's good time.

svaarala:
There's still the possibility of not merging this branch but opting to use a different approach, e.g. where symbols are a distinct internal type. The testcases in this branch would help do so more quickly than it took to develop this approach. Also the code diff indicates where the symbol related changes are needed.

However, the alternatives are looking quite unappealing at present:
- This approach requires essentially no mandatory C API changes because symbols are strings. Small additions might be useful (e.g. `duk_is_symbol()` and maybe a primitive to push a unique local symbol).
- Existing internal properties become "hidden symbols" and are not exposed via e.g. `Object.getOwnPropertyNames()`. But they still otherwise behave like symbols (typeof etc) which is nice.
- There's no need for an additional internal type or type tags.
- Ecmascript code doesn't really see the difference between the approaches, so changing it later is not difficult (for Ecmascript code).

So, this approach makes sense to me for Duktape 2.x. If it turns out a different approach is better, it would be relatively easy to change for Duktape 3.x -- that would mean breaking API changes, but since there are very few API changes here, the calling code upgrade cost is not that much higher overall.

fatcerberus:
Found some warnings:

```
duk_js_ops.c(526): warning C4101: 'tv_tmp': unreferenced local variable (compiling source file ..\src\shared\duktape.c)
duk_js_ops.c(670): warning C4101: 'rc': unreferenced local variable (compiling source file ..\src\shared\duktape.c)
duk_js_ops.c(677): warning C4101: 'rc': unreferenced local variable (compiling source file ..\src\shared\duktape.c)
duk_js_ops.c(688): warning C4101: 'rc': unreferenced local variable (compiling source file ..\src\shared\duktape.c)
duk_js_ops.c(697): warning C4101: 'rc': unreferenced local variable (compiling source file ..\src\shared\duktape.c)
d:\src\minisphere\msvs\duk_bi_object.c(697): warning C4700: uninitialized local variable 'enum_flags' used
```

Unreferenced variable warnings are surely harmless, but "uninitialized local variable used" is concerning.

svaarala:
There are still warnings - I haven't cleaned that part up yet. That `enum_flags` warning has been fixed and should disappear if you pull again.

fatcerberus:
Yeah, I wasn't too surprised by the warnings - I know this is still a WIP :)  If not for the `enum_flags` warning I would have kept quiet.

svaarala:
I'll fix the warnings when I do a code read-through tomorrow. That `enum_flags` thing was an actual bug (caused by a line being dropped in a rebase) BTW :)

svaarala:
I'm off to sleep for a bit - let me know how it turned out :) If you want to create symbols from C code, this should be relatively up-to-date w.r.t. the internal formats: https://github.com/svaarala/duktape/pull/980/files.

fatcerberus:
One thing I noticed looking through the symbol doc is this bit:
> Symbols are visible to the public C API as strings: ``duk_is_string()`` is true, ``duk_get_string()`` returns a pointer to the symbol internal string representation, etc.

What of `duk_require_string()`?  As an API consumer using `duk_require_xxx()` to validate arguments, I'd want it to throw a TypeError.  `duk_is_string()` being true is a bit inconvenient too, for the same reasons.

fatcerberus:
Hm, I don't think this is intended behavior:

```
duk> String(Symbol.toStringTag)
= "Symbol(Symbol.toStringTag)"
duk> Symbol.toStringTag
= undefined
```

fatcerberus:
I notice `Symbol.iterator` is currently commented out in builtins.yaml.  Since iterators are duck typed, could that symbol be enabled?  For built-in objects such as Array it would be undefined, but I could still define it on my own objects.

svaarala:
> What of duk_require_string()? As an API consumer using duk_require_xxx() to validate arguments, I'd want it to throw a TypeError. duk_is_string() being true is a bit inconvenient too, for the same reasons.

`duk_require_string()` accepts symbols now - same as it does for internal strings at present. I know that's a bit inconvenient, but it would also be weird if symbols had `DUK_TYPE_STRING' but wouldn't be accepted by `duk_require_string()`.

Regarding `String(Symbol.toStringTag)` coming out as "undefined" in an interactive session, that's a consequence of using JX for console stringification. Since symbols are not allowed in JSON and JX doesn't serialize them either, the result is undefined. The result is a symbol though:

```
duk> typeof Symbol.toStringTag
= "symbol"
duk> String(Symbol.toStringTag)
= "Symbol(Symbol.toStringTag)"
```

Re `Symbol.iterator`: sure, I'll add it. I did enable a few other symbol properties for testing them out, e.g. `Math[Symbol.toStringTag]` is set just to see that the properties come out right through tooling. I'll add `Symbol.iterator` too.

svaarala:
@fatcerberus So this pull uses the approach where C API doesn't know symbols as such - they are strings because they behave like strings for property access. This carries over the Duktape 1.x behavior into symbols: Duktape 1.x internal keys already behave like that, they're just strings as far as C code is concerned. I know that's not ideal but it's also not changing how the API works, although it's maybe more exposed to users if they deal with symbols more than with internal keys.

Adding a `duk_is_symbol()` to the API would allow user code to distinguish symbols from other strings. But changing `duk_is_string()`, `duk_get_string()`, etc, would make the API a little bit weird: `duk_get_type()` would return `DUK_TYPE_STRING` for symbols but `duk_get_string()` etc would not accept them.

The reason why adding a type tag is not straightforward is that there are a lot of small implications from that. I'll write out some of that in a bit, so you'll get a better idea why it's not as straightforward as one might them.

The main issue comes from using `duk_hstring` also for symbols, but then having a separate type tag for them. That's not an issue when unpacking a symbol from a `duk_tval`. But after that, packing a `duk_hstring` back into a `duk_tval` needs a symbol check to figure out the correct tag. Not using `duk_hstring` for symbols would be very awkward for property code because it passes around the untagged string pointer which would no longer work; it would also mean string table changes because the same byte representation may be used by both a string and a symbol if they're actually distinct types.

Also using a separate API tag for symbols but `duk_hstring` internally for both symbols and strings means that you can no longer `duk_push_string()` arbitrary byte sequences. If they have a symbol representation, that push must either throw or produce a tagged symbol value (so that `duk_is_string()` would be false right after a `duk_push_string()`), neither very ideal.

The cleanest conceptual change would be a separate public tag, a separate internal tag, and a separate internal heap type (say `duk_hsymbol`) for symbols. But the downside of that is a lot more footprint, and awkwardness in the property access code which must genuinely deal with two different key types. When symbols are `duk_hstring`s the property code is almost unaware of symbols; it just needs to take symbols into account when coercing the property key, and it can then use `duk_hstring` as is.

One somewhat odd approach would be to add a `duk_hsymbol` internal type but with the exactly same struct as `duk_hstring`s use. However, their `duk_heaphdr` tag could be different so that (1) they can be used as `duk_hstring`s anywhere, but (2) they can also be distinguished as symbols independent of their byte representation. I'll have to mull over this a bit if it makes sense or not.

So in summary, I like that this pull keeps the Duktape 1.x behavior. I would like for the C API to be more symbol friendly, but what I particularly want to avoid making a hasty API choice without having time to experiment with it. The merit with the approach in this pull is that there's no conceptual change from existing behavior.

svaarala:
Ok, the branch now has the following:

A placeholder for JX/JC serialization so that it's more intuitive (if not clean) in the console. I'm not sure if this should stay or not, and if JX/JC should in general serialize enumerable symbols (doesn't now):

```
duk> Duktape.enc('jx', { [Symbol()]: 'foo' })  // not serialized as property keys now
= "{}"
duk> Duktape.enc('jx', Symbol.for('foo'))  // but is serialized as value
= "{\"_sym\":\"\\x80foo\"}"
duk> Duktape.enc('jx', { prop: Symbol.for('foo') })
= "{prop:{\"_sym\":\"\\x80foo\"}}"
```

This makes it more intuitive for your case:

```
duk> Symbol()
= {"_sym":"\x81\xff0-2\xff"}
duk> Symbol.for('foo')
= {"_sym":"\x80foo"}
```

No "pretty" JX syntax yet. On the upside, you can see the internal representation quickly there :-). JX and JC syntax is the same for now.

`Symbol.iterator` is present:

```
duk> Symbol.iterator
= {"_sym":"\x81Symbol.iterator\xff"}
```

svaarala:
Regarding JX/JC serialization:
- For serialization it's not a big deal to serialize symbols as e.g. `{_sym:"<internal representation>"}`, with `_sym` for JX and `"_sym"` for JC (to keep it JSON compatible).
- For deserialization actually parsing a string representation into a symbol instance is a bit dangerous. In particular, that would allow for creation of hidden symbols using JX parsing.

It might be a bit odd if the JX serialization/deserialization was asymmetric, i.e. symbols wouldn't parse back as symbols but rather as some kind of objects (say, `{_sym:"something"}`). Then again, functions already do that and they JX-serialize as `{_func:true}`. Maybe symbols could behave similar to that in JX.

The other major alternative is that symbols are entirely ignored by JX and JC, just as they are in JSON, because they're not considered "visible" properties and values. The side effect of this is a bit weird behavior in "duk" console, at least for now, because it uses JX for formatting eval results.

fatcerberus:
I wasn't suggesting to go to the trouble of an all-new type tag etc.  I just want an API like `duk_require_string()` that throws if it encounters a symbol because despite the concrete representation being a string, from Ecmascript point of view it's not one and it doesn't make much sense to pass a symbol as e.g. the handle of a map engine entity... :)

If you want to keep the existing string APIs alone that's fine, but then maybe we could have also `duk_require_nonsymbol_string()` or something?

fatcerberus:
For JX, I guess it would be okay to allow it to reconstitute symbols if the sandboxing pull (I don't remember its number) is also merged?

fatcerberus:
`Reflect.ownKeys()` is broken currently: It doesn't include symbols.  duk session:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-1043-g97c3ffe)
duk> var s = Symbol("eaty pig!")
= undefined
duk> s
= {"_sym":"\x81eaty pig!\xff0-1"}
duk> Symbol.iterator
= {"_sym":"\x81Symbol.iterator\xff"}
duk> Symbol.toStringTag
= {"_sym":"\x81Symbol.toStringTag\xff"}
duk> Symbol
= {_func:true}
duk> var o = { [s]: "there's a pig behind you." }
= undefined
duk> o
= {}
duk> o[s]
= "there's a pig behind you."
duk> Reflect.ownKeys(o).length
= 0
duk> 
```

`ownKeys()` is defined as a union of `getOwnPropertyNames()` + `getOwnPropertySymbols()`.

svaarala:
> I wasn't suggesting to go to the trouble of an all-new type tag etc.

That was discussed earlier in this pull so that's why I brought it up.

> I just want an API like duk_require_string() that throws if it encounters a symbol because despite the concrete representation being a string, from Ecmascript point of view it's not one and it doesn't make much sense to pass a symbol as e.g. the handle of a map engine entity... :)

Right, but I think you're not accepting that the C API doesn't recognize symbols and you're trying to apply the Ecmascript type model to the C API. That's logical of course, but the only API model that would *really* work with that is an explicit symbol type in the API. Anything else, at least as far as I've been able to figure, will lead to some inconsistencies.

For example, if `duk_require_string()` throws on a symbol, it means that you can `duk_push_string()` and then say `duk_require_string()` and it will throw, depending on what data you pushed.

It will also mean that `duk_get_type(ctx, -1) == DUK_TYPE_STRING` may still mean `duk_require_string(ctx, -1 )` throws which is highly confusing.

But how do you deal with internal strings in your current calling side code? You're *already* most likely accepting internal strings where ordinary strings are intended?

svaarala:
> If you want to keep the existing string APIs alone that's fine, but then maybe we could have also duk_require_nonsymbol_string() or something?

It's not so much that I want to leave it alone, but to me it seems that changing the current contract (where e.g. `duk_get_type() == DUK_TYPE_STRING` is 1:1 equivalent to `duk_require_string()` not throwing) would be a change for the worse.

But adding a `duk_require_nonsymbol_string()` would be possible, and the internals actually already do that. The potential trouble there is that if the implementation approach is changed, there's API cruft to deal with.

So my main point really is that with Duktape 1.x code you're *already* dealing with internal strings somehow. There's absolutely no change in that respect, only that symbols are treated like internal strings are treated in Duktape 1.x. In other words, C code should just deal with them like any other property keys as far as almost any operation is concerned.

I understand it makes it harder to validate string arguments. But that's *already* the case in Duktape 1.x. And while I agree it's not ideal, this pull isn't making things any worse. I'd particularly like to avoid adding API bindings that get pulled because they are too quickly thought out and baggage going forwards. So any new bindings added due to symbol support need to be carefully thought out so that they're stable against tweaking the representation.

fatcerberus:
True, but before this pull `typeof(internalString)` was still "string" :) (also if one pretends the nonstandard bindings don't exist there's at present no way to create one from Ecmascript code)

I do see what you're getting at with the inconsistencies, so no worries there.  Like I said I'm fine with the existing APIs being kept the way they are, I just want a convenience function that only accepts non-symbol strings.  For Duktape/C functions which back public API entrypoints (like the minisphere API), aligning with the Ecmascript type model makes most sense.

I guess I'll just go implement that validation helper myself for now. :)

svaarala:
> True, but before this pull typeof(internalString) was still "string" :) (also if one pretends the nonstandard bindings don't exist there's at present no way to create one from Ecmascript code)

But that had absolute no impact on C code :-)

svaarala:
> I guess I'll just go implement that validation helper myself for now. :)

I'm OK with adding API bindings for checking symbols conveniently as long as we can be fairly certain they won't be awkward baggage going forward. IOW, if the representation changes, they can be supported via macros without breaking compatibility.

I guess `duk_require_string_notsymbol()` or whatever would be quite easy to support in the end. If symbols get a specific type, that function would then be the same as `duk_require_string()`.

But there's still the overhead of adding API documentation, explaining why they are deprecated etc, so I'd like to keep this set minimal :)

How about adding a `duk_is_symbol()` specifically? That wouldn't be very convenient for API validation purposes though.

fatcerberus:
Regarding the internal strings thing: I guess the difference in my mind is that with internal strings, it was generally safe (ignoring sandboxing concerns) to assume you would never get passed one.  Ecmascript code can now create symbols, so the landscape has changed so to speak.

To answer your question, `duk_is_symbol()` would be fine - I could then write the validation logic on top of that.

svaarala:
Right, but Ecmascript code can create internal strings in Duktape 1.x quite easily by decoding some data (say base64) and coercing it to string. So it's not necessarily very difficult to do by accident.

But I know what you mean: now that "normal" code is supposed to do so, it'll be more common.

My only real concern is that merging this in for 2.0, *and* adding a bunch of API bindings without feeling them out carefully, runs the risk of doing API mistakes that can only be undone in 3.0.

svaarala:
Added a note on test coverage (and fix) for Reflect.keys().

fatcerberus:
> My only real concern is that merging this in for 2.0, and adding a bunch of API bindings without feeling them out carefully, runs the risk of doing API mistakes that can only be undone in 3.0.

That's understandable.

I think `duk_is_symbol()` is a pretty safe bet in this regard - it remains relevant even if (hypothetically) the internal representation of symbols were to change later on.

svaarala:
Also adding a bunch of provisional bindings tagged experimental is an approach used before. But I don't think it will be much less jarring if they're then removed. Experimental API documentation shows up in the API page as a red experimental tag but it should be more prominent to have an actual effect of ensuring that the application writer understands this API may go away.

So that doesn't really seem a very useful approach at least thinking about it right now.

svaarala:
But how about this: I'll leave the API bindings out of this particular pull. We can then see what bindings are really the crucial ones and add them in separate pulls over the next week or so?

fatcerberus:
Alright, no problem here. :)  Since the internal representation of symbols, in particular the initial byte values, is documented I can write a validation function against that for now.

svaarala:
And thanks for test driving this pull, really appreciate it :)

fatcerberus:
> And thanks for test driving this pull, really appreciate it :)

No problem at all.  If you're wondering what my use case was, it was to avoid the need to use a closure for private properties (yes, symbols are visible through reflection, but that's true even in e.g. C#).  I try to put methods and such on `ctor.prototype` whenever I can, but that doesn't work when the methods directly reference closure variables.

svaarala:
That's a common use case I'd imagine. For C code having reliably separated hidden symbols is another that will make application writing a bit easier. It will also isolate Duktape's hidden symbols from tampering. But that's actually provided by #979, not this pull.

svaarala:
Ok, Reflect.ownKeys() should now be fixed.

fatcerberus:
Here's my first foray into using symbols:
https://github.com/fatcerberus/minisphere/blob/master/assets/system/modules/from.js#L85-L116

(Also note that I made good use of `Reflect.construct()` here allowing me to concisely define query operators--this would have been way more verbose with `new`!)

That experiment was successful so I'll see later if there are any other places where I'm using closures for private data that could use symbols instead, this way this pull gets more of a workout before merging.

svaarala:
Minor fix: Object.getOwnPropertyDescriptor() didn't accept a symbol Object as a key.

svaarala:
Removed the JX/JC serialization of symbols for now, and moved it to be considered in #1120.

fatcerberus:
Let me just say though, I find it annoying that one can't do `new Symbol("desc")`.  Function call notation makes me intuitively think that `Symbol("foo")` returns the same value each time, like `Number("812")` does.

Maybe `Symbol()` and `Symbol.for()` could have been reversed.  But alas, that ship has sailed... :)

svaarala:
I find the Symbol API also a bit unintuitive or at least different from other Ecmascript APIs. I think the idea is to generally avoid Object-wrapped symbols so that while `Object(Symbol('foo'))` works, the `new Symbol('foo')` shorthand isn't supported. Otherwise people might use `new Symbol('foo')` when a plain symbol from `Symbol('foo')` would suffice.

svaarala:
Performance tests show no visible impact of the additional Symbol flag checks; in fact property accesses seem a bit faster because the `ToPropertyKey()` now has a fast path (granted, that could also be implemented in master).

svaarala:
@fatcerberus I'll try to do a code read-through and complete the missing documentation tonight (tomorrow at latest). This would then be merge ready, unless some blocker pops up. (If a blocker popped up later, it's still possible to revert in master of course.)

fatcerberus:
By the way, if `duk_require_string()` etc. ultimately retain their current behavior of accepting symbols, then that's an important thing to mention in the migration notes I'd think.  Probably a lot of Duktape 1.x code will be using `duk_require_string()` or `duk_is_string()` for validation purposes, and that code, if not changed, will now accept symbols as well.

svaarala:
Agree, the description could be something along the lines of:
- Symbols behave like internal strings in Duktape 1.x for the C API: they are strings with a specific invalid CESU-8/UTF-8 representation.
- Their type is `DUK_TYPE_STRING`, they test true for `duk_is_string()`, are accepted by `duk_require_string()`, etc.
- Symbols have non-string behaviors for Ecmascript code, like internal strings did in Duktape 1.x.
- However, you're more likely to encounter symbols in C code *if* you use Symbols in your Ecmascript code base.

Previously you would encounter internal keys if you explicitly created them, or passed them in from sloppy C code - this might have happened e.g. if you passed a ISO-8859-1 string to `duk_push_string()` as is.

svaarala:
If the C API behavior is ultimately not good enough one approach would be to skip merging this for 2.0.0 release and try to figure out a compatible way to merge this for 2.1.0.

Note that I'm still toying around with C API alternatives. My main concern is finding something that's likely to be workable for a while. The change in this pull has the merit of being patterned after 1.x so there's no real change in the semantics, only a "renaming of concept". Code *not* using symbols wouldn't be impacted at all essentially.

Anyway, one approach that seems doable (though I'm not overly excited about it) is:
- Add a symbol type to the public API, i.e. `DUK_TYPE_SYMBOL`, `duk_is_symbol()`, etc. Symbols would be `const char *` for C code, exposing the internal representation.
- *Don't* add an internal type tag for `duk_tval`, so that internally symbols are `DUK_TAG_STRING` and `duk_hstring` heap type. (Btw, note that C and Ecmascript types are not 1:1 even disregarding symbols: plain buffers and lightfuncs have a specific C type but they appear as virtual ArrayBuffer and Function objects in the Ecmascript side.)
- For example `duk_is_string()` would check for internal tag `DUK_TAG_STRING`, look up the `duk_hstring *`, and then check for the symbol flag `DUK_HSTRING_FLAG_SYMBOL` before returning.
- Internal call sites using `duk_is_string()`, `duk_get_type()` etc need checking. There are maybe ~50-100 of them so it's some work but still doable. If a call site accepts both strings and symbols, they would then use an internal helper like `duk_is_string_or_symbol()`.
- Most internal call sites switch based on the internal tag, so the Symbol semantics implementations in this pull could remain as is (after checking the call sites of changed public string API calls).

This would be a bit more natural and could keep the current internal mechanics, at least for now. The upside would be that the C and Ecmascript API would be in sync type-wise, and it'd be more likely that internals could be reworked (if needed) without needing to touch the API right away.

But there are still some open questions for this approach. For example, what should `duk_push_string()` do if the string has an internal representation? Perhaps it should throw to prevent a symbol from being created by accident. This would make sense for e.g. sandboxed code, because one wants to avoid accidentally creating internal keys (at least if #979 is not merged).

However, this would break code dealing e.g. with ISO-8859-1 characters or arbitrary byte strings -- code that used to work in 1.x (although quite oddly for the script side).

If symbols are rejected from `duk_push_string()` one would then need some kind of `duk_push_symbol()`. Going that route, it'd then actually be natural to hide the internal representation entirely from C code so that symbols would not be `const char *` but one would push specific symbol kinds (giving their description, and possibly a unique suffix that might be autogenerated), query symbol types, extract symbol descriptions, etc. There's a lot of potential to go overboard here, just to support a single type which isn't all that central to much practical code.

So... concrete, fully fledged API proposals welcome :-)

svaarala:
This API / typing discussions reminds me that it'd be useful to have a detailed type table that provided both the internal and external type tags, related internal structures, etc for all types (including object classes). The Guide has an API oriented "stack types" table but that's not complete enough to work with internals.

fatcerberus:
I'll think some on the C API over the next few days, but since I can never keep quiet :)

> Btw, note that C and Ecmascript types are not 1:1 even disregarding symbols: plain buffers and lightfuncs have a specific C type but they appear as virtual ArrayBuffer and Function objects in the Ecmascript side.

This is true, but the mentioned cases also generally don't affect validation ;-)  In nearly all cases if I accept an ArrayBuffer on the C side, a plain buffer works just as well, and a lightfunc is likewise just as good as a full-fledged function.  It only *sometimes* makes sense to allow symbols in place of strings.  For example if a symbol is passed in place of a string which is destined for display, the Ecmascript ToString() throwing behavior is then preferred.

fatcerberus:
By the way, I think a big part of the reason I see the C/Ecmascript barrier so much differently is that, for minisphere, Duktape is my *frontend*.  So there's a very fine line between C and Ecmascript as control flows seamlessly between the two domains constantly just as part of normal operation.  Thus, the more difficult it is to reconcile the differences in typing between the two sides, the more of a source of pain that becomes for me as an embedder.

svaarala:
> This is true, but the mentioned cases also generally don't affect validation ;-)

The situation is indeed reversed in the sense that multiple C types have a single Ecmascript type in the lightfunc/buffer case.

It does affect validation, just in a different way: you're liable to reject a valid type by accident. For example, if you check for a lightfunc, you won't accept a function. Or if you check for `duk_is_buffer()` you won't accept a buffer object.

The function case works out because `duk_is_function()` returns true for both; there's no type tag for functions because they're a subclass of objects anyway, so there's no direct clash at the type tag level.

The buffer case somewhat works out if you use `duk_require_buffer_data()` for example, because that also accepts plain buffers. But there is no "is" checker for that case.

svaarala:
> By the way, I think a big part of the reason I see the C/Ecmascript barrier so much differently is that, for minisphere, Duktape is my frontend. So there's a very fine line between C and Ecmascript as control flows seamlessly between the two domains constantly just as part of normal operation. Thus, the more difficult it is to reconcile the differences in typing between the two sides, the more of a source of pain that becomes for me as an embedder.

I can understand that - and the typing differences between the C API and the Ecmascript side are obviously not a goal.

In fact, if the public type and the internal type will become more decoupled, it might be useful to make lightfuncs and plain buffers behave like their fully fledged counterparts in the C API too (besides pushing). So, they'd have `duk_is_object()` true for example.

The downside of this is that the C API philosophy would then be different: it'd mimic the Ecmascript side more closely, and hide internals that you might sometimes be interested in. Perhaps introspection can then replace some of those use cases, e.g. reveal if a certain value is a lightfunc.

fatcerberus:
@svaarala No problems so far with the Symbol support by the way - it's been working great.

fatcerberus:
The thing I've found I really like about symbols is that any symbol created with `Symbol()` is guaranteed to be unique within its scope.  Thus you never have to worry about someone accidentally overwriting them by accident.

You can also use them for tagging arbitrary objects and likewise not worry about clobbering something on an object you don't own... which now that I think of it could be used to make a WeakMap polyfill... hmm...

svaarala:
I'd advise against trying to create a weak map without the engine supporting weak references directly. It's quite hard to get right.

fatcerberus:
ES6 WeakMap is interesting in that no reference to the key is created at all, weak or otherwise.  That's why you can't enumerate it.  It actually just creates a *strong* reference of key -> value, which can be done in pure JavaScript.  The only thing having built-in engine support buys you is the ability to use non-extensible objects as keys.

But anyway, that's unrelated to this pull, just something I thought of is all :)

svaarala:
I'm not sure what you mean: WeakMap keys are objects that must be tracked and be weakly referenced so that the key (and the value) can be collected?

fatcerberus:
WeakMap keys are not enumerable, so the ES6 requirements can be met simply by storing the value on the key object itself.  This can be polyfilled easily, so long as the key is extensible.

svaarala:
Hmm. So while a single key (object) may be used in multiple WeakMaps, you would just use a separate symbol for tracking the value related to each WeakMap?

But what if the WeakMap itself becomes unreachable? How do you remove the key/values related to that WeakMap stored in the objects which are *still* reachable?

fatcerberus:
In other words: `WeakMap.set(key, value)` is semantically equivalent to `key.uniqueProp = value` if `uniqueProp` were invisible to outside code.

svaarala:
I still don't understand how you remove all the `uniqueProp` properties from objects in the heap if I say `weakMap = null` and that was the last reference to the WeakMap itself.

svaarala:
You could keep a list of objects in the WeakMap for that purpose - but that would create a strong reference and prevent collection.

fatcerberus:
No reference is held to the WeakMap itself - `uniqueProp` is directly assigned the mapped value so that WeakMap.get(key) just means `key.uniqueProp`.

svaarala:
I know but consider this:

```js
var buf = new ArrayBuffer(8);
var obj = {};
var wm = new WeakMap();  // allocates a symbol S
wm.set(obj, buf);  // sets obj[S] = buf
wm = null;  // remove reference to the *WeakMap* itself
buf = null;
```

On that last line, the *only* strong reference left is `obj` and `buf` should be collectable.

However, who removes the `obj[S] = buf` reference? If it's not removed, `buf` will remain reachable as long as `obj` itself exists - which may be indefinitely.

fatcerberus:
Hm, it may prevent the value from being collected if the key outlives it though.  I'd have to check ES6 to see what it says about that case.

svaarala:
I'm quite sure it should be collected - that's the whole point of weak references in a reference graph.

fatcerberus:
Well, I'll stop derailing the symbol pull discussion now and we can maybe resume this discussion sometime later.

svaarala:
To get the 2.0.0 release finalized the easiest approach (with least rework down the line) would be:
- Merge this for 2.0.0 after rebasing
- No API changes yet, which has the downsides discussed above and they need some documentation
- Compatible API work for 2.1.0 only (there's lot of that left over, like filling in built-in bindings, etc)
- Better symbol support for 3.0.0 with maybe symbols exposed in the API

As discussed above this doesn't really change things much for the application unless you want to use symbols specifically (if you do, then you run into some validation use cases which you brought up). If you don't, the important change is that if you push non-CESU-8/non-UTF-8 strings from the C API they may come out as symbols for script code. But in the end I think there are relatively few applications that really do that and rely on specific behavior, and the updated documentation can discourage doing so for compatibility with changes later on.

fatcerberus:
> As discussed above this doesn't really change things much for the application unless you want to use symbols specifically

Not _entirely_ true. :)  As we *also* discussed above, symbols will be accepted by `duk_require_string()`, which application writers have to be aware of, particularly when writing Duktape/C bindings.

svaarala:
> Not entirely true. :) As we also discussed above, symbols will be accepted by duk_require_string(), which application writers have to be aware of, particularly when writing Duktape/C bindings.

... But I specifically said "unless you want to use symbols specifically"?

svaarala:
In other words, if you *do not* use symbols, this pull has minimal changes to the way you use the C API. Internal strings are already present, and already accepted by e.g. `duk_require_string()`.

fatcerberus:
If Duktape is your front end (like in my case) it could be an issue since your *end users* can now use symbols. :)

svaarala:
Yes, but that would be the case "you are using symbols", would it not, from Duktape feature point of view?

svaarala:
And if you don't like them doing so, `-UDUK_USE_ES6_SYMBOL`, and they won'.t

fatcerberus:
The feature is enabled by default, those upgrading to Duktape 2.0.0 may not realize it's even there.  I don't object to the semantics, just worth a strong mention in the migration notes is all I'm getting at.

svaarala:
Haven't I said several times that I agree with it needs to be documented? :)

svaarala:
Disabling the feature by default would also be quite OK.

fatcerberus:
Sorry, just want to make sure we don't needlessly surprise anyone.  Can you tell I'm an API designer yet? ;)

fatcerberus:
Anyway, now that that horse is thoroughly beaten to death, glad to hear this is being merged soon, I'm greatly looking forward to playing with this. :)

svaarala:
That merge decision was really my point: either this needs to be merged without C API changes (i.e. symbols behave like internal strings now) or this needs to be postponed to 3.0.0 due to C API changes. I don't see other alternatives; I really want to finalize a 2.0.0 release quite soon. I'm not sure which approach is better unfortunately.

If this is merged, it's easy enough to disable if you don't want the feature. But, the merge obviously creates some technical debt to maybe update the C API later, and will at least require documentation writing in the near future. C code that really wants to use symbols needs to be touched in the 3.0.0 update once more.

It this is not merged, the branch will bit rot a bit before 3.0.0 release (this is probably still quite workable), but there's no option to use symbols until then unless one builds from the branch. On the other hand, no user code will need upgrading because none will exist.

I don't see symbols as a critical feature so I'm fine with either approach as such.

fatcerberus:
Ah, I see.  I'll think about this some and see if I can provide any input.

svaarala:
I have some backlog for improving some of the internal data structures which are becoming relevant for footprint or performance -- but so far they don't really affect this merge decision that much. I think the API issue is the most important one, i.e. doing two major steps on the feature vs. postponing the feature but doing it in one step.

The +700 byte footprint impact (when the Symbol built-ins are disabled) is somewhat relevant for low memory targets and may need some follow-up, but it's a minor detail overall.

svaarala:
To elaborate a bit, internal data structure changes are a bit difficult to plan ahead because they often affect every area of the code in some way. It's also quite difficult to see beforehand what data structure changes will be necessary as types are added (say, what will WeakMap mean?).

For example, one of the simplification steps not yet made for buffers is that views could back to ArrayBuffers directly (now they back to a plain buffer and point to an ArrayBuffer via a .buffer property). This would probably mean that the `duk_hbufobj` structure would be repurposed as a "buffer view" structure, while plain buffers and ArrayBuffers would maybe share some common structure (with plain buffers missing a property table).

This has knock-on effects on the public API; in particular, the tagged buffer type would be removed because plain buffers would behave like ArrayBuffers there too. The API would then need some calls to figure out if a value is an ArrayBuffer or one of the view types, etc. Also some of the non-standard ArrayBuffer virtual properties would be affected. ArrayBuffers would lose the ability the have a view offset (which is never possible in the standard obviously). So there's at least a lot of documentation changes involved in the change.

Similarly, if the symbol internal data structure is adjusted down the road, it will have an effect on what API is clean but still aligned with the internals. It feels risky to make API commitments when the internal representation is maybe still going to change; the API commitments made might then be off-sync with the internal representation which causes a lot of trivia work down the line.

So that's why I'm more inclined to either merge this without API changes or postpone the merge altogether until there's enough time to experiment with the representation alternatives and what they mean for the C API.

fatcerberus:
Here's my proposal: Merge this for 2.0.0 without the API changes, but put it behind a `DUK_USE_SYMBOL_SUPPORT` define (this is already present) and disabled by default.  This way the feature is kind of experimental, and users have to read the documentation (along with any caveats) before enabling it.  If it's also documented as experimental, that gives a bit more leeway to change this later if needed.

fatcerberus:
It's okay I think to disable some ES6 features by default because Duktape, even in master, is still "ES5-ish with some ES6 features".  If Duktape were a full ES6 engine then it'd be strange to disable it, but that's not the case yet.

svaarala:
That approach works for me. There's one detail though: even with the Symbol support there is a small impact on the C API because internal properties (even with 0xFF) prefix will behave as symbols for script code. While that's not likely to cause issues, it does interfere with e.g. string coercion of internal strings. So the feature cannot be fully disabled.

It'd be possible to change that so that with the symbols support disabled `typeof internalString` would be `string` as before (and not `symbol`) but that seems a bit of wasted work.

fatcerberus:
Actually it probably makes sense to leave the `typeof` for internal strings as `symbol` even with support disabled since they behave differently than other strings.  Especially if #979 gets merged too.

svaarala:
Right, but I meant that as a shorthand for them getting all the symbol behavior. In particular, String(x) is a TypeError.

fatcerberus:
I guess that's acceptable.  It's really only an issue if C code pushes such strings, since Ecmascript code can no longer create them by default (?).

By the way I thought about it some more, and the `duk_require_string()` behavior does make sense if one views the contract as "require something which is compatible with a C string" instead of specifically "require an Ecmascript string".  So I'm willing to concede on that point. :)

PR #983, opened 2016-09-30
Remove internal support for codepoints above U+10FFFF

svaarala:
Duktape currently uses codepoints up to 32 bits for RegExp bytecode, but if that dependency is removed, supporting codepoints only up to U+10FFFF (4 byte sequences) would be enough. This pull is to see what the overall changes are:

- [ ] Add a different variable integer encoding for RegExp bytecode; it can be simpler than UTF-8 because it doesn't have the same decoding resynchronization etc goals.
- [ ] Remove extended codepoint support from duk_unicode_support.c. CESU-8 needs to remain, and since UTF-8 4-byte sequences technically encode up to U+1FFFFF maybe keep them rather than rejecting?
- [ ] Documentation updates

It would be nice if this change would make it easier to support strict UTF-8 encoding and decoding, and conversions between CESU-8 and UTF-8 needed in #975, with little code overlap.

### DIFF
--- a/src-input/duk_regexp.h
+++ b/src-input/duk_regexp.h
@@ -2,6 +2,8 @@
  *  Regular expression structs, constants, and bytecode defines.
  */
 
+/* FIXME: avoid extended UTF-8 for regexps. */
+
 #ifndef DUK_REGEXP_H_INCLUDED
 #define DUK_REGEXP_H_INCLUDED
 
--- a/src-input/duk_unicode_support.c
+++ b/src-input/duk_unicode_support.c
@@ -30,6 +30,8 @@ DUK_INTERNAL const duk_int8_t duk_is_idchar_tab[128] = {
  *  XUTF-8 and CESU-8 encoding/decoding
  */
 
+/* FIXME: remove extended codepoint support but keep CESU-8 support. */
+
 DUK_INTERNAL duk_small_int_t duk_unicode_get_xutf8_length(duk_ucodepoint_t cp) {
 	duk_uint_fast32_t x = (duk_uint_fast32_t) cp;
 	if (x < 0x80UL) {


svaarala:
While Unicode is not designed to encode arbitrary data, Duktape now uses the extended UTF-8 encodings to allow all 32-bit codepoints to be represented. It's used internally for RegExp bytecode.

Also user code can in essence encode arbitrary 32-bit codepoint sequences into extended UTF-8 and access them as Ecmascript string characters in Ecmascript code. I'm not sure if that's used in practice by applications, but my guess is it isn't because iterating codepoints in Ecmascript code is not terribly convenient, as you need to e.g. `x.charCodeAt(i)` for each offset. Anyway the capability is there at present and would be removed by this pull.

fatcerberus:
`.charCodeAt()` is also inconvenient for another reason: For standard strings, supplemental characters are represented as a surrogate pair.  Which is why ES6 introduced `.codePointAt()`.

svaarala:
Right, but when you use that extended UTF-8 format, .charCodeAt() will return the 32-bit codepoints as is with no surrogate handling. That's why it's quite convenient if you need a compact 32-bit sequence, and is also the reason I used it for the RegExp bytecode.

svaarala:
This is still quite incomplete. I'll try to complete this pull next, as it would simplify the public API conceptually: strings would be UTF-8 with CESU-8 additions to conform to Ecmascript requirements. Unnecessary decoding support for higher codepoints can be removed, and at the same a table driven approach can be added to improve performance. The only consumer inside Duktape was RegExp bytecode and applications are (most likely) not using the extended codepoint support (beyond UTF-8) very much.

fatcerberus:
Would regexp support benefit in either performance or footprint by using a more purpose-built approach?

svaarala:
Maybe a little bit, but that should be pretty minor. Right now the pull simply encodes regexp arguments as 32-bit values with no shorter encoding but for practical regexps it might be useful to have a few variable length encodings, so it becomes similar to UTF-8 (although it can be a bit simpler to decode).

svaarala:
What would benefit RegExp performance more would be the ability to rely on the bytecode being correct during execution. Right now as regexp bytecode is stored as an internal property (and thus mutable from Ecmascript code) the executor needs to check every access to guarantee memory safe behavior. But if #979 and symbols are merged, the regexp bytecode internal property would be safe from script modification and the executor could just use asserts and otherwise avoid bounds checking everywhere.

Issue #271, opened 2015-08-10
Feature request: Support ES6 Object Property Enhancements


kphillisjr:
I figured I would create a few Issue statements covering some of the new Features in ECMAScript in a hope that it will make the changes a little easier to handle.  All of this is copied (with slight modifications ) from the "[ECMAScript 6](http://es6-features.org/): New Features: Overview & Comparison" web site ( they are also on [github](https://github.com/lukehoban/es6features/blob/master/README.md)). That said, I will go ahead and get right to the new features...
- [ ] **Property Definition Short Hand**
 ```JavaScript
// New Method:
obj_new_method = { x, y };
// Old method: ECMAScript 5
obj = { x: x, y: y };
```

- [ ] **Simplified Function Definition**
 ```JavaScript
// New Method:
obj_simplified = {
    foo (a, b) { … },
    bar (x, y) { … }
};
// Old Method:
obj = {
    foo: function (a, b) { … },
    bar: function (x, y) { … }
};
```
- [ ] **Computed Property Names** 
 ```JavaScript
// New Method:
obj_new_Method = { foo: "bar", [ "prop_" + foo() ]: 42 };
// Old Method:
obj = { foo: "bar" }; obj[ "prop_" + foo() ] = 42;
```

- [ ] **Advanced Function Definition** - I Think this can be put on hold for now.
 ```JavaScript
// New Method:
obj_simplified = {
    foo (a, b) { … },
    *quux (x, y) { … }
};
// Old Method:
obj = {
    foo: function (a, b) { … }
    /* quux has no equivalent. */
};
```

svaarala:
I'll be doing an ES6 gap analysis later this autumn to get a full list of changes needed. The main thing implementation-wise is to figure out what internal structure needs rework (non-incremental change) to get to ES6. Smaller feature work can then be incremental.

svaarala:
Yeah, there's a need for two kinds of issues: (1) to outline the internal changes and (2) to describe the external changes. Most users are probably interested in specific features in (2), while issues in (1) are preconditions to the features getting implemented. Some overlap between issues is not a problem.

It's probably good to organize the external features by topics like you've done here.

PR #985, opened 2016-10-02
Add ES6 computed property name support

svaarala:
Add support for ES6 computed property names like:

```js
var obj = { [1+2]: 'three' };
print(obj[3]);  // three
```

This syntax is quite useful with symbols because one can then:

```js
var obj = {
    foo: 'normal property',
    [ Symbol.for('mySymbol') ]: 'symbol property'
};
```

Implements part of #271.

Tasks:
- [x] Compiler change
- [ ] Config option?
- [x] Testcase
- [ ] Website ES6 section
- [ ] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1800,6 +1800,9 @@ Planned
   won't backtrack on e.g. an invalid hex escape and treat it literally
   (GH-926)
 
+* Add support for ES6 computed property names in object literals, e.g.
+  "{ [1+2]: 'three' }" (GH-985)
+
 * Remove no longer needed platform wrappers in duk_config.h: DUK_ABORT(),
   DUK_EXIT(), DUK_PRINTF(), DUK_FPRINTF(), DUK_FOPEN(), DUK_FCLOSE(),
   DUK_FREAD(), DUK_FWRITE(), DUK_FSEEK(), DUK_FTELL(), DUK_FFLUSH(),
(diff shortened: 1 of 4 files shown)


fatcerberus:
I've been hoping for these for a while, good to see Duktape 2.x will get some ES6 language features after all.  Although I am surprised this could be done without redoing the parser/compiler.

svaarala:
In general RHS expressions are pretty straightforward, the main issues with ES6 parsing are for LHS expressions like destructuring.

fatcerberus:
It surprises me how little new code was needed - and the diff is confined entirely to the compiler even.  Somehow no executor changes were required at all?  I'm in amazement.

fatcerberus:
While we're on the topic of (potentially) easy-to-implement language features, ES7 was finalized in June and includes a new exponentiation operator (`**`):
https://tc39.github.io/ecma262/2016/#sec-exp-operator

How difficult would it be to update the parser and compiler to support that operator?  `x ** y` is so much more readable and concise than `Math.pow(x, y)` :-)

Actually it seems there are only two new features in ES7 compared to ES6: The aforementioned exponentiation operator, and `Array.prototype.includes()`.

svaarala:
Exponentation would be quite easy, with some bytecode opcode changes.

But I'll try to focus on closing open pulls, I just implemented this because it complements the symbol feature so that literals with symbols are possible.

fatcerberus:
Fair enough.  I might look into doing it myself if I get sufficiently motivated. :)

PR #987, opened 2016-10-02
Implement ES7 exponentiation operator (`**`)

fatcerberus:
Ecmascript 7 introduces a new right-associative exponentiation operator, which avoids the need for a function call (`Math.pow()`) as in previous Ecmascript versions.

See: http://www.ecma-international.org/ecma-262/7.0/index.html#sec-exp-operator

Providing this operator has concrete benefits for Ecmascript code:

* The compiler can inline expressions like `2 ** 16` at compile time, avoiding a `Math.pow()` call at runtime when both operands are constant.

* Calculating powers on the fly is more performant as you save the cost of up to two property lookups (`Math` -> `Math.pow`) and a function call for every operation.

Usage:
```js
var z = x ** y;  // exponentiation
z **= 2;         // exponentiation assignment
```

Checklist:
- [x] Implement support for exponentiation operator (`**`)
- [x] Implement support for exponentiation assignment (`**=`)
- [x] Add new opcodes to `duk_opcodes.yaml`
- [x] Add constant folding support for new operator, e.g. `2 ** 16` would compile down to `65536`
- [x] Implement special cases to align with ES7 specification
- [x] Add regression test(s) to ensure full compliance
- [x] RELEASES entry

### DIFF
--- a/src-input/duk_js_executor.c
+++ b/src-input/duk_js_executor.c
@@ -84,6 +84,10 @@ DUK_LOCAL DUK__INLINE_PERF duk_double_t duk__compute_mod(duk_double_t d1, duk_do
 	return (duk_double_t) DUK_FMOD((double) d1, (double) d2);
 }
 
+DUK_LOCAL DUK__INLINE_PERF duk_double_t duk__compute_exp(duk_double_t d1, duk_double_t d2) {
+	return (duk_double_t) duk_js_arith_pow((double) d1, (double) d2);
+}
+
 DUK_LOCAL DUK__INLINE_PERF void duk__vm_arith_add(duk_hthread *thr, duk_tval *tv_x, duk_tval *tv_y, duk_small_uint_fast_t idx_z) {
 	/*
 	 *  Addition operator is different from other arithmetic
@@ -319,6 +323,10 @@ DUK_LOCAL DUK__INLINE_PERF void duk__vm_arith_binary_op(duk_hthread *thr, duk_tv
 	}
 	case DUK_OP_MOD >> 2: {
 		du.d = duk__compute_mod(d1, d2);
+		break;
+	}
+	case DUK_OP_EXP >> 2: {
+		du.d = duk__compute_exp(d1, d2);
 		break;
 	}
 	default: {
@@ -3143,7 +3151,11 @@ DUK_LOCAL DUK_NOINLINE void duk__js_execute_bytecode_inner(duk_hthread *entry_th
 		case DUK_OP_MOD_RR:
 		case DUK_OP_MOD_CR:
 		case DUK_OP_MOD_RC:
-		case DUK_OP_MOD_CC: {
+		case DUK_OP_MOD_CC:
+		case DUK_OP_EXP_RR:
+		case DUK_OP_EXP_CR:
+		case DUK_OP_EXP_RC:
+		case DUK_OP_EXP_CC: {
 			/* XXX: could leave value on stack top and goto replace_top_a; */
 			duk__vm_arith_binary_op(thr, DUK__REGCONSTP_B(ins), DUK__REGCONSTP_C(ins), DUK_DEC_A(ins), op);
 			break;
@@ -3211,6 +3223,22 @@ DUK_LOCAL DUK_NOINLINE void duk__js_execute_bytecode_inner(duk_hthread *entry_th
 		}
 		case DUK_OP_MOD_CC: {
 			duk__vm_arith_binary_op(thr, DUK__CONSTP_B(ins), DUK__CONSTP_C(ins), DUK_DEC_A(ins), DUK_OP_MOD);
+			break;
+		}
+		case DUK_OP_EXP_RR: {
+			duk__vm_arith_binary_op(thr, DUK__REGP_B(ins), DUK__REGP_C(ins), DUK_DEC_A(ins), DUK_OP_EXP);
+			break;
+		}
+		case DUK_OP_EXP_CR: {
+			duk__vm_arith_binary_op(thr, DUK__CONSTP_B(ins), DUK__REGP_C(ins), DUK_DEC_A(ins), DUK_OP_EXP);
+			break;
+		}
+		case DUK_OP_EXP_RC: {
+			duk__vm_arith_binary_op(thr, DUK__REGP_B(ins), DUK__CONSTP_C(ins), DUK_DEC_A(ins), DUK_OP_EXP);
+			break;
+		}
+		case DUK_OP_EXP_CC: {
+			duk__vm_arith_binary_op(thr, DUK__CONSTP_B(ins), DUK__CONSTP_C(ins), DUK_DEC_A(ins), DUK_OP_EXP);
 			break;
 		}
 #endif  /* DUK_USE_EXEC_PREFER_SIZE */
@@ -4895,8 +4923,8 @@ DUK_LOCAL DUK_NOINLINE void duk__js_execute_bytecode_inner(duk_hthread *entry_th
 		}
 
 #if !defined(DUK_USE_EXEC_PREFER_SIZE)
-		case DUK_OP_UNUSED194:
-		case DUK_OP_UNUSED195:
+		case DUK_OP_UNUSED190:
+		case DUK_OP_UNUSED191:
 		case DUK_OP_UNUSED196:
 		case DUK_OP_UNUSED197:
 		case DUK_OP_UNUSED198:
--- a/tools/configure.py
+++ b/tools/configure.py
@@ -475,6 +475,7 @@ def main():
         'duk_internal.h',
         'duk_jmpbuf.h',
         'duk_exception.h',
+        'duk_js_arith.c',
         'duk_js_bytecode.h',
         'duk_js_call.c',
         'duk_js_compiler.c',
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1809,6 +1809,10 @@ Planned
 * Add support for ES6 computed property names in object literals, e.g.
   "{ [1+2]: 'three' }" (GH-985)
 
+* Add support for ES7 exponentiation and exponentiation assignment operators,
+  e.g. "2 ** 10" evaluates to 1024, avoiding the cost of an Ecmascript call to
+  Math.pow() while also being more readable (GH-987)
+
 * Remove no longer needed platform wrappers in duk_config.h: DUK_ABORT(),
   DUK_EXIT(), DUK_PRINTF(), DUK_FPRINTF(), DUK_FOPEN(), DUK_FCLOSE(),
   DUK_FREAD(), DUK_FWRITE(), DUK_FSEEK(), DUK_FTELL(), DUK_FFLUSH(),
--- a/debugger/duk_opcodes.yaml
+++ b/debugger/duk_opcodes.yaml
@@ -915,10 +915,26 @@ opcodes:
       - ABC_I
   - name: UNUSED190
   - name: UNUSED191
-  - name: UNUSED192
-  - name: UNUSED193
-  - name: UNUSED194
-  - name: UNUSED195
+  - name: EXP_RR
+    args:
+      - A_R
+      - B_R
+      - C_R
+  - name: EXP_CR
+    args:
+      - A_R
+      - B_C
+      - C_R
+  - name: EXP_RC
+    args:
+      - A_R
+      - B_R
+      - C_C
+  - name: EXP_CC
+    args:
+      - A_R
+      - B_C
+      - C_C
   - name: UNUSED196
   - name: UNUSED197
   - name: UNUSED198
--- a/src-input/duk_bi_math.c
+++ b/src-input/duk_bi_math.c
@@ -127,70 +127,6 @@ DUK_LOCAL double duk__round_fixed(double x) {
 	return DUK_FLOOR(x + 0.5);
 }
 
-DUK_LOCAL double duk__pow_fixed(double x, double y) {
-	/* The ANSI C pow() semantics differ from Ecmascript.
-	 *
-	 * E.g. when x==1 and y is +/- infinite, the Ecmascript required
-	 * result is NaN, while at least Linux pow() returns 1.
-	 */
-
-	duk_small_int_t cx, cy, sx;
-
-	DUK_UNREF(cx);
-	DUK_UNREF(sx);
-	cy = (duk_small_int_t) DUK_FPCLASSIFY(y);
-
-	if (cy == DUK_FP_NAN) {
-		goto ret_nan;
-	}
-	if (DUK_FABS(x) == 1.0 && cy == DUK_FP_INFINITE) {
-		goto ret_nan;
-	}
-#if defined(DUK_USE_POW_NETBSD_WORKAROUND)
-	/* See test-bug-netbsd-math-pow.js: NetBSD 6.0 on x86 (at least) does not
-	 * correctly handle some cases where x=+/-0.  Specific fixes to these
-	 * here.
-	 */
-	cx = (duk_small_int_t) DUK_FPCLASSIFY(x);
-	if (cx == DUK_FP_ZERO && y < 0.0) {
-		sx = (duk_small_int_t) DUK_SIGNBIT(x);
-		if (sx == 0) {
-			/* Math.pow(+0,y) should be Infinity when y<0.  NetBSD pow()
-			 * returns -Infinity instead when y is <0 and finite.  The
-			 * if-clause also catches y == -Infinity (which works even
-			 * without the fix).
-			 */
-			return DUK_DOUBLE_INFINITY;
-		} else {
-			/* Math.pow(-0,y) where y<0 should be:
-			 *   - -Infinity if y<0 and an odd integer
-			 *   - Infinity otherwise
-			 * NetBSD pow() returns -Infinity for all finite y<0.  The
-			 * if-clause also catches y == -Infinity (which works even
-			 * without the fix).
-			 */
-
-			/* fmod() return value has same sign as input (negative) so
-			 * the result here will be in the range ]-2,0], 1 indicates
-			 * odd.  If x is -Infinity, NaN is returned and the odd check
-			 * always concludes "not odd" which results in desired outcome.
-			 */
-			double tmp = DUK_FMOD(y, 2);
-			if (tmp == -1.0) {
-				return -DUK_DOUBLE_INFINITY;
-			} else {
-				/* Not odd, or y == -Infinity */
-				return DUK_DOUBLE_INFINITY;
-			}
-		}
-	}
-#endif
-	return DUK_POW(x, y);
-
- ret_nan:
-	return DUK_DOUBLE_NAN;
-}
-
 /* Wrappers for calling standard math library methods.  These may be required
  * on platforms where one or more of the math built-ins are defined as macros
  * or inline functions and are thus not suitable to be used as function pointers.
@@ -274,10 +210,10 @@ DUK_LOCAL const duk__one_arg_func duk__one_arg_funcs[] = {
 DUK_LOCAL const duk__two_arg_func duk__two_arg_funcs[] = {
 #if defined(DUK_USE_AVOID_PLATFORM_FUNCPTRS)
 	duk__atan2,
-	duk__pow_fixed
+	duk_js_arith_pow
 #else
 	DUK_ATAN2,
-	duk__pow_fixed
+	duk_js_arith_pow
 #endif
 };
 
--- a/src-input/duk_js.h
+++ b/src-input/duk_js.h
@@ -41,6 +41,9 @@ DUK_INTERNAL_DECL duk_bool_t duk_js_instanceof(duk_hthread *thr, duk_tval *tv_x,
 DUK_INTERNAL_DECL duk_bool_t duk_js_in(duk_hthread *thr, duk_tval *tv_x, duk_tval *tv_y);
 DUK_INTERNAL_DECL duk_small_uint_t duk_js_typeof_stridx(duk_tval *tv_x);
 
+/* arithmetic */
+DUK_INTERNAL_DECL double duk_js_arith_pow(double x, double y);
+
 #define duk_js_equals(thr,tv_x,tv_y) \
 	duk_js_equals_helper((thr), (tv_x), (tv_y), 0)
 #define duk_js_strict_equals(tv_x,tv_y) \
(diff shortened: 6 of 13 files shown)


fatcerberus:
There's currently a bug in this pull and I can't pinpoint the cause: If one evaluates, e.g. `2 ** 10`, the result is a NaN.  Running under debugger control, it appears the second operand (10 in this case) is coming out as NaN by the time it reaches the executor.

svaarala:
One useful feature would be adding `**` support to the compiler constant folding (it happens when processing ivalues) so that e.g. `2 ** 16` would be evaluted during compile time. That would be an actual benefit for e.g. expressing constants.

fatcerberus:
Good idea.  I have to figure out the cause of that NaN bug first though. :-/  Any ideas?

fatcerberus:
My hunch is that the lexer treats `**` as two tokens, but the binary operand parsing code only expects one, so that one of the asterisks is being interpreted as an operand.

svaarala:
I can take a look later tonight. From the diff it seems you added the necessary lexer lookahead to detect the `**` token and that looked correct to me. Debug dumping the token stream would make sure of that relatively quickly.

fatcerberus:
I realize now that I will also have to add a new binding power, since I believe `**` has higher precendence than other multiplicative operators (as would be intuitive - PEMDAS).

fatcerberus:
Scratch that: Exponentiation is a MultiplicativeExpression with the same precedence as e.g. multiplication.  Kind of strange, but them's the breaks.

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

fatcerberus:
Alright, I'm home now.  Back to bug hunting :)

fatcerberus:
Okay, so parsing is fine.  The `**` token is lexed correctly and parsed as a MultiplicativeOperator.  My next step was to set a breakpoint in `duk__vm_arith_binary_op()` and trace the stack backwards from there - lo and behold, the expression `2 ** 10` is compiled as `DUK_OP_EXP_CR`.  Both operands are constants, but it was compiled as constant ** register.

The thing I can't figure out yet is why that happens, since I just reused the compiler code for the other multiplicative operators for exponentiation.

fatcerberus:
Found the bug: Opcodes which have reg/const variants must be divisible by 4, while `DUK_OP_EXP` was #defined to be 190.  I must have missed that bit somehow.

fatcerberus:
MDN says ambiguous unary operators are forbidden when used with exponentiation, e.g., `-2**16` is a SyntaxError:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Exponentiation_(**)

You must parenthesize to resolve the ambiguity.  Is this restriction possible to implement with the current parser, and if so, how?

svaarala:
Btw, MDN is usually right but it'd be best to check the requirements from the specification directly as they can differ a bit (compare Proxy for example). Just at a quick glance at http://www.ecma-international.org/ecma-262/7.0/ it's probably a SyntaxError based on the grammar production targets (ExponentiationExpression takes a UpdateExpression rather than an UnaryExpression). I'd have to think a bit what that means for the compiler.

svaarala:
Re: that opcode limitation, I think there are asserts in the compiler for that but I'm not 100% sure.

fatcerberus:
I actually tried reading the spec, but couldn't find a clause mentioning the restriction, so it must just be implied by the productions like you said.  In any case, I used MDN only as a convenient reference here, I first found out about it here:
* https://esdiscuss.org/topic/exponentiation-operator-precedence

svaarala:
Ok, I'll try to check the ES7 specification also when I have time to look into this more closely.

I don't mind linking to whatever references by the way, the specification is not always very easy to read. But it's still the best source when deciding on what should be the compliant behavior so that if there's a code comment saying "XXX is required" it's best to say "XXX is required by E7 Section X.Y.Z.W". I usually then follow by comparing to Firefox and Chrome web console to see what the existing engines do, and if they disagree, I'll re-read the specification ;)

fatcerberus:
Chrome V8 seems to agree it's a syntax error:
```
-2**16
Uncaught SyntaxError: Unexpected token **
-(2**16)
-65536
```

Is it possible to backtrack in the token stream during parsing?  I had an idea that I could look backwards two tokens from the `**` operator to see if there is a unary operator there and throw an error if so.

svaarala:
For much of the compiler the previous token is available. But that's not enough here, because you'd need to see more than one token back I guess?

Also note that `-2` is, at least in ES5, technically two tokens: an unary minus and a positive constant. That unary minus is generally executed during compilation to yield a negative constant. But technically `-2**4` would be four tokens: `-`, `2`, `**`, `4`.

fatcerberus:
Right, that's why I had the idea to go back two tokens: From `**`, the previous one would be `2`, and the one before that is `-`, a unary minus, which is disallowed.  This idea is just a hack though, a cleaner solution would of course be preferable. :)

Anyway I'll try to get the rest of the pull cleaned up and we can worry about this particular issue later.

svaarala:
Sounds good, I'd be OK with adding that SyntaxError check in a separate pull.

svaarala:
There's a mechanism to rewind the lexer but it's relatively heavyweight as it'll reinitialize the lexer window, etc. It's used for implementing multiple passes over a function but not really on a per-token level.

If that SyntaxError is very difficult to implement it might be best to just leave it out (document it as a known issue) and revisit it when the compiler is reworked for ES6. It would be trivial to check for with an IR tree.

fatcerberus:
ES7 also includes an exponentiation assignment operator (`**=`):
https://tc39.github.io/ecma262/2016/#sec-assignment-operators

The checklist keeps getting larger. :)

fatcerberus:
@svaarala I like the way you implemented assignment in the compiler: It's a shared handler for all assignment variants.  Adding support for a new operator was trivial, I didn't even have to add any opcodes. :+1:

svaarala:
That reuse is mostly a result of trying to squeeze footprint, good to hear it works for this operator too :)

fatcerberus:
I assume this is where constant folding is done:
https://github.com/svaarala/duktape/blob/master/src-input/duk_js_compiler.c#L2177-L2183

svaarala:
Yeah, it's a narrow check for operations that can be done at compile time. Really just for the most obvious cases like string concetenation (which allows efficient multiline strings combined with `+`) and number constants.

tests/ecmascript/test-expr-expo.js
```diff
@@ -0,0 +1,44 @@
...
+print(-1 ** -Infinity);
+
+/*===
+exponentiation assignment
+1000
+===*/
+
+print("exponentiation assignment");
+w = 10;
+w **= 3;
```
svaarala: Maybe a test for `x **= y **= 3`?

fatcerberus:
On thinking about this some more, `Infinity` is probably not treated as a compile-time constant.  I'll likely have to use an expression like `1 ** (1 / 0)` to test compile-time behavior.

svaarala:
That's right, there's no literal for infinity so you'd need to use something like `1/0` for compile time arithmetic - at least for the time being (until more intelligent inlining is supported).

fatcerberus:
Same reason I prefer `void null` over `undefined` - it avoids a property lookup at runtime.

fatcerberus:
This is nearly merge ready, but `test-expr-expo.js` is currently failing.  Investigating now to find out what's wrong.

fatcerberus:
Hm, `**` should intuitively outrank `*` in precedence, but in Duktape the expression `2 * 2**10` is evaluated as `(2 * 2)**10`, causing the test to fail.

The confusing thing is that the operator precedence table on MDN gives `**` the same rank as `* / %`, implying Duktape's behavior is correct; however, Chrome disagrees.  I can't test in Firefox without installing a nightly build.  If I could understand the grammar productions in the Ecmascript spec that would help, but those are gibberish to me :)

fatcerberus:
Babel agrees that `2 * 2**10` is 2048, not 1048576.  Likewise for TypeScript.  So it definitely looks like Duktape's behavior is wrong.  I'll give the spec another look to see if I can figure out what the correct interpretation is.

fatcerberus:
Okay, looked at the spec and 2048 is indeed the correct value of `2 * 2**10`.  The higher precedence for `**` is implied by the grammar - MultiplicativeExpression takes ExponentialExpressions as argument, so that the latter gets evaluated first.

Fixed by adding a new binding power above multiplicative.  I guess this is one of those cases where MDN turns out to be wrong. ;-)

fatcerberus:
Hm, the testcase still seems to be failing.  I'll have to look into it in the morning.

svaarala:
I think the issue might be that it is right associative which needs handling described (a bit vaguely) in https://github.com/svaarala/duktape/blob/master/doc/compiler.rst#operator-precedences-binding-powers.

fatcerberus:
Correct, right-associative operators should have 1 subtracted from their binding power, which I have done:
https://github.com/fatcerberus/duktape/blob/exponentiation-operator/src-input/duk_js_compiler.c#L3957

And `duk` seems to agree that it works correctly:
```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-689-g17b36a2)
duk> 2**2**4
= 65536
duk> (2**2)**4
= 256
duk> 
```

So I think the issue is somewhere else.  I'll run the testcase manually and see what's wrong, it might just be an error in the expect string.

fatcerberus:
Hm, I can't run "make qecmatest", I get the following output:
```
fatcerberus@pigcult-vm:~/src/duktape$ make qecmatest
### qecmatest
"/usr/bin/nodejs" runtests/runtests.js --prep-test-path util/prep_test.py --minify-uglifyjs2 UglifyJS2/bin/uglifyjs --util-include-path tests/ecmascript --known-issues doc/testcase-known-issues.yaml --run-duk --cmd-duk=/home/fatcerberus/src/duktape/duk --num-threads 4  --log-file=/tmp/duk-test.log tests/ecmascript/
Executing 859 testcase(s) with 1 engine(s) using 4 thread(s), total 827 task(s)
{ [Error: Command failed: python util/prep_test.py --minify-uglifyjs2 UglifyJS2/bin/uglifyjs --util-include-path tests/ecmascript --input tests/ecmascript/test-arguments-access-inner.js --output /tmp/tmp-runtests-8883eW4GtSupfmRi --prologue /tmp/tmp-runtests-888345vFI51oi5hn
/usr/bin/env: ‘node’: No such file or directory
  File "util/prep_test.py", line 193, in <module>
    main()
  File "util/prep_test.py", line 190, in main
    preparer.prepareTestcase(fn_in, fn_out, fn_prologue)
  File "util/prep_test.py", line 148, in prepareTestcase
    res = self.prepEcmaTest(fn_in, fn_prologue, data)
  File "util/prep_test.py", line 136, in prepEcmaTest
    lines[0] = self.prepEcmaPrologue(fn_prologue) + ' /*...*/ ' + lines[0]
  File "util/prep_test.py", line 111, in prepEcmaPrologue
    return stripTrailingNewlines(self.minifyOneLine(fn))
  File "util/prep_test.py", line 106, in minifyOneLine
    return self.minifyUglifyJS2(fn)
  File "util/prep_test.py", line 91, in minifyUglifyJS2
    raise Exception('uglifyjs2 minify failed')
Exception: uglifyjs2 minify failed
]
  killed: false,
  code: 1,
  signal: null,
  cmd: 'python util/prep_test.py --minify-uglifyjs2 UglifyJS2/bin/uglifyjs --util-include-path tests/ecmascript --input tests/ecmascript/test-arguments-access-inner.js --output /tmp/tmp-runtests-8883eW4GtSupfmRi --prologue /tmp/tmp-runtests-888345vFI51oi5hn' }
fs.js:60
      throw err;  // Forgot a callback but don't know where? Use NODE_DEBUG=fs
      ^

Error: ENOENT: no such file or directory, unlink '/tmp/tmp-runtests-8883eW4GtSupfmRi'
    at Error (native)
Makefile:485: recipe for target 'qecmatest' failed
make: *** [qecmatest] Error 1
```

At this point I'll just run the test with `duk`, it's easier. :)

fatcerberus:
This is the failing test:
```
y = 2;
y **= y **= 2;
print(y);
```

The expect string is "16", but what's actually printed is 256.  Is that correct for an assignment operator?  That implies `y **= 2` is evaluated first, so that that overall effect of the expression is that `y` is set to `4**4`.

svaarala:
Normal assignment is right associative IIRC:

    y = 10;
    print(y += y *= 4);

would be same as:

    y = 10
    print(y += (y *= 4));

So assuming that's the case for exponentiation too (it's evident from the lexer productions - I can check that later on when I have time):

- y **= y **= 2 <=> y **= (y **= 2)
- With y=2, y**=2 evaluates to 2^2 = 4 which gets assigned to y and also becomes the expression result
- y **= 4 will then be 4^4 = 256 which gets assigned to y and becomes the expression result

fatcerberus:
Yeah, I edited my post above, I guess you didn't see it.  I remembered assignment was right associative, but assumed the variable wouldn't be written to until the whole expression had been evaluated.

svaarala:
```js
x = 5;
y = 2;
print(x **= y **= 3);
print(x);
print(y);
```

Would evaluate as follows:
- With y=2, `y **= 3` would evaluate to 2^3 = 8, which becomes `y` and expression value.
- With x=5, `x **= 8` would evaluate to 5^8 = 390625, which becomes `x` and final expression value.

So the printout would be, maybe (;-):

```
390625
390625
8
```

svaarala:
Yes, each expression is evaluated on its own and each `x <op>= y` expression updates its LHS as part of evaluating that individual (sub)expression.

svaarala:
Hmm, I'll actually need to check that against the specification.

svaarala:
It may be that the LHS previous value should be evaluated before evaluating the RHS, in which case 2^4 = 16 would be correct. So the `y` *is* updated by the intermediate expression (which would be apparent and matter in practice if there side effects somewhere in the chain, e.g. a `with` binding for `y`) but that update might not matter as far as the outermost assignment is concerned.

There may be a bug in master related to this if that's the case.

svaarala:
Yeah, I think there's a bug in master. Test case:

```js
if (typeof print === 'undefined') { print = console.log; }

var obj = {};
var my_x;
Object.defineProperty(obj, 'x', {
    set: function (v) { print('Write x:', v); my_x = v; return true; },
    get: function () { print('Read x:', my_x); return my_x; }
});

with (obj) {
    x = 10;
    print('TEST');
    x += (x *= 4);
}
```

Node.js (with comments):

```
Write x: 10
TEST
Read x: 10    // LHS evaluation for *outermost* expression
Read x: 10    // LHS evaluation for *innermost* expression
Write x: 40   // x *= 4 => 40, write to x
Write x: 50   // 10 + 40 => write to x
```

Duktape 1.5.0 and master:

```
Write x: 10
TEST
Read x: 10    // LHS evaluation for innermost expression
Write x: 40   // x * 4 => 40, write to x, becomes expression result
Read x: 40    // LHS evaluation for outermost expression - this is too late
Write x: 80   // 40 + 40 = 80 => write to x
```

So Duktape evaluates the `x <op>= y` LHS too late at the moment. I'll fix that in master and you can rebase against that change?

svaarala:
The fix is relatively tricky: the right order is evident but inefficient even in the common case. I'll try to fix it tonight in any case, so hang on :)

fatcerberus:
So what's the correct value of `y` in `y = 2; y **= y **= 2`?  Is it 16 or 256?

svaarala:
With the apparently correct evaluation order 16 would be correct: the final exponentiation `y **= 4` uses the initial value for `y`, 2. The assignment is still evaluated `y **= (y **= 2)` but the top level `y` is evaluated and stored in a temporary before the top level RHS is evaluated.

svaarala:
Btw, looking at the revised ES6 semantics for assignment: http://www.ecma-international.org/ecma-262/6.0/#sec-assignment-operators-runtime-semantics-evaluation. There's some .name magic for function expressions... :8ball:

fatcerberus:
Chrome gives a result of 16 as well.

I remember reading about this a few years ago: While assignment is right associative, it's supposed to be evaluated left-to-right.  As a result of that, a lot of JavaScript implementations get it wrong.  Jurassic had a similar bug a while back, don't know if it's been fixed.

svaarala:
Yep, I fixed an earlier bug related to this too, but there's a leftover bug in that fix, see #992. The dilemma is that it's easy to implement the correct version, but it generates inefficient code even for simple cases like `x += 4` (generating an unnecessary temporary variable load). Trying to shortcut the most obvious cases would be nice, but the current shortcut is (still) unsafe.

fatcerberus:
I guess while I'm waiting for the fix for the assignment LHS issue I'll squash some of the commits, I seem to have made a mess of the branch. :)

fatcerberus:
This is now merge-ready, pending the fix in #992.

Overall it wasn't difficult at all to add a new operator, it just took me a bit to figure out where the changes needed to be made.  The compiler looked a lot more complicated before I started actually working with it.  Once I figured out that ivalues represent "expressions" and not "values" per se, it become a lot easier to grasp how everything worked.

svaarala:
The compiler document describes the ivalues a little bit with examples (in case you didn't stumble upon it): https://github.com/svaarala/duktape/blob/master/doc/compiler.rst#compiler.

fatcerberus:
Hm, it seems that the required semantics for `**` match those of `Math.pow()`.  Maybe they should share a helper?

svaarala:
Agree. There are two logical places for that:
- duk_js_ops.c currently houses a bunch of operations that are related to Ecmascript semantics but which are not obviously part of any specific object or API. It's not a full set; some operations are implemented directly in duk_api_stack.c for example.
- Some operations which are directly related to some built-in are provided by the built-in files, e.g. duk_bi_date.c provides some Date-related helpers to other functions.

So I guess this helper could be in duk_bi_math.c or in duk_js_ops.c.

The other arithmetic operations are DUK_LOCAL in duk_js_executor.c. I placed them there to ensure they're most likely to be inlined and they're not called from other places right now. That might change if the compiler constant folding is improved though (e.g. bitwise ops) so they'd need a new home.

A third option would be adding something like duk_js_arith.c for arithmetic operations specifically.

fatcerberus:
I went with adding `duk_js_arith.c`  it didn't seem to fit well in either of the other two files, so this seemed cleanest.

Now `**` always behaves identically to `Math.pow()`, both at runtime and compile time.

svaarala:
Sounds good to me :)

svaarala:
The diff looks good to me - hopefully rebase to master (now that #992 is merged) fixes that expect string issue.

svaarala:
The website will need a small update, maybe the "ES6 features" needs to be renamed etc. I can look into that separately.

tools/configure.py
```diff
@@ -475,7 +475,8 @@ def default_from_script_path(optname, orig, alternatives):
         'duk_internal.h',
         'duk_jmpbuf.h',
         'duk_exception.h',
-        'duk_js_bytecode.h',
+        'duk_js_arith.c',
+		'duk_js_bytecode.h',
```
fatcerberus: Whoops, notepad++ usually preserves the proper whitespace when adding lines, not sure what happened here.  I'll fix it when I rebase in a bit.

fatcerberus:
Maybe "features borrowed from ES6" could instead be "borrowed from newer editions".

tools/configure.py
fatcerberus: Would it be feasible to add a code policy check to catch mistakes like this?

fatcerberus:
Okay, rebased to master and... Travis passed, yay!

fatcerberus:
I expanded the testcase to cover all special cases in E7 Section 12.7.3.4.  Even though most of the cases match expected IEEE behavior, the test should be comprehensive.

It turns out a shared helper was exactly the right thing to do.  "`Math.pow()` == `**`" is made explicit in the ES7 standard:
> 20.2.2.26 Math.pow ( *base*, *exponent* )
>
> Return the result of **Applying the ** operator** with *base* and *exponent* as specified in 12.7.3.4.

fatcerberus:
What is your method to measure footprint?  I compared the size of `duk` (x64) between master and this branch and there's something like a ~12kB difference in size.  No way this ultimately small change expanded the footprint by that much.

svaarala:
I'll try to see if I can fix the webhook whitelist so it'd trigger the runs for your commits too. Basically the tests just compile a test program (with various size optimization flags given to gcc) and run `size` on it.

fatcerberus:
Alright.  This is merge ready, `test-expr-exponentiation.js` seems to be passing now thanks to #992. :)

svaarala:
Cool :) I'll look into the whitelist issue before merging.

fatcerberus:
Fancy automated tests, yay! :)

fatcerberus:
"qecmatest-assert" sounds like it will take a while. ;-)

svaarala:
Yeah, current test runs take a long time and it often takes a while before assert runs even start (they work through the backlog slowly). It's not necessary to wait for all jobs to pass prior to merging though. Tests serve development, not the other way around :)

fatcerberus:
Valgrind test is STILL going 12 hours later, wow.  Looks like everything else passed though.

svaarala:
Yep, it takes a lot of time. For valgrind runs it'd be useful to break the test set into smaller chunks or even run the tests individually, but that's a rework for a later time :)

fatcerberus:
In any case, looks like all the tests have passed.  Anything else I need to do before merge?

svaarala:
So ~+170 bytes for x86 minimum size build and +330 bytes for x86 stripped size build.

The difference is probably explained by the fact that before the change, the stripped size build has `Math` built-in removed so no code for pow() is included in the build. This brings the pow() code back.

svaarala:
No, I'll merge tonight, once I get some free time.

src-input/duk_js_executor.c
```diff
@@ -4895,8 +4923,8 @@ DUK_LOCAL DUK_NOINLINE void duk__js_execute_bytecode_inner(duk_hthread *entry_th
 		}
 
 #if !defined(DUK_USE_EXEC_PREFER_SIZE)
-		case DUK_OP_UNUSED194:
-		case DUK_OP_UNUSED195:
+		case DUK_OP_UNUSED190:
+		case DUK_OP_UNUSED191:
 		case DUK_OP_UNUSED196:
```
fatcerberus: See I *do* pay attention to comments. ;)

fatcerberus:
Regarding the SyntaxError for e.g. `-2**16`, I might have an idea for how to implement that.  I'll look into that either later on or tomorrow, after I figure out what's going on with #996.

Issue #956, opened 2016-09-07
Remove function properties when built-in bindings are disabled


svaarala:
Now with an atomic config-and-prepare (tools/configure.py) implement proper optional built-in bindings:
- When a certain binding is disabled (say `DUK_USE_FINALIZER_SUPPORT` is false), remove corresponding properties too.

One practical implementation is for `genbuiltins.py` to be aware of active config options (defaults + forced overrides) and certain bindings can be conditional on config options. For example, the syntax could be:

    - key: "fin"
      value: ...
      condition:
        DUK_USE_FINALIZER_SUPPORT: true

or more simply:

    - key: "fin"
      value: ...
      present_if: DUK_USE_FINALIZER_SUPPORT
      # similarly, absent_if: ...

Convert existing conditional built-ins to this model.

svaarala:
#988 adds `present_if` which is enough for now.

svaarala:
This in in place now. There's definitely room to improve the metadata syntax incrementally.

PR #988, opened 2016-10-03
Add support for truly optional built-ins

### LINKED ISSUE #874: Add better support for optional built-ins
Currently when built-ins (for example, buffer objects) are disabled via `duk_config.h` the top level objects and their methods will be present but will throw an error. This is not desirable for several reasons:

- It's a common idiom to test for features by presence of certain objects or methods; for example, `typeof Uint8array !== 'undefined'` would indicate that typed array (or at least Uint8Array) support was present. Having the objects and methods present defeats this.
- Footprint-wise it'd be preferable for the built-ins to be entirely absent.

Genbuiltins.py already has support for deleting built-ins as part of creating a new dist package. One can use YAML files to modify and/or delete the built-in objects and then create a new dist package (this process is used by several low memory targets) The object and method code will still be present (as stubs) and modern linkers will be able to mostly optimize that code away.

However, this process is separate from config options so it's not very convenient. It'd be preferable if the same configuration process encapsulated both config option header and genbuiltins.py "metaprogramming".

svaarala:
Add support for objects and properties not needed by the active configuration to be entirely dropped from the build:

- [x] Add integration between genconfig.py and genbuiltins.py which allows the set of "active config options" to be conveyed to genbuiltins.py.
- [x] Add genbuiltins.py support for `present_if: DUK_USE_xxx` notation for objects and properties.
- [x] Add genbuiltins.py support for removing explicitly disabled from the "needs bidx" list. If Duktape source code then refers to the related bidx, compilation will fail.
- [x] Add builtins metadata for dropping Proxy when not supported
- [x] Add builtins metadata for dropping bufferobjects when not supported
- [x] Add builtins metadata for dropping Math when not supported
- [x] Add builtins metadata for dropping coroutine methods when not supported
- [x] Add builtins metadata for dropping Section B functions when not supported
- [x] Add builtins metadata for dropping RegExp when not supported
- [x] Add builtins metadata for dropping Duktape.fin when not supported
- [x] Remove the throwing built-in functions for disabled config which has metadata support, they're now dead code unless there's a config mismatch
- [x] Some source fixes for dropping bufferobjects entirely; current implementation depended on DUK_BIDX_ARRAYBUFFER_PROTOTYPE even when bufferobjects were disabled. (This could be reverted and the ArrayBuffer prototype made an empty object instead, which would allow user code to add inherited methods for buffers.)
- [x] Maybe add configure.py option to skip dropping unused objects/properties, may be useful with fixups. => Not an option if the "throwing" variants are no longer present when a certain feature is disabled.
- [x] Solution or documentation for fixup limitation. => Documentation.
- [x] Documentation updates.
- [x] 2.x migration note: disabled objects no longer present.
- [x] Releases entry.

There are some limitations still in the pull:
- When using a fixup to enable/disable a feature covered by `present_if` the result will be a build failure. Genconfig.py cannot reliably figure out the final active configuration when fixups are present - even if the final header including the fixup were scanned, a fixup may include further headers which are not known at genconfig time.

Fixes #874. Implements part of #956.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1772,6 +1772,12 @@ Planned
   [[DefaultValue]] specification algorithm removed in ES6 (it was folded
   to ToPrimitive()), use duk_to_primitive() instead (GH-984)
 
+* Add support for dropping built-in bindings entirely when they are disabled
+  in configuration, e.g. the Proxy and buffer object bindings will be absent
+  instead of being replaced by functions throwing an error; this is more
+  in line with how application code detects active features and reduces
+  footprint (GH-988)
+
 * Add convenience API calls duk_get_prop_lstring(), duk_put_prop_lstring(),
   duk_del_prop_lstring(), duk_has_prop_lstring(), duk_get_global_lstring(),
   duk_put_global_lstring() (GH-946, GH-953)
--- a/config/examples/low_memory_strip.yaml
+++ b/config/examples/low_memory_strip.yaml
@@ -26,3 +26,6 @@ DUK_USE_ES6_PROXY: false
 
 # Don't support non-BMP characters in source code (UTF-8 otherwise OK).
 DUK_USE_SOURCE_NONBMP: false
+
+# Remove Math built-in.
+DUK_USE_MATH_BUILTIN: false
--- a/doc/duk-config.rst
+++ b/doc/duk-config.rst
@@ -136,8 +136,7 @@ but also ``duktape.c`` and ``duktape.h`` to be included in your build.
 
 You can override individual defines using in several ways (see "Option
 overrides" section below for more details): C compiler format (-D and -U
-options), YAML config through a file or inline, or verbatim fixup header
-through a file or inline.
+options) and YAML config through a file or inline.
 
 If you're building Duktape as a DLL, you should use the ``--dll`` option::
 
@@ -186,8 +185,7 @@ in config option metadata files in ``config/config-options/``.
 
 You can override individual defines using in several ways (see "Option
 overrides" section below for more details): C compiler format (-D and -U
-options), YAML config through a file or inline, or verbatim fixup header
-through a file or inline.
+options) or YAML config through a file or inline.
 
 Some changes such as reworking ``#include`` statements cannot be represented
 as override files; you'll need to edit the resulting config header manually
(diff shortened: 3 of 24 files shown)


svaarala:
Impact on x86 "stripsize" build which removes Proxy and bufferobjects (and coroutines etc) is -1.3kB. Not bad, considering that no essential functionality is changed and the not having non-functional Proxy, Uint8Array, etc around is actually an improvement :)

svaarala:
The difficulty with fixup headers is that there's no way to know how they alter DUK_USE_xxx options because the tweaks may happen even in external headers not known at genconfig time (this happens e.g. when a fixup uses `#include "my_duk_custom.h"`).

Object/property pruning based on config options is also not just a footprint optimization as it has a visible impact on e.g. the `Proxy` binding being absent when it's not actually supported. So, if pruning was automatically disabled when fixup headers were present, the resulting build would behave differently compared to a build where the option tweaks were made using configure.py options and YAML metadata files.

I don't think there's a very good solution for this. The current state can be improved a bit by basing the set of active options on scanning the final header file instead of just the config metadata defaults and metadata-based overrides, so that e.g. `#undef DUK_USE_BUFFEROBJECT_SUPPORT` in the fixup is taken into account and causes bufferobjects to be pruned.

Related improvements:
- If a `present_if` is used and the active option set doesn't have the key at all, emit a warning indicating the object will be kept. This can happen if the final header is scanned and there's no unambiguous value for an option.
- Emit a warning about pruning if `#include` is present in a fixup file.

Overall I'd like to just say manipulating `DUK_USE_xxx` options in fixup headers is not supported. They're still useful for pulling in custom include files and declaring functions needed by macros, etc.

svaarala:
Added some more dropping of disabled built-ins, x86 stripsize build is now -3.2kB from original. Let's see with a few final additions.

svaarala:
Ok, final x86 stripsize change is -3.5kB, and the stripsize build is now around 113kB (for a command line eval tool).

fatcerberus:
Considering you haven't actually removed any functionality, that's impressive.

svaarala:
Also this is one missing piece in allowing Duktape to grow ES6 (and maybe other like WHATWG) features without compromising the low memory targets, at least as far as the built-in objects are concerned, which is often where most of the footprint lies.

svaarala:
One thing I'll add in a separate pull is to make *all* built-ins configurable and optional. This will give a useful indication of what's the absolute minimum footprint to get core scripting to work, and you can then pick what built-ins (e.g. Array.prototype functions) you need in practice. I hope that build is significantly below 100kB so that it'll be possible to work on 128kB flash devices comfortably (though it's never really comfortable with that amount of flash ;).

Startup RAM usage will also be much less (probably <10kB) with that, so 128kB system flash + 64kB system RAM would then be a workable configuration, even without ROM built-ins.

PR #989, opened 2016-10-03
Make (almost) all built-ins optional

svaarala:
Make almost all built-in objects optional and configurable via config options (e.g. `DUK_USE_ARRAY_BUILTIN`).

When a certain built-in is disabled, the global property (`Array` for example) is removed.

Prototype objects are kept when a primitive type relates to it (e.g. Array prototype is kept, but `Array.prototype` doesn't exist). The prototype object is also scrubbed of any function properties so it's empty except for a few mandatory properties. This approach allows user code to attach custom properties to e.g. array prototype (it can be accessed as `Object.getPrototypeof([]).customMethod = ...` or via the C API) and they will work for array instances. An alternative approach would be to remove the prototypes too and make e.g. array instances inherit directly from Object.prototype (or if that's removed, from nothing). But the prototypes are pretty small and quite useful to leave behind so that approach is now used.

- [x] Global; catch-all for stuff like encodeURIComponent(), escape() etc, global object itself remains and e.g. `globalObject.undefined` remains because it's so commonly assumed
- [x] Object
- [x] Array
- [x] Function
- [x] Math
- [x] Date
- [x] JSON
- [x] Boolean
- [x] Number
- [x] String
- [x] Duktape
- [x] Make JSON actually optional, so that C API will throw "unsupported" when JSON support is disabled (this should be separate from the JSON built-in object)
- [x] Leave Error constructors and prototypes alone for now; there'd be various good alternatives to them (e.g. all constructors pointing to a shared `Error` constructor) but better done separately.
- [x] Resolve FIXMEs
- [x] Need support for compound "present if" expressions, maybe just add some generic evaluable syntax
- [x] Low memory documentation update, 128kB system flash and 32kB system RAM should now be possible (barely, of course)
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1988,7 +1988,8 @@ Planned
   GH-973); minor RegExp compile/execute optimizations (GH-974)
 
 * Miscellaneous footprint improvements: RegExp compiler/executor (GH-977),
-  internal duk_dup() variants (GH-990)
+  internal duk_dup() variants (GH-990), allow stripping of (almost) all
+  built-ins for low memory builds (GH-989)
 
 * Internal change: rework shared internal string handling so that shared
   strings are plain string constants used in macro values, rather than
--- a/config/config-options/DUK_USE_ARRAY_BUILTIN.yaml
+++ b/config/config-options/DUK_USE_ARRAY_BUILTIN.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_ARRAY_BUILTIN
+introduced: 2.0.0
+default: true
+tags:
+  - ecmascript
+description: >
+  Provide an Array built-in.
--- a/config/config-options/DUK_USE_BOOLEAN_BUILTIN.yaml
+++ b/config/config-options/DUK_USE_BOOLEAN_BUILTIN.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_BOOLEAN_BUILTIN
+introduced: 2.0.0
+default: true
+tags:
+  - ecmascript
+description: >
+  Provide a Boolean built-in.
--- a/config/config-options/DUK_USE_DATE_BUILTIN.yaml
+++ b/config/config-options/DUK_USE_DATE_BUILTIN.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_DATE_BUILTIN
+introduced: 2.0.0
+default: true
+tags:
+  - ecmascript
+description: >
+  Provide a Date built-in.
--- a/config/config-options/DUK_USE_DUKTAPE_BUILTIN.yaml
+++ b/config/config-options/DUK_USE_DUKTAPE_BUILTIN.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_DUKTAPE_BUILTIN
+introduced: 2.0.0
+default: true
+tags:
+  - duktape
+description: >
+  Provide a Duktape built-in.
--- a/config/config-options/DUK_USE_FUNCTION_BUILTIN.yaml
+++ b/config/config-options/DUK_USE_FUNCTION_BUILTIN.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_FUNCTION_BUILTIN
+introduced: 2.0.0
+default: true
+tags:
+  - ecmascript
+description: >
+  Provide a Function built-in.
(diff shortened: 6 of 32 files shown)


fatcerberus:
Hehe, variants of JS with `Object`, `Array` and `Function` removed seem pretty useless to me :)

svaarala:
More useless than not being able to run at all on a very low flash target? :)

fatcerberus:
Well, true... but it would have to be extremely well-documented if nothing else, otherwise it will confuse whoever has to write the code.  JS programmers usually take for granted that those exist. :wink:

svaarala:
Just to be clear, the config options can be used to remove the global `Object` etc bindings. The types will still remain and that's often enough for very low memory targets which do callbacks for sensor events, etc. But obviously one would only strip e.g. the Array built-in when that mattered, it'd be very preferable to keep it around for `.forEach()` and such.

It's only really a useful choice when you're working on such a low memory target that getting more space for application code is more important than niceties like `.forEach()`. You can still loop through arrays with a for-index loop of course e.g.:

```js
function onSensorEvent(sensors) {
    for (var i = 0; i < sensors.length; i++) {
        ...
    }
}
```

svaarala:
> Well, true... but it would have to be extremely well-documented if nothing else, otherwise it will confuse whoever has to write the code. JS programmers usually take for granted that those exist.

Anyone who disabled these would probably be pretty aware of what they were doing and why... :)

fatcerberus:
No worries, to be sure I think this is a great idea, I just found the idea of JS without `Object.prototype` amusing is all. :)

svaarala:
It's actually already used with some other engines. It's used for those tiny environments where one really just wants to be able to write simple callbacks and such.

Btw, the diff keeps the prototype in place, just that the global reference to it is not present. The prototype can be accessed e.g. via C code (push an object and `duk_get_prototype()`) so that you can add stuff there. This still allows you to add whatever bindings you feel are useful. Or you can of course edit the builtins.yaml file to add back bindings you feel are necessary.

svaarala:
Ok, with most built-ins stripped (JSON and Duktape still remain) the stripsize build is -20kB, for an x86 binary size of roughly 91kB. It'll drop just a bit more with Duktape and JSON disabled. However, for JSON, for example, the core JSON algorithms will remain in the footprint because the C API calls them so the footprint impact won't be that dramatic.

So this definitely makes it possible to run a bare metal minimal OS, startup code, Duktape, and some application code in 128kB flash. It's then just a matter of how much application code fits, and how much you can switch back more Duktape functionality (such as Object built-in binding).

svaarala:
Final-ish numbers for footprint impact are roughly:
- x64 stripsize: -30kB footprint change, binary size ~93kB.
- x86 stripsize: -27kB footprint change, binary size ~85kB.

svaarala:
There are a couple of annoying trivia issues in this pull before it can be merged, I'll fix them during this week probably before merging, there's no particular hurry to get this in.

svaarala:
Hmm, actually, I'll add an option to disable JSON functionality from the C API too - it's not always needed and is relatively large. With that in place C API would link but JSON functions would throw "unsupported" errors (same as what happens when bufferobjects are disabled). That'll probably get the minimum footprint closer to 80kB.

svaarala:
Current globals with every (implemented) option disabled, except Object so that getting property names is possible ;):

```js
var arr = Object.getOwnPropertyNames(this);
for (var i = 0; i < arr.length; i++) { print(arr[i]); }
```

Prints:

```
undefined
Object
Error
EvalError
RangeError
ReferenceError
SyntaxError
TypeError
URIError
Duktape
// (from this point on bindings are registered by 'duk' so they're not built-in automatically)
print
alert
console
require
readFile
writeFile
arr
i
```

The only bindings left are Errors which I'm very hesitant to remove (perhaps everything but the generic `Error` could be stripped). `Duktape` remains also for now due to a technicality in error augmentation. The global `undefined` binding remains because it's very commonly assumed (often people don't even realize it's not a literal, and I agree it's a very confusing feature of ES).

svaarala:
Ok, with that x86 stripsize is 83kB. It'd probably be closer to 81-82kB if specific error subtypes were strippable. Maybe I'll stop here though :)

fatcerberus:
Presumably `undefined` is a variable and not an real keyword for historical reasons, e.g. it used to be possible to do `undefined = "not undefined anymore";` and break everything. :)  That said, `undefined` was made non-writable at some point, so why they didn't make it a keyword at the same time is beyond me.

PR #990, opened 2016-10-04
Use some internal duk_dup() helper variants for smaller footprint

svaarala:
A little footprint experimentation for using specialized dup helpers internally, e.g. `duk_dup_m2(ctx)` == `duk_dup(ctx, -2)`. Saves around 100-200 bytes depending on configuration.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1980,7 +1980,8 @@ Planned
   opcode handler optimizations (GH-903); refcount optimizations (GH-443,
   GH-973); minor RegExp compile/execute optimizations (GH-974)
 
-* Miscellaneous footprint improvements: RegExp compiler/executor (GH-977)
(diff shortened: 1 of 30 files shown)


svaarala:
While GCC could theoretically create a shared helper for e.g. `duk_dup(ctx, -2)` it won't do so in practice, so adding an internal helper for `duk_dup_m2(ctx)` and a few others buys ~100-200 bytes (depending on configuration and platform). While the call sites are not as pretty, it's still relatively readable and there are no adverse effects otherwise. Hiding the helper behind a macro check is difficult in practice (`duk_dup(ctx)` is part of the public API, and macros with constant comparisons cause warning issues). A performance oriented build will inline the helpers so there's no performance impact (and in fact the change opens a possibility to special case a few of the dup indices for performance).

PR #991, opened 2016-10-04
Reconsider duk_push_string() and duk_push_lstring() NULL pointer handling

svaarala:
Currently duk_push_string() with NULL pointer pushes a `null` while duk_push_lstring() pushes an empty string.

With duk_push_string() an empty string can be pushed as `duk_push_string(ctx, "")` while `null` is pushed with `duk_push_string(ctx, NULL)`.

The duk_push_lstring() behavior could be changed so that `duk_push_lstring(ctx, NULL, 0)` also pushes a `null` while `duk_push_lstring(ctx, <non-NULL>, 0)` pushes an empty string. But this is rather awkward because you need to use a dummy-non-NULL pointer whereas with duk_push_string() the empty string is naturally non-NULL and points to a NUL character.

The other more consistent alternative would be to remove the `null` special case from duk_push_string() so that `duk_push_string(ctx, NULL)` would also push an empty string. This seems like the easier choice. This case was originally added because it made some idioms more convenient for calling code. It also matches `lua_pushstring()` which pushes a `nil` for a C NULL pointer (though maybe `undefined` would be more appropriate than `null` with Duktape).

### DIFF
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -3239,6 +3239,7 @@ DUK_EXTERNAL const char *duk_push_lstring(duk_context *ctx, const char *str, duk
 	 * a bit dubious.  This is unlike duk_push_string() which pushes a
 	 * 'null' if the input string is a NULL.
 	 */
+	/* FIXME: NULL handling is different from duk_push_string() */
 	if (!str) {
 		len = 0;
 	}
@@ -3264,6 +3265,7 @@ DUK_EXTERNAL const char *duk_push_string(duk_context *ctx, const char *str) {
 	if (str) {
 		return duk_push_lstring(ctx, str, DUK_STRLEN(str));
 	} else {
+		/* FIXME: NULL handling is different from duk_push_string() */
 		duk_push_null(ctx);
 		return NULL;
 	}


andoma:
FWIW I'd prefer that duk_push_string(ctx, NULL) pushes undefined as you say in the last comment.

It's quite common (at least for me) to do stuff like:

`duk_push_string(ctx, find_something(foo, bar));`

where find_something() returns NULL if nothing is found.

Pushing an empty string instead is typically not the right thing to do in that case.

svaarala:
@andoma Would `undefined` be preferable to `null` in that case?

No worries, I won't change the current `duk_push_string()` behavior if it's useful :)

andoma:
I'm mostly allergic to having NULL C pointer values automagically turn into empty strings. It would even be better to assert()/crash than that. But for something like Duktape I think it make perfect sense to push something that's not a string, though I don't really have a strong opinion about `null` vs `undefined` in that case. In fact I'm not 100% sure when to use `null` and when to use `undefined` :-)

svaarala:
Hehe :) Yeah, the current behavior is to push `null`. It would probably be a bit more in line with usual Ecmascript semantics to push `undefined` though. But maybe it's not worth changing.

svaarala:
@andoma What's your opinion on these BTW:
- `duk_push_lstring(ctx, NULL, 0)`: current behavior is to push an empty string.
- `duk_push_lstring(ctx, NULL, 123)`: the length is rather odd, but current behavior is still to push an empty string (i.e. pretend length is zero).

fatcerberus:
With regard to the `null`/`undefined` dichotomy: It's a bit of a weird setup (I prefer just having one "null" value, like in e.g. Lua or C#), but usually in JS I consider `undefined` as meaning "this variable hasn't been assigned a value yet" or "this argument was not provided" - I rarely use the value directly, preferring `null` if it's deliberate.  In other words, I treat "undefined" in the same sense as "undefined behavior".

svaarala:
But then, reading an unset property returns `undefined`.

fatcerberus:
Right, because the property hasn't been assigned to yet. :)  Basically if I read a variable or property or get back undefined, I consider it a bug (in the case of the hypothetical undefined property I'd check for `'prop' in obj` first).  Obviously there are exceptions, but that's my general rule of thumb.

fatcerberus:
To clarify, I was responding to @andoma's comment that he wasn't sure when to use `null` vs. `undefined` in code.

svaarala:
I guess I view it a bit differently. Accessing an undefined property is relatively common in code I've worked with, and if you do something like:

    var description = msg.description;

it gets assigned `undefined` if the property is not set. So in some ways a NULL string pointer could be considered a "missing" string (similarly to a missing property). With this line of thought it'd be logical if this pushed an undefined:

    duk_push_string(ctx, get_my_description());

Comparing to the Lua `lua_pushstring()` binding it pushes a `nil` which is closer to `undefined` than `null` (Lua doesn't make the distinction) I think.

In any case the current behavior is to push a `null` and we can change it or keep it, whatever works most intuitively.

fatcerberus:
Kind of off-topic, but I guess my main hangup is that there's a distinction between:
```
var x = { undef: undefined };
```
And
```
var x = {};
```

In both cases reading `x.undef` gives back a value of `undefined`.  So I prefer to make an explicit check for `'undef' in x` first, and only if that passes do I read the property.

Mostly though I just like having a distinction while debugging between "deliberate" `null` values and "accidental" undefined ones :)

fatcerberus:
Anyway I don't really have a horse in this race, I actually didn't even realize `NULL` was a legal value to pass to duk_push_string(). :)

fatcerberus:
Having thought about this a bit since yesterday, I think the current behavior for `duk_push_lstring()` (pushing an empty string) makes sense.  The semantics match those of C malloc(), which can return NULL if size == 0.  For the same reason it should probably be considered an error if str==NULL and size is nonzero.

It's less clear-cut for `duk_push_string()`.  I don't think it should push an empty string though -- because C strings are nul terminated, there's no reason for any valid string pointer (even an empty one) to be NULL.

svaarala:
Yeah, I had a note that the current behavior is a bit inconsistent but it might be it's the most reasonable approach.

Though at least `duk_push_lstring(ctx, NULL, 123)`, i.e. NULL with non-zero length, is a bit odd. It could be an API error just as well; now if a NULL is given the length is considered always to be zero so that an empty string is pushed.

svaarala:
Closing as unmerged, as there's no clear idea of how to improve the behavior.

PR #997, opened 2016-10-06
Cleanups for exponentiation operator

svaarala:
- [x] Code comment trivia
- [x] Add try-catch wrappers and remove tab indent in testcase
- [x] Reorder EXP opcode to be right after MOD
- [x] Fix debugger metadata typo for `GETPROP_RC`
- [x] Check debugger bytecode dump to ensure it still looks correct


svaarala:
There were a few tabs in the .js source file, I'll see if I can change the codepolicycheck to reject tab indent for `*.js`.

svaarala:
There are some existing offenders so I'll do that in a separate pull.

fatcerberus:
Sorry about the tabs.  I assumed code policy for JS was the same as that for C.

By the way I don't know if you want to renumber opcodes here - I tried to do so and broke everything, so I ended up just sticking it in the middle of the unused block.

svaarala:
Opcode numbering doesn't usually matter much - it seems there's sometimes some benefit to commonly used opcodes being nearby (presumably the switch-case jump table will then be more likely cached) but for uncommon operations like exponentiation, it doesn't really matter.

svaarala:
> Sorry about the tabs. I assumed code policy for JS was the same as that for C.

It's actually not documented for anything else than C (which is what really matters), so more my bad. For Python the repo now follows PEP8, at least as far as indents are concerned (indent by 4 spaces). For Ecmascript indent by 4 spaces is also used, mostly because tab indent for .js seems rare and since it matches the Python convention it's easy to remember.

The indent by tabs in C is of course quite arbitrary. I like it from previous habit, but indent by 4 would be equally good and maybe better to reduce the total indent level.

Obligatory ;) - https://www.youtube.com/watch?v=SsoOG6ZeyUI

svaarala:
Hehe, it's sure enough "odd" from aesthetics point of view. The logical place for EXP opcodes would be right after MOD for example (and just before bitwise operations). I can do that reordering here.

svaarala:
The only real requirement is for the base opcode to be 0 mod 4 - this is not assumed by the executor, but rather the compiler when it flips the two lowest bits for reg/const operands.

svaarala:
The dump/load API test broke before it prints the bytecode dump which is affected with the renumbering. I'll fix shortly.

svaarala:
That testcase is actually a bit annoying because any optimization or opcode renumbering breaks it. On the other hand, I want some test case to detect if opcode or bytecode dump output changes for no intended reason.

svaarala:
Ah, forgot that pre/post inc/dec opcodes also have some constraints, these in particular:

```
                        DUK_ASSERT((DUK_OP_PREINCP_RR & 0x0c) == 0x00);
                        DUK_ASSERT((DUK_OP_PREDECP_RR & 0x0c) == 0x04);
                        DUK_ASSERT((DUK_OP_POSTINCP_RR & 0x0c) == 0x08);
                        DUK_ASSERT((DUK_OP_POSTDECP_RR & 0x0c) == 0x0c);
```

fatcerberus:
That explains why things started breaking when I tried renumbering it myself. :(

svaarala:
Yep - asserts are in place and will catch these cases but it requires running the related opcodes.

fatcerberus:
> Obligatory ;) -
> https://www.youtube.com/watch?v=SsoOG6ZeyUI

Without starting a religious war :) I mostly prefer tabs in my own code because text editors tend to treat them consistently when navigating.  I use a bunch of different editors and nothing's more annoying than having muscle memory betray me when switching due to subtle differences in behavior.

For example when backspacing one editor might delete an entire indent level (= 4 spaces) at a time, while another will delete single spaces.  If there's a tab I know I can just press backspace once and it will always do what I expect.

That said, I'm adaptable and willing to work with whatever the project I'm working on dictates.  No sense getting into bikeshed debates. :)

svaarala:
I like tabs for similar reasons but it's not a big deal to me either. PEP8 suggests avoiding tabs though; I did use tabs for Python code before but several people pointed out it was not the common practice so I changed it. Indent by 8 (tabs or not) seems rare for JS code while indent by 4 spaces seems common, so I chose that (it also matches PEP8 indent so it's easy to remember).

svaarala:
Oh, and mixing tabs and spaces for indent level (not for continuation lines) is an absolute horror ;)

fatcerberus:
I thought that went without saying. :)

fatcerberus:
It was surprising to see a new operator added to ES7, I remember rumors of exponentiation being in ES6, but then it didn't show up.

This should prove useful for minisphere games.

PR #999, opened 2016-10-06
Document gap to ES6 for typed arrays

svaarala:
Add the gap to buffers.rst future work, based on a quick scan of ES6.

### DIFF
--- a/doc/buffers.rst
+++ b/doc/buffers.rst
@@ -946,6 +946,117 @@ there must be no side effects between the check and the operation:
 Future work
 ===========
 
+Missing ES6 features
+--------------------
+
+General semantics:
+
+* ToLength() coercion allows ArrayBuffer and typed array length up to
+  ``2^53 - 1``.
+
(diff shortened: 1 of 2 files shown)


svaarala:
The initial implemention for some of the Array-like methods (like .forEach) could actually be the equivalent `Array.prototype` methods as is; they're generic. ES6 specifies a difference in how `.length` is accessed: it should be read directly without invoking an actual property get.

Right now `.length` is a virtual own property (rather than inherited as in ES6) so this would work well. Even at the moment:

```js
var u16 = new Uint16Array([ 1, 2, 3, 4, 5, 6 ]);
u16.__proto__.join = Array.prototype.join;
print(u16.join(' - '));
```

prints:

```
1 - 2 - 3 - 4 - 5 - 6
```

This might also be a good baseline for low memory targets; it's not ideal but better than implementing the same algorithm manually. Default build configuration could use typed array optimized variants of course when they've been implemented.

This doesn't work for all methods, e.g. `.map()` result is not a generic array.

PR #1000, opened 2016-10-07
Document gap to Node.js v6.7.0 Buffer

svaarala:
Documenting gap between current implementation (subset of v0.12.1) and v6.7.0.  To be rebased and merged after #999.

### DIFF
--- a/doc/buffers.rst
+++ b/doc/buffers.rst
@@ -1057,6 +1057,140 @@ should be accessed directly without invoking side effects.  For now this
 would not be an issue because typed array ``.length`` is a virtual own
 property, and accessing it has no side effects.
 
+Update to newer Node.js Buffer API version
+------------------------------------------
+
+Current:
+
+* https://nodejs.org/docs/v0.12.1/api/buffer.html
+
+Latest at time of writing:
+
+* https://nodejs.org/docs/v6.7.0/api/buffer.html
+
+Gap between current implementation and latest:
+
+* ``Buffer.prototype`` should inherit from ``Uint8Array.prototype`` now;
+  Buffer instances are now Uint8Array instances, and
+  ``new Uint8Array([1,2,3]).__proto__ == Uint8Array.prototype`` is true.
+
+* Buffers can be for-of iterated; buf.values(), buf.keys(), and
+  buf.entries() create iterators.  For-of iteration requires @@iterator
+  support.
+
+* ``new Buffer(...)`` is deprecated in favor of ``Buffer.from()``,
+  ``Buffer.alloc()``, and ``Buffer.allocUnsafe()``.
+
+* ``new Buffer(arrayBuffer[, byteOffset [, length]])`` variant is not
+  supported.  This variant is already deprecated.
+
+* ``new Buffer(string[, encoding])`` does not handle encoding correctly.
+  The internal string representation (CESU-8, extended UTF-8, or even
+  invalid UTF-8) is used as buffer bytes as-is.  This is also incorrect
+  for Node.js Buffer v0.12.1 (and already incorrect in master).
+
+* ``Buffer.alloc(size[, fill[, encoding]])`` is missing.
+
+* ``Buffer.allocUnsafe(size)`` is missing.  In practice could be implemented
+  as ``Buffer.alloc(size)`` (ignoring any further parameters) so that a
+  zero-filled buffer is allocated.
+
+* ``Buffer.allocUnsafeSlow(size)`` is missing.  Could be implemented as
+  ``Buffer.alloc(size)`` (ignoring any further parameters).
+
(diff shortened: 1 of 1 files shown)


doc/buffers.rst
```diff
@@ -946,6 +946,203 @@ there must be no side effects between the check and the operation:
...
+* ``new Buffer(string[, encoding])`` does not handle encoding correctly.
+  The internal string representation (CESU-8, extended UTF-8, or even
+  invalid UTF-8) is used as buffer bytes as-is.  This is also incorrect
+  for Node.js Buffer v0.12.1 (and already incorrect in master).
+
+* ``Buffer.alloc(size[, fill[, encoding]])`` is missing.
+
+* ``Buffer.allocUnsafe(size)`` is missing.  In practice it should be
+  implemented as ``Buffer.alloc(size)`` (ignoring any further parameters)
+  so that a zero-filled buffer is allocated.
```
fatcerberus: This is incorrect.  `allocUnsafe()` is equivalent to C `malloc()`, that is, it doesn't initialize the buffer at all.  Hence "unsafe".  It's mostly meant for performance sensitive code.

svaarala: We already discussed this earlier :) It's not incorrect to zero the memory. The API says there are no guarantees, so zeroing the data is fine. You shouldn't rely on the data *not* being zeroed.

svaarala: Note that the current implementation zeroes data for `new Buffer()` which doesn't have zeroing guarantees either. Because Duktape is not a high performance engine, I think the best default is to zero allocated buffers as it has the least potential for surprise. I'd be OK with not zeroing the data if explicitly configured to do so, and there's such a config option already for plain buffers.

svaarala: Hmm, come to think of it, with Buffers now inheriting from Uint8Array (and eventually an ArrayBuffer) there's no way to allocate non-zeroed data without using an actual "fast buffer pool": ArrayBuffers are always zero initialized. So short of implementing an actual "fast buffer pool", the buffers would come out unavoidably as zeroed as far as I can tell.

svaarala: > [...] is the ONLY correct way to implement Node.js guarantees, which is not entirely accurate. :)

Right, that text is just saying what the probable implementation in Duktape should be. These are not specifications, just rough future work notes.

> Interestingly allocUnsafe() still returns an uninitialized buffer in Node 6.7.0. Since it inherits from Uint8Array now that's technically a spec violation.

It's not - `allocUnsafe()` is backed by a larger buffer and it can return a view to a section of it which contains previous data (the larger buffer will be zeroed on allocation but since it's reused over and over again, they may be older data). But since Duktape doesn't implement a buffer pool, that situation would never arise.

svaarala:
Overall I'd like to eliminate the whole Node.js Buffer binding -- except for the core parts which can't be easily implemented on top of ArrayBuffer and typed arrays. A Node.js Buffer binding could then be included as an extra where it would have much less impact on the internals, version guarantees, etc. It'd then make more sense, for example, to implement a "fast buffer pool" which feels very odd for core functionality.

This approach would be mostly viable but to keep performance reasonable, at least the index properties and read/write accessors would need to be implemented by Duktape. Not sure if there'd be much else that absolute needed to be in the core source.

The downside of this would be that there'd be more overlap between the internals and the extra providing bindings such as `.indexOf()`, `.fill()`, etc. Much of the low level code is shared between Node.js and typed arrays. It's not a huge amount but it would impact any low memory targets specifically wanting/needing Node.js Buffer bindings.

Not sure what to do :)

doc/buffers.rst
svaarala: Btw, I guess to be really pedantic, ES6 doesn't say that there isn't some out-of-spec way of creating ArrayBuffers which are not zeroed. IIRC it just says that *if* you create one using `new ArrayBuffer()` it's zero initialized.

And in fact Duktape allows this too: you can create an external buffer and create an ArrayBuffer over it, in effect creating an ArrayBuffer over whatever data you want (initialized or not).

Finally, the Node.js API just says Buffers are Uint8Arrays, but doesn't say where the underlying ArrayBuffer *must* come from.... :)

fatcerberus:
Buffer bindings are tricky - it's quite a low level primitive.  You *can* implement it from outside of the black box, but it's probably not ideal to do so.

svaarala:
Ok, so I did one pass over the v6.7.0 specification and compared it to v0.12.1 and the current limitation. Probably missed something but should be fairly complete.

svaarala:
One change I'm really glad about is the `Buffer.concat()` behavior for a 1-element array: it *didn't* create a copy in v0.12.1 so depending on the array size the result would either have shared memory backing or not which was quite error prone.

PR #1001, opened 2016-10-07
Add ES6 \u{H+} escape

svaarala:
- [x] Add to source lexer
- [x] Code comment consistency for `\u{H+}`
- [x] Config option for the new syntax (disable in low memory config)
- [x] Testcase for Ecmascript source case
- [x] Testcase for regexp case
- [x] Website update
- [x] 2.0 migration note - no special note (JX/JC changes may require one, but separate pull)
- [x] Releases
- [x] Future work: non-BMP in regexps #1030
- [x] Remove support for `\u{H+}` for regexps, as it's only enabled with `/u` regexp flag
- [ ] JX/JC changes in separate pull

For JX/JC:
- Add to JX
- Consistency for JC
- Make the \Unnnnnnnn escape optional for JX (just in case someone needs backward compatibility()


svaarala:
Just a bunch of FIXMEs for now.

fatcerberus:
From the ES6 specification:
> * It is a [static] SyntaxError if the MV of HexDigits > 1114111.

If Duktape continues to support the extended UTF-8 range that restriction could of course be relaxed.

svaarala:
I think the full 32-bit codepoint range should be allowed for now, to remain consistent with current extended codepoint support. But if #983 is merged that should also remove support for extended range from this escape format.

svaarala:
Hmm, then again there's no syntax for extended codepoints now so maybe that'd be a better default here too. It would also discourage creating such strings from Ecmascript code, and you'd only do so from C code.

svaarala:
Having thought it over the lexer should reject `\u{H+}` above U+10FFFF to match current behavior for source code: source code is interpreted as UTF-8 and any codepoints above BMP are encoded as surrogate pairs (CESU-8) and anything above U+10FFFF is already rejected.

So even the current parser rejects (unescaped) codepoints above U+10FFFF so that there's no way to generate literals, identifier names, etc, which are not originally valid UTF-8 and which are then CESU-8 encoded.

svaarala:
Reworked the escape parsing into a shared helper, and also reorganized some throw sites into shared goto labels with error handling. The lexer now supports `\u{NNNNNN}` but is smaller.

svaarala:
I'll split off the JX/JC changes to a separate pull. They don't impact compliance and I'm not yet sure what change would be right for them.

svaarala:
Hmm, it seems that neither Firefox nor V8 accepts this:

```js
> x = /\u{41}/
/\u{41}/
> x.test('A')
false
```

Based on my reading of http://www.ecma-international.org/ecma-262/6.0/#sec-patterns this should be accepted match the input. I'll keep that in the testcase unless someone points out why I'm wrong.

svaarala:
RegExp non-BMP handling won't be correct with this pull yet. For character atoms the pull emits a surrogate pair matcher which is close to correct (handled as if there was a surrogate pair in the RegExp source instead of the non-BMP escape) I think. For codepoint ranges non-BMP codepoints are kept as is and because RegExp input processing doesn't combine surrogates, it doesn't do anything useful. I'll have to read the ES6 with more thought to figure out the necessary changes, so leaving this intentionally outside the scope of this pull.

svaarala:
Ok, I'm done with the actual work here. I'll read through the changes tomorrow and merge unless there are suggestions for improvement. JX/JC stuff will be a separate pull (not sure if changes make sense) and a future work issue tracks non-BMP support in regexps.

fatcerberus:
I checked ES7 for the `\u{41}` regexp issue figuring maybe the semantics had been updated (ES7 does seem to have changed some things, just as a heads-up), but I couldn't find anything there that would prevent the example regexp from matching either.  Odd.

svaarala:
>[...] ES7 does seem to have changed some things, just as a heads-up [...]

Good point, ES7 is sometimes described as just adding "two new features" but there are some more subtle changes. It's good to take them into account right away when possible, but also OK to go ES6 first and then make ES7 fixes separately. The differences are luckily quite minor.

fatcerberus:
Did some more testing.  The example regexp matches (at least in SpiderMonkey) if you add the "u" flag:
```js
C:\src\mozilla-central\js\src\build_OPT.OBJ>dist\bin\js
js> var re = /\u{41}/
js> re
/\u{41}/
js> re.test('A')
false
js> var re = /\u{41}/u
js> re
/\u{41}/u
js> re.test('A')
true
js>
```

fatcerberus:
That goes for Chrome too - the 'u' flag must be present for the regexp to match.

svaarala:
That's still a bit odd, why would an ASCII escape `\u{41}` be dependent on the regexp match mode.

svaarala:
Maybe that syntax is indeed enabled only for the `u` match mode - I'll need to check that later tonight. If so, I'll change this pull so that the `\u{H+}` format won't be supported by RegExp at all because the match mode is also not supported now.

fatcerberus:
The only thing I can think of is that `\u` is interpreted as an identity escape and `{41}` as a quantifier.  Adding the `u` flag disambiguates.

svaarala:
That's probably the case - but why?

svaarala:
In other words, why would it be OK to add a new escape format to string literals (which is enabled by default) but not for RegExps (not enabled by default)? That may be the case though, I'll check later on.

fatcerberus:
I was reading through the Ecmascript spec to try to verify, but that wasn't much help.  I did find this though which seems to confirm that `u` is required:
https://mathiasbynens.be/notes/es6-unicode-regex

Without it `\u` is interpreted as an identity escape.

fatcerberus:
As for the "why" - no idea.  Perhaps for compatibility with code which assumes identity escapes are OK.

svaarala:
Ok, so it seems the syntax is only enabled with `/u`. First, in general pattern syntax (http://www.ecma-international.org/ecma-262/6.0/#sec-grammar-notation):

> A Pattern is either a BMP pattern or a Unicode pattern depending upon whether or not its associated flags contain a "u".

The relevant RegExp rule is:

```
[+U] u{ HexDigits }
```

Where "+U" means the unicode flag must be enabled for the rule to be enabled: http://www.ecma-international.org/ecma-262/6.0/#sec-regexpinitialize.

fatcerberus:
So if I understand this correctly, this is something you can do now?

```js
var pign\u{1F43D}se = "*munch*";
```

Kind of weird. :)

svaarala:
Yep, identifiers can already be escaped in ES5 with `\uHHHH` e.g.:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-793-ga2812f7-dirty)
duk> A = 'foo';
= "foo"
duk> \u0041
= "foo"
```

This pull then extends that to work with a new escape form.

```
duk> \u{00000000041}
= "foo"
```

Though IIRC Ecmascript doesn't allow surrogate pairs as valid identifier characters (not sure, would have to check). Sadly, at least the classic U+1F4A9 doesn't seem to be allowed by V8:

```
> \u{41} = 'bar'
'bar'
> A
'bar'
> \u{1f4a9} = 'quux'
\u{1f4a9} = 'quux'
^^^^^^^^^
SyntaxError: Invalid or unexpected token
```

svaarala:
@fatcerberus As a side note, things are starting to work a bit more nicely:

```
$ ./duk
((o) Duktape [linenoise] 1.99.99 (v1.5.0-793-gf3b7281)
duk> new TextEncoder().encode('foo\u{1f4a9}bar')
= |666f6ff09f92a9626172|
duk> 
Cleaning up...

$ python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> '666f6ff09f92a9626172'.decode('hex')
'foo\xf0\x9f\x92\xa9bar'
>>> '666f6ff09f92a9626172'.decode('hex').decode('utf-8')
u'foo\U0001f4a9bar'
```

In this case surrogate pairs are nowhere to be seen :-)

PR #1002, opened 2016-10-08
Make exponentiation operator (** and **=) optional

svaarala:
The ES7 exponentiation operator pulls in native `pow()` which has a significant footprint impact on bare metal low memory targets. Removing Math built-ins is not enough to remove the dependency when the operator is supported, so make the operator itself optional.

### DIFF
--- a/config/config-options/DUK_USE_ES7_EXP_OPERATOR.yaml
+++ b/config/config-options/DUK_USE_ES7_EXP_OPERATOR.yaml
@@ -1,0 +1,10 @@
+define: DUK_USE_ES7_EXP_OPERATOR
+introduced: 2.0.0
+default: true
+tags:
+  - ecmascript
+  - ecmascript7
+description: >
+  Support the ES7 exponentation operator (** and **=).  This is optional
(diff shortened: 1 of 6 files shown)


fatcerberus:
Operator is ES7, not 6, in case it matters for documentation.

I was considering making this optional in the initial merge, but it seemed weird to have a core language feature be optional.  Then again, minimal build is ES5, so this does make sense.

svaarala:
Oops, I'll fix the comments, ES7 indeed.

I wouldn't be worried about a 300 bytes difference (which this roughly was on a desktop system), but on bare metal platforms the actual math functions get pulled in. For example, transcendental functions were >20kB total on one target I worked with, which is pretty much when the build is otherwise ~100kB.

svaarala:
Btw, I realize option naming would need a little improvement. At one point it made sense to name options e.g. `DUK_USE_ES6_PROXY` but it might not make sense to include the specification version going forward (so `DUK_USE_PROXY` instead?).

fatcerberus:
I'd agree re: removing specification version.  Especially now that the specification is being revised annually.

PR #1004, opened 2016-10-10
Align Buffer.concat() behavior for one-element arrays with Node.js v6.7.0

svaarala:
`Buffer.concat()` always returns a copy, even for an one-element array argument.

- [x] Fix behavior
- [x] Fix testcases
- [x] Migration notes
- [x] Release notes

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1674,6 +1674,10 @@ Planned
   methods, and String.fromBuffer()); this reduces code footprint by around
   1.2 kB (GH-889)
 
+* Incompatible change: Node.js Buffer binding aligned with Node.js v6.7.0
+  (from v0.12.1): Buffer.concat() special case for 1-element array removed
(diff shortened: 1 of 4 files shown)


fatcerberus:
That pre-6.7.0 behavior looks like a clear case of premature optimization to me: Someone figured they could improve performance by not making a copy if the caller only passed in one buffer.  Trouble is that also changes the semantics... :-)

svaarala:
What I still find puzzling is the special handling for a zero length input but explicit totalLength:

```
$ nodejs
> process.version
'v6.7.0'
> x = Buffer.concat([], 10)
<Buffer >
> x.length
0
> x = Buffer.concat([ new Buffer('') ], 10)
<Buffer 00 00 00 00 00 00 00 00 00 00>
> x.length
10
```

But I've kept that behavior because it's apparently intentional.

Issue #1005, opened 2016-10-10
Remove default bindings converting a buffer 1:1 to an internal string representation


svaarala:
In Duktape 1.x and in current master there are several Ecmascript built-in bindings which accept a buffer and convert it to a string by simply using bytes as-is for the string internal representation.

Because standard strings are represented internally as CESU-8, this conversion allows Ecmascript code to create non-standard strings involving UTF-8 codepoints, invalid UTF-8/CESU-8, etc. It also allows creation of internal strings from script code alone. It also exposes the internal string representation to Ecmascript code which can sometimes be confusing (and is unportable).

In some sense this has been a feature: Duktape allows C code to push arbitrary byte sequences into the value stack and then later read them out. Unless one operates on those strings with built-ins Duktape internals don't really care about the bytes in the strings, and it's supported to, for example, push an ISO-8859-1 string as a property value and later read it out.

The proposed change would be to remove this ability by default from the Ecmascript environment but leave it for C code which would:
- Avoid exposing the internal string representation to Ecmascript code via default built-ins. User code would of course be free to add such bindings if they made sense for the application.
- Prevent Ecmascript code from creating internal strings via default built-ins. This would be a sandboxing improvement regardless of #979. Again, application would be free to add bindings "subverting" this feature.

The necessary concrete changes would be:
- [x] Go through all internal occurrences of `duk_buffer_to_string()` which is the only direct API call converting a string 1:1 to a buffer. Clear cases are:
  - [x] `Buffer.toString()`, which should be changed to decode the buffer as strict UTF-8 and encode the codepoints into CESU-8, emitting surrogate pairs for non-BMP codepoints and using replacement characters for invalid sequences. #1020
  - [x] `String.fromBuffer()`, which can be removed entirely.
  - [x] Standard string-to-buffer conversions would be provided by the Encoding API.
- [x] Go through any other internal places where strings are pushed and ensure the data is guaranteed to be valid CESU-8.
- [x] For consistency, go through all places where a string is converted into a buffer. While that's not dangerous as such, comment in code why that's the case. Another option is that it should be the inverse of a buffer-to-string conversion for consistency. For example:
  - [x]  `new Buffer(strValue)` should be changed to decode the string as CESU-8, combining surrogates and using replacement characters for invalid sequences, and encode the resulting codepoints as strict UTF-8.
- [x] To allow unit tests to run, `duk` should provide some global binding to do the equivalent of current `String.fromBuffer()`.
[...]

PR #1006, opened 2016-10-11
Remove String.fromBuffer() and implement it in "duk" for testcases

svaarala:
Implements parts of #1005.


svaarala:
Hmm, it seems Github just recently changed the status check layout so that the descriptive string is really short now. Hopefully this is not permanent, otherwise I'll need to rework the descriptions to say something useful in a much shorter space :/

fatcerberus:
That status check issue looks like it might be a UI bug.  Tests with longer names (like travis) have practically no room at all for the status string.

PR #1008, opened 2016-10-11
Change Node.js Buffer to inherit from Uint8Array

svaarala:
- [x] Change internal prototype of Buffer.prototype from Object.prototype to Uint8Array.prototype
- [x] Fix `nodejsBuffer.subarray()` which should create a Uint8Array but now creates another Node.js Buffer; similarly for all other inherited Uint8Array built-ins
- [x] Fix Node.js Buffer creation sites to create Uint8Arrays which inherit from Buffer.prototype
- [x] Fix initializer handling for Node.js buffer constructor
- [x] Fix Node.js .slice() handling in shared slice code
- [x] Remove internal Node.js Buffer class, it's no longer a real distinction (just the inheritance chain)
- [x] Keep the `DUK_BUFOBJ_NODEJS_BUFFER` constant in C API to allow pushing Node.js Buffers
- [x] Recheck `Object.prototype.toString.call(new Buffer('abcd'))` - Node.js v6.7.0 prints `[object Uint8Array]`
- [x] Create a `.buffer` property for Node.js Buffers, also needs to create an ArrayBuffer instance
- [x] Add testcase for handling of objects inheriting from typed array views
- [x] 2.0 migration notes
- [x] Releases entry


svaarala:
After trivially changing the inheritance chain:

```
duk> buf = new Buffer('ABCD')
= {type:"Buffer",data:[65,66,67,68]}
duk> console.log(buf instanceof Buffer, buf instanceof Uint8Array)
true true
= undefined
```

However this is incorrect:
```
duk> view = buf.subarray(1)
= {type:"Buffer",data:[66,67,68]}
duk> console.log(view instanceof Buffer, view instanceof Uint8Array)
true true
= undefined
```

The internal `.subarray()` handling should walk the prototype chain to find that we're actually dealing with an `Uint8Array` so that an Uint8Array is returned from .subarray(). Here's Node.js v6.7.0 output:

```
> buf = new Buffer('ABCD')
<Buffer 41 42 43 44>
> console.log(buf instanceof Buffer, buf instanceof Uint8Array)
true true
undefined
> view = buf.subarray(1)
Uint8Array [ 66, 67, 68 ]
> console.log(view instanceof Buffer, view instanceof Uint8Array)
false true
undefined
```

In fact, current typed array code doesn't handle objects inheriting from typed array instances correctly. Specifically, when the `this` binding is such an object, the code assumes we can use the class of the `this` binding to figure out what kind of typed array view to create e.g. in `.subarray()`. This needs a separate bug issue but probably no fix for 1.x if the fix involves a lot of diff.

svaarala:
Hmm. ES6 seems to require that e.g. `.subarray()` does a `[[TypedArrayName]]` internal slot lookup for the "this" binding (http://www.ecma-international.org/ecma-262/6.0/#sec-%typedarray%.prototype.subarray). Internal slots are not inherited, so calling a `.subarray()` on a subclass instance should not work. And it doesn't in Node.js v6.7.0:

```
> x = new Uint8Array(10)
Uint8Array [ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ]
> y = {}; Object.setPrototypeOf(y, x);
Uint8Array {}
> y.subarray(1)
TypeError: Method get TypedArray.prototype.subarray called on incompatible receiver [object Object]
    at Uint8Array.subarray (native)
    at repl:1:3
    at sigintHandlersWrap (vm.js:22:35)
    at sigintHandlersWrap (vm.js:96:12)
    at ContextifyScript.Script.runInThisContext (vm.js:21:12)
    at REPLServer.defaultEval (repl.js:313:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:513:10)
    at emitOne (events.js:101:20)
```

So, Duktape behavior seems to be correct after all:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-717-ge9bb3e2-dirty)
duk> x = new Uint8Array(10)
= |00000000000000000000|
duk> y = {}; Object.setPrototypeOf(y, x);
= {}
duk> y.subarray(1)
TypeError: not buffer
    at [anon] (duk_bi_buffer.c:171) internal
    at subarray () native strict preventsyield
```

So apparently what Node.js does is that Buffers are Uint8Array instances directly (they have the same typed array name slot):

```
> Object.prototype.toString.call(new Buffer('ABCD'))
'[object Uint8Array]'
```

But most likely their internal prototype is forced to be `Buffer.prototype`.

svaarala:
The reason why this doesn't happen at present is that Node.js Buffers are implemented at the same level as typed arrays in Duktape, i.e. they're all `duk_hbufobj` instances. The `.subarray()` lookup thus finds a valid `duk_hbufobj` whose internal class number is "Node.js Buffer" instead of "Uint8Array".

One fix approach would be to remove the internal Node.js Buffer type distinction:
- When pushing Node.js Buffers, use the Uint8Array internal class but set internal prototype to Buffer.prototype.
- Remove the `DUK_HOBJECT_CLASS_BUFFER` internal class number.

svaarala:
This is close to complete now:
- Removed internal class for Node.js Buffer
- Changed creation sites to create Uint8Array instance with Buffer.prototype as internal prototype (Buffer constructor now shares typed array constructor code), and a .buffer property pointing to an ArrayBuffer as required
- Changed Buffer.prototype inheritance chain

Still needs release notes etc.

With this change plain buffers and ArrayBuffers are the only actual buffer types while the rest are views: Node.js Buffer, DataView, and typed arrays. This makes it more relevant to maybe rework the implementation so that buffers and views are distinct internally and that the `.buffer` reference is internal rather than an explicit property (it needs to be an inherited getter in ES6 anyway). Doing so would improve memory efficiency a little bit.

fatcerberus:
> But most likely their internal prototype is forced to be Buffer.prototype.

Changing the internal prototype is not the same thing as `Object.setPrototypeOf(x, newProto)`?

svaarala:
> Changing the internal prototype is not the same thing as Object.setPrototypeOf(x, newProto)?

That's the same thing and sets the internal prototype.

fatcerberus:
I looked at the Node.js source and it seems that `Buffer` inherits from `FastBuffer` which inherits directly from Uint8Array:
https://github.com/nodejs/node/blob/master/lib/buffer.js#L9-L16

Figuring out the exact semantics will likely require looking at the ES6 spec to see how class inheritance works.

svaarala:
There's really not much alternative in practice: unless Buffer is internally a Uint8Array directly (and not an instance inheriting from an Uint8Array *instance*) some of the Uint8Array.prototype methods will reject the "this" binding as they require the "this" binding to be a typed array directly (instead of something that has a typed array in its internal prototype chain).

This seems clear to me from the ES6 specification because the methods look up an internal typed array name slot - and internal slots are not inherited (they're not actual properties).

svaarala:
The inheritance chain is apparent from this:

```
> x = new Buffer('abcd')
<Buffer 61 62 63 64>
> x.__proto__ === Buffer.prototype
true
> x.__proto__.__proto__ === Uint8Array.prototype
true
```

So: instance -> Buffer.prototype -> Uint8Array.prototype. From that point on there's a generic typed array prototype (which is unnamed, at least it seems so, in ES6) and finally Object.prototype::

```
> x.__proto__.__proto__.__proto__
TypedArray {}
> x.__proto__.__proto__.__proto__.__proto__ === Object.prototype
true
```

So with these two the only remaining approach is to create a normal Uint8Array and forcibly set its internal prototype to Buffer.prototype. The Duktape internals do this in one step in practice (in this pull, that is).

fatcerberus:
Unrelated, I like the way Node.js summarizes buffers, with the bytes separated by spaces.  Any chance `duk` could do the same?  Well, I guess it's technically JX that handles that...

svaarala:
Yep, it's JX now. As discussed when the change was made what "duk" should really use is some sort of fancier introspection.

Rather than build that into Duktape itself (which isn't really feasible because every context and user will have a different idea of what the correct/best introspected string ig :-), there should be enough APIs for "duk" to do a good job of that in whatever manner suits it best (coloring output etc).

For the summary API call the plan would be to group the bytes in buffers and views based on the element size. So, for example, for Uint16Array:

```
> new Uint16Array([ 0xdead, 0xbeef ])
[Uint16Array adbe efbe]
```

svaarala:
Turns out the changes broke at least Node.js Buffer `.slice()` which needs some special attention to get the result internal prototype to come out right. I'll fix this soon.

PR #1009, opened 2016-10-12
Allow RAM built-in accessors to have a missing setter/getter

svaarala:
- [x] Genbuiltins change to allow missing setter/getter, reserve natidx 0 for that
- [x] RAM init code change to special case missing function
- [x] Enable DUK_USE_ASSERTIONS for dukd config (mistakenly disabled at some point)


fatcerberus:
@svaarala Thanks again for the quick fix, it's appreciated as always. :D

svaarala:
Hopefully this fixes the Encoding API issue - I did test this manually by removing the setter and/or getter from `Error.prototype.stack` and seems to work OK.

PR #1010, opened 2016-10-12
Remove internal helper for defining property accessors

svaarala:
Use `duk_def_prop()` instead which reduces footprint.


svaarala:
Another 170 bytes bites the dust with no loss of functionality ;)

PR #1013, opened 2016-10-14
Minor cleanups and improvements to Encoding API implementation

svaarala:
Work in progress.


svaarala:
Here's a performance baseline from master:

```
test-textencoder-ascii.js           : duk.O2  5.69
test-textencoder-nonascii.js        : duk.O2 11.27
test-textdecoder-ascii.js           : duk.O2  2.96
test-textdecoder-nonascii.js        : duk.O2  3.39
```

Encoder testcases process a string of 16M codepoints 50 times (= 800M codepoints). Decoder testcases process a Uint8Array of 16M bytes 50 times (= 800MB). Encoding is much slower than decoding now because encoding relies on `duk_decode_string()`.

fatcerberus:
Realistically, going forward we will probably need some dedicated CESU-8<->UTF-8 transcoding logic.

tests/ecmascript/test-bi-textencoder.js
```diff
@@ -4,7 +4,7 @@
 
 // determine if two arrays contain the same values, used here to compare buffers
 function equiv(arr1, arr2) {
-    if (arr1.prototype !== arr2.prototype) {
+    if (Object.getPrototypeOf(arr1) !== Object.getPrototypeOf(arr2)) {
```
fatcerberus: Oops, meant to fix this before the merge.  Good that you caught it.

src-input/duk_bi_encoding.c
```diff
@@ -242,39 +283,35 @@ DUK_INTERNAL duk_ret_t duk_bi_textdecoder_constructor(duk_context *ctx) {
...
 	duk_get_prop_string(ctx, -1, "\xff" "Context");
 	dec_ctx = (duk__decode_context *) duk_require_buffer(ctx, -1, NULL);
 	DUK_ASSERT(dec_ctx != NULL);
+	return dec_ctx;
+}
 
-	/* XXX: change to look up from dec_ctx once other encodings are supported */
-	DUK_UNREF(dec_ctx);
+DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_encoding_getter(duk_context *ctx) {
+	/* Encoding is now fixed, so no need to look up _Context. */
```
fatcerberus: I originally left that fake `dec_ctx` lookup in just in case someone calls the getter directly, I still wanted it to TypeError if `this` was something other than a TextDecoder instance.  I guess that's not very footprint friendly though, and admittedly a bit of a contrived use case.

src-input/duk_bi_encoding.c
```diff
@@ -115,34 +131,53 @@ DUK_LOCAL void duk__utf8_encode_char(void *udata, duk_codepoint_t codepoint) {
...
+			} else {
+				/* consecutive high surrogates, consider first one unpaired */
+				codepoint = DUK_UNICODE_CP_REPLACEMENT_CHARACTER;
+			}
 		} else {
-			/* unpaired low surrogate */
-			enc_ctx->out += duk_unicode_encode_xutf8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, enc_ctx->out);
+			/* low surrogate */
+			if (enc_ctx->lead != 0x0000L) {
+				codepoint = 0x010000L + ((enc_ctx->lead - 0xd800L) << 10) + (codepoint - 0xdc00L);
```
fatcerberus: Does `+` generate better code than `|`?  Genuinely curious.

src-input/duk_bi_encoding.c
```diff
@@ -197,15 +235,17 @@ DUK_INTERNAL duk_ret_t duk_bi_textencoder_prototype_encode(duk_context *ctx) {
...
-		enc_ctx.out += duk_unicode_encode_xutf8(DUK_UNICODE_CP_REPLACEMENT_CHARACTER, enc_ctx.out);
+		DUK__EMIT_UTF8_REPL_CHAR(&enc_ctx);
 		DUK_ASSERT(enc_ctx.out <= output + (3 * len));
 	}
 
+	/* The output buffer is usually very much oversized, so shrink it to
+	 * actually needed size.
+	 */
+	DUK_ASSERT_TOP(ctx, 2);
+	duk_resize_buffer(ctx, -1, (duk_size_t) (enc_ctx.out - output));
```
fatcerberus: At first I thought this was an unnecessary step if bufferobjects are enabled.  But it's actually a memory hog if you don't since the plain buffer will be the concrete backing for the Uint8Array.

src-input/duk_bi_encoding.c
svaarala: In this case seemed to be the case, most likely it allows the 0x10000 and -0xd800 to be computed at compile time. For OR that might not be possible because the OR counterparts are variables whose range is not necessarily known by the compiler.

svaarala: A quick test program shows the difference; with bitwise ops:

```
  400461:       2d 00 d8 00 00          sub    $0xd800,%eax
  400466:       81 ee 00 dc 00 00       sub    $0xdc00,%esi
  40046c:       c1 e0 0a                shl    $0xa,%eax
  40046f:       81 ce 00 00 01 00       or     $0x10000,%esi
  400475:       09 c6                   or     %eax,%esi
```

With addition:

```
  400461:       c1 e0 0a                shl    $0xa,%eax
  400464:       8d b4 02 00 24 a0 fc    lea    -0x35fdc00(%rdx,%rax,1),%esi
```

The key issue is that with addition the addition and subtractions can be precomputed. With OR the components themselves involve subtraction which doesn't combine well during compile time.

svaarala:
So far around 150 bytes of footprint improvement.

I won't do anything very radical in this branch, just some low hanging fruits especially for footprint. Actual improvement for encoding/decoding performance means a table driven approach, but it'd be nice to have more comprehensive test case coverage first.

fatcerberus:
I think I did something wrong - under minisphere debugger (SSJ) TextDecoder/Encoder.prototype methods show as non-enumerable.  Might be worth double checking that while this pull is open.

svaarala:
```
duk> Object.getOwnPropertyDescriptor(TextEncoder.prototype, 'encode')
= {value:{_func:true},writable:true,enumerable:false,configurable:true}
```

It's more likely to be the tooling though, the metadata seems correct.

svaarala:
Yep, seems to be the RAM built-ins limitations. With a ROM build:

```
duk> Object.getOwnPropertyDescriptor(TextEncoder.prototype, 'encode')
= {value:{_func:true},writable:true,enumerable:true,configurable:false}
```

(It's not configurable because ROM objects are forced to be non-configurable to signal their read-only nature; they can't be made non-writable however because that would affect also writes to inheriting RAM objects.)

src-input/duk_bi_encoding.c
```diff
@@ -9,91 +9,103 @@
...
-#define DUK__U8_CONTINUE   0  /* continue to next byte */
-#define DUK__U8_CODEPOINT  1  /* have codepoint */
-#define DUK__U8_ERROR      2  /* decoding error */
-#define DUK__U8_RETRY      3  /* retry last byte, constraint: RETRY > ERROR */
+/* The signed duk_codepoint_t type is used to signal a decoded codepoint
+ * (>= 0) or various other states using negative values.
+ */
+#define DUK__CP_CONTINUE   (-1)  /* continue to next byte, no completed codepoint */
+#define DUK__CP_ERROR      (-2)  /* decoding error */
+#define DUK__CP_RETRY      (-3)  /* decoding error; retry last byte, constraint: RETRY > ERROR */
```
svaarala: I was a bit on the fence about this change: it doesn't affect performance on modern processors (but probably does improve a little bit on ARM etc), but it does reduce footprint a bit (total footprint reduction is around 350 bytes now on x64), so it seemed to make sense.

But assuming that the same code would be used to handle CESU-8 decoding too, which would be good for footprint, it would make sense to revert this particular change. Otherwise two decoders or some other hackery would be required which might increase footprint overall.

For performance oriented builds the decoding should be table oriented anyway (and thus handled by different code) so IMO the master baseline should be as footprint friendly as possible, and then have best performance possible within that constraint.

svaarala:
@fatcerberus This is still work in progress - there's no hurry to merge this unless it improves master overall. I'll finish up tomorrow (for the time being) and will open an issue with various issues and notes for future improvement.

The pull also has some Encoding API unrelated improvements at present, e.g. better error message for non-constructor calls to built-ins requiring one:

```
# Master
((o) Duktape [linenoise] 1.99.99 (v1.5.0-758-g84f462f)
duk> Proxy()
TypeError: type error (rc -6)
    at Proxy () native strict preventsyield
duk> 

# This pull
((o) Duktape [linenoise] 1.99.99 (v1.5.0-762-ge340b9b)
duk> Proxy()
TypeError: constructor must use 'new'
    at [anon] (duk_api_call.c:483) internal
    at Proxy () native strict preventsyield
```

fatcerberus:
Oh good, you did fix that.  If you hadn't I was eventually going to open an issue about it when I got annoyed enough with the generic type errors. ;-)

fatcerberus:
While I'm thinking about it, how feasible would it be (with the current parser/compiler), to implement support for ES6 `new.target`?

svaarala:
To be sure, the generic error messages are not a bug but a feature for low memory targets ... :)

In fact the best underlying idiom for e.g. a non-constructor call to a constructor requiring one would be something like:

```c
if (!duk_is_constructor_call(ctx)) {
#if !defined(DUK_USE_VERBOSE_ERRORS)
    return DUK_RET_TYPE_ERROR;
#else
    DUK_ERROR_TYPE(thr, "bla bla");
    return 0;  /* unreachable */
#endif
}
```

The idiom is unfortunately different for a Duktape/C function compared to an ordinary function because the return value shorthand is only possible inside Duktape/C functions.

Packaging that idiom into easy-to-use (and easy-to-read) macros is the main question in serving both low memory and verbose builds at the same time.

svaarala:
Re: `new.target`, it should be quite straightforward for normal functions, it would require a new bytecode opcode though. Bound functions might pose some issues (not sure) with current data model.

However, it's also visible to outer scopes (from Mozilla):

> The new.target property is a meta property that is available to all functions. In arrow functions, new.target refers to the new.target of the surrounding function.

Arrow functions pose other issues for compiling too, but the above means `new.target` cannot be simply a function local feature, it has some scope visibility too.

svaarala:
To clarify my point re: generic error messages, I'd be very much in favor of improving them but simultaneously ensuring that the Duktape/C error return shorthand would be used for low memory builds wherever reasonably possible (this would improve footprint from current because the shorthand is not always used even where possible right now).

So, with a good macro wrapper there would be little impact on call sites but simultaneously better messages for default build and smaller footprint for low memory build.

fatcerberus:
> To be sure, the generic error messages are not a bug but a feature for low memory targets ... :)

Yes, I knew that; we discussed it at length during the error message reworks a while back.  It was just weird to see such a minimal error with verbose errors enabled. :)

fatcerberus:
@svaarala Anyway, no rush on getting this merged.  I can get by perfectly fine with with what's in master now.  I mainly needed TextDecoder so I can rewrite the Sphere legacy map engine, which uses binary file formats, in JS.

src-input/duk_strings.h
```diff
@@ -61,7 +62,6 @@
 #define DUK_STR_DECODE_FAILED                    "decode failed"
```
fatcerberus: Since this string will now be used by TextDecoder and not only internally in e.g. `duk_compile()`, maybe it could be clarified: "text decoding error".  Then there could also be a verbose version that includes the name of the encoding, e.g. "unable to decode text as utf-8"

svaarala:
>  It was just weird to see such a minimal error with verbose errors enabled. :)

In practical terms any code that doesn't specifically provide separate verbose and minimal errors will default to "something" which will be inappropriate for the other environment. So any *verbose* error is a surprise for a minimal target... :)

At present most call sites without explicit verbose/minimal variants are mostly implemented so that:
- If the error is common and expected, the message is relatively verbose (a string etc).
- If the error is rare and not expected, shorthands are used. For example, if you call `ArrayBuffer.prototype.slice()` with a `Date` instance, there isn't much point in defaulting to a verbose error.

Of course ideally every throw site would provide both. But code is rarely ideal because there isn't an infinite amount of effort to spend on it ;)

svaarala:
Current macro setup is pretty OK for non-return-shorthand cases so that:

```c
DUK_ERROR_TYPE(thr, "foobar");
```

Compiles to a helper which assumes the error type (`DUK_ERR_TYPE_ERROR`) so that it doesn't copied to every call site. For low memory targets that essentially compiles to a direct noreturn function call with only one argument, `thr`, which is about as good as it gets assuming a longjmp will require access to `thr` which is hard to avoid.

So any call sites using the `DUK_ERROR_xxx()` macros generate relatively good code for verbose and minimal targets.

What is missing is some macro setup for the return shorthand. While the different isn't huge, a `return -1;` is still more efficient than setting up for a function call etc. But as an intermediate measure it might make sense to:
- Ensure all internal `duk_error()` call sites are converted to macro based to provide verbose/minimal behavior. I think that's actually the case now (I fixed a few calls in this pull).
- Remove return shorthands throughout in favor of `DUK_ERROR_xxx()` calls and reintroduce them when a macro setup is available. Or implement the shorthand support at the same time.

fatcerberus:
Something to check: Is TextDecoder behavior correct when `DUK_USE_ZERO_BUFFER_DATA` is disabled?  I don't remember if all values are properly initialized.

svaarala:
I have a cleanup note of that too - I think they are, but the initialization is split into decoding parameters (ignoreBOM, fatal) and decoding state. It deserves a comment but seemed correct to me.

svaarala:
Remaining issues described in #1017.

svaarala:
I'm finished with this pull - apart from the duk__decode_next() return value convention change. Testcase improvements etc in separate pulls. Final footprint improvement for x64 default build is around 350 bytes, and the Encoding API is now disabled by default for minimal and stripped build (which saves around 2kB).

svaarala:
Changed the constructor error message from:

    constructor must use 'new'

to:

    constructor requires 'new'

which seems more intuitive because it's caller (not the constructor) that must use 'new'.

fatcerberus:
Agree with that change to the error message - the original wording seemed somehow awkward but I couldn't come up with anything better while still being concise.

src-input/duk_bi_encoding.c
```diff
@@ -165,47 +208,80 @@ DUK_INTERNAL duk_ret_t duk_bi_textencoder_prototype_encoding_getter(duk_context
...
+		/* The output buffer is usually very much oversized, so shrink it to
+		 * actually needed size.
+		 */
+		DUK_ASSERT_TOP(ctx, 2);
+		final_len = (duk_size_t) (enc_ctx.out - output);
+		duk_resize_buffer(ctx, -1, final_len);
 	}
 
+	/* Pointer stability assumption. */
+	DUK_ASSERT(output == (duk_uint8_t *) duk_get_buffer_data(ctx, -1, NULL));
```
fatcerberus: This assert is likely to fail because you just resized the buffer above.

src-input/duk_bi_encoding.c
```diff
@@ -215,11 +291,10 @@ DUK_INTERNAL duk_ret_t duk_bi_textdecoder_constructor(duk_context *ctx) {
...
 	duk_bool_t ignore_bom = 0;
 
-	if (!duk_is_constructor_call(ctx)) {
-		duk_error(ctx, DUK_ERR_TYPE_ERROR, "constructor must use 'new'");
-	}
+	DUK_ASSERT_TOP(ctx, 2);
+	duk_require_constructor_call(ctx);
 	if (!duk_is_undefined(ctx, 0)) {
-		/* XXX: for now, 'label' argument is ignored */
+		/* XXX: For nor ignore 'label' (encoding identifier). */
```
fatcerberus: "For nor" - typo

src-input/duk_bi_encoding.c
```diff
@@ -231,81 +306,77 @@ DUK_INTERNAL duk_ret_t duk_bi_textdecoder_constructor(duk_context *ctx) {
...
+	dec_ctx = duk__get_textdecoder_context(ctx);
+
 	if (!duk_is_undefined(ctx, 0)) {
 		input = (const duk_uint8_t *) duk_require_buffer_data(ctx, 0, &len);
+	} else {
+		input = NULL;  /* NULL is OK if 'len' is 0. */
+		DUK_ASSERT(len == 0);
 	}
 	if (!duk_is_null_or_undefined(ctx, 1)) {
 		if (duk_get_prop_string(ctx, 1, "stream")) {
```
fatcerberus: As we figured out in #1017, this should have a check for Type(arg2) = Object, and throw a TypeError if not.

fatcerberus: Also, same deal in TextDecoder constructor for the second argument.

src-input/duk_bi_encoding.c
svaarala: Oops :) I'd like to keep the assert so I'll relocate it a bit.

src-input/duk_bi_encoding.c
svaarala: Right, I'll fix these two.

svaarala: By the way, there's this internal helper for this just case:

    DUK_INTERNAL_DECL duk_hobject *duk_require_hobject_accept_mask(duk_context *ctx, duk_idx_t idx, duk_uint_t type_mask);

It will accept an object (returning its pointer) *or* accept any types in the accept mask (returning NULL).

fatcerberus: Thanks, that'll come in handy in the future.  I guess the closest thing to that in the public API would be to check if `duk_get_heapptr()` returns NULL?

svaarala: That's close but not correct: plain strings are heap allocated but still not objects so they wouldn't be handled correctly.

I've tried to add internal helpers where some idiom repeats a lot - the primary purpose is to reduce code size/duplication and thus footprint, but a secondary purpose is to get an idea what's convenient and what's not. That may then help deciding what new public APIs would make sense. What I don't want to do is expose stuff prematurely :)

svaarala: Bah, actually that helper won't help here because there are three different handling cases: special treatment or undefined/null, accept certain other types, and TypeError the rest.

svaarala:
Ok, added testcase coverage for `.decode() first and second arguments and fixed the implementation for the second argument.

svaarala:
For `.encode()` Firefox and Chrome behavior seems to be that `undefined` has special handling, and anything else is `ToString()` coerced. For example:

```
> new TextEncoder().encode(12)
Uint8Array [ 49, 50 ]
> new TextEncoder().encode(new ArrayBuffer(10))  // ToString: [object ArrayBuffer]
Uint8Array [ 91, 111, 98, 106, 101, 99, 116, 32, 65, 114, 10 more… ]
```

Current implementation matches that so I'll just add tests to check for the behavior.

fatcerberus:
I suspect the Firefox and Chrome implementations just do the equivalent of:

```js
if (input === undefined) {
    input = "";
}
input = String(input);
```

Which will be true both if the argument isn't present or if the caller actually passes `undefined`.  Not sure I would call that "special handling", as it's a very common JS idiom. ;)

Agree with adding a test for it in any case.

svaarala:
I don't think that's very common in ES5/ES6 bindings - they more commonly do strict type checking or do a blind ToString() coercion. In some places null and undefined are treated the same (e.g. rejected). For example:

```
> new String(void 0)
[String: 'undefined']
```

svaarala:
It's also relatively common in ES5 to treat undefined and no-argument differently, such as String:

```
> new String()
[String: '']
> new String(void 0)
[String: 'undefined']
```

But then:

```
> new RegExp()
/(?:)/
> new RegExp(void 0)
/(?:)/
> new RegExp(null)
/null/
```

which behaves like .encode().

svaarala:
But in any case, I meant "special handling" as in check for `undefined` explicitly rather than doing a ToString() for the argument ... :)

svaarala:
@fatcerberus Once the tests pass I'll merge this to master - the pull doesn't address all of the issues #1017 (which isn't intended to be a short term "must fix" list either) but does fix some of them. Some issues will still need fixing for a 2.x release (like documentation and more testcase coverage).

Changing the duk__decode_next() argument convention back to its original form (or maybe some other form) can be done as CESU-8 support is added. The form in this pull has a smaller footprint so I prefer that as a baseline in master.

svaarala:
Once this is in master, I think I'll tackle Node.js Buffer `ToString()` coercion next so that there'd be no more 1:1 coercion primitives in the binding (at least not obvious ones). The quickest initial implementation would be to essentially call `duk_bi_textencoder_prototype_encode()`.

In fact that function could otherwise be used directly to provide the `.toString()` method except that with Encoding API the argument is an explicit one, while for Node.js .toString() it is the `this` binding. This is easy to fix with a wrapper call (push this, tail call the encode() method) or by adding a magic value to .encode() controlling where the argument is read from.

With that `.toString()` would accept any buffer-like types as the `this` binding: this is not how Node.js operates but Duktape currently allows lenient mixing of types. Node.js behavior is strict:

```
> Buffer.prototype.toString.call(new Buffer('abcd'))
'abcd'
> Buffer.prototype.toString.call(new Uint8Array([ 0x41, 0x42 ]))
TypeError: this.utf8Slice is not a function
    at Uint8Array.Buffer.toString (buffer.js:483:19)
    at repl:1:27
    at sigintHandlersWrap (vm.js:22:35)
    at sigintHandlersWrap (vm.js:96:12)
    at ContextifyScript.Script.runInThisContext (vm.js:21:12)
    at REPLServer.defaultEval (repl.js:313:29)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:513:10)
    at emitOne (events.js:101:20)
```

More details in a separate pull :)

svaarala:
Above meant .decode() rather than .encode() :)

fatcerberus:
Looking at that REPL log it seems that Node.js isn't being deliberately strict but rather that the plain Uint8Array doesn't have the necessary method(s) for it to work.  In other words it doesn't "quack like a duck" enough :)

Duktape being lenient about mixing types is definitely one of its strong points IMO.  Duck typing is one of Javascript's greatest assets and what makes it such a joy to use for me.

PR #1016, opened 2016-10-15
Verbose errors also for return shorthand sites

svaarala:
Current macro setup handles verbose/minimal errors pretty well for cases such as:

```c
DUK_ERROR_TYPE_INVALID_ARGS(thr);
DUK_ERROR_TYPE(thr, "Your AMIGA is alive !!!");
```

For verbose errors this generates a good message. For minimal errors this generates a noreturn function call which only takes `thr` as its argument which is about as good as it gets.

However, Duktape/C functions using return shorthand (e.g. `return DUK_RET_TYPE_ERROR`) don't provide good verbose messages.

This pull adds some error macros to improve the situation so that the macro resolves to an actual error call when verbose errors are used but uses e.g. `return DUK_RET_TYPE_ERROR` when minimal errors are used.


svaarala:
This is what a minimal footprint `DUK_ERROR_TYPE()` call site looks like:

```
  406a21:       48 89 df                mov    %rbx,%rdi
  406a24:       e8 20 58 00 00          callq  40c249 <duk_err_type>
```

The return shorthand is:

```
  402b87:       b8 fa ff ff ff          mov    $0xfffffffa,%eax
  402b8c:       c3                      retq   
```

The difference is only 2 bytes on x64 so the current advantage of using them is 100 bytes (50 return sites). Not very much.

In fact, if the whole return value shorthand feature was removed (I'm not sure it's a very good feature because it cannot be used in helper functions that are called by a Duktape/C function for example) it'd most likely make up that 100 bytes and more.

So one approach would be to remove the shorthand entirely and ensure current shorthand sites compile to tiny error helper calls for low memory builds.

svaarala:
The pull replaces all direct internal uses of `DUK_RET_xxx` with error macros `DUK_DCERROR_xxx` which either map to a shorthand error return or to an actual error call. This improves error messages for the default build, e.g. from:

```
duk> Proxy()
TypeError: type error (rc -6)
    at Proxy () native strict preventsyield
duk> String.prototype.toString.call(1)
TypeError: type error (rc -6)
    at toString () native strict preventsyield
    at call () native strict preventsyield
```

to:

```
duk> Proxy()
TypeError: invalid args
    at [anon] (duk_bi_proxy.c:13) internal
    at Proxy () native strict preventsyield
duk> String.prototype.toString.call(1)
TypeError: invalid args
    at [anon] (duk_bi_string.c:128) internal
    at toString () native strict preventsyield
    at call () native strict preventsyield
```

The affected errors are mostly (but not always) in rarely executed type check code paths etc. The error messages are not necessarily very descriptive but should be better than before.

The pull also simplifies the automatic error messages generated for `return DUK_RET_xxx` sites from e.g. `uri error (rc XXX)` to `error (rc XXX)`: the shorthand error message doesn't matter much as it's only relevant for low memory code anyway. There are also some trivial error message text changes.

Footprint impact is around +370 bytes for x64 default build, -200 bytes for x64 minimal build, and -100 bytes for x64 stripped build.

PR #1019, opened 2016-10-16
Implement ES6 Promise support

fatcerberus:
_Work in progress._

Implement support for ES6+ Promises.  Promises allow async code to be written in a linear fashion like sync code.

Checklist:
- [ ] Implement `Promise`, all ES7 bindings
- [ ] C API for promises?
- [ ] `Promise.prototype.done()`?  Investigate pros and cons, as many devs dislike it
- [ ] If async resolve implemented here, need an API to dispatch Jobs, e.g. `duk_perform_jobs()`
- [ ] Testcases
- [ ] RELEASES entry

### DIFF
--- a/config/config-options/DUK_USE_PROMISE_BUILTIN.yaml
+++ b/config/config-options/DUK_USE_PROMISE_BUILTIN.yaml
@@ -1,0 +1,9 @@
+define: DUK_USE_PROMISE_BUILTIN
+introduced: 2.0.0
+default: true
+tags:
+  - ecmascript
+  - ecmascript6
+description: >
+  Provide a Promise built-in which allows Ecmascript code to create and use ES6
+  Promise objects.
--- a/src-input/builtins.yaml
+++ b/src-input/builtins.yaml
@@ -265,6 +265,12 @@ objects:
         #present_if: DUK_USE_DUKTAPE_BUILTIN
 
       # ES6
+      - key: "Promise"
+        value:
+          type: object
+          id: bi_promise_constructor
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
       - key: "Proxy"
         value:
           type: object
@@ -2629,6 +2635,98 @@ objects:
   #
   #  ES6
   #
+
+  - id: bi_promise_constructor
+    class: Function
+    internal_prototype: bi_function_prototype
+    native: duk_bi_promise_constructor
+    callable: true
+    constructable: true
+    es6: true
+    bidx: true
+    present_if: DUK_USE_PROMISE_BUILTIN
+
+    properties:
+      - key: "length"
+        value: 1
+        attributes: ""
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+      - key: "prototype"
+        value:
+          type: object
+          id: bi_promise_prototype
+        attributes: ""
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+      - key: "name"
+        value: "Promise"
+        attributes: ""
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+      - key: "all"
+        value:
+          type: function
+          native: duk_bi_promise_all
+          length: 1
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+      - key: "race"
+        value:
+          type: function
+          native: duk_bi_promise_race
+          length: 1
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+      - key: "reject"
+        value:
+          type: function
+          native: duk_bi_promise_resolve_shared
+          magic: 1
+          length: 1
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+      - key: "resolve"
+        value:
+          type: function
+          native: duk_bi_promise_resolve_shared
+          magic: 0
+          length: 1
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+
+  - id: bi_promise_prototype
+    class: Object
+    internal_prototype: bi_object_prototype
+    es6: true
+    bidx: true
+    present_if: DUK_USE_PROMISE_BUILTIN
+
+    properties:
+      - key: "constructor"
+        value:
+          type: object
+          id: bi_promise_constructor
+        attributes: "wc"
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+      - key: "catch"
+        value:
+          type: function
+          native: duk_bi_promise_prototype_catch
+          length: 1
+        attributes: "wc"
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+      - key: "then"
+        value:
+          type: function
+          native: duk_bi_promise_prototype_then
+          length: 2
+        attributes: "wc"
+        es6: true
+        present_if: DUK_USE_PROMISE_BUILTIN
+
 
   - id: bi_proxy_constructor
     class: Function
--- a/src-input/duk_bi_promise.c
+++ b/src-input/duk_bi_promise.c
@@ -1,0 +1,307 @@
+/*
+ *  Promise built-ins (ES7 Section 25.4)
+ *  http://www.ecma-international.org/ecma-262/7.0/#sec-promise-objects
+ *
+ *  Also refer to the Promises/A+ specification, which Ecmascript Promises are
+ *  based on:
+ *  https://promisesaplus.com/
+ */
+
+#include "duk_internal.h"
+
+#define DUK__PS_PENDING (0)
+#define DUK__PS_FULFILL (1)
+#define DUK__PS_REJECT  (2)
+
+typedef struct {
+	duk_uint8_t handled;
+	duk_uint8_t state;
+} duk__promise;
+
+DUK_LOCAL void duk__perform_promise_then(duk_context *ctx);
+DUK_LOCAL void duk__trigger_promise_reactions(duk_context *ctx, duk_idx_t idx);
+
+#if defined(DUK_USE_PROMISE_BUILTIN)
+DUK_LOCAL void duk__push_promise(duk_context *ctx)
+{
+	/*
+	 *  [ ... ] -> [ ... promise ]
+	 */
+
+	duk__promise *prom;
+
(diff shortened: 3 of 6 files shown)


svaarala:
Cool stuff :)

fatcerberus:
One feature I'm considering beyond the core ES6 functionality is `Promise#done()` support.  Normally if an exception is thrown within a promise chain and it is not handled by a `.catch()` then it gets eaten.  Putting `.done()` at the end of the chain ensures any uncaught error will propagate out of the promise.  It's useful enough to be a de facto standard feature in third-party promise libraries, despite not appearing in any formal specification.

svaarala:
Hmm, at least Bluebird recommends against using .done():

> Like .then, but any unhandled rejection that ends up here will crash the process (in node) or be thrown as an error (in browsers). The use of this method is heavily discouraged and it only exists for historical reasons.

Not sure why, at first glance it seems like a useful feature for making "oops" situations more apparent.

fatcerberus:
I personally have had to debug situations in C# using async/await where thrown errors have gotten eaten by the async harness.  It's not very much fun.  I'd rather have the process crash than not see the effects of the bug until it's too late to trace the cause.

svaarala:
I don't disagree about that, but I wonder why it's not recommended in Bluebird - has it maybe been replaced by something newer?

fatcerberus:
Found this after a bit of Googling:
https://github.com/domenic/promises-unwrapping/issues/19

It seems that `.done()` was going to be in ES6, but was ultimately left out.  The expectation is that dev tools will add functionality to report and/or intercept rejected promises so that `.done()` is not needed.

svaarala:
Seems to be there are several issues with `.done()`, and I'd need to dig deeper to understand them. One issue seems to be that "no one handled an error" is not known reliably at any point in time because handlers may be attached at any point, even after the Promise has been resolved.

There *is* one point where that is reasonably well defined, and that is at GC time: if an error hasn't been handled by then, it never will (modulo rescuing etc).

The idea that promises are debugged via a debug console doesn't work all that well for embedded environments. So if any promise diagnostics are implemented they should probably be e.g. custom callbacks which are called when interesting events (like unhandled errors) occur.

fatcerberus:
I assume that hypothetical GC check is the reason why ES7 (but not ES6, interestingly) specifies a [[PromiseIsHandled]] internal slot (= whether anyone ever called `.then()`).

fatcerberus:
I'm going to have to go back to the drawing board on this one.  While Promises can readily be implemented within the constraints of an ES5 engine, it will need some thought as to the best way to do so in practice.

The main issue is that Promises rely pretty heavily on closures (conceptually at least), which is difficult to mimic with the value stack mechanism.  It's clear that the monolithic design currently used in this pull is not going to work.

What might work better is use the specification primitives as a starting point - only combining them later on if that makes sense.

svaarala:
Hmm, so what is the concrete problem with closures - aren't Promise function arguments ordinary function instances, and those reference can be stored for later calling?

fatcerberus:
Right, the functions have to refer back to the promise object so that calling them resolves it.  For `.then()` you also need to link the two promises so that the resolution of the first one ultimately resolves the second.  In JS with real closures all that is implicit.  But in C, you end up doing a lot of manual value stack balancing and you don't want to inline all that or else footprint starts going up quickly.

fatcerberus:
If it helps to illustrate, this is a fairly typical Promises/A+ implementation in JS:
https://github.com/fatcerberus/minisphere/blob/master/assets/system/polyfills.js#L5-L164

svaarala:
I don't understand what you mean by "it is implicit" - for example, if `.then()` is called it gets a `this` binding and arguments. The `.then()` implementation isn't aware of the scope where the call occurs and cannot see variables there.

svaarala:
Or do you mean that an Ecmascript `.then()` implementation itself can take advantage of closures?

svaarala:
If the latter, I think a natural implementation in C would be quite different. For example, the connections between promises would be handled as internal properties rather than variables in a scope.

fatcerberus:
Right, JS natively has closures so then() can take advantage of it.

> For example, the connections between promises would be handled as internal properties rather than variables in a scope.

Indeed, which is what I meant above that constantly inlining all those internal property accesses, with associated stack balancing, was bad for footprint.

All I was getting at above was that I wanted to start over and reevaluate the best way to divide up the helpers so that redundant code is eliminated.  I started out with a very monolithic design that got way out of hand very quickly.

To be clear I wasn't soliciting advice - just giving a progress update :)

svaarala:
Fair enough :) I'm quite interested in getting the promises integrated as well as possible (especially for footprint, as you might imagine). It only really makes sense to write promise support in C if it's significantly faster and/or smaller.

svaarala:
Also interesting is how the C API should support promises (if at all). That might have some impact on the implementation. But we can merge an initial implementation and then rework if it turns out to be necessary.

svaarala:
I know this is work in progress but just a quick comment on the internal `duk__promise` struct: `duk_tval` values in structs are not visible to GC so if they're used, they must be borrowed references (not reference counted) and their lifeycle must be exactly correct (= shorter than the strongly reachable counterparts). Internal properties are a bit easier because they're visible to GC.

It's of course possible to make the GC understand the structure. So far I've tried to avoid this (opting to use internal properties instead) but it may be that the GC needs some modularity to more flexibly adapt to structs like this.

fatcerberus:
I was indeed aware the tvals are borrowed.  They are stored as internal properties on the promise object to keep them reachable, I just keep a copy of the duk_tval around to avoid having to duk_push_this() + duk_get_prop_string() all the time.

That said, it would be convenient if one could "pin" a duk_tval without having to actually store it as a property... :)

svaarala:
I'd be a bit hesitant in using borrowed references like that: if the internal properties are edited from C code it may cause a segfault. There are some other cases like this in the current code base too (e.g. type assumptions) but I'd like to avoid making more and fixing the existing assumptions too. It'd be fine to misbehave in a memory safe manner if the properties are edited though.

svaarala:
Pinning is also very dangerous because unless the unpinning is automatic and intuitive, it's then very easy to leak memory. Which is not as bad as a straight segfault, but close especially on low memory targets. Leaking is even more difficult to figure out in the field. That's why I'm not very excited about pinning as an API design because: it will get misused.

But there are various solutions besides borrowed references and pinning, e.g. some form of GC interaction which allows you to extend the reachability graph in a modular fashion. Or requiring that when something is pinned, the "pinning" must be associated with an object and unpinning is then handled automatically (which is not far from GC interaction but implemented in a different way).

fatcerberus:
I'll probably get rid of the borrowed references in the rewrite, mostly it was done to reduce a lot of repeated property accesses, but helpers are probably a much better solution for that.

Regarding this:
> Pinning is also very dangerous because unless the unpinning is automatic and intuitive, it's then very easy to leak memory.

That's true, but then again the stash(es) have this issue as well - if an app stashed something for later use and then forgets to remove it, it will be leaked.  And manipulating stashed properties requires some value stack movement that a simple pin/unpin API wouldn't... :)

svaarala:
I agree about the stashes - though it is a bit more simple to ask (and answer) "do you have a reference in the stash?" vs. "have you pinned an object and potentially failed to unpin it at some point?". For example, if you do a debug dump of stash keys you can easily see if a reference exists. How do you dump "active pins"? That might need another API, and so on... :)

But yes, there are no easy alternatives I'm afraid. "Automatic unpinning" is an odd concept because if it could be automated, why would pinning be needed in the first place. Doing automatic unpinning when a base reference is released is essentially doing weak references so it's better to implement them rather than some form of clever pinning.

Note that I'm not at all trying to argue *something* shouldn't be added to make things easier to manage. It's just that the most obvious easy solutions are not actually easy :)

fatcerberus:
Speaking of weak references, that reminds me that WeakMap was another ES6 thing I wanted to look into.

svaarala:
Maps and sets are not an easy change because they have GC interactions, and there's also some potential to reuse some code between maps, sets, and plain objects (or not, depending on what works best). I've done some prototyping on these but so far no approach has been the obviously right one; for example, there are designs which work well when there are relatively few maps and sets (minimizes footprint) but don't work well when there are a lot of them (non-optimal RAM usage).

It'd also be good to get some of the pending work merged before taking on more ES6 stuff.

fatcerberus:
> It'd also be good to get some of the pending work merged before taking on more ES6 stuff.

Agreed.  For my part I should try to get at least #1025 polished off, that'll save me from having to write a `Reflect` polyfill for minisphere. :)

fatcerberus:
Now that my attention is not divided between this and Reflect, maybe I'll see if I can draft up a cleaner solution for this tonight.

fatcerberus:
I revisited the pull and cleaned some things up.  Now that I have some more experience implementing built-ins, this is going much more smoothly. :)

It's still buggy though.  For example here only the first `.then()` will execute:

```js
Promise.resolve("It's 8:12, do you know where your existence is? *munch*")
	.then(function(value) { console.log(value); return "STUPID PIG!"; })
	.then(function(value) { console.log(value); });
```

But it's a start.

fatcerberus:
Hm, it seems that ES6/7 is defined in terms of the implementation being fully async.  Not only Promise settlement but script execution is assumed to be handled in the event loop, see:
http://www.ecma-international.org/ecma-262/7.0/#sec-jobs-and-job-queues

Implementing these semantics requires some architectural changes that I don't think I'm prepared to take on at this point.  I'll close this pull; I believe it'd be best to wait until some other ES6 reworks have been done first.

fatcerberus:
@svaarala I'll keep hacking on this in my spare time, but I don't want there to be any pressure to merge, so I closed the pull until I have something more substantial to show.

I might open a separate issue to discuss implementation options and such for Promises, though.

svaarala:
Ok, sounds good.

Re: that eventloop thing, it'd be interesting to figure out if the need for an event queue could be generalized to some extent so that it would handle Promises but would also make it easy to integrate timers and maybe user custom events into the same mechanism. The event queue model is also used by browsers so while it's not a part of the core Ecmascript model, it's still fairly common.

For the C API there's maybe be an API call to handle one queued event or indicate all events are done. There would also need to be a way to push events-to-be-handled into the queue, preferably from C code (which would be exercised by the Promise implementation).

PR #1020, opened 2016-10-16
Remove 1:1 buffer-to-string coercion in Node.js Buffer.prototype.toString()

svaarala:
- [x] Rework encoding API to provide a shared buffer-to-string helper for Node.js Buffer code (even when Encoding API is disabled)
- [x] Change `Buffer.prototype.toString()` to decode the buffer as UTF-8 and encode output using CESU-8. This prevents buffer code from creating internal strings, and also aligns the Buffer .toString() operation with Node.js.
- [x] Website updates for buffer coercions
- [x] Website: document difference to Node.js replacement character behavior
- [x] Internal document updates, e.g. sandboxing, buffer future work status
- [x] 2.0 migration note
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1677,7 +1677,9 @@ Planned
 
 * Incompatible change: Node.js Buffer binding aligned with Node.js v6.7.0
   (from v0.12.1): Buffer.concat() special case for 1-element array removed
-  (GH-1004), Buffer now inherits from Uint8Array (GH-1008)
+  (GH-1004); Buffer now inherits from Uint8Array (GH-1008);
+  Buffer.prototype.toString() does UTF-8 decoding (previously buffer data
+  was copied into internal string representation as is) (GH-1020)
 
 * Incompatible change: plain pointer values now test true in instanceof
   (plainPointer instanceof Duktape.Pointer === true) (GH-864)
--- a/doc/buffers.rst
+++ b/doc/buffers.rst
@@ -1146,7 +1146,8 @@ Gap between current implementation and latest:
 
 * ``buf.swap16()``, ``buf.swap32()``, ``buf.swap64()`` missing.
 
-* ``buf.toString()`` doesn't implement encoding.
+* ``buf.toString()`` always decodes the buffer using UTF-8 (with replacement
+  characters for invalid sequences), and ignores the encoding argument.
 
 * ``buf.values()`` missing.
 
--- a/doc/release-notes-v2-0.rst
+++ b/doc/release-notes-v2-0.rst
@@ -231,7 +231,8 @@ changes below.  Here's a summary of changes:
   and string operations) and can expose such a binding to Ecmascript code.
 
 * Node.js Buffer binding has been aligned more with Node.js v6.8.1 (from
-  Node.js v0.12.1).
+  Node.js v0.12.1) and some (but not all) behavior differences to actual
+  Node.js have been fixed.
 
 * Disabling ``DUK_USE_BUFFEROBJECT_SUPPORT`` allows use of plain buffers in
   the C API, and allows manipulation of plain buffers in Ecmascript code via
@@ -255,7 +256,7 @@ To upgrade:
     has the default prototype for the result type (e.g. ``Uint8Array.prototype``)
     rather than being copied from the argument.
 
-* If you're using the Node.js Buffer binding, XXX.
+* If you're using the Node.js Buffer binding, review the following:
 
   - Node.js Buffer ``.slice()`` handling of arguments inheriting from a Buffer
     (rather than being a direct instance) has been fixed so that the result has
@@ -264,6 +265,9 @@ To upgrade:
 
   - Node.js Buffer ``.concat()`` always returns a buffer copy, even for a
     one-element input array which had special handling in Node.js v0.12.1.
+
+  - Node.hs Buffer.prototype ``.toString()`` now decodes the input buffer
+    using UTF-8, emitting replacement characters for invalid UTF-8 sequences.
 
   - Review Buffer code for Node.js Buffer changes between Node.js versions
     v0.12.1 and v6.8.1 in general.
--- a/doc/sandboxing.rst
+++ b/doc/sandboxing.rst
@@ -25,9 +25,8 @@ carefully written with these sandboxing goals in mind.
 
 This document describes best practices for Duktape sandboxing.
 
-.. note:: This document is in a rough draft state and describes the current
-          sandboxing status (which is not complete).  Sandboxing shortcomings
-          will be fixed piece by piece in later versions.
+.. note:: This document described the current status of sandboxing features
+          which is not yet a complete solution.
 
 Suggested measures
 ==================
@@ -37,10 +36,10 @@ Isolation approaches
 
 There are two basic alternatives to sandboxing Ecmascript code with Duktape:
 
-* Use a separate Duktape heap for each sandbox
+* Use a separate Duktape heap for each sandbox.
 
 * Use a separate Duktape thread (with a separate global environment) for
-  each sandbox
+  each sandbox.
 
 Pros and cons of using a Duktape heap for sandboxing:
 
@@ -56,9 +55,11 @@ Pros and cons of using a Duktape heap for sandboxing:
   heap, however.
 
 * One downside is that there is some per-heap overhead which accumulates for
-  each sandbox.  If the memory pool for a heap is pre-allocated, some of the
-  pool will be unused.  If memory is not pre-allocated, actual memory usage
-  is quite tightly bound because of reference counting.
+  each sandbox; ROM built-ins can mitigate some of that effect.  If the memory
+  pool for a heap is pre-allocated and not shared between heaps (in a thread
+  safe manner), some of the pool will be unused.  If memory is not
+  pre-allocated, actual memory usage is quite tightly bound because of
+  reference counting.
 
 Pros and cons of using a Duktape thread for sandboxing:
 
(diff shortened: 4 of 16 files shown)


svaarala:
The pull uses a simple eval code hack to get the desired effect; this allows testcases to be updated before an actual implementation.

svaarala:
@fatcerberus There's a slight difference in replacement character handling in Duktape and V8 TextDecoder() at least for truncated codepoints.

From Duktape master:
```
duk> new TextDecoder().decode(new Uint8Array([ 0xec, 0xab ]))
= "\ufffd"
```

So, the two bytes generate a single replacement character. This matches Firefox behavior:

```
new TextDecoder().decode(new Uint8Array([ 0xec, 0xab ]))
"�"
new TextDecoder().decode(new Uint8Array([ 0xec, 0xab ])).length
1
```

However, Chrome behaves differently:

```
new TextDecoder().decode(new Uint8Array([ 0xec, 0xab ])).length
2
new TextDecoder().decode(new Uint8Array([ 0xec, 0xab ])).charCodeAt(0)
65533
new TextDecoder().decode(new Uint8Array([ 0xec, 0xab ])).charCodeAt(1)
65533
```

I noticed this because it carries over (maybe because of shared code - maybe not) to Node.js v6.7.0 Buffer .toString():

```
> buf = new Buffer(2); buf[0] = 0xec; buf[1] = 0xab; str = buf.toString();
'��'
> str.charCodeAt(0)
65533
> str.charCodeAt(1)
65533
```

If Duktape's Node.js Buffer .toString() just uses the TextDecoder algorithm as is, it will generate a single replacement character for the test above; this is from the current (hacked) branch:

```
duk> buf = new Buffer(2); buf[0] = 0xec; buf[1] = 0xab; str = buf.toString();
= "\ufffd"
```

What's your opinion, is Chrome or Firefox correct?

svaarala:
Both behaviors seem to be acceptable but http://www.unicode.org/review/pr-121.html recommends the current approach in master if I understood it correctly:

```js
var u8 = new Uint8Array([ 0x61, 0xF1, 0x80, 0x80, 0xE1, 0x80, 0xC2, 0x62 ]);
var res = new TextDecoder().decode(u8);
print(Array.prototype.map.call(res, function (v) { return v.charCodeAt(0); }).join(' '));
```

prints:

```
97 65533 65533 65533 98
```

which matches "recommendation 2" to:

> Replace each maximal subpart of the ill-formed subsequence by a single U+FFFD.

I'll add this test to the TextDecoder test set too with a link to the recommendation.

fatcerberus:
The WHATWG algorithm emits replacement characters only when the decode helper returns "error".  Since that error isn't known until you encounter an invalid byte (or EOS), one replacement char seems correct to me.

svaarala:
Sounds good to me too, #1021 adds an explicit testcase with explanation.

svaarala:
The replacement character behavior will differ from Node.js v6.7.0 but I think it's OK - it's not something that hopefully anyone would depend on. The only concrete thing that comes to mind is any kind of canonicalization where that might matter.

svaarala:
Found a small issue with BOM. Node.js v6.7.0:

```
> new Buffer(new Uint8Array([ 0xef, 0xbb, 0xbf ])).toString()
'﻿'
> new Buffer(new Uint8Array([ 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).toString()
'﻿AB'
> new Buffer(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).toString()
'@﻿AB'
```

BOM character is stripped even when it is inside the buffer. Current TextDecoder() implementation (which I presume is correct):

```
duk> new Buffer(new Uint8Array([ 0xef, 0xbb, 0xbf ])).toString()
= ""
duk> new Buffer(new Uint8Array([ 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).toString()
= "AB"
duk> new Buffer(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).toString()
= "@\ufeffAB"
```

fatcerberus:
That looks like a bug (or oversight).  WHATWG sets the "BOM seen" flag after the first codepoint, even if it is not a BOM.

See the serialization algorithm here:
https://encoding.spec.whatwg.org/#interface-textdecoder

I studied the specification very closely while implementing this. :)

svaarala:
I don't doubt that and I think the handling in master is correct. I'm just wondering why Node.js v6.7.0 behaves differently - and if it's not a clear bug I'll document it in the list of differences between Duktape and Node.js Buffer implementation.

svaarala:
To be precise there's no reason `Buffer.toString()` would follow the TextDecoder() requirements. So it might be desired functionality and that's not a bug on their part if so. But it does come up when trying to reuse the TextDecoder() implementation for Node.js Buffer .toString(), which is why I encountered the issue.

I'll retest with Node.js v6.8.1. Sidenote: I'll need to do another gap analysis for Node.js v6.8.1 :o

fatcerberus:
Actually what concerns me is that Chrome's TextDecoder has the same bug - since they both seem to be using the same V8 helper for the actual decoder.

svaarala:
Same here on my Chrome version:

```
> new TextDecoder().decode(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ]))
"@﻿AB"
```

Interesting :)

svaarala:
Ah, it's just the printout which doesn't show the BOM at all (at least in a non-initial position).

For example, on Firefox:

```
> new TextDecoder().decode(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ]))
"@﻿AB"
new TextDecoder().decode(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).length
4
new TextDecoder().decode(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).charCodeAt(1)
65279
```

On Chrome:

```
new TextDecoder().decode(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ]))
"@﻿AB"
new TextDecoder().decode(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).length
4
new TextDecoder().decode(new Uint8Array([ 0x40, 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).charCodeAt(1)
65279
```

So it's not supressed in these cases.

svaarala:
Ok, so it seems Node.js also doesn't strip BOMs in the middle but it also won't strip BOM at the beginning:

```
> new Buffer(new Uint8Array([ 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).toString()
'﻿AB'
> new Buffer(new Uint8Array([ 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).toString().length
3
> new Buffer(new Uint8Array([ 0xef, 0xbb, 0xbf, 0x41, 0x42 ])).toString().charCodeAt(0)
65279
```

svaarala:
So for Buffer .toString() ignoreBOM=1 seems correct.

svaarala:
@fatcerberus So to clarify: my bad, it was just that Node.js console and the web consoles don't display a BOM present in a string at all so I misread the output.

But it did reveal that Node.js Buffer doesn't seem to strip an initial BOM (which is assumed it would). So not a total loss.

fatcerberus:
That makes sense.  It surprised me that a BOM would ever be removed from the middle of a string unless it was a bug.

svaarala:
There's a relatively ugly solution now in the pull. It adds ~100 bytes to x64 default footprint and ~600 bytes to low memory (minsize) footprint. The minsize footprint increases because Encoding API is disabled in that configuration but the Node.js Buffers are supported, which pulls in the TextDecoder decode() helper which seems to be around 500 bytes.

svaarala:
@fatcerberus I'll give this one more read-through tomorrow and then merge unless you have suggestions for improvement - while it's not perfect it removes the last direct 1:1 buffer-to-string binding from master which allows some other stuff to proceed.

fatcerberus:
@svaarala Sounds good to me. :)

svaarala:
Hmm, seems that Node.js v6.8.1 still uses two replacements characters for this case:

```
new Buffer(new Uint8Array([ 0xec, 0xab ])).toString()
```

whereas Duktape (and WHATWG TextDecoder) uses one.

svaarala:
```
$ bin/node 
> process.version
'v6.8.1'
> new Buffer(new Uint8Array([ 0xec, 0xab ])).toString()
'��'
> new Buffer(new Uint8Array([ 0xec, 0xab ])).toString().length
2
```

svaarala:
This pull doesn't yet fix the reverse operation - creating a buffer from a string. Node.js will encode the input string (from the logical codepoint sequence) into UTF-8. Duktape will currently just copy the bytes from the internal representation as is which is *not* a sandboxing issue (it won't allow creation of internal keys) but it does differ from Node.js.

Node.js:

```
> b = new Buffer('\ud800'); console.log(b.length, b[0], b[1], b[2])
3 239 191 189
```

Duktape:

```
duk> b = new Buffer('\ud800'); console.log(b.length, b[0], b[1], b[2])
3 237 160 128
```

Node.js replaces the unpaired surrogate with a replacement character but Duktape outputs it as is. In essence, Duktape encoding for standard strings is CESU-8, not UTF-8. And it passes internal non-CESU-8/non-UTF-8 sequences as is into the output - which has no compliance impact because such strings are outside any standard behavior.

In any case, this will be fixed in a separate pull.

fatcerberus:
Chrome 54 TextDecoder still has that nonconformant replacement character behavior, too.  If that ever gets fixed then Node.js should eventually converge to the WHATWG required behavior.

PR #1021, opened 2016-10-16
Add test for replacement character policy

svaarala:


fatcerberus:
@svaarala You're really going all-in on this feature aren't you? :)

svaarala:
If you look into the test cases, you'll notice that other features have pretty detailed test case harnesses too. I like to document current behavior when there are several alternatives, so that the behavior is known and won't change by accident.

svaarala:
Btw, proper testcase coverage is one reason it takes some time to work in changes into master, even for seemingly small features. Quite typically it takes at least as long or even significantly longer to write tests that actually cover all relevant code paths.

I *really* dislike features that don't have active test coverage - it makes refactoring dangerous and bug hunting for any fallout is also very costly. Similarly changes to behavior after release often pose versioning issues which take a lot of cumulative time. As a whole there's then a tendency to avoid refactoring which is very bad in the longer term (I've seen this effect enough times).

In this case, replacement character handling is one of the rare differences in (otherwise correct) UTF-8 decoders, so I think it's definitely worth documenting what alternative is used and why - I also wasn't aware of the alternatives and recommendation myself :)

That said I don't expect pulls to always provide comprehensive testcase coverage - I'll add more tests into master afterwards until I'm satisfied with the coverage. There's a minor backlog for 2.x already in that department, e.g. for suspend/resume API calls and now Encoding API.

PR #1022, opened 2016-10-17
Fix a potential memory safety issue in TextDecoder decode()

svaarala:
Avoid potential for memory unsafe behavior if parsing decode options or pushing a fixed output buffer cause a side effect which invalidates the input buffer pointer.

The fix assumes that argument type check order is not strictly in order; the WHATWG specification doesn't seem to guarantee a specific order which is useful here. If this isn't correct, the type check needs to happen first, while the actual data pointer is looked up just before the decoding process.

If the algorithm is later converted into a chunked + bufwriter based process, the input base pointer needs to be looked up on every resize. The input length also needs to be revalidated at every resize because a side effect may indeed resize the input buffer to be smaller so that a previously looked up length is no longer valid.

### DIFF
--- a/src-input/duk_bi_encoding.c
+++ b/src-input/duk_bi_encoding.c
@@ -366,12 +366,20 @@ DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_decode(duk_context *ctx) {
 
 	dec_ctx = duk__get_textdecoder_context(ctx);
 
-	if (!duk_is_undefined(ctx, 0)) {
-		input = (const duk_uint8_t *) duk_require_buffer_data(ctx, 0, &len);
-	} else {
-		input = NULL;  /* NULL is OK if 'len' is 0. */
-		DUK_ASSERT(len == 0);
+	/* Careful with input buffer pointer: any side effects involving
+	 * code execution (e.g. getters, coercion calls, and finalizers)
+	 * may cause a resize and invalidate a pointer we've read.  This
+	 * is why the pointer is actually looked up at the last minute.
+	 * Argument validation must still happen first to match WHATWG
+	 * required side effect order.
+	 */
+
+	if (duk_is_undefined(ctx, 0)) {
+		duk_push_fixed_buffer(ctx, 0);
+		duk_replace(ctx, 0);
 	}
+	(void) duk_require_buffer_data(ctx, 0, NULL);  /* For side effects only. */
+
 	if (duk_is_null_or_undefined(ctx, 1)) {
 		/* Use defaults. */
 	} else {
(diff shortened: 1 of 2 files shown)


svaarala:
This issue has been in master but not released so no releases entry.

svaarala:
These issues are currently difficult to catch automatically, but quite easy to demonstrate in practice with crafted code. One torture helper would be to force a mark-and-sweep on every allocation (this already exists) *and* when walking through the heap, reallocate every dynamic buffer so that its base pointer changes to ensure any invalid assumptions re: the base pointer are caught.

This would be an improvement but wouldn't catch issues with external buffers and side effects, or side effects resizing a buffer to a smaller size (which may cause an overrun even if the base pointer is not assumed to be static).

svaarala:
@fatcerberus I'm off to sleep now, I'll merge this tomorrow assuming you agree with the change (in particular the WHATWG argument coercion order part).

fatcerberus:
Web IDL seems to require arguments be coerced left to right:
https://www.w3.org/TR/WebIDL/#es-operations

In particular:
> Let <operation, values> be the result of passing S and arg0..n−1 to the overload resolution algorithm.

That overload resolution algorithm is here:
https://www.w3.org/TR/WebIDL/#dfn-overload-resolution-algorithm

It's pretty involved, but it seems apparent from skimming over it that arguments are processed left to right and "converted to IDL values" (which usually involves coercion).

svaarala:
Ok, I'll figure out another way to fix it without affecting the argument coercion order.

fatcerberus:
As an aside, Web IDL is interesting but sometimes confusing: it's effectively a metaspecification and many of the "algorithms" described outside of the Ecmascript mapping are pretty abstract, having little relation to how any real implementation would function.  Interpreting the document is a challenge to say the least. :)

svaarala:
Yep, that's why I think the best approach would be to figure out what is required, compare to other engines, and then document it in testcases. At least there's then a baseline against which to make changes. It's easy to change stuff in refactoring - e.g. when introducing new internal helpers for more convenience, some coercion or validation aspect may subtly change and it's nice if the change is at least conscious and not an accident.

svaarala:
@fatcerberus Does this look good for a merge?

fatcerberus:
On the subject of this pull: I presume this is a big reason behind standard ArrayBuffers not being resizable.  It makes things challenging for the engine if it has to guarantee safety in the face of potentially moving pointers, even more so in engines like SpiderMonkey where GC is done on a separate thread and can start at any time.

svaarala:
Maybe, but even without resizing there can be pointer changes. For example, a compacting garbage collector will move objects around to reduce fragmentation and data pointers will change at the same time. It may be that they handle the actual data allocation separately without relocation - not sure. But that'd then be prone to fragmentation so it'd be an odd approach.

The ability to resize buffers is pretty central to C coders (`realloc()` for example) so not having the ability would be quite awkward IMO.

fatcerberus:
Everything looks good to me.

svaarala:
There are a few current call sites which disable potential side effects for some critical operation and then restore the relevant heap flags on their way out. That's rather awkward and error prone because you need to catch errors, keep track of the flags etc.

So one internally useful helper would be a `duk_safe_call()` variant which disabled side effects for the duration of the call and then automatically restored the relevant GC flags. While not super convenient, it'd still avoid the heap flag tweaks in several call sites.

fatcerberus:
@svaarala No issues with the diff, and it seems all the checks have passed, so this looks good to merge to me.

PR #1023, opened 2016-10-17
Fix memory unsafe behavior in TextDecoder decode()

svaarala:
Previous fix was incorrect: the coercion check for `.decode()` first argument didn't read "len" so that output buffer allocated was always zero length. This caused memory unsafe behavior for almost any input (but wasn't caught by non-valgrind tests).

Fix by looking up the length once to size the output buffer. Then, when looking up the input data pointer recheck that the length is still the same. If it isn't, fail: it's not safe continue. If the input became longer, the output buffer wouldn't (always) be large enough.

No release entry because bug hasn't been released.

### DIFF
--- a/src-input/duk_bi_encoding.c
+++ b/src-input/duk_bi_encoding.c
@@ -358,6 +358,7 @@ DUK_INTERNAL duk_ret_t duk_bi_textdecoder_prototype_decode(duk_context *ctx) {
 	const duk_uint8_t *input;
 	duk__decode_context *dec_ctx;
 	duk_size_t len = 0;
+	duk_size_t len_tmp;
 	duk_bool_t stream = 0;
 	duk_codepoint_t codepoint;
 	duk_uint8_t *output;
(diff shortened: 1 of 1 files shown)


fatcerberus:
Hm, at first I was thinking you could just defer allocating the output buffer until after the input buffer pointer was retrieved, but that won't actually work because `duk_push_fixed_buffer()` may indirectly cause side effects up to and including finalizer calls.

svaarala:
Yep - and on the other hand you need the input length to allocate the output buffer.

svaarala:
This is one place where a chunked algorithm would actually be a bit easier: you'd process input in fixed chunks and append it to an output buffer resized on each step. The output reserve would be a fixed amount rather then depending on the input length. At each step the input pointer *and* length would be rechecked.

This would be memory safe but would be subject to side effects modifying the input buffer so that the output might reflect two "versions" of the input. This is not really a problem as such, because it's still memory safe, and you'd have to quite purposefully modify the input.

PR #1024, opened 2016-10-17
Update Node.js Buffer gap for Node.js v6.8.1

svaarala:
- [x] Recheck gap description from v6.7.0 to v6.8.1
- [x] Update website, documentation, and tests referring to Node.js snapshot (a bit premature but some behaviors have already been changed)


svaarala:
Checked the diff between v6.7.0 and v6.8.1 buffer docs and there doesn't seem to be any changes. Also checked v6.8.0 and v6.8.1 release notes and there seem to be only internal buffer changes.

PR #1027, opened 2016-10-20
Add internal support for [[Get]]/[[Set]] receiver

fatcerberus:
Prerequisite for #1025, this pull adds a `tv_recv` argument for `duk_hobject_{get|put}prop()` which corresponds to the _Receiver_ argument for [[Get]] and [[Set]], needed by several ES6+ algorithms and methods:
- http://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-get-p-receiver
- http://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-set-p-v-receiver

Checklist:
- [x] Add support for _Receiver_ argument of [[Get]]
- [ ] Add support for _Receiver_ argument of [[Set]]
- [x] RELEASES entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2019,3 +2019,6 @@ Planned
 * Internal change: rework tagged value (duk_tval) fastint/integer handling
   macros to avoid multiple evaluation of argument(s) and for easier mixing
   of fastint and non-fastint aware code (GH-702)
+
+* Internal change: Add support for a receiver object in get/putprop operations,
+  needed by several ES6+ algorithms (GH-1027)
--- a/src-input/duk_api_object.c
+++ b/src-input/duk_api_object.c
@@ -27,7 +27,7 @@ DUK_EXTERNAL duk_bool_t duk_get_prop(duk_context *ctx, duk_idx_t obj_idx) {
 	tv_obj = duk_require_tval(ctx, obj_idx);
 	tv_key = duk_require_tval(ctx, -1);
 
-	rc = duk_hobject_getprop(thr, tv_obj, tv_key);
+	rc = duk_hobject_getprop(thr, tv_obj, tv_key, tv_obj);
 	DUK_ASSERT(rc == 0 || rc == 1);
 	/* a value is left on stack regardless of rc */
 
@@ -114,7 +114,7 @@ DUK_LOCAL duk_bool_t duk__put_prop_shared(duk_context *ctx, duk_idx_t obj_idx, d
 	tv_val = duk_require_tval(ctx, idx_key ^ 1);
 	throw_flag = duk_is_strict_call(ctx);
 
-	rc = duk_hobject_putprop(thr, tv_obj, tv_key, tv_val, throw_flag);
+	rc = duk_hobject_putprop(thr, tv_obj, tv_key, tv_val, tv_obj, throw_flag);
 	DUK_ASSERT(rc == 0 || rc == 1);
 
 	duk_pop_2(ctx);  /* remove key and value */
--- a/src-input/duk_bi_json.c
+++ b/src-input/duk_bi_json.c
@@ -1952,7 +1952,7 @@ DUK_LOCAL duk_bool_t duk__enc_value(duk_json_enc_ctx *js_ctx, duk_idx_t idx_hold
 	DUK_ASSERT(DUK_TVAL_IS_OBJECT(tv_holder));
 	tv_key = DUK_GET_TVAL_NEGIDX(ctx, -1);
 	DUK_ASSERT(DUK_TVAL_IS_STRING(tv_key));
-	(void) duk_hobject_getprop(thr, tv_holder, tv_key);
+	(void) duk_hobject_getprop(thr, tv_holder, tv_key, tv_holder);
 
 	/* -> [ ... key val ] */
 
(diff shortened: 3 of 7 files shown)


fatcerberus:
ES6 [[Get]] is very straightforward, *Receiver* is used for is the `this` binding of a getter call (if one is made) and nothing else.  Data properties will always be read from *O*, never *Receiver*.  So overall the changes for getprop are trivial.

Semantics for [[Set]] when *Receiver* !== *O*, however, are quite strange.  The property descriptor is taken from *O*, but the value is written to *Receiver*.  I guess this is nice in that it allows for virtualization property writes, but it seems inconsistent with [[Get]].

So in the end the necessary changes for putprop are decidedly less trivial and will need a bit more time and thought to fully implement.  Please stand by. :)

fatcerberus:
For ES6 [[Set]], two property descriptor lookups are required: Once for the target object, then again for the receiving object.  If either or both has `{writable:false}` then [[Set]] fails.

I think it would be best to optimize for the case where Receiver === O, as that will be the common case and can avoid the second writability check entirely.  For now the only time this won't hold is when a receiver is passed to `Reflect.set()` which probably won't occur that often in real-world code.

svaarala:
I'll need to think about this a bit. Adding the extra parameter to every property get/set may not be the best approach because the parameter is very often not needed and will impact the performance of every operation.

I'm not sure what the best alternative is though - and if there is none, this is probably a good opportunity to clean up the internal property access APIs which are not ideal in their argument/value stack handling. In particular, the internal calls take `duk_tval`s but in practice need to make copies of them to stabilize them against side effects.

svaarala:
So maybe a good intermediate step (regardless of what happens with this pull) would be to list the ES6 semantic changes and internal rework needs for property handling. This would give a better idea what needs to be done. Some rework approaches I've been prototyping with involve rewriting around 50% of the property code to make it easier to work with so it'd be best to address multiple issues at once.

Also Sets and Maps (especially WeakSet and WeakMap) are somewhat related: they need a key/value structure but the keys are objects. For footprint it might be best to implement them using the same data structure as plain objects, so that they key space would be extended from strings (just like in Lua BTW). On the other hand it would be more modular and easier for maintaining property code if they were separate.

Not saying this can't be merged BTW, just want to think about this topic a bit more before doing so :)

fatcerberus:
I actually didn't realize when I first opened this pull how deep the rabbit hole went... :)

MDN documentation for `Reflect.set()` says the `receiver` argument is used only for `this` in setter calls and doesn't mention the redirection behavior for data properties at all.  I know MDN is sometimes wrong but this strikes me as a major documentation deficiency that is very likely to trip someone up.

As a result I got a nasty surprise when I cross-referenced against ES6 (well... actually 7) and found that I had to do two property descriptor checks and write the property to the receiver, NOT the target object.

So no huge pressure to merge this if it doesn't make sense to do that now. :)

fatcerberus:
Skimming through ES7 real quick there seem to be only two cases where [[Set]] is called with a *Receiver* different from the target object:

* Proxy exotic [[Set]]: http://www.ecma-international.org/ecma-262/7.0/#sec-proxy-object-internal-methods-and-internal-slots-set-p-v-receiver

* `Reflect.set()`: http://www.ecma-international.org/ecma-262/7.0/#sec-reflect.set

Both are uncommon enough I'd imagine that taking a performance hit for even an extra branch for all property writes (fast path check) is probably unacceptable.

svaarala:
For the Proxy case the problem is we never know beforehand if a property read will eventually hit a Proxy. If it does, the argument needs to be available somehow.

But it *is* possible to optimize it so that if the receiver == target (which is almost always the case if I've understood correctly) the call site doesn't need to give it explicitly and the Proxy call site can assume the value is the same. The other call sites would need a different API which allows the values to be different.

In practice this might be unworkable because there would be two different property access paths. The argument could also be NULL and a modern compiler would technically be able to optimize the whole code path with that assumption (specific function variants with constant argument values are created by compilers) -- in practice that will almost certainly not happen for something as large as property access code and there's no standard way to "nudge" a compiler to do it.

svaarala:
Assuming a bit more rework:

The current internal property access API takes values as `duk_tval` pointers but needs to copy them to locals to stabilize the pointers against side effects. This is a workaround because I didn't first realize it's going to be so difficult to avoid with essentially any side effect potentially affecting the value stack. So it might be cleaner (and even faster) to just pass the values via the value stack instead.

Another potential need is to pass around a flags field to control property access; without that it'd be awkward to implement e.g. primitives to read properties without side effects, reading only own properties, and so on.

These two taken together mean that the receiver wouldn't need to be provided on the value stack by default and a flag `DUK_PROP_RECEIVER_PRESENT` could then *only* be checked in the Proxy case to see which value to use.

For the value stack indexing to work reasonably the receiver would need to be pushed first (if present), followed by key and value (for "set"). This way property code could assume key is at -2 and value is at -1, and receiver, if present, is at -3. Property code would only pop two keys and for the call sites where an explicit receiver is used, the call site would be responsible for popping it off.

fatcerberus:
It actually surprised me that Duktape passes tagged values around by reference internally - I'd expect them to be passed by value in most cases.  Although I guess a pointer *is* usually half the size on most platforms.

svaarala:
Passing them by value quite often leads to them being passed via the C stack rather than registers which produces pretty awkward code at least on (some) 32-bit platforms.

fatcerberus:
@svaarala If you think this change doesn't make sense in its current form, I'll close the pull to be revisited later (perhaps in a Proxy rework).

PR #1028, opened 2016-10-20
Align argument policy for Object.keys(), Object.seal(), etc. with ES6

fatcerberus:
ES6 changes the argument policy for functions such as `Object.keys()` and `Object.freeze()` to accept primitive values, treating them as though they were non-extensible objects with no properties, rather than rejecting them with a `TypeError` as in ES5.  This pull implements that change of behavior.

Affected functions:
- [x] `Object.freeze`
- [x] `Object.getOwnPropertyDescriptor`
- [x] `Object.getOwnPropertyNames`
- [x] `Object.getPrototypeOf`
- [x] `Object.isExtensible`
- [x] `Object.isFrozen`
- [x] `Object.isSealed`
- [x] `Object.keys`
- [x] `Object.preventExtensions`
- [x] `Object.seal`

Checklist:
- [x] Review ES 2015 to see exactly which functions are affected
- [x] Implement the argument policy change
- [x] This got a bit awkward in spots, review changes and clean up
- [x] Fix testcases to align with new argument policy
- [x] Verify against ES6 to ensure other behavior matches
=> *Possible issue with freeze/seal, best tackled separately*
- [x] Squash commits and rebase against latest master
- [x] Releases entry

### DIFF
--- a/src-input/duk_bi_object.c
+++ b/src-input/duk_bi_object.c
@@ -233,6 +233,8 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_seal_freeze_shared(duk_context 
 	duk_hobject *h;
 	duk_bool_t is_freeze;
 
+	DUK_ASSERT_TOP(ctx, 1);
+
 	is_freeze = (duk_bool_t) duk_get_current_magic(ctx);
 	if (duk_is_buffer(ctx, 0)) {
 		/* Plain buffer: already sealed, but not frozen (and can't be frozen
@@ -255,8 +257,12 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_seal_freeze_shared(duk_context 
 	h = duk_require_hobject_promote_mask(ctx, 0, DUK_TYPE_MASK_LIGHTFUNC | DUK_TYPE_MASK_BUFFER);
 #endif
 
-	h = duk_require_hobject(ctx, 0);
-	DUK_ASSERT(h != NULL);
+	h = duk_get_hobject(ctx, 0);
+	if (h == NULL) {
+		/* ES6 Sections 19.1.2.5, 19.1.2.17 */
+		return 1;
+	}
+
 	if (is_freeze && DUK_HOBJECT_IS_BUFOBJ(h)) {
 		/* Buffer objects cannot be frozen because there's no internal
 		 * support for making virtual array indices non-writable.
@@ -282,7 +288,6 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_seal_freeze_shared(duk_context 
 DUK_INTERNAL duk_ret_t duk_bi_object_constructor_is_sealed_frozen_shared(duk_context *ctx) {
 	duk_hobject *h;
 	duk_bool_t is_frozen;
-	duk_bool_t rc;
 	duk_uint_t mask;
 
 	is_frozen = duk_get_current_magic(ctx);
@@ -293,9 +298,13 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_is_sealed_frozen_shared(duk_con
 		                          1 :               /* lightfunc always frozen and sealed */
 		                          (is_frozen ^ 1)); /* buffer sealed but not frozen (index props writable) */
 	} else {
-		h = duk_require_hobject(ctx, 0);
-		rc = duk_hobject_object_is_sealed_frozen_helper((duk_hthread *) ctx, h, is_frozen /*is_frozen*/);
-		duk_push_boolean(ctx, rc);
+		/* ES6 Sections 19.1.2.12, 19.1.2.13: anything other than an object
+		 * is considered to be already sealed and frozen.
+		 */
+		h = duk_get_hobject(ctx, 0);
+		duk_push_boolean(ctx, (h == NULL) ?
+		                      1 :
+		                      duk_hobject_object_is_sealed_frozen_helper((duk_hthread *) ctx, h, is_frozen /*is_frozen*/));
 	}
 	return 1;
 }
@@ -377,13 +386,20 @@ DUK_INTERNAL duk_ret_t duk_bi_object_getprototype_shared(duk_context *ctx) {
 	duk_hobject *h;
 	duk_hobject *proto;
 	duk_tval *tv;
+	duk_int_t magic;
 
-	if (duk_get_current_magic(ctx) == 0) {
+	magic = duk_get_current_magic(ctx);
+
+	if (magic == 0) {
 		DUK_ASSERT_TOP(ctx, 0);
 		duk_push_this_coercible_to_object(ctx);
 		tv = DUK_HTHREAD_THIS_PTR(thr);
 	}
 	DUK_ASSERT(duk_get_top(ctx) >= 1);
+	if (magic < 2) {
+		/* ES6 Section 19.1.2.9, step 1 */
+		duk_to_object(ctx, 0);
+	}
 	tv = DUK_GET_TVAL_POSIDX(ctx, 0);
 
 	switch (DUK_TVAL_GET_TAG(tv)) {
@@ -603,7 +619,15 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_define_property(duk_context *ct
 
 #if defined(DUK_USE_OBJECT_BUILTIN) || defined(DUK_USE_REFLECT_BUILTIN)
 DUK_INTERNAL duk_ret_t duk_bi_object_constructor_get_own_property_descriptor(duk_context *ctx) {
+	DUK_ASSERT_TOP(ctx, 2);
+
+	/* ES6 Section 19.1.2.6, step 1 */
+	if (duk_get_current_magic(ctx) == 0) {
+		duk_to_object(ctx, 0);
+	}
+
 	/* [ obj key ] */
+
 	duk_hobject_object_get_own_property_descriptor(ctx, -2);
 	return 1;
 }
@@ -611,19 +635,35 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_get_own_property_descriptor(duk
 
 #if defined(DUK_USE_OBJECT_BUILTIN) || defined(DUK_USE_REFLECT_BUILTIN)
 DUK_INTERNAL duk_ret_t duk_bi_object_constructor_is_extensible(duk_context *ctx) {
+	/*
+	 *  magic = 0: Object.isExtensible()
+	 *  magic = 1: Reflect.isExtensible()
+	 */
+
 	duk_hobject *h;
 
-	h = duk_require_hobject_accept_mask(ctx, 0, DUK_TYPE_MASK_LIGHTFUNC | DUK_TYPE_MASK_BUFFER);
+	if (duk_get_current_magic(ctx) == 0) {
+		h = duk_get_hobject(ctx, 0);
+	} else {
+		/* Reflect.isExtensible(): throw if non-object, but we accept lightfuncs
+		 * and plain buffers here because they pretend to be objects.
+		 */
+		h = duk_require_hobject_accept_mask(ctx, 0, DUK_TYPE_MASK_LIGHTFUNC | DUK_TYPE_MASK_BUFFER);
+	}
+
 	duk_push_boolean(ctx, h != NULL && DUK_HOBJECT_HAS_EXTENSIBLE(h));
 	return 1;
 }
 #endif  /* DUK_USE_OBJECT_BUILTIN || DUK_USE_REFLECT_BUILTIN */
 
 #if defined(DUK_USE_OBJECT_BUILTIN) || defined(DUK_USE_REFLECT_BUILTIN)
-/* Shared helper for Object.getOwnPropertyNames() and Object.keys().
- * Magic: 0=Object.getOwnPropertyNames() or Reflect.ownKeys(), 1=Object.keys().
- */
 DUK_INTERNAL duk_ret_t duk_bi_object_constructor_keys_shared(duk_context *ctx) {
+	/*
+	 *  magic = 0: Object.getOwnPropertyNames()
+	 *  magic = 1: Reflect.ownKeys()
+	 *  magic = 2: Object.keys()
+	 */
+
 	duk_hthread *thr = (duk_hthread *) ctx;
 	duk_hobject *obj;
 #if defined(DUK_USE_ES6_PROXY)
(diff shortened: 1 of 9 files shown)


fatcerberus:
Yep, I thought that would happen.  All I did was to change `Object.defineProperty()` to coerce its argument instead of throwing a TypeError and I already broke a test. :)

fatcerberus:
Annex E lists the functions affected by this.  It also lists other behavioral differences between E6 and E5[.1]:
http://www.ecma-international.org/ecma-262/6.0/#sec-additions-and-changes-that-introduce-incompatibilities-with-prior-editions

svaarala:
That's an interesting list, hadn't read that in detail before. I see that `ToNumber()` coercion needs octal support for example:

Node.js:
```
> +'0o123'
83
```

Duktape:
```
duk> +'0o123'
= NaN
```

svaarala:
Interestingly, there are also now *runtime* SyntaxErrors:

> In ECMAScript 2015, a runtime SyntaxError is thrown if a Catch clause evaluates a non-strict direct eval whose eval code includes a var or FunctionDeclaration declaration that binds the same Identifier that appears as the Catch clause parameter.

fatcerberus:
A runtime syntax error is a very odd concept indeed.  It makes sense though since eval() can be thought of as a kind of lazy evaluation.

fatcerberus:
@svaarala I'll look at finishing this up in the next few days so it'll be ready to go if you decide it should be merged.  The changes are straightforward, it's just a matter of fixing the testcases.

fatcerberus:
Weird, for some reason I thought `Object.defineProperty()` was included but it turns out that's not the case.  No idea why I thought that, it doesn't make much sense to pass a primitive value to it.

svaarala:
Many operations create a temporary object which is modified and then lost - that may seem pointless but it does cause side effects and the operation may be rejected e.g. by an inherited property.

But yes, Object.defineProperty() requires an object and won't coerce. Not sure why that's different from e.g. `Object.freeze()` which also required an object in ES5.1 but is a silent success in ES6.

fatcerberus:
I assume the rationale for `Object.freeze()` etc. is that one may do, e.g. `return Object.freeze(val);` at the end of a function and this allows that to work generically without an explicit type check.  For `Object.defineProperty()`  it doesn't make much sense to coerce - the defined properties would be lost which would probably lead to misbehavior anyway.

fatcerberus:
Well, I seem to have broken everything :)

I'll work on this tonight; I'm gearing up to start minisphere 5.0 development, so it'll take me a while to take stock of everything that needs to be done.  In the meantime I'm free to finish up this pull finally.

fatcerberus:
The ES6 argument policy changes have been implemented for all affected functions.  Next: fixing tests.

fatcerberus:
The testcases should all be fixed now.  I have to do some cleanup of the implementation before this is merge ready though; for the shared helpers it got a bit messy in spots because of the subtle behavioral differences between variants.

fatcerberus:
On closer inspection it seems that `Object.keys()` is still failing (this is probably a bug in the test), and Reflect arg policy is broken too.  So object coercion seems to be applied too broadly now.

I'll fix it tonight, no more time right now.

fatcerberus:
The Object.keys() fail was indeed a test bug.  `Object.keys('foo')` returns 3 keys: the character indices.  I forgot that strings were array-like and assumed the returned array would be empty, like with `Object.keys(123)`.

All tests should be passing now.  Just have to give the code a once-over to see if it can be cleaned up, verify the behavior matches ES6, then squash the commits and this will be ready to merge.

fatcerberus:
Hm, the algorithms for `Object.seal()` and `Object.freeze()` in ES6 differ slightly from ES5.  ES5 sets [[Extensible]] to false last:
http://www.ecma-international.org/ecma-262/5.1/#sec-15.2.3.8

ES6 does it first:
http://www.ecma-international.org/ecma-262/6.0/#sec-setintegritylevel

Not sure if that's observable.

fatcerberus:
@svaarala This is finally merge ready :)

fatcerberus:
Weird, all the tests passed except for linux-x86-packed-tval.  I don't think that's related to this pull?

svaarala:
I updated the testrunner to expect an object-based Duktape.info() which is needed for the packed duk_tval test. So it will now be broken until the object-based Duktape.info() is merged, and can just be ignored until then.

svaarala:
@fatcerberus Could you rebase to master? Let's see if the test error gets fixed.

This adds around +150 bytes to footprint for a relatively little change, so I'll open a follow-up to see if I can optimize the footprint a bit.

src-input/duk_bi_object.c
```diff
@@ -255,6 +257,11 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_seal_freeze_shared(duk_context
 	h = duk_require_hobject_promote_mask(ctx, 0, DUK_TYPE_MASK_LIGHTFUNC | DUK_TYPE_MASK_BUFFER);
 #endif
 
+	/* ES6 Sections 19.1.2.5, 19.1.2.17 */
+	if (!duk_is_object(ctx, 0)) {
+		return 1;
+	}
+
 	h = duk_require_hobject(ctx, 0);
```
svaarala: The above duk_is_object() call can be avoided by using duk_get_hobject() here and checking for NULL.

src-input/duk_bi_object.c
```diff
@@ -255,6 +257,11 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_seal_freeze_shared(duk_context
 	h = duk_require_hobject_promote_mask(ctx, 0, DUK_TYPE_MASK_LIGHTFUNC | DUK_TYPE_MASK_BUFFER);
 #endif
 
+	/* ES6 Sections 19.1.2.5, 19.1.2.17 */
+	if (!duk_is_object(ctx, 0)) {
+		return 1;
+	}
```
svaarala: How should plain buffers and lightfuncs behave here? They mimic objects, so they're generally accept anywhere where objects are.

src-input/duk_bi_object.c
```diff
@@ -293,9 +300,16 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_is_sealed_frozen_shared(duk_con
 		                          1 :               /* lightfunc always frozen and sealed */
 		                          (is_frozen ^ 1)); /* buffer sealed but not frozen (index props writable) */
 	} else {
-		h = duk_require_hobject(ctx, 0);
-		rc = duk_hobject_object_is_sealed_frozen_helper((duk_hthread *) ctx, h, is_frozen /*is_frozen*/);
-		duk_push_boolean(ctx, rc);
+		h = duk_get_hobject(ctx, 0);
+		if (h != NULL) {
+			rc = duk_hobject_object_is_sealed_frozen_helper((duk_hthread *) ctx, h, is_frozen /*is_frozen*/);
+			duk_push_boolean(ctx, rc);
```
svaarala: Footprint-wise better would be:

    duk_push_boolean(ctx, h == NULL ? 1 : duk_hobject_object_is_sealed_frozen_helper(...));

src-input/duk_bi_object.c
```diff
@@ -705,6 +760,11 @@ DUK_INTERNAL duk_ret_t duk_bi_object_constructor_prevent_extensions(duk_context
 
 	magic = duk_get_current_magic(ctx);
 
+	/* ES6 Section 19.1.2.15, step 1 */
+	if (magic == 0 && !duk_is_object(ctx, 0)) {
+		goto done;  /* silent success */
+	}
+
 	if (duk_check_type_mask(ctx, 0, DUK_TYPE_MASK_LIGHTFUNC | DUK_TYPE_MASK_BUFFER)) {
```
svaarala: The above duk_is_object() test can be implemented here, by choosing the type mask based on the magic value. For readability the type mask can be computed to a variable first.

src-input/duk_bi_object.c
fatcerberus: Those are handled above here it looked like?

svaarala: Ah, that's true. There wasn't enough context in the diff. Never mind.

src-input/duk_bi_object.c
fatcerberus: I saw that trick used in a couple other places, just didn't think to apply it here.

svaarala:
Regarding footprint, what I'm mainly concerned about for the short term is the low memory footprint for the ES5-ish subset. That's what's mostly used on low memory devices and literally every kilobyte saved in Duktape footprint allows a kilobyte more of actual application code.

So, I'm mostly trying to make sure the "minsize" build doesn't grow in footprint, at least for obvious reasons. It's of course not a blocker for making changes if additional footprint cannot be avoided and the change otherwise makes sense.

fatcerberus:
I guess most of the extra footprint is due to the extra checks, since there's now behavioral variance between, e.g. Reflect and Object.  I'll see if I can streamline it somewhat.

fatcerberus:
The packed duk_tval test passed now :)

fatcerberus:
I made the suggested changes, anything else I can do here?

svaarala:
I'll give it another read through in a bit and I can then merge the change.

In practice this merge means Duktape no longer targets ES5 compliance but rather provides an ES5-ish subset for low memory targets, with the "-ish" caused by semantic changes (to bindings present in ES5) in ES6 and later.

So, I'll need to open a pull for updating the website and other documentation to reflect this compliance goal (or rather, non-goal).

fatcerberus:
Well if we're being *technical* about it you already did a couple such merges, for example #1135... ;-)

svaarala:
Hehe, that's true - but those were really technical ones compared to changing the Object built-ins.

I'm fine with that change but it needs to be changed in the documentation.

svaarala:
The practical issue is that it's fine to merge this to master, and it will commit master more to the "ES5-ish" approach. It's still possible, of course, to add a "use strict ES5 semantics" config option and go over all the places with the changed semantics (that's been the basic idea for now). But the more changes are merged, the more extra work that becomes if it's done afterwards (for instance, testcases will need fixing, etc).

I'm leaning towards the ES5-ish approach for practical reasons, though. So it's not an issue for this pull; I just have to update the documentation as master will be more committed to this approach now.

svaarala:
Regarding foorprint, to illustrate the footprint changes in recent pulls, here's a graph of x64 low memory build (gcc -Os with link time removal of unused symbols) over the last 120 pulls (not commits); master on the right; the last uptick is caused by Duktape.info() getting an object format - quite noticeable here:

output

In last 120 pulls the footprint improvement has been roughly 1kB, for roughly the same (or increased) functionality. The low-memory configuration is roughly a minimal ES5 (or ES5-ish in master) subset. ARM Thumb builds are roughly speaking around 10% smaller IIRC.

As long as the footprint of this configuration doesn't increase compared to an earlier release, a project using Duktape can upgrade to a newer version and won't need to remove or size optimize application code. If the footprint does increase, say by 1kB, the application writer must shave off 1kB somewhere which can be a significant undertaking if the application is already, say, 40kB. Conversely, 1kB more space means the application scope can be increased.

So right now I'm tracking this ES5-ish low memory configuration w.r.t. footprint as that's a useful subset, and since the subset is fixed it's meaningful to compare build sizes. ES6 is a bit difficult to track because as new features are added, footprint will naturally grow. Technically it'd still be interesting to somehow measure "per feature footprint", e.g. the footprint of enabling Proxy over the ES5-ish baseline. Maybe later on comparing footprints of certain ES6 subsets (over versions) may be useful.

With the above in mind, I hope it's clearer why I'm frowning on a +200 bytes minsize footprint impact: it represents losing around 20% of the gains made in the last 120 pulls. It's a constant struggle but very meaningful for people doing low memory stuff :)

fatcerberus:
And now you have me wondering what that massive spike was around 90 pulls ago... :)

svaarala:
Quite likely it was some new feature (non-ES5) added to master, but without a line in `low_memory.yaml` to disable it for the low memory configuration. That was then fixed and the spike went away. That's the most likely cause anyway :)

fatcerberus:
It might be possible to reverse most of the footprint impact here by refactoring things a bit, e.g. rather than a large shared Duktape/C helper with branching code paths based on a magic value, maybe the base functionality itself could be moved into helpers - and then Object and Reflect methods can have separate Duktape/C functions which call the shared primitives after ensuring the value stack state is correct etc.  This way when Reflect support is not present you don't pay for having the related code paths.

svaarala:
Earlier when I was experimenting with footprint, generally a function with a magic value has better footprint than two entrypoints that call a shared helper with flags. But doing so does have the upside that when only one call site remains, the compiler can inline the call and remove any flags checks as having a constant value. With a magic value the compiler can't know what values are possible, so, as you said, the code paths remain.

The other approach is to keep a single function but #ifdef the relevant magic value checks.

If this pull is otherwise functional, I'm fine with merging it and experimenting with footprint improvement as a follow-up pull; if that doesn't work out with a reasonable effort then the added footprint is fine too.

fatcerberus:
Alright, sounds good to me :)  Do I have to rebase again?

svaarala:
Doesn't seem to conflict and changes in master are unrelated, so no need. I'll merge this tomorrow - I've got to sleep for a bit :)

fatcerberus:
Whoops, found an unreferenced variable warning and couldn't leave it unfixed :)

fatcerberus:
@svaarala Not related to this pull specifically but I see your minisphere snapshot is a bit old (May 16).  As various useful enhancements like #1128 get merged into master, I try to take advantage of them so it might be good to keep the minisphere testrunner up to date to help ensure that the newer features aren't broken either.

Six months doesn't sound like much, but for a project as actively developed as minisphere it's an eternity.  Looking at my release history real quick, on 05/16 the current release was v3.1,2.  I'm at v4.3.8 now. ;)

svaarala:
I'll build a few graphs nightly here for fun: http://duktape.org/graphs.html.

svaarala:
@fatcerberus I froze the Minisphere version in the builds because there wasn't support for configuring via DUK_USE_xxx options and `tools/configure.py` at that point - could I now use Minisphere master?

fatcerberus:
No automated support no, but currently I do use configure.py internally to build the bundled duktape.c.  I still need a fixup for `DUK_OPT_HAVE_CUSTOM_H` though because I still support Duktape 1.x (mostly at your request :)

What would I need to do to enable you to update?

svaarala:
Hmm, so maybe if I make the test run tools/configure.py and overwrite the bundled duktape.c - and then just build normally?

fatcerberus:
Yeah, that should work.  I actually thought that's what that test did already...?

svaarala:
Close but it needs a fixup file:

```python
def context_linux_x64_minisphere():
    cwd = os.getcwd()

    execute([ 'make', 'dist' ])

    # Unpack minisphere snapshot and copy Duktape files over.
    unpack_targz(os.path.join(repo_snapshot_dir, 'minisphere-20160516.tar.gz'))

    prep(options=[ '--fixup-file', os.path.join(cwd, 'minisphere', 'src', 'engine', 'duk_custom.h') ])

    for i in [ 'duktape.c', 'duktape.h', 'duk_config.h' ]:
        execute([
            'cp',
            os.path.join(cwd, 'prep', i),
            os.path.join(cwd, 'minisphere', 'src', 'shared', i)
        ])

    # sudo apt-get install liballegro5-dev libmng-dev
    os.chdir(os.path.join(cwd, 'minisphere'))
    return execute([ 'make' ], catch=True)['success']
```

Would this work in Minisphere master?

fatcerberus:
Assuming that prep() runs configure.py, that should work in master.  For my part I'll see if I can phase out the `duk_custom.h` for minisphere 5.0.

PR #1032, opened 2016-10-26
Small cleanups to ES6 Reflect built-in

svaarala:
Some compile warning and other small cleanups to the ES6 Reflect built-in. Some of the changes are to code already in master prior to Reflect, but were convenient to improve at the same time.

### DIFF
--- a/dist-files/README.rst
+++ b/dist-files/README.rst
@@ -9,8 +9,8 @@ spirit to Lua's.
 Duktape supports the full E5/E5.1 feature set including errors, Unicode
 strings, and regular expressions, a subset of Ecmascript 2015 (E6) and
 Ecmascript 2016 (E7) features (e.g. computed property names, Proxy objects,
-exponentiation operator), Khronos/ES6 ArrayBuffer/TypedView, Node.js Buffer,
-and WHATWG Encoding API living standard.
+exponentiation operator, Reflect), Khronos/ES6 ArrayBuffer/TypedView, Node.js
+Buffer, and WHATWG Encoding API living standard.
 
 Duktape also provides a number of custom features such as error tracebacks,
 additional data types for better C integration, combined reference counting
--- a/src-input/duk_api_internal.h
+++ b/src-input/duk_api_internal.h
@@ -198,6 +198,8 @@ DUK_INTERNAL_DECL void duk_unpack(duk_context *ctx);
 
 DUK_INTERNAL_DECL void duk_require_constructor_call(duk_context *ctx);
 
+DUK_INTERNAL_DECL void duk_require_constructable(duk_context *ctx, duk_idx_t idx);
+
 /* Raw internal valstack access macros: access is unsafe so call site
  * must have a guarantee that the index is valid.  When that is the case,
  * using these macro results in faster and smaller code than duk_get_tval().
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -1605,6 +1605,18 @@ DUK_EXTERNAL void duk_require_function(duk_context *ctx, duk_idx_t idx) {
 	}
 }
 
+DUK_INTERNAL_DECL void duk_require_constructable(duk_context *ctx, duk_idx_t idx) {
+	duk_hobject *h;
+
+	h = duk_require_hobject_accept_mask(ctx, idx, DUK_TYPE_MASK_LIGHTFUNC);
+	if (h == NULL) {
+		return;  /* Lightfuncs are constructable. */
+	}
+	if (!DUK_HOBJECT_HAS_CONSTRUCTABLE(h)) {
+		DUK_ERROR_REQUIRE_TYPE_INDEX((duk_hthread *) ctx, idx, "constructable", DUK_STR_NOT_CONSTRUCTABLE);
+	}
+}
+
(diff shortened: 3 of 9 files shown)


svaarala:
@fatcerberus I'll add some small fixes here by tomorrow, but this is a placeholder for now.

svaarala:
@fatcerberus About this btw: https://github.com/svaarala/duktape/pull/1032/commits/988fd9c4ba92116cbd36bcff12124d957d7f5b9d, when you know the value stack contains the index, `DUK_GET_TVAL_{POSIDX,NEGIDX}` generates code for e.g. `tv = thr->valstack_bottom + 3` which is just a few instructions. But you do need to know the index exists; this is often the case for built-ins whose function `nargs` is fixed.

This kind of direct access will be eventually possible in the public API macros. It will make pushing much faster because e.g. `duk_push_number(ctx, 123)` would then, with "unsafe value stack" option enabled, compile down to just:

    DUK_TVAL_SET_NUMBER(thr->valstack_top++, 123);  /* no refcount update */

There are a few pending issues to make this work, for example the public API header also needs internal struct definitions so that API macros can look into them. Several API operations can then avoid an unnecessary call, for example `duk_get_top(ctx)` will just be:

    (duk_idx_t) (thr->valstack_top - thr->valstack_bottom)

which avoids a call.

fatcerberus:
> `DUK_GET_TVAL_{POSIDX,NEGIDX}`

Thanks, I had forgotten about those.  We went through this a while back when I worked on AppRequest :)

svaarala:
Ok, that's about all the trivia - I'll read the Reflect section from ES6 and check property attributes once more and then this should be done.

fatcerberus:
I think the argument check in `Reflect.construct()` is wrong - it should check that the function passed in is constructable, not just callable.  For Ecmascript functions there's no difference (IIRC), but it may matter for native calls.

svaarala:
That would only matter for coercion / check order I guess because `duk_new()` will fail if the target is not constructable?

svaarala:
There's actually no way to create non-constructable Ecmascript or C functions now - but the internals do support the difference and some built-in functions are magically non-constructable:

```
```

Also what would be convenient is the ability to create functions that are *only* constructable, not callable as plain functions. That would remove a bunch of `duk_is_constructor_call()` checks from the internals and also probably from some user code. This would be easy to add with an additional flags field for `duk_push_c_function()` (or maybe aliasing some flags into the `nargs` field.

fatcerberus:
As long as `duk_new()` throws a TypeError that may be adequate.  I'll double check the specification after work.

svaarala:
So current behavior is, for example:

```
duk> Reflect.construct(Math.cos,[])
TypeError: [object Function] not constructable
    at [anon] (duk_api_call.c:431) internal
    at construct () native strict preventsyield
```

svaarala:
The current behavior is only incorrect in coercion order: step 1 of http://www.ecma-international.org/ecma-262/6.0/#sec-reflect.construct already checks for `IsConstructor()` so that should fail before the argument array is processed. As that may have side effects, it matters whether the constructor check happens before the array handling or in `duk_new()`.

I'm happy to write this up into a Reflect cleanup issue, along with newTarget.

svaarala:
If this is fixed now, I'll add a `duk_require_constructable()` internal helper call to make it reusable elsewhere.

fatcerberus:
I'm not sure but I think ES6 arrow functions might not be constructable, so that check will become more relevant in the future. :)

svaarala:
Sure, the check exists of course in `duk_new()` at the latest. And there are already functions that are not constructable like `Math.cos` so it's already visible (though maybe not very common) behavior.

I can add that `duk_require_constructable()` in this pull. I'll need to add a testcase to demonstrate the current wrong behavior first though. I'll do that in a bit.

svaarala:
Okay, `duk_require_constructable()` is in (but not called by Reflect yet) and testcase extended. So it should now fail.

Also checked that V8 respects the coercion order as discussed above.

svaarala:
Yes, broke as expected.

fatcerberus:
The constructable test should probably be moved into `test-bi-reflect-arg-policy.js`.  Other than that the changes look good.

svaarala:
The style I've used for switches is that braces are used if there are locals (which cannot otherwise be declared portably). And if one clause has a brace, then I've usually added it to all of them. But in this case no clause needs braces.

svaarala:
Ok, moved the test.

svaarala:
Here's an example of a switch-case which doesn't need braces: https://github.com/svaarala/duktape/blob/master/src-input/duk_error_misc.c#L43-L61.

But, it might simpler to just require braces in every case. However, that looks funny when there's a fallthrough, as in this case :-/

fatcerberus:
@svaarala Just a heads up: `linux-x64-qecmatest-assert` is failing:
```
test-dev-regexp-negative-jump-offset: fail; 9 diff lines
```

Fallout from #1033, looks like.

svaarala:
I re-ran the test and it's just a test timeout followed by the test being killed and thus failing to match the expect string.

svaarala:
Forgot to push the squashed and cleaned up commits so these went to master as is. Looks ugly in the commit log but not a big deal otherwise.

PR #1037, opened 2016-10-27
Return 1 from duk_push_boolean() and variants

svaarala:
Currently `duk_push_boolean()` returns void; change it to return `(duk_ret_t) 1` so that one can:

```c
return duk_push_boolean(ctx, x == y);
```

This allows tailcalls to also be used internally which should reduce footprint. It's also convenient in user code, so the change could be added to the public API. At least the following API functions which now return void could return 1 to facilitate tail calls:

- duk_push_undefined()
- duk_push_null()
- duk_push_boolean()
- duk_push_true()
- duk_push_false()
- duk_push_number()
- duk_push_nan()
- duk_push_int()
- duk_push_uint()
- duk_push_pointer()

Also related (often requested) change is to add a return value to `duk_error()` and `duk_throw()` prototype so that one can:

```c
return duk_error(ctx, ...);
```

The downside of the change is some additional compiler output for loading the (often unused) result. The result values of the pusher API calls will also be a little bit inconsistent in that some return a value related result (for example, a `const char *` for a pushed string) while others return a dummy tailcall-facilitating value. The result values are thus driven by convenience rather than being strictly consistent.

This pull is a small test in what the impact of making such a change is. If it works internally to reduce footprint, the API change can be avoided by simply using an API macro and casting to void. So the API decision is separate.


svaarala:
Moving to 3.0.0 for consideration. The most important change is for `duk_error()` and variants to allow the `return duk_error(...)` idiom; that's already in master for 2.0.0 release.

PR #1038, opened 2016-10-28
Add duk_ret_t return value to duk_{throw,error,error_va,fatal}() prototype

svaarala:
While the functions never return, a `duk_ret_t` return value allows application code to do something like:

```c
if (argvalue < 0) {
    return duk_error(ctx, DUK_ERR_TYPE_ERROR, "invalid arg: %d", (int) argvalue);
}
```

If this is not the case and the compiler doesn't support noreturn functions, application code needs an explicit return after a `duk_error()` to prevent potential compiler warnings (e.g. use of uninitialized variables) when the compiler thinks execution will resume after the error.

Same treatment for `duk_error_va()`, `duk_throw()`, and `duk_fatal()`.

Tasks:
- [x] API changes
- [x] Testcases
- [x] API documentation changes
- [x] 2.x migration note (just a note there's a return value now)
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1785,6 +1785,10 @@ Planned
   in line with how application code detects active features and reduces
   footprint (GH-988)
 
+* Change return from of duk_throw(), duk_error(), and duk_fatal() from void
+  to duk_ret_t which allows them to be called using the idiom
+  "return duk_error(ctx, DUK_ERR_TYPE_ERROR, "invalid argument");" (GH-1038)
+
 * Add convenience API calls duk_get_prop_lstring(), duk_put_prop_lstring(),
   duk_del_prop_lstring(), duk_has_prop_lstring(), duk_get_global_lstring(),
   duk_put_global_lstring() (GH-946, GH-953)
--- a/doc/release-notes-v2-0.rst
+++ b/doc/release-notes-v2-0.rst
@@ -601,6 +601,23 @@ To upgrade:
   to accept also ``TypeError`` or ``RangeError``.  (In general depending on a
   specific error type should be only be done when it's absolute necessary.)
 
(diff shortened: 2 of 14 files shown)


svaarala:
There's one more convenience issue in this idiom:

```c
if (argvalue < 0) {
    return duk_error(ctx, DUK_ERR_TYPE_ERROR, "invalid arg: %d", (int) argvalue);
}
```

The error type argument is a bit awkward, so I'll add a few type specific calls in a separate pull so that one can:

```c
if (argvalue < 0) {
    return duk_type_error(ctx, "invalid arg: %d", (int) argvalue);
}
```

After this change the error functions are a bit nicer to use.

fatcerberus:
Hm, MSVC is apparently not a big fan of this change:
```
1>..\src\shared\duktape.c(19863): warning C4646: function declared with 'noreturn' has non-void return type
1>..\src\shared\duktape.c(19905): warning C4646: function declared with 'noreturn' has non-void return type
1>..\src\shared\duktape.c(19934): warning C4646: function declared with 'noreturn' has non-void return type
1>..\src\shared\duktape.c(19944): warning C4646: function declared with 'noreturn' has non-void return type
```

svaarala:
That would be easy to fix by removing the noreturn. But without noreturn the generated call sites are much worse (and larger).

I could change the calls to be macros like:

```c
#define duk_error(ctx,err_code,...) \
    (duk_error_raw(<args>), 0) 
```

and then change `duk_error_raw()` back to returning void.

That would run the risk of creating warnings for "unreachable code" but it's worth a try.

fatcerberus:
I don't think that works.  The comma operator requires a value in all positions, a void-returning function isn't valid in that context I think...?

PR #1041, opened 2016-10-29
Another duk_error() et al return value approach

svaarala:
Try to fix MSVC warning of a noreturn function having a return value in its prototype.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1787,7 +1787,8 @@ Planned
 
 * Change return from of duk_throw(), duk_error(), and duk_fatal() from void
   to duk_ret_t which allows them to be called using the idiom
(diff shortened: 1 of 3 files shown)


svaarala:
Just `duk_error()` change in this test for now, let's see what MSVC thinks about it.

svaarala:
That would seem to fix it for MSVC. I'll complete this pull in a while.

fatcerberus:
Weird, I didn't think this would work.  The last time I tried using a void-returning function in a comma operator expression I remember getting a compile error.

svaarala:
Something like `(void, void, 123)` won't compile but this will:

```c
#include <stdio.h>
void dummy(void) {
}
void main(void) {
	printf("test %d\n", (dummy(), dummy(), 123));
}
```

PR #1042, opened 2016-10-30
Faster debugger paused check in refzero handling

svaarala:
When debugging is enabled the check to skip refzero handling is:

```c
(DUK_HEAP_HAS_MARKANDSWEEP_RUNNING(heap) || DUK_HEAP_IS_PAUSED(heap))
```

But because "mark-and-sweep running" is a heap flag and "debugger is paused" is a separate `duk_heap` field, this compiles into awkward code:

```
  4037aa:       f6 03 01                testb  $0x1,(%rbx)
  4037ad:       75 31                   jne    4037e0 <duk_hstring_refzero+0x40>
  4037af:       8b 93 14 01 00 00       mov    0x114(%rbx),%edx
  4037b5:       85 d2                   test   %edx,%edx
  4037b7:       75 27                   jne    4037e0 <duk_hstring_refzero+0x40>
```

This pull changes `heap->dbg_paused` into a heap flag which improves the check to:

```
  4037aa:       f6 03 41                testb  $0x41,(%rbx)
  4037ad:       75 27                   jne    4037d6 <duk_hstring_refzero+0x36>
```

It would be better if the entire dbg_paused check was omitted (it should be implied by mark-and-sweep being disabled, at least right now). But for processors with a bit field test primitive there should be no practical difference.

Tasks:

- [x] Convert dbg_paused flag
- [x] Convert other heap binary flags too if it improves footprint (the remaining flags are performance insensitive) => separate pull
- [ ] Manual performance test run
- [x] Manual debugger smoke test
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2012,7 +2012,8 @@ Planned
   Array .length coercion (GH-862); value stack operation optimization
   (GH-891); call related bytecode simplification (GH-896); minor bytecode
   opcode handler optimizations (GH-903); refcount optimizations (GH-443,
-  GH-973); minor RegExp compile/execute optimizations (GH-974, GH-1033)
+  GH-973, GH-1042); minor RegExp compile/execute optimizations (GH-974,
+  GH-1033)
 
 * Miscellaneous footprint improvements: RegExp compiler/executor (GH-977);
   internal duk_dup() variants (GH-990); allow stripping of (almost) all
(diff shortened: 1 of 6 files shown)


svaarala:
I'll merge this now and post the performance results later when the test is done.

svaarala:
Since this is just removing one check-and-branch from refzero handling (which accounts for a tiny fraction of execution time) the impact is not easy to measure and pretty small as expected. Very specific tests like addition, interning, etc, show small improvement or decline, but overall there's a small improvement. Tests involving more functionality generally show a small improvement, e.g.

```
test-mandel.js                      : duk.O2.dbg  3.27 duk.O2.master  3.53 duk.O2.150  4.47
```

Then again, this may just be due to code being reorganized by the compiler compared to master (which then affects e.g. caching performance).

But overall, having one less compare-and-branch with no trade-off involved should be generally better, and reduces branch prediction state which has an indirect potential effect elsewhere.

PR #1043, opened 2016-10-30
Add ES6 String.prototype.codePointAt()

svaarala:
Add support for `String.prototype.codePointAt()`. Also revises `duk_char_code_at()` and `String.charCodeAt()` behavior for decode error: instead of an error, a U+FFFD is returned. This behavior only affects non-standard strings so it has no compliance effect.

I'll add `String.fromCodePoint()` in a separate pull, it's a counterpart to this function. The quite useful `for-of` iteration for strings won't be supported yet because there's no iterator or for-of support in general.

- [x] Add binding
- [x] Testcase updates
- [x] Migration notes
- [x] Releases entry


svaarala:
This binding isn't conditional to any config option for now. I'm not sure what to do with individual bindings like these yet w.r.t. config, entire built-ins (like Reflect) are easier from a config perspective.

fatcerberus:
IMO it's probably not worth it to have config options for individual methods.  Maybe we could have some kind of a catch-all option for "post-ES5 support" and then all the ES6/7/etc. options would depend on that.

svaarala:
I'm leaning towards some kind of catch-all approach for small features that are version dependent but unlikely to be enabled/disabled individually. One simple approach is what you propose:
- A config option for enabling post-ES5 support.

Another is to just provide generic defines for each release, e.g.:
- DUK_USE_ES6: enable all ES6 functionality not controlled by other config options
- DUK_USE_ES7: enable all ES7 functionality not controlled by other config options
- Etc.

That's better in some ways, e.g. it's easier to see where feature were introduced which may be sometimes interesting.

But then, with yearly specification releases that's unlikely to matter very much at least beyond ES7.

Right now ES5 is a useful baseline for low memory targets but going forward a few years it may be that one would like to have e.g. Symbols even in such a target (and perhaps remove some ES5 functionality). So a "pure ES5" configuration is most likely going to become less and less useful as a starting point over time.

Ponder.

svaarala:
By the way, the problem with adding a lot of individual defines is not really the defines themselves, at least in my view. The current tooling handles that pretty transparently, and up-to-date example YAML config files provide relatively stable logical starting points for different environments so that one doesn't need to deal with all the individual options in general.

The most important practical issue is the `#ifdef`s in the code base, and there specifically the proper conditions for including or excluding some shared helper which often gets an interesting multipart condition.

That, in itself, is not really an issue except for compiler warnings. Modern compilers will strip out any unnecessary functions (at least if you tell them to). But it's still a bit awkward to maintain, and many users (rightly or wrongly) want to treat all warnings as errors.

svaarala:
Added `DUK_USE_ES6` for generic, uncategorized options for now.

fatcerberus:
> The most important practical issue is the #ifdefs in the code base, and there specifically the proper conditions for including or excluding some shared helper which often gets an interesting multipart condition.

Agreed, the tooling wasn't my concern either, it was the necessary manual maintenance of `#ifdef`s which could get out of hand fast if there's a lot of them and/or too many complicated conditions to keep track of.  ES7 exponentiation being a good example - it turned out that it needed to pull in `pow()` which previously was only needed for the `Math` builtin.

svaarala:
For the pow() case a "post-ES5" define would have had the same issue BTW :)

fatcerberus:
I know, I was just using it as an example, in other words, "take this case and extrapolate out to hundreds of such options" :)  A catch-all "post-ES5" define seems to me like it would make for fewer maintenance cost in the long term vs. a whole bunch of granular ones.

fatcerberus:
That said, I do think it's worthwhile to continue adding individual defines for entire builtins like Reflect.  It's nice to be able to opt in and out of individual components, like some kind of JS ala carte. 😄

svaarala:
>  A catch-all "post-ES5" define seems to me like it would make for fewer maintenance cost in the long term vs. a whole bunch of granular ones.

I agree with that, it's certainly less maintenance to have fewer options. In fact, without low memory targets it'd be quite logical to have essentially *no* feature options (at least as far as built-ins, operators, etc are concerned): you can just compile them in, and initialization code can delete the bindings if they're not wanted.

So granular config options really serve the low memory targets where it's maybe not possible to include all bindings. When that's the case, it's preferable, all other things being equal, to be able to control with precision what gets compiled in. The downside to that is maintenance, testing configuration combinations (must be automated in practice), documentation, and so on.

The easy solution which produces pretty good compiled results (with modern compilers) is to just not worry about #ifdefing helpers: they'll be left out at link stage if they're not needed. The only hitch is that it generates compiler warnings, and many users won't be happy with that.

I merged that `DUK_USE_ES6` catch-all into master for now - I don't mean it to be final. There should be one round of config option cleanup for 2.0 release, so it can be changed easily up until that point.

fatcerberus:
Off-topic, it'd annoy me to have the compiler warn me about dead code.  IMO that's not the compiler's job (beyond simple things anyway; warnings for e.g. unreachable code within a function do make sense) - if one needs to find dead functions then some kind of static analysis tool is called for.

PR #1046, opened 2016-10-30
Fix String.fromCharCode() coercion to ToUint16() when standard behavior enabled

svaarala:
When standard behavior is enabled (DUK_USE_NONSTD_STRING_FROMCHARCODE_32BIT) ToUint32() + CESU-8 was used which emits two characters for each non-BMP codepoint. The standard behavior is disabled by default (ToUint32() + UTF-8 is the default) so this only has an effect if specifically enabling the standard behavior.

- [x] Fix bug
- [x] Manual test (no regression coverage now for this non-default option)
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1934,6 +1934,11 @@ Planned
   seed mixing; previous algorithm (Shamir's three-op algorithm) is still
   used for low memory targets and targets without 64-bit types (GH-970)
 
+* Fix String.fromCharCode() behavior for non-BMP characters when standard
+  behavior is enabled (DUK_USE_NONSTD_STRING_FROMCHARCODE_32BIT disabled):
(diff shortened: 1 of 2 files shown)


svaarala:
Behavior with default options (= non-standard String.fromCharCode()):

```
duk> x = String.fromCharCode(0x123456);
= "\U00123456"
duk> x.length
= 1
duk> x.charCodeAt(0)
= 1193046
```

Behavior with compliant behavior enabled, before fix:

```
duk> x = String.fromCharCode(0x123456);
= "\ud84d\udc56"
duk> x.length
= 2
duk> x.charCodeAt(0)
= 55373
duk> x.charCodeAt(1)
= 56406
```

And after fix:

```
duk> x = String.fromCharCode(0x123456);
= "\u3456"
duk> x.length
= 1
duk> x.charCodeAt(0)
= 13398
```

PR #1050, opened 2016-10-31
Add ES6 String.prototype.repeat()

svaarala:
- [x] Rebase once #1049 is merged
- [x] Basic implementation
- [x] Small footprint variant
- [x] Internal fixed buffer zeroing optimization (very often zeroing not needed)
- [x] Note on UTF-8 safety: if input is UTF-8, output is also UTF-8; if input is not UTF-8 it's possible for initial and last bytes of the repeated string to combine in the repetition => noted in test case
- [x] Testcase
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1836,8 +1836,8 @@ Planned
   and identifiers, no RegExp support yet (requires RegExp /u Unicode mode)
   (GH-1001)
 
-* Add support for ES6 String.prototype.codePointAt() and String.fromCodePoint()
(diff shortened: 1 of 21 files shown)


fatcerberus:
There's a (probably) harmless warning in master related to this change:
```
1>duk_bi_string.c(1345): warning C4267: '=': conversion from 'size_t' to 'duk_uint_t', possible loss of data (compiling source file ..\src\shared\duktape.c)
```

svaarala:
Thanks, I'll fix that.

PR #1051, opened 2016-11-01
Small IEEE double handling improvements

svaarala:
Minor footprint and performance improvements for IEEE double handling:
- Avoid explicit `signbit()` calls here and there where it benefits both footprint and performance.
- Rework some `duk_js_ops.c` algorithms like comparison for better footprint and performance.
- Other odds and ends

Dropping dependency entirely to e.g. `signbit()` and `fpclassify()` would simplify porting and avoid pulling them in from a libc provider on bare metal targets.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2021,12 +2021,13 @@ Planned
   (GH-891); call related bytecode simplification (GH-896); minor bytecode
   opcode handler optimizations (GH-903); refcount optimizations (GH-443,
   GH-973, GH-1042); minor RegExp compile/execute optimizations (GH-974,
-  GH-1033)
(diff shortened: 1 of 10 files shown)


svaarala:
Final score seems to be around 300 bytes off x86 minsize and 220 bytes off x86 stripsize. Performance impact for default configuration is small, but places where a lot of comparisons or coercions are made are improved a bit, e.g.:

```
test-empty-loop-slowpath.js         : duk.O2.double  1.86 duk.O2.master  1.95
test-empty-loop.js                  : duk.O2.double  1.58 duk.O2.master  1.68
test-mandel.js                      : duk.O2.double  2.92 duk.O2.master  3.22
test-string-compare.js              : duk.O2.double  1.93 duk.O2.master  2.12
```

PR #1054, opened 2016-11-03
Revise enumeration order to match common ES6 engine behavior

svaarala:
- [x] Add support for ES6 enumeration order as a sorting post-step (similarly to existing array index sort)
- [x] Performance optimization for "get enumerated keys", write keys directly into dense array part
- [x] Check if duk__sort_array_indices() actually works for ES6 for now (without symbols!)
- [x] Check duk_enum() flags, can some flags be removed?
- [x] Avoid explicit sorting when not needed
- [x] Config option, low memory ES5 build won't need it => Not necessary, footprint impact is < 100 bytes and existing sort algorithms works for now (at least without symbols)
- [x] Website note on enumeration order
- [x] Clarify on website that `for-in` etc are not guaranteed by ES6 or ES7 but Duktape follows the order like most other engines
- [x] Clarify code and comments that enumeration order matches ES6/ES7 (not just ES6)
- [x] Internal document fixes
- [x] Testcases and testcase fixes
- [x] Testcase coverage: random orders and sorting
- [x] Performance test: enumeration
- [x] JSON enumeration order (own properties for object, inherited properties for array!)
- [x] 2.0 migration notes
- [x] Releases entry
- [x] Future work issue for `duk_enum()` flag cleanup, `DUK_ENUM_SORT_ARRAY_INDICES` is not very descriptive now => #1060

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1843,6 +1843,11 @@ Planned
   Ecmascript code to read and write text stored in an ArrayBuffer or a plain
   buffer (GH-975)
 
+* Respect ES6 enumeration order (array index keys, other keys in insertion
+  order) for Object.getOwnPropertyNames(), also use the same order in
+  for-in, Object.keys(), and duk_enum() even though that's not strictly
+  required by ES6 or ES7 (GH-1054)
+
 * Remove no longer needed platform wrappers in duk_config.h: DUK_ABORT(),
   DUK_EXIT(), DUK_PRINTF(), DUK_FPRINTF(), DUK_FOPEN(), DUK_FCLOSE(),
   DUK_FREAD(), DUK_FWRITE(), DUK_FSEEK(), DUK_FTELL(), DUK_FFLUSH(),
--- a/doc/hobject-enumeration.rst
+++ b/doc/hobject-enumeration.rst
@@ -2,9 +2,7 @@
 Object enumeration issues
 =========================
 
-.. note:: This document is quite incomplete.  Tests should be expanded
-          so that the same enumeration test is executed at least in
-          Rhino and V8.
+This document provides some design notes for pre-ES6 and ES6/ES7 enumeration.
 
 Key order during enumeration
 ============================
@@ -21,7 +19,7 @@ However, some existing code apparently relies on some ordering behavior:
     order in which they were defined. Because of this the Chrome team has
     deemed this to be a bug and will be fixing it.
 
-Apparently the behavior relied on is roughly:
+User code seems to rely roughly on the following order:
 
 * For arrays, return all used array index keys in ascending order first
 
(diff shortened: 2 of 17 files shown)


fatcerberus:
Incidentally this guarantee is likely the reason `Reflect.enumerate()` was dropped in ES7.  It makes things very difficult for the engine if an object can mutate while there's a live iterator but key order still has to be guaranteed.

svaarala:
Hmm, but a plain `for-in` enumeration has the same issue: the live iterator is used for a while and it must deal consistently with mutation. I tried to implement for-in initially without an explicit iterator state (which is initialized once and then won't rely on the objects not mutating) but that turned out to be very difficult: one would need to at least freeze objects against key reordering, it has GC impact, etc, etc.

svaarala:
This won't be as easy as I thought: ES6 enumeration order is not strictly "array indices followed by keys in insert order (followed by symbols)" but rather that rule is used for *own properties* only. Inherited non-duplicate properties then follow, even if they are array indices. Node.js example:

```
> var parent = { 4: 'foo' }
undefined
> var object = { foo: 1, 2: 'bar', bar: 'quux' }
undefined
> object.__proto__ = parent;
{ '4': 'foo' }
> Object.keys(object)
[ '2', 'foo', 'bar' ]
> for (var k in object) { console.log(k); }
2
foo
bar
4
undefined
```

So, it's not correct to form the enumeration sequence (including parents) and then sort it. One should rather gather keys for one inheritance level, sort them, and then repeat the process for each inheritance level in turn, appending any non-duplicate keys, sorted using the ES6 rule, for each level.

svaarala:
Another example:

```
> var a = { 1: 'foo' }; var b = { 2: 'foo' }; var c = { 3: 'foo' }; var d = { 4: 'foo' }; var e = [ 'foo' ];
undefined
> e.__proto__ = d; d.__proto__ = c; c.__proto__ = b; b.__proto__ = a;
{ '1': 'foo' }
> for (var k in e) { console.log(k); }
0
4
3
2
1
undefined
```

Array related:

```
> var a = [ 1,1,1,1,1 ]; var b = { 3: 'foo' }; b.__proto__ = a;
[ 1, 1, 1, 1, 1 ]
> for (var k in e) { console.log(k); }
0
undefined
> for (var k in b) { console.log(k); }
3
0
1
2
4
undefined
```

Array gaps:

```
> var a = [ 1,1,1,1,1 ]; var b = [1,,,1]; b.__proto__ = a;
[ 1, 1, 1, 1, 1 ]
> for (var k in b) { console.log(k); }
0
3
1
2
4
undefined
```

fatcerberus:
It can never be easy can it? :(

svaarala:
Similar but shows how parent properties are again sorted (separately):

```
> var a = { 4: true, 0: true, 3: true, 1: true, 2: true }; var b = [1,,,1]; b.__proto__ = a;
{ '0': true, '1': true, '2': true, '3': true, '4': true }
> for (var k in b) { console.log(k); }
0
3
1
2
4
```

svaarala:
I think I found a reasonable way to implement this so not to worry :-) The enumeration code will need overhauling for ES6 anyway but it'd be nice to get the enumeration order correct for ordinary objects and arrays.

svaarala:
Heh, it turns out the original array index sort algorithm is actually ES6 compatible as long as it is applied per inheritance level. The insertion sort maintains key order for keys with the same array index, and all non-array-indices have a marker index 0xFFFFFFFF so they sort after array indices but keep their order.

This won't work with symbols later on so a more flexible algorithm will be better later.

svaarala:
@fatcerberus I'll read the ES6 requirements carefully before merging, but it might be that enumeration order is not guaranteed for e.g. `for-in` or `Object.keys()`, but only for some bindings like `Object.getOwnPropertyNames()`:

- http://stackoverflow.com/questions/30076219/does-es6-introduce-a-well-defined-order-of-enumeration-for-object-properties
- https://esdiscuss.org/topic/property-ordering-of-enumerate-getownpropertynames

If this is indeed the case no change to `for-in` etc would be needed to conform to ES6, only specific bindings with guarantees would need changing.

Even if not required I think this change makes sense because it's least surprising, and matches other implementations. It also removes the outward difference when enumerating dense vs. sparse arrays.

fatcerberus:
Also might be worth cross-referencing to ES7 as it revises some of the semantics.

svaarala:
It seems there are still no specific guarantees in ES6 for `for-in` or `Object.keys()` if I read the specification correctly:
- `for-in` and `Object.keys()` don't have any stronger guarantees compared to ES5 (I'm not 100% sure but the requirements are more or less the same). The internal algorithm calls `[[Enumerate]]` which has no guarantees for order (but does for some other things like no duplicate keys, mutation stuff etc): "The mechanics and order of enumerating the properties is not specified but must conform to the rules specified below.".
- `Object.getOwnPropertyNames()` calls the internal `GetOwnPropertyKeys()` specification method which in turn calls `[[OwnPropertyKeys]]` which *does* provide the "ES6 enumeration order", i.e. array indices first, non-array-index keys in insertion order, symbols in insertion order.

I'll update the documentation, code comments etc, to be clear that while no specific requirements exist, `for-in` etc follow the same order as `Object.getOwnPropertyNames()`.

svaarala:
For ES7 `for-in` calls `EnumerateObjectProperties()` which has:

> The mechanics and order of enumerating the properties is not specified but must conform to the rules specified below.

So the situation seems unchanged from ES6 at least for enumeration order.

svaarala:
Same thing for ES7 `Object.keys()`:

ES7 `Object.keys()` calls `EnumerableOwnNames()`, http://www.ecma-international.org/ecma-262/7.0/#sec-enumerableownnames, whose step 5 says:

> Order the elements of names so they are in the same relative order as would be produced by the Iterator that would be returned if the EnumerateObjectProperties internal method was invoked with O.

So the guarantees are the same as for `for-in`, which calls `EnumerateObjectProperties()`.

svaarala:
I'll fix the enumeration order references in code comments, documentation, website, etc to reflect that the guarantees are stronger than required (and match most other engines). But I'll merge the behavior as is because it's more predictable and less surprising: a lot of people assume ES6 guarantees ordering (and I was under that impression too before checking).

fatcerberus:
So if the guarantees are basically the same as ES5, where did this idea of "array indices, keys in insert order, symbols in insert order" come from?  Not being difficult, just genuinely curious.

fatcerberus:
Hm, the specification seems to be very ambiguous here.  While the text you quoted ("the mechanics and order are not specified" etc.) suggests it's implementation-defined, the rules for EnumerateObjectProperties() *also* say this:
> EnumerateObjectProperties must obtain the own property keys of the target object by calling its [[OwnPropertyKeys]]

Since the order for this operation is unspecified, the most natural thing for an implementation to do is to return properties in the same order as [[OwnPropertyKeys]] -- for which the order is well-defined.  Reordering them here, e.g. by sorting them, would be doing extra work for no reason.

svaarala:
Optimized the code a little bit using a shared helper. Also added a simple check: if entry part has not array-index keys there's no need to ES6 sort because the keys are already in correct order (for now, because there are no symbols). This avoids an explicit sort in all common cases but falls back to sorting when the keys indicate it is actually needed.

svaarala:
> So if the guarantees are basically the same as ES5, where did this idea of "array indices, keys in insert order, symbols in insert order" come from? Not being difficult, just genuinely curious.

Several people (you included) told me ES6 requires that bevhavior so I took it at face value :-)

svaarala:
> Since the order for this operation is unspecified, the most natural thing for an implementation to do is to return properties in the same order as [[OwnPropertyKeys]] -- for which the order is well-defined. Reordering them here, e.g. by sorting them, would be doing extra work for no reason.

But that's not how implementations work in practice. There may not be a concrete `[[OwnPropertyKeys]]` function inside the engine, that's just a specification concept. For example if you have hidden classes, the actual implementation will be quite different. It's also the case for Duktape, the enumeration code isn't calling some explicit helper but is monolithic.

But even if you did follow that order, the `[[Enumerate]]` section doesn't say how keys from different inheritance levels should mix. You could argue it would be natural to just append them to the result (checking for duplicates) but there are many other viable choices.

Also, it's not guaranteed an implementation will create the sequence beforehand: it can also be done at run time if the engine internals permit it.

In any case, what matters is the guarantees made, and unfortunately the guarantees for `for-in` and `Object.keys()` re: ordering are basically non-existent.

svaarala:
> > So if the guarantees are basically the same as ES5, where did this idea of "array indices, keys in insert order, symbols in insert order" come from? Not being difficult, just genuinely curious.
> Several people (you included) told me ES6 requires that bevhavior so I took it at face value :-)

Hmm, in case you meant whether that order exists in the specification anywhere - it does, it is required by the ES6 `GetOwnPropertyKeys()` internal algorithm. But that is only used for `Object.getOwnPropertyNames()` etc, not for `for-in` and `Object.keys()`.

I updated the discussion here: https://github.com/svaarala/duktape/blob/d3caed902c0be193822728e0c6a44688da0eb13a/doc/hobject-enumeration.rst.

fatcerberus:
Oh, trust me, I know specification devices and actual implementations are different things - Ecmascript calls them "abstract operations" for a reason :)

I was just pointing out that the way the rules are worded comes off as ambiguous to me (and seemingly everyone else, from the looks of things).  It starts out saying the order is unspecified but then the way the rules are worded suggests otherwise, at least to me :)

svaarala:
I think the reference to `[[OwnPropertyKeys]]` just means that the enumeration result should contain the same keys (as a set) as `[[OwnPropertyKeys]]`. In other words, for consistency you shouldn't leave out keys or introduce keys not part of `[[OwnPropertyKeys]]` result which might not otherwise be obvious.

However, since the order is explicitly not guaranteed right at the beginning of the section, I find it hard to read the requirements as specifying that order in the enumeration result.

In any case it's a bit academic because the least surprising behavior is to apply that order to `for-in` and `Object.keys()` too (and mimic V8 for inheritance handling).

So it's just a matter of ensuring Duktape documentation and code comments don't lead yet more people astray w.r.t. what guarantees are ES6 vs. engine specific :-)

fatcerberus:
> as a set

I guess that makes sense, yeah.  There'd be nothing stopping an implementation from getting clever and, e.g. parallelizing the enumeration process, and that would be allowed by the enumeration rules.

svaarala:
Ok, I'm now done and will merge once the tests pass. If anything needs to be fixed after that, I'll use a separate pull.

fatcerberus:
Sounds good, I'm just glad to know `Reflect.ownKeys()` order is guaranteed by the spec since I rely on it for my binary file module. :)

PR #1056, opened 2016-11-04
Add precomputed arridx to duk_hstring

svaarala:
Add back a precomputed `arridx` field to `duk_hstring`. This may be preferable if flash is critical but RAM isn't, or when (small) performance improvements matter. The arridx is computed during interning in any case to properly set the `DUK_HSTRING_HAS_ARRIDX()` flag. With a precomputed arridx RAM usage is higher but footprint for call sites is smaller.

- [x] Add config option, default true (low memory default false, because RAM is usually more restricting)
- [x] Releases entry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,11 +73,11 @@ CONFIGOPTS_NONDEBUG_SCANBUILD=--option-file util/makeduk_base.yaml --option-file
 CONFIGOPTS_NONDEBUG_PERF=--option-file config/examples/performance_sensitive.yaml
 CONFIGOPTS_NONDEBUG_SIZE=--option-file config/examples/low_memory.yaml
 CONFIGOPTS_NONDEBUG_AJDUK=--option-file util/makeduk_base.yaml --option-file util/makeduk_ajduk.yaml --fixup-file util/makeduk_ajduk_fixup.h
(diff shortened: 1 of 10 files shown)


fatcerberus:
I looked at the diff but I still can't figure out what this pull is about.

svaarala:
It's about how to handle the "duk_hstring to array index" conversion.

In current master that conversion is done once during string intern so that we know whether to set/clear the `DUK_HSTRING_HAS_ARRIDX` flag. That flag is needed so that we can quickly check for array index special handling in various internal places. However, the conversion result itself (the array index) is discarded and if it is needed after interning, it is re-computed on-the-fly. This happens relatively rarely, e.g. when you do `obj['123']`, but it's important to be able to do the array-index-or-not check itself quickly.

This pull changes the default behavior so that when the intern time check is done, the resulting array index (or "no array index" indicator) is stored into `duk_hstring`. The runtime lookup is then just a field lookup which simplifies the lookup sites and reduces footprint. The cost is 4 bytes more RAM usage per string, which is not appropriate in all environments but probably a good choice by default.

fatcerberus:
Ah, okay, it's a "is this a canonical array index or not?" flag, since in Ecmascript all keys are strings (but integer strings are special anyway).

I was confused because I thought it had to do with indexing the string itself, e.g. `str[123]` but that didn't make sense because it's only a single integer being cached.

Issue #274, opened 2015-08-10
Feature request: Support ES6 Extended Literals


kphillisjr:
This is not a lot of features, but this is general features revolving around strings. (This and more examples can be found at: http://es6-features.org/ )
- [ ] Unicode String & RegExp Literal
 ```JavaScript
// New Code:
"𠮷".length === 2;
"𠮷".match(/./u)[0].length === 2;
"\u{20BB7}" === "𠮷" === "\uD842\uDFB7";
"𠮷".codePointAt(0) == 0x20BB7;
for (let codepoint of "𠮷") console.log(codepoint);

// Old Code
"𠮷".length === 2;
"𠮷".match(/(?:[\0-\t\x0B\f\x0E-\u2027\u202A-\uD7FF\uE000-\uFFFF][\uD800-\uDBFF][\uDC00-\uDFFF][\uD800-\uDBFF](?![\uDC00-\uDFFF])(?:[^\uD800-\uDBFF]^)[\uDC00-\uDFFF])/)[0].length === 2;
"𠮷" === "\uD842\uDFB7";
//  no equivalent in ES5
//  no equivalent in ES5
```

- [ ] Binary and Octal Literals.
 ```JavaScript
// New Code:
0b111110111 === 503;
0o767 === 503;

// old Code:
parseInt("111110111", 2) === 503;
parseInt("767", 8) === 503;
0767 === 503; // only in non-strict, backward compatibility mode
```

svaarala:
Regarding Unicode literals, the following work in master:

```
duk> "𠮷".length
= 2
duk> "\u{20BB7}" === "𠮷"
= true
duk> "𠮷" === "\uD842\uDFB7"
= true
duk> "𠮷".codePointAt(0) == 0x20BB7;
= true
```

What doesn't yet work is:

```
// No RegExp /u mode yet
"𠮷".match(/./u)[0].length === 2;

// No for-of iteration yet
for (let codepoint of "𠮷") console.log(codepoint);
```

PR #1057, opened 2016-11-04
Add ES6 octal and binary literals, more lenient decimal literal (e.g. "09" = 9)

svaarala:
- [x] Allow octal-like decimal literals like `09` which are still parsed as decimal (09 -> 9) when they cannot be parsed as octal
- [x] Add ES6 `0oNNN` octal support
- [x] Add ES6 `0bNNN` binary support
- [x] Testcase coverage for plain octal legacy literal `077`
- [x] Testcase coverage for plain octal-lookalike decimal literal `078`
- [x] Testcase coverage for ES6 octal literal `0oNNN` (this is not Annex B so should always be enabled, even in strict mode)
- [x] Testcase coverage for ES6 binary literal `0bNNN` (this is not Annex B so should always be enabled, even in strict mode)
- [x] Testcase coverage for string octal literals
- [x] Testcase coverage for parseInt() octal-lookalike values (parsed purely as decimal)
- [x] Testcase fixes
- [x] Website ES6 feature note
- [x] Documentation
- [x] 2.0 migration notes
- [x] Releases entry

Implements part of #274.


svaarala:
`parseInt('0<xx>')` already parses everything as decimal and matches V8.

fatcerberus:
Ah, so nothing to fix there.  I wonder if that's the standard behavior now.

svaarala:
IIRC it was one of the clear cases.

All of this octal stuff is ES6 Annex B BTW - i.e. not recommended to implement unless you must, and never recommended to be used in new code ... :-)

svaarala:
String literal octals have a difference to V8 now.  V8:

```
> '\099bar'
'\u000099bar'
> '\088bar'
'\u000088bar'
```

So, the `\099` gets decoded as `\0` and 9 is treated as garbage, becoming literal characters. ES5 says `\0` with a digit lookahead is a SyntaxError () which Duktape matches now:

```
duk> '\078bar'
SyntaxError: invalid escape (line 1)
    at [anon] (input:1) internal
    at [anon] (duk_lexer.c:904) internal
```

Quickly checking it seems the lookahead condition is no longer any digit but "non-octal digit" so that "9" can follow an octal sequence directly.

This too is Annex B stuff, like everything else related to octal support.

fatcerberus:
Even the `0oNNN` literal notation is Annex B?  One would think at least that would be part of the main specification since the semantics are unambiguous.

svaarala:
Ah, that's correct, the `0oXXX` literal is standard. I'll need to ensure that is outside `DUK_USE_OCTAL_SUPPORT`.

(However, Annex B contains many legacy features which are unambiguous, for example the added RegExp variants, it's not a mandatory criterion as far as I can tell :-)

fatcerberus:
`09` = 9 is Annex B and thus normal behavior.  This is the relevant grammar production:

```
NonOctalDecimalIntegerLiteral ::
0 NonOctalDigit
LegacyOctalLikeDecimalIntegerLiteral NonOctalDigit
NonOctalDecimalIntegerLiteral DecimalDigit
```

svaarala:
Yeah, I found the same production so it seems supported by Annex B (which I would call legacy behavior rather than normal behavior ;)

fatcerberus:
Note also that the lenient Annex B semantics are not allowed in strict mode.

svaarala:
Yep, that's been the case in master already in 1.x:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-869-g660a49a-dirty)
duk> function foo () { 'use strict'; var dummy = 077; }
SyntaxError: invalid number literal (line 1)
    at [anon] (input:1) internal
    at [anon] (duk_lexer.c:1683) internal
duk> function foo () {  var dummy = 077; }
= undefined
```

fatcerberus:
Would it be possible while you're at it to add support for ES6 binary literals, e.g. `0b1111` = 15.  Maybe in a separate pull though.

svaarala:
Sure, I can do that. Probably separate pull though.

svaarala:
Hmm, actually that binary literal is easy to add so I'll include it here after all.

fatcerberus:
Really I've often wished C had a binary notation like that, kind of funny that JavaScript got it first. :-p

svaarala:
The legacy octal behavior corner cases are fun ... :)

```
> 07778.123
7778.123
> 07777.123
07777.123
     ^^^^
SyntaxError: Unexpected number
```

If the '8' in the literal triggers decimal parsing which also allows fractions, why doesn't a period trigger decimal parsing even if the integer part doesn't contain 8 or 9?

svaarala:
Also:

```
> 0777.toString
[Function: toString]
```

But...

```
> 0778.toString
0778.toString
^^^^^
SyntaxError: Invalid or unexpected token
```

Duktape matches this:

```
duk> 0777.toString
= {_func:true}
duk> 0778.toString
SyntaxError: invalid number literal (line 1)
    at [anon] (input:1) internal
    at [anon] (duk_lexer.c:1719) internal
```

fatcerberus:
Seems dumb, but I guess that's why it's Annex B.  JavaScript legacy behavior rarely makes sense.  For example `typeof null` == "object" :/

svaarala:
Hmm, I can't find support for this V8 behavior in ES6:

```
> (function () { "use strict"; return "\9"; })()
'9'
> (function () { "use strict"; return "\7"; })()
(function () { "use strict"; return "\7"; })()
                                      
SyntaxError: Octal literals are not allowed in strict mode.
```

It seems to me that the NonEscapeCharacter escape doesn't allow EscapeCharacters which includes all digits (\0 is handled by a separate rule). The Annex B legacy octal rules don't apply to \8 or \9. Further, V8 and Spidermonkey both accept \8 and \9 in strings even in strict mode. The situation seems the same in ES7.

But I think I'll add this behavior because it's dominant. Maybe I parsed ES6/ES7 incorrectly...

svaarala:
This is now complete but I'll re-read the code before merging. The pull removes the config option for disabling octal support because it became awkward to maintain and the code footprint impact is small. Overall this pull reduces x64 footprint by about 100 bytes.

fatcerberus:
As an interesting aside, I don't know whether you already knew this, but it seems not to be necessary to wrap strict mode code in an IIFE for the Node REPL - it's enough just to start the line with a strict mode directive to get the desired behavior:

```
> 'use strict'; "\7"
'use strict'; "\7"

SyntaxError: Octal literals are not allowed in strict mode.

> "\7"
'\u0007'
```

Very convenient.

svaarala:
Hadn't thought of that for some reason, thanks :)

It works with Duktape too because the command line entry is eval()'d:

```
duk> 'use strict'; "\7"
SyntaxError: invalid escape (line 1)
    at [anon] (input:1) internal
    at [anon] (duk_lexer.c:914) internal
duk> "\7"
= "\x07"
```

fatcerberus:
By the way `parseInt()` not supporting octal is actually ES5.1, not ES6 like I assumed.  From ES5.1 Annex E:
> 15.1.2.2: The specification of the function parseInt no longer allows implementations to treat Strings beginning with a 0 character as octal values.

My recollection of `parseInt()` doing octal parsing is likely from using Sphere 1.5 (minisphere's predecessor) for so long, as that used a really old SpiderMonkey version with only ES3 support.

svaarala:
> By the way parseInt() not supporting octal is actually ES5.1, [...]

Yep, that's why it was already correct in master and covered here: https://github.com/svaarala/duktape/blob/master/doc/number-conversion.rst.

svaarala:
When I wrote that parseInt() code V8 actually did parse octal, here's a not from number-conversion.rst:

> [3b]: Leading zeroes may trigger automatical octal mode in some implementations. E.g. in V8, parseInt("0009") returns 0 because V8 switches to octal mode, and treats '9' as garbage; parseInt("0009", 10) returns the correct value 9.

PR #1058, opened 2016-11-05
Use duk_known_{hstring,hbuffer,hobject}() helpers

svaarala:
These can be used whenever we're 100% certain that the value stack index exists and the type matches expected type.  When these are true, a duk_hstring, duk_hbuffer, or duk_hobject pointer fetch can be inlined to small code.

- [x] Internal helpers and call site fixes
- [x] Releases entry


svaarala:
Around -150 for x86 stripsize and -300 bytes for x86 minsize.

PR #1062, opened 2016-11-07
Support ARM64 platform detection on Android.

beevik:
Use the Android __aarch64__ #define as a valid ARM64 signifier when detecting the compiler platform.

Rationale: the library currently fails to configure correctly for some Android ARM64 platforms.  See [here](http://www.phoronix.com/scan.php?page=news_item&px=MTY5ODk) for a longer explanation.


svaarala:
Thanks!

beevik:
If you'd like me to wordwrap the #ifdefs to be more like the surrounding code, please just let me know.

svaarala:
That's not necessary, it's not overly long and quite readable as is.

PR #1064, opened 2016-11-10
Implement Object.assign()

fatcerberus:
This is a fully functional implementation of ES6 `Object.assign()`.

The changes in #1054 turned out to be convenient here.  Since `duk_enum()` respects ES6 [[OwnPropertyKeys]] ordering I was able to use it to implement `Object.assign()` without breaking compliance.

### DIFF
--- a/tests/ecmascript/test-bi-object-assign.js
+++ b/tests/ecmascript/test-bi-object-assign.js
@@ -1,0 +1,207 @@
+/*
+ *  Object.assign() (E7 19.1.2.1)
+ *  http://www.ecma-international.org/ecma-262/7.0/#sec-object.assign
+ */
+
+if (typeof print === 'undefined') {
+    print = function() {
+        console.log([].map.call(arguments, function (v) { return String(v); }).join(' '));
+    };
+}
+
+/*===
+basic tests
+true
+pig
+cow
+ape
+0
+1
+2
+badger
+whale
+maggie Kittycow Machel Fatty Whale 812 1208 9001
+0,1,2,pig,cow,ape,badger,whale
+maggie 1208 9001 Vegeta
+cow,pig,ape,saiyan
+TypeError
+foo read-only unchanged undefined
+foo,bar,baz
+undefined
+this belongs here
+undefined
+prop1
+get
+set 812
+===*/
+
+function basicTest() {
+    var obj, src, dest;
+
+    // If no coercion is needed, target object is returned (i.e. not a copy).
+    obj = {};
+    dest = Object.assign(obj, { test: "test" });
+    print(dest === obj);
+
+    // Properties are assigned in [[OwnPropertyKeys]] order, moving from left
+    // to right in the argument list:
+    //     1. Array indices (sequential, ascending)
+    //     2. String keys (creation order)
+    //     3. Symbol keys (creation order)
+    // The only reliable way to verify this is by forcing Object.assign() to
+    // call getters and watch the side effects.
+    var obj1 = {}, obj2 = {}, arr = [];
+    Object.defineProperties(obj1, {
+        pig: { enumerable: true, get: function() { print("pig"); return "maggie"; } },
+        cow: { enumerable: true, get: function() { print("cow"); return "Kittycow"; } },
+        ape: { enumerable: true, get: function() { print("ape"); return "Machel"; } }
+    });
+    Object.defineProperties(arr, {  // note: This abandons the array part.
+        badger: { enumerable: true, get: function() { print("badger"); return "pig food"; } },
+        0: { enumerable: true, get: function() { print("0"); return 812; } },
+        2: { enumerable: true, get: function() { print("2"); return 9001; } },
+        1: { enumerable: true, get: function() { print("1"); return 1208; } }
+    });
+    Object.defineProperties(obj2, {
+        whale: { enumerable: true, get: function() { print("whale"); return "Fatty Whale"; } }
+    });
+    obj = Object.assign({}, obj1, arr, obj2);
+    print(obj.pig, obj.cow, obj.ape, obj.whale, obj[0], obj[1], obj[2]);
+    print(Reflect.ownKeys(obj));
+
+    // Existing properties in the target object can be overwritten.
+    obj = Object.assign({ cow: 1208, pig: 812, ape: 9001 },
+        { pig: "maggie", saiyan: "Vegeta" });
+    print(obj.pig, obj.cow, obj.ape, obj.saiyan);
+    print(Reflect.ownKeys(obj));
+
+    // If a property write fails, an error is thrown (strict mode semantics)
+    // and the object is left partially updated:
+    obj = Object.create(Object.prototype, {
+        foo: { writable: true, value: "unchanged" },
+        bar: { writable: false, value: "read-only" },
+        baz: { writable: true, value: "unchanged" },
+    });
+    try {
+        Object.assign(obj, { foo: "foo", bar: "bar", baz: "baz", quux: "quux" });
+        print("never here");
+    } catch (e) {
+        print(e.name);
+        print(obj.foo, obj.bar, obj.baz, obj.quux);
+        print(Reflect.ownKeys(obj));
+    }
+
+    // Only enumerable own properties of the source(s) are used.
+    src = Object.create({ inherited: "something stupid happened" }, {
+        prop1: { enumerable: true,  value: "this belongs here" },
+        prop2: { enumerable: false, value: "something stupid happened" }
+    });
+    obj = Object.assign({}, src);
+    print(obj.inherited);
+    print(obj.prop1);
+    print(obj.prop2);
+    print(Reflect.ownKeys(obj));
+
+    // Accessors in both target and source object(s) are called.
+    obj = Object.create(Object.prototype, {
+        prop: { set: function(value) { print("set", value); }}
+    });
+    src = Object.create(Object.prototype, {
+        prop: { enumerable: true, get: function() { print("get"); return 812; }}
+    });
+    Object.assign(obj, src);
+}
+
+try {
+    print("basic tests");
+    basicTest();
+} catch (e) {
+    print(e.stack || e);
+}
+
+/*===
+argument policy
+object 812
+*MUNCH*
+object cows eat kitties
+moooo *munch*
+TypeError
+TypeError
+foo,bar
+baz quux
+===*/
+
+function argumentTest() {
+    var obj;
+
+    // The first argument to Object.assign() is ToObject() coerced.  The
+    // coerced value is what's returned by the function.
+    obj = Object.assign(812, { test: "*MUNCH*" });
+    print(typeof obj, obj);
+    print(obj.test);
+
+    // XXX: ToObject() coercion is specified for all arguments, is there a way
+    //      to test for this?
+
+    // Coercion is performed even if no sources are provided:
+    obj = Object.assign("cows eat kitties", { test: "moooo *munch*" });
+    print(typeof obj, obj);
+    print(obj.test);
+
+    // null and undefined are not object coercible!
+    try {
+        Object.assign(null);
+        print("never here");
+    } catch (e) {
+        print(e.name);
+    }
+    try {
+        Object.assign(undefined);
+        print("never here");
+    } catch (e) {
+        print(e.name);
+    }
+
+    // However, both null and undefined are accepted as sources.  They are
+    // treated as though they were empty objects.
+    obj = Object.assign({}, undefined, { foo: "baz" }, null, { bar: "quux" });
+    print(Reflect.ownKeys(obj));
+    print(obj.foo, obj.bar);
+}
+
+try {
+    print("argument policy");
+    argumentTest();
+} catch (e) {
+    print(e.stack || e);
+}
+
+/*===
+virtual properties
+0 255 44
+0 undefined
+0,1,2
+0 255 44
+===*/
+
+function virtualPropTest() {
+    var u8 = new Uint8Array(3);
+    var arr = [ 9001 ], obj;
+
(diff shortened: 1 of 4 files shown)


svaarala:
Good stuff :) This should work fine as a first implementation. Later on the algorithm should probably be optimized a bit via the internal representation, but I'm not yet sure of the best approach for that: every property may be duplicate and may invoke a side effect in both the source and the target, so there are few shortcuts that work without issues.

It'd be good if the testcases had coverage for plain objects, plain arrays, and some objects with a lot of virtual properties like ArrayBuffer. Setter side effects are already covered by the tests which is nice.

Related issue (and separate pull): `duk_assign()` or `duk_object_assign()` C API?

svaarala:
That assert test still fails due to test timeout on:

```
              test-dev-regexp-negative-jump-offset: fail; 9 diff lines
```

I think I'll reduce the coverage of that test a bit to make it run faster and avoid timeouts.

fatcerberus:
Yeah, I don't think it's actually okay to skip duplicates: ES6/7 explicitly describes the algorithm as processing each property of the source(s) in turn, assigning to the target object each time--which may invoke getters/setters, Proxy traps, etc.  So I don't know how much more it can really be optimized. :(

svaarala:
Right, that's what I meant that all properties technically need to be processed. And since side effects may be involved, it's not safe to e.g. walk the source property table because it may be arbitrarily modified by any setter/getter call.

But it *is* possible to start the process with the optimistic assumption that no side effects will be involved, and when you encounter any (setter or getter), abandon the process and start over with the current implementation. It should be OK to leave the already copied values in place: they'd just be re-copied and they can't invoke side effects because we only processed them up until a side effect was detected.

This is obviously slower for source/target objects with setters and getters, but they're not the common case.

Anyway, future work :-)

svaarala:
A possible optimized approach is actually somewhat similar to the JSON fast path: it walks internal structures without side effects but when it encounters any problematic properties or objects, it bails out and falls back to the generic algorithm. Being very narrow and tailored for the "common case only" the fast path can avoid a lot of unnecessary checks.

The nice thing about this approach is that the fast path is intentionally very narrow and can always fall back to the slow path. The fast path can also be disabled from low memory builds because it's always safe to use the fallback which must always be present anyway for compliance.

fatcerberus:
That fast path could also be skipped entirely (preventing duplicate effort) if the target is a Proxy.

svaarala:
Right, that'd be a quick reject for the fast path. Also if an object kept track of "I have at least one setter/getter" that would be a useful fast path reject. No such flag exists now though, and I'm not sure if adding one would be worth it (it might have use in other contexts too though).

fatcerberus:
Any suggestions for how the current test script could be expanded?  It's somewhat basic right now and only covers the common cases.

svaarala:
Maybe cover a small mix of objects with virtual properties? E.g.
- Assign buffer indices from a source object
- Assign Array `.length` from a source object
- Assign virtual properties from a `Uint8Array` to a plain target object

fatcerberus:
Hm, adding `print = console.log;` breaks the test because arrays are coerced differently.

fatcerberus:
Is ToObject() coercion observable from Ecmascript?  I'd like to test that argument coercion behavior but I couldn't figure out a way to do it.

fatcerberus:
One possible compliance issue is for Step 4c(ii):
```
c. Repeat for each element nextKey of keys in List order,
    i. Let desc be ? from.[[GetOwnProperty]](nextKey).
    ii. If desc is not undefined and desc.[[Enumerable]] is true, then
        1. Let propValue be ? Get(from, nextKey).
        2. Perform ? Set(to, nextKey, propValue, true).
```

Particularly "if *desc* is not `undefined`", or in plain English "if the property still exists".  In this step *desc* may be `undefined` if an earlier side effect deleted it.  I've assumed `duk_enum()` covers this case too, but I'm not 100% sure.

svaarala:
If you want to provide print for Node.js, this should do the trick:

```
if (typeof print === 'undefined') {
    print = function () {
        console.log(Array.prototype.map.call(arguments, function (v) { return String(v); }).join(' '));
    };
}
```

Now that console.log is also provided by the `duk` harness by default, I could maybe write a pull which removes all "print" bindings and replaced them with "console.log". That'd make it easier to do cross engine comparisons.

svaarala:
> Is ToObject() coercion observable from Ecmascript? I'd like to test that argument coercion behavior but I couldn't figure out a way to do it.

I don't think it is, unfortunately.

> Particularly "if desc is not undefined", or in plain English "if the property still exists". In this step desc may be undefined if an earlier side effect deleted it. I've assumed duk_enum() covers this case too, but I'm not 100% sure.

It's covered yes: https://github.com/svaarala/duktape/blob/master/src-input/duk_hobject_enum.c#L687-L691.

fatcerberus:
Yeah, that's what I thought re: ToObject coercion.  In any case it is tested for indirectly: There's a test for `Object.assign(obj) === obj`, and later on there are also tests for, e.g., `Object.assign(812) instanceof Number`.  So not perfect coverage, but between the two tests, it's as close as we can get.

fatcerberus:
`duk_enum()` behavior seems completely consistent with the spec description of `Object.assign()` (i.e. [[OwnPropertyKeys]] is called in advance and iterated upon):

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-901-gffacb9e)
duk> var src = { pig: "maggie" };
= undefined
duk> var dest = { set pig(value) { print(value); src.cow = "Kittycow"; } };
= undefined
duk> Object.assign(dest, src)
maggie
= {pig:undefined}
duk> 
```

I was worried that properties added during enumeration would show up later, thus breaking compliance.  Luckily this isn't the case :)

svaarala:
Yeah, that works out well because the enumeration is snapshot based. The `[[Enumerate]]` semantics don't allow on-the-fly deleted properties to be enumerated but *does* allow on-the-fly added properties to be enumerated (though that's not required and Duktape doesn't do it).

fatcerberus:
So yeah, I'm done with this pull, feel free to review whenever it's convenient.

tests/ecmascript/test-bi-object-assign.js
```diff
@@ -0,0 +1,147 @@
...
+    dest = Object.assign(obj, { test: "test" });
+    print(dest === obj);
+
+    // Properties are assigned in [[OwnPropertyKeys]] order, moving from left
+    // to right in the argument list:
+    //     1. Array indices (sequential, ascending)
+    //     2. String keys (creation order)
+    //     3. Symbol keys (creation order)
+    // This behavior is easy to test for by assigning from multiple sources to
+    // an empty object and then calling Reflect.ownKeys() on the result.
```
svaarala: This comment doesn't seem right: if Object.assign() worked in the incorrect order, using Reflect.ownKeys() wouldn't reveal that because *it* would reorder the keys?

So I guess the only way to detect the correct order is to have printing getters with array indices, non-array indices etc, and to see that they're called in the expected order?

tests/ecmascript/test-bi-object-assign.js
```diff
@@ -0,0 +1,147 @@
...
+        [ 812, 1208, 9001 ],
+        { whale: "Fatty Whale" });
+    print(obj.pig, obj.cow, obj.ape, obj.whale, obj[0], obj[1], obj[2]);
+    print(Reflect.ownKeys(obj));
+
+    // Existing properties in the target object can be overwritten.
+    obj = Object.assign({ cow: 1208, pig: 812, ape: 9001 },
+        { pig: "maggie", saiyan: "Vegeta" });
+    print(obj.pig, obj.cow, obj.ape, obj.saiyan);
+    print(Reflect.ownKeys(obj));
```
svaarala: Maybe useful test: if the attempt to overwrite an existing property fails, demonstrate what the result is: error + already written properties updated.

tests/ecmascript/test-bi-object-assign.js
```diff
@@ -0,0 +1,147 @@
...
+true
+TypeError
+TypeError
+===*/
+
+function coercionTest() {
+    // The first argument to Object.assign() is ToObject() coerced.  The
+    // coerced value is what's returned by the function.
+    var obj = Object.assign(812, { test: "something" });
+    print(obj instanceof Number);
```
svaarala: This is enough to check for object-ness but `print(typeof obj)` (expected "object", not "number") might be easier to read :)

tests/ecmascript/test-bi-object-assign.js
fatcerberus: Good idea, thanks :)

tests/ecmascript/test-bi-object-assign.js
svaarala: Right, it's just maybe not intuitive that `1 instanceof Number` is false but `new Number(1) instanceof Number` is true. At least it isn't for me :-)

tests/ecmascript/test-bi-object-assign.js
fatcerberus: I tend to approach unit tests from a black-box perspective, that is, with the assumption that everything other than the unit under test works as designed.

`Reflect.ownKeys()` is specified to list the keys in insertion order, so checking that order is (again, from black-box perspective) a good indicator that `Object.assign()` did the right thing.

That said, even if that correctness assumption holds, array index keys are still sorted to the top always, so this test is flawed after all.  I'll try to come up with a better solution.

svaarala: > That said, even if that correctness assumption holds, array index keys are still sorted to the top always, so this test is flawed after all. I'll try to come up with a better solution.

This is what I meant: at least the test comment is wrong IMO, and the test wouldn't notice if Object.assign() didn't handle the keys in the correct order.

svaarala: So just to be clear:

> Reflect.ownKeys() is specified to list the keys in insertion order, so checking that order is (again, from black-box perspective) a good indicator that Object.assign() did the right thing.

My point here was that for array index inserts Reflect.ownKeys() actually won't list keys in the insertion order. So assuming Reflect.ownKeys() works correctly but Object.assign() handles array indices in an incorrect order, Reflect.ownKeys() (if working correctly) will effectively mask the error.

Reflect.ownKeys() is good for checking that non-array-index keys were handled correctly, agree with that.

fatcerberus: Well, it *would* notice (in theory) if the non-integer keys were done in the wrong order. ;)  But it's still wrong for array indices in any case, so 100% agreed it needs fixing.

svaarala: Sure, agree with that. But handling of array vs. non-array index ordering is quite a likely implementation error so I guess the only working method would be getters with print side effects?

fatcerberus: Not disagreeing with you at all by the way, just wanted to explain my thought process in writing the test.  The array index behavior was an oversight.

fatcerberus: > getters with print side effects

Yeah, that seems the most reliable method to me too.  I'll look at it once I get settled in, just got home from work. :)

tests/ecmascript/test-bi-object-assign.js
fatcerberus: So to be clear, you'd like me to change the test to:
```js
print(typeof obj, obj instanceof Number);
```

Correct?  The instanceof check or something equivalent needs to be there to ensure we got a Number object back and not something else entirely.

tests/ecmascript/test-bi-object-assign.js
svaarala: Unfortunately there's no primitive to enumerate stuff in the *internal order*. Which is good in some sense, but unfortunate for white box (well, more "grey box") testing. So getters is all I can think of right now.

To test likely errors well (which is a bit grey box) I think maybe three cases would be needed:
- Object with a mixed set of array / non-array properties, in an order which requires array index sorting. Getters to notice side effects.
- Array with additional properties. All array index properties should be enumerable (and cannot be getters!) to preserve the internal array part. Getters to notice side effects; but these will only be in place for non-array-index properties.
- Array with some array index properties that are non-enumerable. That causes the array part to be abandoned so the internal structure will be different. Insert the non-enumerable orders in a non-ascending order: they will go into the entry part in that order, but then be reordered for OwnPropertyKeys. Again, getters to notice side effects.

I don't think there's a way to test the array-with-array-part case: using a getter for an array causes the array part to be abandoned.

svaarala: > Yeah, that seems the most reliable method to me too. I'll look at it once I get settled in, just got home from work. :)

No hurry, obviously :) I'm going to sleep anyway, I'll recheck and merge later when I wake up :)

tests/ecmascript/test-bi-object-assign.js
fatcerberus: Nevermind, I realized that `print(new Number(1))` will print "1" (i.e. it coerces back to the original value), so the instanceof check is unnecessary.

tests/ecmascript/test-bi-object-assign.js
fatcerberus: Hm, this comment is still showing as unfixed.  I did make the suggested change, but I guess GitHub doesn't realize it because I didn't modify this particular line.

fatcerberus:
I expanded the testcase, this should be good to go now.

svaarala:
Looks good :)

fatcerberus:
Yep, it's great to have this in master.  I've often wished for a standard way to set multiple properties at once on an existing object, but an `Object.defineProperties()` call site is too verbose for most use cases.  It was quite a pleasant surprise to see this functionality was in ES6, and it seemed easy to implement so I went ahead and did so. :)

PR #1068, opened 2016-11-11
Add ES6 Object.is() support

svaarala:
- [x] Add support for Object.is()
- [x] Add `duk_samevalue()` API binding
- [x] API documentation
- [x] Testcases for Object.is()
- [x] Testcases for duk_samevalue()
- [x] Releases entry

### DIFF
--- a/tests/ecmascript/test-bi-object-is.js
+++ b/tests/ecmascript/test-bi-object-is.js
@@ -1,0 +1,61 @@
+/*
+ *  Object.is()
+ */
+
+/*===
+basic test
+function true false true
+2
+true
+false
+false
+true
+true
+0 0
+1 1
+2 2
+3 3
+4 4
+5 5
+6 6
+7 7
+8 8
+9 9
+10 10
+11 11
+12 12
+===*/
+
+function basicTest() {
+    var pd = Object.getOwnPropertyDescriptor(Object, 'is');
+    print(typeof pd.value, pd.writable, pd.enumerable, pd.configurable);
+    print(Object.is.length);
+
+    // SameValue speciality is to take zero sign into account and to compare
+    // NaNs are true.
+    print(Object.is(-0, -0));
+    print(Object.is(-0, +0));
+    print(Object.is(+0, -0));
+    print(Object.is(+0, +0));
+    print(Object.is(0/0, 0/0));
+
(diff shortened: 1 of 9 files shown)


svaarala:
There are already `duk_equals()` and `duk_strict_equals()` in the public API. As a related issue, maybe add `duk_samevalue()`. SameValue() is usually mostly used internally but if there's an Ecmascript binding for it (Object.is()) then maybe it also belongs in the C API.

fatcerberus:
I was thinking the same thing.

svaarala:
I'll add it in this pull, and remove it if there's a good reason not to add it.

svaarala:
Other comparison operations are also not provided at present, e.g. `duk_gt()`, `duk_ge()`, etc. They would be logical to have with `duk_equals()` etc.

svaarala:
I'll add the testcases later tonight and this should be pretty quickly in master.

fatcerberus:
I'm only surprised that the Ecmascript function is named `Object.is()` rather than `Reflect.is()`.

tests/ecmascript/test-bi-object-is.js
```diff
@@ -0,0 +1,61 @@
...
+8 8
+9 9
+10 10
+11 11
+12 12
+===*/
+
+function basicTest() {
+    var pd = Object.getOwnPropertyDescriptor(Object, 'is');
+    print(typeof pd.value, pd.writable, pd.enumerable, pd.configurable);
```
svaarala: When I wrote the ES5 bindings I wrote a testcase to print the property attributes of all built-ins in a separate test case. That leaves the specific test cases to focus on functionality only. But it's rather annoying to maintain because you need to remember the test case even exists, so I've now tried to include e.g. attribute and `.length` tests in the specific tests themselves.

As a sidenote, luckily test cases don't need to be identical in structure, overlap in tests, or even all that well organized: test cases are not part of the build so they don't have footprint issues, and they can just be deleted/rewritten if something becomes too much of an issue.

What matters is regression coverage and ease of writing new tests to maintain that coverage. For example, when new functionality is written, it doesn't really matter if there's some overlap to existing tests. So I try not to be too critical of test cases beyond their coverage :-)

PR #1069, opened 2016-11-12
Implement support for ES6 Math.hypot()

fatcerberus:
This is an implementation of `Math.hypot()` using an algorithm based on V8 (without sharing code) that avoids overflow and minimizes floating point rounding error.  The algorithm first finds the largest (absolute) value, normalizes all arguments against it, and uses Kahan summation followed by a `sqrt()` to arrive at the result.

All arguments are ToNumber coerced, per ES6 spec, as part of the "find the largest value" step.

Checklist:
- [x] Implement `Math.hypot()` from ES6
- [x] Testcase
- [x] RELEASES entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1841,6 +1841,8 @@ Planned
 * Add support for ES6 String.prototype.codePointAt(), String.fromCodePoint(),
   and String.prototype.repeat() (GH-1043, GH-1049, GH-1050)
 
+* Add support for ES6 Math.hypot() (GH-1069)
+
 * Add support for ES6 Object.assign() (GH-1064)
 
 * Add support for ES6 Object.is() and duk_samevalue() API call (GH-1068)
--- a/src-input/builtins.yaml
+++ b/src-input/builtins.yaml
@@ -2309,6 +2309,14 @@ objects:
           magic:
             type: math_onearg
             funcname: "floor"  # BI_MATH_FLOOR_IDX
+      - key: "hypot"
+        value:
+          type: function
+          native: duk_bi_math_object_hypot
+          length: 2
+          varargs: true
+        es6: true
+        present_if: DUK_USE_ES6
       - key: "log"
         value:
           type: function
--- a/src-input/duk_bi_math.c
+++ b/src-input/duk_bi_math.c
@@ -249,6 +249,74 @@ DUK_INTERNAL duk_ret_t duk_bi_math_object_twoarg_shared(duk_context *ctx) {
 	return 1;
 }
 
+#if defined(DUK_USE_ES6)
+DUK_INTERNAL duk_ret_t duk_bi_math_object_hypot(duk_context *ctx) {
+	/*
+	 *  E6 Section 20.2.2.18: Math.hypot
+	 *
+	 *  - If no arguments are passed, the result is +0.
+	 *  - If any argument is +inf, the result is +inf.
+	 *  - If any argument is -inf, the result is +inf.
+	 *  - If no argument is +inf or -inf, and any argument is NaN, the result is
+	 *    NaN.
+	 *  - If all arguments are either +0 or -0, the result is +0.
+	 */
+
+	duk_idx_t nargs;
+	duk_idx_t i;
+	duk_bool_t found_nan;
+	duk_double_t max;
+	duk_double_t sum, summand;
+	duk_double_t comp, prelim;
+	duk_double_t t;
+
+	nargs = duk_get_top(ctx);
+
+	/* Find the highest value.  Also ToNumber() coerces. */
+	max = 0.0;
+	found_nan = 0;
+	for (i = 0; i < nargs; i++) {
+		t = DUK_FABS(duk_to_number(ctx, i));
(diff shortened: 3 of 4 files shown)


fatcerberus:
AppVeyor x64 build seems to be failing with a link error.  I assume it's a problem with the AppVeyor build script since all the Linux builds (including Travis) are okay and it works for me in MSVC 2015.

```
duktape.obj : error LNK2019: unresolved external symbol fmax referenced in function duk_bi_math_object_hypot
```

Other than that, this should be merge ready.

svaarala:
AppVeoyr builds with multiple MSVC versions and apparently fmin/fmax have been missing in some versions:

- http://stackoverflow.com/questions/19234452/visual-c-express-fmin-and-fmax-identifier-not-found

duk_config.h detects for the situation and defines `DUK_USE_MATH_FMIN/FMAX` which are undefined if C99 isn't supported. The confusing thing is that `DUK_FMIN` and `DUK_FMAX` are still defined.

This is a pretty poor approach: it would be better to simply provide a replacement for it so that internals could just call `DUK_FMAX()`.

I'll fix that in master first.

svaarala:
Ok, the master fix is #1071, I don't have time to finish it right now but I'll do it later today.

fatcerberus:
My pull requests always seem to dig up the weirdest issues :)

svaarala:
That's true :) Most of them have been configuration related and that's to be expected as the need to add ES6 features but maintain an ES5 subset grows.

The current configuration solution now has a decent structure I think (a single `configure.py` with a lot of leeway for doing e.g. code autogeneration based on config) but the specific config options, platform detection, and other internals are still more or less inherited from the previous configuration approach with little changes. So it's definitely to be expected that the configuration stuff will creak a little bit and need incremental improvement going forward.

svaarala:
By the way, does the pull rely on C99 `fmin()` behavior for NaNs, in particular: if one argument is NaN and the other is not, `fmin()` is required to return the non-NaN argument. This differs from `Math.fmin()` which returns NaN in that situation and there's an explicit check for that.

I ask because the `fmin()` replacement does *not* guarantee that behavior so it may matter here too.

svaarala:
I think it shouldn't matter because of the `found_nan` handling.

svaarala:
The testcase could maybe test these explicitly:

```js
Math.hypot(0/0, 1);
Math.hypot(1, 0/0);
Math.hypot(0/0, 0/0);
```

This covers NaN handling for the two-arguments case, and I'd like to add a fast path for that case separately.

By the way, it's great you went the extra mile to use a proper summation algorithm, so there's no backlog to improve it later :)

fatcerberus:
`Math.hypot()` is required to return NaN if any argument is NaN, _unless_ any argument is +/- ∞.  As such `fmax()` is never called for a NaN, because that risks losing the infinity, in case one is present.

So no, I don't think it matters here.

fatcerberus:
I updated the tests to cover the two-argument case more comprehensively as well as the argument coercion behavior.  I guess this is just waiting on #1071 now.

svaarala:
Ok, ready to be rebased.

fatcerberus:
Alright, done.

svaarala:
BTW, that NaN/inf behavior is pretty interesting, compared to Math.max:

```
> Math.max(0/0, 1/0)
NaN
```

Because Math.hypo(0/0, 1/0) is Inf, why wouldn't Math.max() also be? Mysterious :)

fatcerberus:
Yeah, I thought that was odd behavior too.

svaarala:
Generally operations are "NaN tainting" I think. Then again, C99 fmin is required to return the non-NaN argument for `fmin(NAN, 123.0)` and `fmin(123.0, NAN)`, so I guess both behaviors are somewhat normal.

fatcerberus:
Yeah, I know "NaN tainting" is the norm.  This is an odd case though: If any argument is NaN the result is also NaN, unless one or more is +/-inf.  The special casing for infinity is what seems weird to me.

fatcerberus:
In any case this is all academic because we have to conform to the specification.  Just interesting is all :)

svaarala:
> Yeah, I know "NaN tainting" is the norm.

Interestingly `Math.min()` and `Math.max()` are "NaN tainting" but the somewhat analogous C99 `fmin()` and `fmax()` are not (I would expect them to be - but they're not).

> In any case this is all academic because we have to conform to the specification. Just interesting is all :)

Indeed :)

svaarala:
The footprint of the algorithm is around 350 bytes so this is rather large for a single binding. I'll check separately if the proper summation behavior can be kept but with a smaller footprint.

svaarala:
There are maybe a few testcase coverage issues I can address at the same time: e.g. coercion side effects and demonstrating that all arguments are coerced even if there's a special case exit like Inf.

svaarala:
Some coverage notes:
- Coercion side effects in order, especially if there's a special case like +Inf (one might be tempted to stop coercions there).
- For zero sign handling, test all combinations of zero sign for at least two arguments. Again easy to get wrong.
- Function property attributes and .length.

PR #1071, opened 2016-11-12
Ensure DUK_FMIN() and DUK_FMAX() are always available

svaarala:
`fmin()` and `fmax()` are not available in all environments so replacements are needed. However, previous approach required calling code inside Duktape to `#if defined(DUK_USE_MATH_FMIN)` before using `DUK_FMIN()` and also to provide an inline replacement.

This pull simplifies the handling so that internal code can always call `DUK_FMIN()` and it is either provided by the platform or a built-in replacement.

- [x] Change config handling and call sites
- [x] Check MSVC build
- [x] 2.0 migration notes
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2025,6 +2025,9 @@ Planned
 
 * Fix incorrect duk_hbufferobject size in Duktape.info() (GH-804)
 
+* Miscellaneous portability improvements: simplified DUK_FMIN() and DUK_FMAX()
+  handling (GH-1071)
+
 * Internal performance improvement: rework bytecode format to use an 8-bit
   opcode field (and 8-bit A, B, and C fields) to speed up opcode dispatch
   by around 20-25% and avoid a two-level dispatch for EXTRA opcodes; the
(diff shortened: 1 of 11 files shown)


svaarala:
There are actually only two current call sites to `DUK_FMIN()` and `DUK_FMAX()` (and maybe one more with `Math.hypot`). All of the call sites have specific NaN handling so they can cope with a simple `(x < y ? x : y)` solution.

So I may open a follow-up later on to remove the macros altogether: the generated code is smaller and better without a call to a library function, and there's one less worry for porting. Particularly because a replacement is already needed for non-C99 environments.

But let's merge this and #1069 first.

fatcerberus:
Most compilers are probably smart enough to use an intrinsic for fmin/max, but even so I agree it's probably better not to have a library dependency if it can be avoided.

svaarala:
They actually don't :-) At least GCC doesnt:

```
[...]
  404293:       c3                      retq   
  404294:       e9 67 da ff ff          jmpq   401d00 <fmin@plt>
  404299:       c3                      retq   
```

svaarala:
I think part of the reason might be that fmin() and fmax() have those specific NaN handling requirements which may not map easily to a single assembly instruction (I'd have to check, it's been a long time since I actually wrote assembly code manually...). And the compiler may not be clever enough to notice NaN has been checked for before, because that happens using `fpclassify()` or similar which isn't necessarily easy to analyze as reliably checking for NaN at compile time.

PR #1073, opened 2016-11-12
Math.hypot() testcase improvements

svaarala:
Math.hypot() testcase improvements, split Math testcases into separate files.


svaarala:
Two coverage improvements:
- Math.hypot() with a large number of arguments.
- Some argument sequence to demonstrate that the Kahan summation bit works correctly. There's no single right answer but at least the current correct answer should be part of the test coverage so it can be seen if it'd accidentally broken.

fatcerberus:
V8's unit test does some crazy fractal calculations to test the function's accuracy:
https://github.com/v8/v8/blob/master/test/mjsunit/es6/math-hypot.js#L68-L92

At some point it might be good to split up the Math built-in tests like I did when I wrote the tests for Reflect.  It's pretty monolithic now and easy to forget to add something.

svaarala:
Yeah, I added a specific Math.hypot() test in the pull already and I could move the tests there. Splitting up the existing tests into separate testcases might also be good to make the pattern consistent. The math testcase needed those print helpers (for negative zeros and such) but now that there's an include mechanism it's no longer a big issue.

svaarala:
Ok, math tests are now split into separate testcases. That's actually the case for almost all the newer tests; Math tests are very old and during initial development monolithic tests were easier to write (for example there wasn't an include mechanism for test utils). Some other very old tests are monolithic for similar reasons.

In general, most tests are also already organized by binding, for example `test-bi-object-keys.js`, `test-bi-arraybuffer-proto-slice.js`, etc. There are exceptions when the feature under test is not a single binding but rather cuts across a lot of functionality. For example, for fastints there is a single "topic" testcase which demonstrates fastint behavior across various bindings etc. There's necessarily some overlap in that, but if fastint tests were peppered across other testcases (organized mainly by binding) it would be difficult to get a general idea how they are supposed to behave.

So there isn't really a single underlying principle clearly deciding where a certain thing needs to be tested, but competing concerns sometimes creates overlap which is fine.

svaarala:
On the topic of testcases, one of the options forward is to move into an assert based testcase format (`assertEquals(Object.is(-0, +0), false)` etc). That's a more common solution with the big upside that the test and the expected result are on the same line so they're easier to move around etc.

One downside that the tests don't normally print anything so you don't learn much by running them manually. Another is that if an assert fails, you typically only get an error for that assert line; even if you just log and continue, the error may be in the argument expression which causes an uncaught error. This is mostly fine in incremental development when the assumption is that maybe one or two tests fails. The approach also works best when there are a lot of small tests.

Finally, when doing really low level testing like testing the lexer and such, it's awkward to assume that a whole unit test framework runs. That's actually the main reason I started out with print() testing: as long as print() works, you can write out a testcase with very minimal bindings elsewhere. As time goes on, this feature becomes less and less important of course and is not really important for Duktape testing anymore (except maybe when rewriting the lexer and/or compiler).

All in all, moving to assert based testing would be workable by splitting some of the larger tests into separate tests first. With that done it might be quite feasible to use some existing test framework to drive the tests.

svaarala:
Moved the code changes into a separate pull, as the testcase reorganization is significant enough to be done in a separate pull.

fatcerberus:
I do have a tendency to manually run tests to see the output whenever I see that the test suite failed.  Sometimes this illuminates the nature of the bug before even looking at the code.

Assert-based testing necessarily loses some information - you then only know that a given test failed, and not the exact nature of the failure.  With that in mind, I'm naturally hesitant to suggest moving in that direction.

svaarala:
The other possible direction would be to make it easier to write expect strings. For example, maybe adding an inline syntax for it:

```
print(Object.is(-0, +0));  //> false
```

svaarala:
And also related, removing the whole `/*@include ... @*/` syntax and allowing testcases to just assume that a global testing module exists in a convenient binding such as:

```
Test.printExactNumber(-0);  //> -0
```

PR #1074, opened 2016-11-13
Math.hypot() footprint optimization

svaarala:


svaarala:
Not really effective, closing as unmerged.

PR #1077, opened 2016-11-13
Faster duk_get_type() and duk_get_type_mask()

svaarala:
For unpacked duk_tval, do a straight lookup with no tag type range check.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2048,7 +2048,8 @@ Planned
   opcode handler optimizations (GH-903); refcount optimizations (GH-443,
   GH-973, GH-1042); minor RegExp compile/execute optimizations (GH-974,
   GH-1033); minor IEEE double handling optimizations (GH-1051); precomputed
-  duk_hstring array index (GH-1056)
+  duk_hstring array index (GH-1056); duk_get_{type,type_mask}() optimization
+  (GH-1077)
 
 * Miscellaneous footprint improvements: RegExp compiler/executor (GH-977);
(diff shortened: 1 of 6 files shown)


svaarala:
The resulting code for e.g. `duk_get_type()` for unpacked `duk_tval` is something like (with -Os, which doesn't inline the `duk_get_tval_or_unused()` helper):

```
00000000004064a8 <duk_get_type>:
  4064a8:       e8 fa ca ff ff          callq  402fa7 <duk_get_tval_or_unused>
  4064ad:       8b 00                   mov    (%rax),%eax
  4064af:       8b 04 85 d0 b3 43 00    mov    0x43b3d0(,%rax,4),%eax
  4064b6:       c3                      retq   
```

Before the change:

```
00000000004064a8 <duk_get_type>:
  4064a8:       e8 fa ca ff ff          callq  402fa7 <duk_get_tval_or_unused>
  4064ad:       8b 10                   mov    (%rax),%edx
  4064af:       b8 04 00 00 00          mov    $0x4,%eax
  4064b4:       83 ea 02                sub    $0x2,%edx
  4064b7:       83 fa 08                cmp    $0x8,%edx
  4064ba:       77 07                   ja     4064c3 <duk_get_type+0x1b>
  4064bc:       0f be 82 28 9d 43 00    movsbl 0x439d28(%rdx),%eax
  4064c3:       c3                      retq   
```

There's an unnecessary range check for the type tag because the compiler cannot know the type tag range is limited.

PR #1078, opened 2016-11-14
Add linenoise hints to 'duk'

svaarala:
- [x] Convert completion handling to a script
- [x] Add hints handling as a script

### DIFF
--- a/examples/cmdline/duk_cmdline.c
+++ b/examples/cmdline/duk_cmdline.c
@@ -24,7 +24,7 @@
 /* Helper define to enable a feature set; can also use separate defines. */
 #if defined(DUK_CMDLINE_FANCY)
 #define DUK_CMDLINE_LINENOISE
-#define DUK_CMDLINE_LINENOISE_COMPLETION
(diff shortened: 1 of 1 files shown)


svaarala:
Example of hints:

hints

svaarala:
Encountered #1079 in writing the linenoise scripts.

fatcerberus:
Those hints look nifty.  Chrome's JS console has something like that which makes it a joy to use compared to the Node.js REPL.

svaarala:
It's still a bit barebones and only recognizes a maximal sequence like `foo.bar.qu` at the end of the input. But for example `[].leng` won't provide a hint or completion with TAB.

But now that the completion/hint code is scripted it's easier to experiment with it and improve it. Errors are cleanly caught and won't be segfaults.

svaarala:
@fatcerberus Linenoise doesn't support Windows but linenoise-ng does and is intended to be API compatible if I understood correctly. I'll try that out if I have time later tonight.

PR #1084, opened 2016-11-14
Add support for ToNumber() octal and binary conversion

svaarala:
See https://github.com/svaarala/duktape/pull/1028#issuecomment-260497240.

- [x] Change ToNumber() to support ES6 octal and binary (still no legacy octal)
- [x] Testcase coverage
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1836,7 +1836,8 @@ Planned
   and identifiers, no RegExp support yet (requires RegExp /u Unicode mode)
   (GH-1001)
 
-* Add support for ES6 octal (0o123) and binary (0b100001) literals (GH-1057)
+* Add support for ES6 octal (0o123) and binary (0b100001) in source code
+  literals and ToNumber() coercion (e.g. "+'0o123'") (GH-1057, GH-1084)
 
 * Add support for ES6 String.prototype.codePointAt(), String.fromCodePoint(),
   and String.prototype.repeat() (GH-1043, GH-1049, GH-1050)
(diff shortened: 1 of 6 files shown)


fatcerberus:
Also according to ES6, binary, e.g. ToNumber("0b1101") = 13.

svaarala:
Yep, legacy octal is still not recognized (matches V8):

```
> +'01234'
1234
> +'0o1234'
668
> +'0b10001'
17
```

Current Duktape:

```
duk> +'01234'
= 1234
duk> +'0o1234'
= NaN
duk> +'0b10001'
= NaN
```

fatcerberus:
Wait, shouldn't the latter two coerce to zero in Duktape, or does having garbage anywhere in the string automatically = NaN?

svaarala:
Garbage is a NaN yes. That's also noted in Annex E:

> [...] In previous editions such strings were converted to NaN [...]

That's different from parseInt() which stops parsing at garbage but allows the pre-garbage part to be parsed:

```
> parseInt('123foo')
123
```

Here's a related fun table :-): https://github.com/svaarala/duktape/blob/master/doc/number-conversion.rst#lexical-trivia-differences-in-call-sites

fatcerberus:
I notice in that table that `parseInt()` has special handling for hex "0xNNN".  Is this the case in ES6 for octal and binary too?  If so that might be good to fix in this pull also.

svaarala:
I don't think parseInt() supports octal or binary; from Node.js:

```
> parseInt('0x123')
291
> parseInt('0o123')
0
> parseInt('0b10001')
0
```

Duktape master:

```
duk> parseInt('0x123')
= 291
duk> parseInt('0o123')
= 0
duk> parseInt('0b10001')
= 0
```

So parseInt() seems fine.

svaarala:
Ok, now works as expected:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-941-g3b5721a-dirty)
duk> +'01234'
= 1234
duk> +'0o1234'
= 668
duk> +'0b10001'
= 17
```

PR #1089, opened 2016-11-17
Add testcase for replacing global object with Proxy

svaarala:


fatcerberus:
I didn't even know this was a thing you could do. :)

svaarala:
Just a small note on the test case: it demonstrates how the technique works and prints in the traps shows when they're called.

But note that the print() calls in the traps are *not* caught by the Proxy. This is because the eval creating the Proxy is compiled before the global object has been replaced, so that the functions in the expression have environment objects pointing to the original global object. These references in the functions don't change so they keep pointing to the original global object.

Only code compiled and executed after the proxy has been replaced will use the replaced global object.

svaarala:
@fatcerberus It's not very common but does have some useful applications, e.g. allowing some constants or virtual object names or something similar to appear as virtual global bindings.

PR #1090, opened 2016-11-17
Fix a few broken asserts in refcount code

svaarala:
The asserts didn't check for NULL properly.  The code itself is OK, and handles NULL correctly.


svaarala:
This is not present in 1.x so nothing to fix in the maintenance branches.

PR #1093, opened 2016-11-19
Implement Math.cbrt(), Math.log{2,10}(), Math.trunc()

fatcerberus:
This pull adds support for ES6 `Math.cbrt()`, which calculates the cubic root of a number, as well as `Math.log2()` and `Math.log10()` (base 2 and base 10 logarithm).  The C99 standard functions are used if available (i.e. for C99/C++11, MSVC 2013+), otherwise a workaround is used instead.

Checklist:
- [x] Implement support for `Math.cbrt()`
- [x] Implement support for `Math.log2()`, `Math.log10()`
- [x] Implement support for `Math.trunc()`
- [x] Testcases
- [x] Releases entry

### DIFF
--- a/src-input/duk_bi_math.c
+++ b/src-input/duk_bi_math.c
@@ -90,18 +90,18 @@ DUK_LOCAL double duk__fmax_fixed(double x, double y) {
 	return duk_double_fmax(x, y);
 }
 
-DUK_LOCAL double duk__cbrt_fixed(double x) {
+DUK_LOCAL double duk__cbrt(double x) {
 	/* cbrt() is C99.  To avoid hassling embedders with the need to provide a
 	 * cube root function, we can get by with pow().  The result is not
 	 * identical, but that's OK: ES6 says it's implementation-dependent.
 	 */
 
-	duk_small_int_t c = (duk_small_int_t) DUK_FPCLASSIFY(x);
-
 #if defined(DUK_CBRT)
 	/* cbrt() matches ES6 requirements. */
 	return DUK_CBRT(x);
 #else
+	duk_small_int_t c = (duk_small_int_t) DUK_FPCLASSIFY(x);
+
 	/* pow() does not, however. */
 	if (c == DUK_FP_NAN || c == DUK_FP_INFINITE || c == DUK_FP_ZERO) {
 		return x;
@@ -114,7 +114,7 @@ DUK_LOCAL double duk__cbrt_fixed(double x) {
 #endif
 }
 
-DUK_LOCAL double duk__log2_fixed(double x) {
+DUK_LOCAL double duk__log2(double x) {
 #if defined(DUK_LOG2)
 	return DUK_LOG2(x);
 #else
@@ -122,11 +122,19 @@ DUK_LOCAL double duk__log2_fixed(double x) {
 #endif
 }
 
-DUK_LOCAL double duk__log10_fixed(double x) {
+DUK_LOCAL double duk__log10(double x) {
 #if defined(DUK_LOG10)
 	return DUK_LOG10(x);
 #else
 	return DUK_LOG(x) / DUK_LOG(10.0);
+#endif
+}
+
+DUK_LOCAL double duk__trunc(double x) {
+#if defined(DUK_TRUNC)
+	return DUK_TRUNC(x);
+#else
+	return x >= 0.0 ? DUK_FLOOR(x) : DUK_CEIL(x);
 #endif
 }
 
@@ -233,9 +241,10 @@ DUK_LOCAL const duk__one_arg_func duk__one_arg_funcs[] = {
 	duk__sin,
 	duk__sqrt,
 	duk__tan,
-	duk__cbrt_fixed,
-	duk__log2_fixed,
-	duk__log10_fixed
+	duk__cbrt,
+	duk__log2,
+	duk__log10,
+	duk__trunc
 #else
 	DUK_FABS,
 	DUK_ACOS,
@@ -250,9 +259,10 @@ DUK_LOCAL const duk__one_arg_func duk__one_arg_funcs[] = {
 	DUK_SIN,
 	DUK_SQRT,
 	DUK_TAN,
-	duk__cbrt_fixed,
-	duk__log2_fixed,
-	duk__log10_fixed
+	duk__cbrt,
+	duk__log2,
+	duk__log10,
+	duk__trunc
 #endif
 };
 
--- a/src-input/duk_bi_math.c
+++ b/src-input/duk_bi_math.c
@@ -90,6 +90,30 @@ DUK_LOCAL double duk__fmax_fixed(double x, double y) {
 	return duk_double_fmax(x, y);
 }
 
+DUK_LOCAL double duk__cbrt_fixed(double x) {
+	/* cbrt() is C99.  To avoid hassling embedders with the need to provide a
+	 * cube root function, we can get by with pow().  The result is not
+	 * identical, but that's OK: ES6 says it's implementation-dependent.
+	 */
+
+	duk_small_int_t c = (duk_small_int_t) DUK_FPCLASSIFY(x);
+
+#if defined(DUK_CBRT)
+	/* cbrt() matches ES6 requirements. */
+	return DUK_CBRT(x);
+#else
+	/* pow() does not, however. */
+	if (c == DUK_FP_NAN || c == DUK_FP_INFINITE || c == DUK_FP_ZERO) {
+		return x;
+	}
+	if (DUK_SIGNBIT(x)) {
+		return -DUK_POW(-x, 1.0 / 3.0);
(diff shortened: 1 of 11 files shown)


fatcerberus:
MSVC 2012 seems to be missing `cbrt()`.  Not surprising, C99 support in MSVC was abysmal prior to v2015.

svaarala:
The easiest solution for platform_fillins.h would be to define `DUK_CBRT` narrowly when it is known to exist:
- C99 or above
- C++11 or above
- MSVC above a certain version

Otherwise leave it undefined.

fatcerberus:
Okay, MSVC builds (AppVeyor) look like they're passing now.

fatcerberus:
@svaarala This is ready to go, just awaiting final test results and review.

fatcerberus:
I added `log2()` and `log10()` to the pull, since they are generally useful (unlike, say, `Math.clz32()`) and have trivial workarounds (log division).

fatcerberus:
The log2 and log10 tests are failing, I'll have to look into that tonight.

fatcerberus:
Added `Math.trunc()`.  I think I'm done now :)

tests/ecmascript/test-bi-math-cbrt.js
```diff
@@ -0,0 +1,47 @@
...
+===*/
+
+function test() {
+    var pd = Object.getOwnPropertyDescriptor(Math, 'cbrt');
+    print(typeof pd.value, pd.writable, pd.enumerable, pd.configurable);
+    print(Math.cbrt.length);
+
+    printExact(Math.cbrt(0));
+    printExact(Math.cbrt(-0));
+    printExact(Math.cbrt(Infinity));
```
svaarala: This is fine but generally testcases use `1/0`, `-1/0`, and `0/0` for +Inf, -Inf, and NaN. Not important as such, just avoids a dependency on the global binding.

tests/ecmascript/test-bi-math-log2.js
```diff
@@ -0,0 +1,44 @@
...
+    printExact(Math.log2(-1));
+    printExact(Math.log2(0));
+    printExact(Math.log2(-0));
+    printExact(Math.log2(1));
+    printExact(Math.log2(Infinity));
+
+    printExact(Math.log2(2));
+    printExact(Math.log2(1024));
+    printExact(Math.log2(2.73121871170759e+244));
+    printRounded6(Math.log2(1000));
```
svaarala: One interesting test for log2 would be exactness for inputs like 2^n where n is in a reasonable range (at least 0 to 53, i.e. the exact integer range).

I know no guarantees are in the specification but for practical purposes this is probably quite interesting. The guarantee holds for V8 at least for the entire IEEE double exponent range.

svaarala: The guarantee won't hold for the C99 `log2()` replacement but it'd still be nice to document the current behavior when the C99 binding is available. Perhaps in a separate testcase, `log2-2ton.js` or something.

svaarala: Here's an example test which checks for both (prints nothing when everything checks out):

```
> expect = 0; x = 1; while (x !== 1/0) { if (Math.log2(x) != expect) { console.log(x, Math.log2(x)); } x *= 2; expect++; }
1023
> expect = 0; x = 1; while (x !== 0) { if (Math.log2(x) != expect) { console.log(x, Math.log2(x)); } x /= 2; expect--; }
-1074
``'

File: src-input/duk_bi_math.c [hunk 3]
```diff
@@ -90,6 +90,54 @@ DUK_LOCAL double duk__fmax_fixed(double x, double y) {
...
+		return DUK_POW(x, 1.0 / 3.0);
+	}
+#endif
+}
+
+DUK_LOCAL double duk__log2_fixed(double x) {
+#if defined(DUK_LOG2)
+	return DUK_LOG2(x);
+#else
+	return DUK_LOG(x) / DUK_LOG(2.0);
```
fatcerberus: I considered that too, but then I remembered reading that most compilers will optimize e.g. `log(2.0)` to a constant during constant folding anyway.  So I kept it this way for clarity.  Your call, though.

tests/ecmascript/test-bi-math-log2.js
svaarala: (The test prints out the last number because of the way Node.js console evals input; it's not actually console.log()ed by the loop.)

src-input/duk_bi_math.c
svaarala: That's quite likely, but they can't do so when `log()` is provided by a custom library which the compiler doesn't know about. But I can make the change separately.

tests/ecmascript/test-bi-math-cbrt.js
fatcerberus: Hm, I guess that makes sense for purity since Duktape allows you to remove bindings at compile time.  I wrote the above under the assumption that the global constants would always be available (they are non-writable and non-configurable, so this is a reasonable assumption in most environments), but for Duktape that may not always hold.

svaarala: Testcases don't need to support those cases though. This was more a comment on consistency. Functionally it's fine to reference Infinity etc.

svaarala: It's not consistently applied but I've tried to start using `void 0` for `undefined` too; as an idiom it generates better code and doesn't rely on the weird global binding. Again, more of a style choice than anything.

src-input/duk_bi_math.c
```diff
@@ -90,6 +90,54 @@ DUK_LOCAL double duk__fmax_fixed(double x, double y) {
 	return duk_double_fmax(x, y);
 }
 
+DUK_LOCAL double duk__cbrt_fixed(double x) {
+	/* cbrt() is C99.  To avoid hassling embedders with the need to provide a
+	 * cube root function, we can get by with pow().  The result is not
+	 * identical, but that's OK: ES6 says it's implementation-dependent.
+	 */
+
+	duk_small_int_t c = (duk_small_int_t) DUK_FPCLASSIFY(x);
```
fatcerberus: Alright, those are easy fixes :)

tests/ecmascript/test-bi-math-cbrt.js
fatcerberus: Alright, I'll change the tests before the merge and save you some trouble :)

fatcerberus:
I fixed the tests to use computed Infinity and NaN and renamed the helper functions.  Let me know if there are any other changes you want to make in this pull.

svaarala:
I'd be interested in the Math.log2() exact 2^n test coverage - but other than that just squashing the fixes. :)

fatcerberus:
`duk__cbrt()` etc. helpers should probably be hidden behind an `#ifdef DUK_USE_ES6` but I didn't do that here because it was awkward with the lookup table for the onearg/twoarg helpers.

svaarala:
Hmm, right.. They'll then be pulled in even for low memory builds. Let's see how that affects footprint once the tests clear. It's probably going to be ~200-300 bytes if I had to guess, maybe enough to fix in a follow-up (not a blocker here IMO).

That array would need to be structured so that it contains the low memory (ES5) bindings first, followed by ES6 binding which will be absent. Awkward, as you said.

fatcerberus:
By the way it's impossible to run individual testcases directly from the command-line using duk now - the include file doesn't get pulled in.  That was an inconvenient surprise ;-)

svaarala:
Yep, there's `util/prep_test.py` which does the preparation but it's not really friendly to use. I'll try to merge in `tools/runtest.py` soon so there's something easier to use.

svaarala:
Hmm, you should probably be able to use `tools/runtest.py` just by copying it off the branch if you wanted to try it right now. But up to you naturally, I'll try to get it merged and test it on Windows myself too.

fatcerberus:
Thanks, I forgot about #1094, I still have to try that out.  After I prep this pull for merge I'll go look at that.

fatcerberus:
I added a test for `Math.log2(2 ** n)` for exponents 0-53, like you suggested.  Ran the test suite manually (it's way quicker on my i7-6700K than waiting for Travis), and everything passes, so looks like we're good. :)

svaarala:
Ok, I'll wait for Travis to see the footprint results and then merge.

fatcerberus:
I could have sworn I fixed those tabs in the .js file...

I recently had to wipe Notepad++ settings as it was locking up on startup, so I lost all my space/tab settings.  I keep forgetting to change them back.

svaarala:
Impact for minsize build on x86 is 200 bytes, roughly as expected.

fatcerberus:
Okay, code policy is good now, this is ready to be merged.

svaarala:
Great, merged :)

PR #1094, opened 2016-11-20
Add util/runtest.py for running individual testcases

svaarala:
Add `util/runtest.py` which allows running individual testcases manually and from scripts:
- Parse testcase metadata from input file.
- Parse testcase expect string from input file.
- Add a testcase framework prefix (with e.g. print/console.log harmonization) and process includes to create a stand-alone prepared testcase (which can be executed as is).
- Execute the test, with optional valgrind wrapping.
- Analyze test results.
- Print a test summary with a diff snippet.
- Output test results into a JSON file for other tools to process (e.g. when running a test suite the suite summary can be created using this output).

The tool can also just prepare the test without running it, so that it's then easy to run it manually using e.g. Node.js and edit the test.

Tasks:
- [x] Add a minimally functioning version
- [x] Resolve FIXMEs
- [x] Use "diff" by default, and if missing on Windows default to FC
- [x] Check overlap with `util/prep_test.py`, e.g. to complete minifier code
- [x] Create knownissue files for existing known issues
- [x] Test on Windows
- [x] Documentation for revised testcase format

The runtest.py tool supports YAML for metadata test case to allow more compact syntax:

```
/*---
custom: true
---*/
```

and allows an inline expect string syntax:

```
print('aiee');  //>aiee
```

Known issues are now described by individual files in `tests/knownissues/`. The file contains a YAML document which provides a summary and other metadata for the known issue, optionally followed by `---` and the expected output for that known issue (i.e. the exact buggy output):

```
summary: prints 'aiee' incorrectly
description: >
    Prints 'aiee' incorrectly because foo is not bar.
----
foo
bar
aiee
```

The expected (broken) output can also be specified using an MD5 key which is useful for huge outputs:

```
summary: prints 'aiee' incorrectly
description: >
    Prints 'aiee' incorrectly because foo is not bar.
md5: bf8759a584cfe6d800c45de80fc59582
```

This allows known issue to be tracked more precisely; previously `doc/testcase-known-issues.yaml` just listed failing tests with a description but the output itself was not checked. So it was possible for a test to fail differently than before without being noticed.

This pull doesn't change runtests or any existing test infrastructure yet. Later pulls will change runtests to use this tool for executing tests and eventually the Javascript-based runtests.js will be replaced with something a bit better.


svaarala:
@fatcerberus This is very far from complete (lots of FIXMEs, some output formatting bugs, etc) but it'd be interesting to hear how this works for you out of the box:

```
$ python tools/runtest.py tests/ecmascript/test-dev-mandel2-func.js 
SUCCESS test-dev-mandel2-func.js [0.0 sec]
duktape@keppihevonen:~/duktape$ python tools/runtest.py tests/ecmascript/test-regexp-invalid-charclass.js 
```

```
$ python tools/runtest.py tests/ecmascript/test-regexp-invalid-charclass.js 
FAILURE test-regexp-invalid-charclass.js [6 diff lines] [errors: expect-mismatch] [0.0 sec]
- Diff to expected result:
    @@ -1,2 +1,2 @@
    -SyntaxError
    +no error
     SyntaxError
```

```
$ python tools/runtest.py tests/ecmascript/test-bi-number-proto-toexponential.js 
KNOWN   test-bi-number-proto-toexponential.js [74 diff lines] [known: corner case rounding errors in toExponential()] [errors: expect-mismatch,known-issue] [0.0 sec]
- Diff to expected result:
    @@ -77,13 +77,13 @@
     2,19 string -1.0000000000000000000e+21
     2,20 string -1.00000000000000000000e+21
     RangeError
    -3,undefined string -9.8765e+20
    -3,notgiven string -9.8765e+20
    +3,undefined string -9.876500000000001e+20
    +3,notgiven string -9.876500000000001e+20
     RangeError
     3,0 string -1e+21
    [... 62 more lines]
```

To get the prepped test case without executing the test:

```
$ python tools/runtest.py --prepare-only --output-prepared /tmp/prepared.js tests/ecmascript/test-bi-number-proto-toexponential.js
Wrote prepared testcase to: '/tmp/prepared.js'
$ nodejs /tmp/prepared.js
[...]
```

svaarala:
Forgot to mention: the tool searches for includes and known issues automatically related to the testcase location. These will get explicit options before merge.

Similarly, the tool expects "duk" to be in `./duk`. If that's not the case, you can say:

```
$ python tools/runtest.py --duk ./duk.O2.master tests/ecmascript/test-dev-mandel2-func.js 
SUCCESS test-dev-mandel2-func.js [0.0 sec]
```

fatcerberus:
I'll look at this later.  I like that it seems to show the diff now though, that'll save me from having run the testcase manually after a failure.

svaarala:
Yeah, I'm trying to implement a few features which makes it nice to use manually when developing, at least:
- Default paths for utilities so that you don't need to provide them explicitly (like now).
- Show a short diff snippet so it's easy to figure out what went wrong. Maybe color the diff for easier readability.
- Show the line in the original source file which contains the first mismatching expect line - this allows pinpointing the failing test more easily than a plain diff (which can sometimes be confusing for large tests).
- Provide comparison to Node.js automatically.

fatcerberus:
It would be nice if one could say:

```
$ tools/runtest.py test-bi-math-hypot`
```

And it would pick up the correct file from tests/ecmascript.

svaarala:
Ok, I'll add that. These defaultings can assume the repo layout so that it would scan for `../tests/ecmascript` when a plain name is not found.

So ideally the utility will have some heuristics built-in for manual use, but accept explicit paths and dir names for scripted use with minimal heuristics.

svaarala:
Current output when running:

```
for i in tests/ecmascript/test-*; do python tools/runtest.py $i; done
```

is something like:

testrun

There are still some of these indent bugs etc, but it's close to being usable as a test suite executor. Apart from summaries etc of course, which should require much less code than what exists in runtests.js now.

fatcerberus:
Hm, this doesn't seem to work on Windows right now:

```
D:\src\duktape>python tools\runtest.py test-bi-math-hypot.js
Execution failed
Failed to diff!
[1;37;41mFAILURE[0m test-bi-math-hypot.js [1 diff lines] [errors: expect-mismatch,returncode-nonzero,stderr-nonempty] [0.0 sec]
- Test wrote to stderr:
    FIXME
    - Diff to expected result:
```

svaarala:
Ah yes, it expects a `diff` command. What would be appropriate on Windows to compare text outputs and provide a diff-like summary?

svaarala:
The other thing you may need to specify now is path to "duk"; it checks for `<cwd>/duk` but not `<cwd>/duk.exe` which I probably need to add. So for now:

```
python tools\runtest.py --duk duk.exe test-bi-math-hypot.js
```

svaarala:
One alternative seems to be `FC` but the output looks quite poor. But maybe I'm just used to unified diffs :-)

fatcerberus:
Windows has `FC` and `COMP`.  Oddly, they seem to be different tools and not aliases for one another:

```
D:\src\minisphere>fc
FC: Insufficient number of file specifications


D:\src\minisphere>comp
Name of first file to compare:
```

fatcerberus:
The main problem with `FC` is that it only shows the lines that are different.  It doesn't include the surrounding lines like `diff` does.

svaarala:
Would either provide a useful diff output for testcases?

svaarala:
I was initially thinking about implementing a minimal fallback into the runtest.py utility itself. Diffing is trivial until you get to missing or extra lines - at which point you need to actually do work to represent the difference in a useful way.

I don't want to replicate "diff" but maybe the tool could show a few of the changed lines with their line numbers; quite often it's enough to indicate the first (and often only) offender.

fatcerberus:
`COMP` looks like it does a character-by-character/binary comparison, so that's useless.  I'd be okay with using FC for diffs.

fatcerberus:
Ah, COMP only works if the files are the exact same size.  I think that one is a relic from early MS-DOS, so that makes sense.

svaarala:
Ok, so I'll use "diff" if that's in the path (you can install one for Windows, even outside Cygwin) and fall back to FC on Windows if diff is not present.

svaarala:
The tooling works on Cygwin but only if testcase paths in the temporary file are converted to Windows path when executed. Sigh... :)

fatcerberus:
Yeah, I was never able to get the test tooling to run on Cygwin even before this pull.  I finally just set up a Ubuntu VM for that purpose and haven't had a problem since :)

svaarala:
Now seems to work on Windows command prompt and Cygwin. For command prompt `fc` is used and ANSI colors are disabled (they seem to be supported out of the box only in Windows 10). For Cygwin paths given to the "duk" command need to be converted to Windows paths but ANSI colors are fine. CR LF normalization is needed for both.

svaarala:
There seem to be a few Math built-in tests that fail on Cygwin right now (atan2 test for example) - I'll open issues/pulls for them once I've confirmed it's the tests and not the tooling.

fatcerberus:
> ANSI colors are disabled (they seem to be supported out of the box only in Windows 10)

Yeah, and there's also a regression in Windows 10 Anniversary Edition (build 14393) that prevents them from working again, so it's better that you disabled them.

fatcerberus:
Hm, this still doesn't seem to work:
```
D:\src\duktape>python tools\runtest.py test-bi-math-hypot.js
Execution failed
FAILURE test-bi-math-hypot.js [36 diff lines] [errors: expect-mismatch,returncode-nonzero,stderr-nonempty] [0.0 sec]
- Test wrote to stderr:
    FIXME
    - Diff to expected result:
    Comparing files C:\USERS\FATCE\APPDATA\LOCAL\TEMP\TMP-DUK-RUNTEST-L8MENM\expect  [... 68 more chars]
    ***** C:\USERS\FATCE\APPDATA\LOCAL\TEMP\TMP-DUK-RUNTEST-L8MENM\expect
    0
    NaN
    NaN
    NaN
    Infinity
    Infinity
    NaN
    Infinity
    [... 26 more lines]
```

I notice a FIXME in the output - Is this not intended to be functional yet?

svaarala:
Did you give `--duk` option to provide the executable path?

svaarala:
The failure reporting is still broken, that's the FIXME.

fatcerberus:
> Did you give --duk option to provide the executable path?

No, that would make sense what went wrong then.  Guess I'm just too used to Linux builds - "make qecmatest" builds duk automatically :)

svaarala:
Yep, this is just the raw tool so it can't really do that :) But I'll make it print out more useful error messages though.

svaarala:
Error messages should now be a bit better, and the tool now tries both `duk` and `duk.exe`. So maybe a plain:

    python tools\runtest.py test-dev-mandel2-func.js

should now work.

svaarala:
Moved the utility to `util/runtest.py` because it won't be part of the dist.

fatcerberus:
Hm, there doesn't seem to be a way to get it to print the entire output of a test, so I can compare manually?  The diffs often don't have enough context for me to gauge where things went wrong.

svaarala:
`--output-stdout <somefile.txt>` should do that.

svaarala:
While this utility is still pretty rough, I'm now using it for most test runs because it shows a useful diff directly and the colored output is easy to follow at a glance. It's also easy to run test subsets using a `for i in <something>; do python util/runtest.py ... ; done` loop.

So maybe in a while this could replace most (or all) of runtests.js and make the testing tools more accessible than they are now.

fatcerberus:
Ah, thanks.  I thought that option was just for redirecting the normal output (i.e. the diffs)

Regarding coloring: I don't seem to get that on my Ubuntu VM.  Is there a command-line option to enable it?

svaarala:
@fatcerberus The options are quite likely to change when the tool is improved towards replacing runtests.js, but in the meanwhile:

```
python util/runtest.py --help
```

shows available options with hopefully reasonable descriptions.

svaarala:
The coloring should be automatically enabled if running on Linux and output is a tty. This is the exact condition:

```
windows = platform.system() == 'Windows'
cygwin = 'CYGWIN' in platform.system()
istty = sys.stdout.isatty()
use_colors = istty and not windows
```

It's not really that good so maybe `istty` is incorrect for you, for some reason.

Note that redirecting causes the output to become not a tty and lose coloring, e.g.:

```
$ python util/runtest.py tests/ecmascript/test-dev-mandel2-func.js | less
SUCCESS test-dev-mandel2-func.js [0.0 sec]
```

would not be colored.

svaarala:
I could add a command line option to force ANSI coloring even if the detection doesn't work for some reason.

fatcerberus:
Hm, on further inspection it seems the coloring is working for the test tool itself (FAILURE with red bg), but Ubuntu's `diff` isn't colored for some reason.

svaarala:
The diff clip isn't colored: it's just plain output from a `diff -u`.

If you wanted to view a colored diff, a roundabout way of doing that would be to:
- Use `--output-diff <somewhere>` for runtest.py
- Then `colordiff < <somewhere>`.

fatcerberus:
Ah, that makes sense.  I got too spoiled by `git diff` always being colored I guess. :)

PR #1095, opened 2016-11-21
Minor cleanups to Math.cbrt() etc

svaarala:
- [x] Make function pointers in math table conditional to `DUK_USE_ES6` or replace them with `duk__math_nop()` when ES6 not enabled (footprint)
- [x] Extend Math.log2() 2^N testcase for all IEEE double exponents (separate testcase)
- [x] Precomputation for `/ DUK_LOG(2)` and `/ DUK_LOG(10)` in fallback path

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1847,7 +1847,7 @@ Planned
   and String.prototype.repeat() (GH-1043, GH-1049, GH-1050)
 
 * Add support for ES6 Math.hypot(), Math.cbrt(), Math.log2(), Math.log10(),
-  Math.trunc() (GH-1069, GH-1093)
+  Math.trunc() (GH-1069, GH-1093, GH-1095)
 
 * Add support for ES6 Object.assign() (GH-1064)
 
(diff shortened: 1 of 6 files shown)


svaarala:
```
> expect = 0; x = 1; while (x !== 1/0) { if (Math.log2(x) != expect) { console.log(x, Math.log2(x)); } x *= 2; expect++; }
1023
> expect = 0; x = 1; while (x !== 0) { if (Math.log2(x) != expect) { console.log(x, Math.log2(x)); } x /= 2; expect--; }
-1074
```

fatcerberus:
Don't forget to make `Math.hypot()` conditional too.

fatcerberus:
Regarding that test above, why do positive exponents only go up 1023 but negative exponents go to -1074?  Based on my (somewhat limited) knowledge of IEEE floating point representation I'd have thought it would be symmetric in both directions.

svaarala:
The exponent range doesn't but in the negative direction there are denormals which make up the difference.

svaarala:
Math.hypot() is already conditional and it's not part of the function pointer tables so it should already be OK.

fatcerberus:
Ah, that's right, I forgot it has its own function because it's vararg.

PR #1099, opened 2016-11-23
Fix a few math issues on Cygwin (MinGW)

svaarala:
- [x] Bug testcase
- [x] Fix
- [x] Add asserts for ISO C assumptions
- [x] Generalize NetBSD workaround config option
- [x] `NaN ** 0` should be 1, but is NaN
- [x] `(-1e100) % (-1/0)` should be -1e100, but is NaN
- [x] `(1/0) % (1/0)` (and any inf sign combination) should be NaN (not +/- inf)
- [x] `0 % x` should be 0, preserving sign, for any x not a NaN or 0
- [x] `0 % (0/0))` and `0 % 0` should be NaN
- [x] `Math.atan2(1/0, 1/0)` should be 0.785... but is NaN; same NaN issue for infs of any sign
- [x] `Math.pow(0/0, 0)` and `Math.pow(0/0, -0)` should be 1, but is NaN
- [x] 2.0 migration notes
- [x] Releases entry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,13 +59,20 @@ DUKTAPE_CMDLINE_SOURCES = \
 	examples/alloc-logging/duk_alloc_logging.c \
 	examples/alloc-torture/duk_alloc_torture.c \
 	examples/alloc-hybrid/duk_alloc_hybrid.c \
-	examples/debug-trans-socket/duk_trans_socket_unix.c \
 	extras/print-alert/duk_print_alert.c \
 	extras/console/duk_console.c \
 	extras/logging/duk_logging.c \
 	extras/module-duktape/duk_module_duktape.c
-LINENOISE_SOURCES = \
-	linenoise/linenoise.c
+ifdef SYSTEMROOT  # Windows
+DUKTAPE_CMDLINE_SOURCES += examples/debug-trans-socket/duk_trans_socket_windows.c
+else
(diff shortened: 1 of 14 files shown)


fatcerberus:
I assume this is an issue with the CRT used in Cygwin.

svaarala:
Most likely, unless intrinsics were used (for `%` that might be possible). The easiest fix is to special case a few of the required behaviors for Cygwin (same as for NetBSD pow()). It might be a bit overbroad but pinpointing a very specific condition is also quite difficult.

svaarala:
No time to finish the fix right now but there's a specific bug testcase in the pull now.

svaarala:
For simplicity it's probably best to widen the set of NetBSD fixes to include the MinGW issues too, and then change `DUK_USE_POW_NETBSD_WORKAROUND` to be a bit more generic like `DUK_USE_POW_WORKAROUNDS`.

fatcerberus:
As an aside, I find the specified behavior of `NaN ** 0 = 1` to be odd - returning NaN for such operation seems more intuitive.

svaarala:
Quite a lot of small things broken with MinGW. I'm not sure if all of them are actual ISO/ANSI C guarantees or just guarantees that are usually in place. There are asserts for the assumptions now however so easier to catch if some other runtime violates them.

svaarala:
After maybe 10 attempts I think I finally got all the cases handled. Phew. :)

PR #1102, opened 2016-11-25
Runtest.py improvements

svaarala:
Some follow-ups:

- [ ] Command line option cleanup
- [ ] Automatic Node.js comparison

### DIFF
--- a/util/runtest.py
+++ b/util/runtest.py
@@ -8,6 +8,8 @@
 #  linking.
 #
 
+# FIXME: nodejs comparison
+
 # XXX: encoding issues
 # XXX: external minifier
 # XXX: use strict support; perhaps via test metadata


svaarala:
@fatcerberus After trying this out a bit, any specific improvements that you'd like to see? I'll be doing the above at least.

fatcerberus:
One option I'd like is the ability to print the raw test output to stdout, rather than the diff-to-expected.  I tried doing `--output-stdout=-` but that didn't work.  Currently it's a two-step process:

```
python util/runtest.py --output-stdout=out test-foo.js
cat out
```

svaarala:
So maybe `--print-stdout` which would print the direct output - but still display the summary line?

fatcerberus:
If by summary line you mean the Success/Fail line then yes.

With a lot of tests I don't find the diffs to be that useful because much of the output is very similar.  diff output of course includes line numbers but since those aren't related to the line numbers in the actual test script, that doesn't help.

svaarala:
> With a lot of tests I don't find the diffs to be that useful because much of the output is very similar. diff output of course includes line numbers but since those aren't related to the line numbers in the actual test script, that doesn't help.

Agree there, two things that would make this less of an issue:

1. Smaller testcases. As I touch tests, I'm trying to split them up into smaller tests.
2. Identifying the source testcase line related to the problem.

For the latter it's not really possible to identify the line of code that prints the failing expect output; though that would be possible for the new shorthand notation:

    print('hello');  //>hello

one could then assume quite safely that the expected output is related to the line.

But lacking that, what might be useful is to identify the source line of the *expect string* that fails. That's often enough to find the offending line.

It would be relatively easy to figure out the first erroring expect line and pinpoint that; follow-up lines are more difficult because it needs some understanding of "edit distance" to make useful suggestions. But the first erroring line would be easy and often useful.

fatcerberus:
> But lacking that, what might be useful is to identify the source line of the expect string that fails. That's often enough to find the offending line.

Agree, that'd be a nice feature to have.  Assuming that failing tests and passing tests don't differ in the number of lines they print (which should be true most of the time, barring some really epic failure case), then it would be possible to pinpoint the followup lines too right?

svaarala:
Actually quite many testcases will have a different number of lines for failing tests vs. successful tests. For example, if a test function does three things and the first errors out, three lines of output are replaced with a single line of error description. So I think that it's a minority of testcases where the output line count is constant regardless of failures.

One of the benefits of the assert-based model is that it will pinpoint the exact offending line accurately.

fatcerberus:
> One of the benefits of the assert-based model is that it will pinpoint the exact offending line accurately.

That's true.  The tradeoff is that you maybe lose information on *how* the assert failed.  So I'm not really sure what's better.

svaarala:
How much information is lost depends a bit on the assert model. For example:

    assertEqual(-0, +0);

can fail and indicate:

- Assertion failed at line N.
- Required value '-0', actual value '+0'.

But if the assertion has the form:

    assert(Object.is(-0, +0));

then information is indeed lost.

PR #1104, opened 2016-11-25
Fix a few debug/assert compile warnings

svaarala:


fatcerberus:
Speaking of compile warnings, there's been a pointer-to-int coercion warning in the x86 build (on linux) for a while now but I keep forgetting to mention it.  I'm not near a computer now so I can't tell you where exactly it is unfortunately.  I think it might have been in `duk_heap_alloc.c` maybe?

svaarala:
It's probably the pointer-to-seed-value which was intentionally left as part of the random number rework: https://github.com/svaarala/duktape/blob/master/src-input/duk_heap_alloc.c#L899-L914.

The pointer can be removed from the seeding. It's not strong but also not insignificant especially with memory layout randomization.

fatcerberus:
Hm, ES6 seems to specifically require `Math.random()` to produce different sequences for different realms:
http://www.ecma-international.org/ecma-262/6.0/#sec-math.random

> Each Math.random function created for distinct code Realms must produce a distinct sequence of values from successive calls.

svaarala:
In a sense that is fulfilled with a heap-wide generator: because the global environments (realms) share the sequence, they won't see identical sequences. If one thread reads numbers, those specific numbers won't be given to another thread.

The intent of the specification is more likely that the PRNGs should be independent. Duktape doesn't do that at least now.

svaarala:
So I don't think the current PRNG violates that code realm requirement: no code sequence will see an identical stream because they consume a shared stream which is not repeated (except after a long, long period).

In any case this is out of scope of this pull :grin:

fatcerberus:
> In any case this is out of scope of this pull 😁

Indeed, I just wanted to mention the warning before I forgot about it for the umpteenth time :)

svaarala:
I meant more the ES6 PRNG requirements which are unrelated to the pull (or even the heap pointer being mixed in - a heap is shared across multiple code realms so including or excluding the pointer makes no difference to the ES6 requirement).

PR #1109, opened 2016-11-26
Debugger callstack index change cleanups

svaarala:
- [x] Fix Eval segfault
- [x] Fix compile warning for unused variable
- [x] Shared helper to read and validate callstack index
- [x] 2.0 migration notes on affected commands
- [x] Debugger documentation X or Y argument notation consistency
- [x] Test affected commands using JSON proxy

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1721,7 +1721,7 @@ Planned
   version incompatible protocol changes in the 2.0.0 release (GH-756)
 
 * Incompatible change: make callstack level mandatory for most debugger
-  commands which accept one (GH-747)
+  commands which accept one (GH-747, GH-1109)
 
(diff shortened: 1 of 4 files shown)


svaarala:
Notation even prior to the change in #747 wasn't very consistent: sometimes `[foo]` was used to indicate alternatives and sometimes optionality/repetition. I changed to use the following conventions:

```
# Alternatives, e.g. one integer or two strings:
(<int: foo> | <str: bar> <str: quux>)

# Repetition, e.g. 0-N integers:
[<int: foo>]*

# Repetition, e.g. 1-N values, each string or integer:
[<str: foo> | <int: bar>]+
```

What I don't really like is putting multiple values inside single angle brackets, because each `<foo>` is intended to represent a single dvalue.

PR #1114, opened 2016-11-27
Add some new Array methods from ES6+

fatcerberus:
This is a work-in-progress pull to implement `Array.prototype.includes()` and any other easy-to-implement `Array` built-ins.  Not yet merge ready.

- [x] Implement `Array.prototype.includes()` built-in
- [ ] Review ES6 for other low-hanging fruit w.r.t. `Array` builtins
- [x] Basic testcase
- [ ] Improve coverage in testcase
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1910,6 +1910,8 @@ Planned
 * Add support for ES6 octal (0o123) and binary (0b100001) in source code
   literals and ToNumber() coercion (e.g. "+'0o123'") (GH-1057, GH-1084)
 
+* Add support for ES6 Array.prototype.includes() (GH-1114)
+
 * Add support for ES6 String.prototype.codePointAt(), String.fromCodePoint(),
   and String.prototype.repeat() (GH-1043, GH-1049, GH-1050)
 
--- a/src-input/builtins.yaml
+++ b/src-input/builtins.yaml
@@ -878,6 +878,17 @@ objects:
           length: 1
           varargs: true
         present_if: DUK_USE_ARRAY_BUILTIN
+      - key: "includes"
+        value:
+          type: function
+          native: duk_bi_array_prototype_indexof_shared
+          length: 1
+          varargs: true
+          magic: 2  # magic: includes()
+        es6: true
+        present_if:
+          - DUK_USE_ARRAY_BUILTIN
+          - DUK_USE_ES6
       - key: "indexOf"
         value:
           type: function
--- a/src-input/duk_bi_array.c
+++ b/src-input/duk_bi_array.c
@@ -1287,14 +1287,24 @@ DUK_INTERNAL duk_ret_t duk_bi_array_prototype_unshift(duk_context *ctx) {
 }
 
 /*
- *  indexOf(), lastIndexOf()
+ *  indexOf(), lastIndexOf(), includes()
  */
 
 DUK_INTERNAL duk_ret_t duk_bi_array_prototype_indexof_shared(duk_context *ctx) {
+	/*
+	 *  magic == +1: Array.prototype.indexOf()
+	 *  magic == +2: Array.prototype.includes()
+	 *  magic == -1: Array.prototype.lastIndexOf()
+	 */
+
+	duk_small_int_t magic;
 	duk_idx_t nargs;
 	duk_int_t i, len;
 	duk_int_t from_idx;
-	duk_small_int_t idx_step = duk_get_current_magic(ctx);  /* idx_step is +1 for indexOf, -1 for lastIndexOf */
+	duk_small_int_t idx_step;  /* idx_step is +1 for indexOf, -1 for lastIndexOf */
+
+	magic = duk_get_current_magic(ctx);
+	idx_step = magic >= 0 ? 1 : -1;
 
 	/* lastIndexOf() needs to be a vararg function because we must distinguish
 	 * between an undefined fromIndex and a "not given" fromIndex; indexOf() is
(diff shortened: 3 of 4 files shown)


fatcerberus:
I would have liked to implement `Array.of()` and `Array.from()` here (especially the latter), but they have strange semantics for constructing the new Array and also seem to be contingent on both symbol *and* iterator support being present.  So those might have to wait :)

svaarala:
So if the iterator path is skipped for now, the other interesting stuff seems to be:
- If `this` is a constructor, it is called to create the result; otherwise an ordinary array is created.
- There's a mapping function, which doesn't seem too odd.
- The `length` is set explicitly at the end which is necessary at least if an external constructor was called.

Looking at that algorithm it would be tempting to add an internal helper to convert an array-like object into an array for internal processing purposes. It's not the best memory-wise but would allow faster implementation of initial bindings. Unfortunately that doesn't work here because some of the side effects need to happen "as you go". It might be possible to create a "walk and array like" helper with a C callback though.

If you wanted to implement a subset of the required behavior (at least without iterator support and maybe without an external constructor support) that'd be fine by me, as long as some test or issue tracks the remainder.

fatcerberus:
Would `duk_enum()` with ARRAY_INDICES_ONLY work?  Or does that only work for real arrays?

svaarala:
Unfortunately eval doesn't respect `.length` specially as required for "array like" objects in the specifications. Essentially they're processed by looking up `.length`, and then walking all indices [0,length-1[ by looking up the properties explicitly.

svaarala:
Another related issue is that it would be nice if "array like" call sites would handle actual dense arrays as a fast path because in 99+% cases they're dense arrays. This can be implemented case by case (and sometimes that's the only option) but for iterating it'd be nice if a helper could handle that specially. Even though it wouldn't be as fast as a fast path crafted for a specific call site, it'd still improve performance without call site specific handling.

fatcerberus:
Huh, that constructor redirection behavior is interesting.  A quick test in Node.js:

```
> Array.from([1, 2, 3])
[ 1, 2, 3 ]
> Reflect.apply(Array.from, Object, [ [ 1, 2, 3 ] ])
{ '0': 1, '1': 2, '2': 3, length: 3 }
```

fatcerberus:
One nice thing about ES6 iterators is that it's mostly all duck typed, there's no actual "iterator class" as far as I'm aware.  So for bindings which *consume* iterators (like `Array.from()`, that really only relies on symbol support: To be iterable, an object must have `typeof obj[Symbol.iterator] === 'function'`, and from there it's just a matter of calling the function to get the iterator, and then calling `iter.next()` successively.  `duk_enum()` could be adapted to do that pretty easily, I think.

The harder part is adding support to _produce_ iterators, e.g. when one calls `plainArray[Symbol.iterator]()`.  But that's of course a problem to be solved in a separate issue.

svaarala:
Hmm, but wouldn't the iterator `.next()` call need to be in `duk_next()` rather than `duk_enum()`? That wouldn't be really difficult either, but it's a bit different control-wise than the current enumerate-ahead-of-time approach.

It may be possible to implement this so that `duk_next()` and its internal variants would accept either the current enumerator state object created by `duk_enum()` or otherwise expect an iterator object. The former can be detected on each `duk_next()` from its internal properties which are missing from the iterator. If the target is an iterator, just call `.next()`, otherwise use the current behavior.

fatcerberus:
Ah right, `duk_enum()` is not atomic but works like an iterator itself, i.e. you have to manually advance it.  That's probably even better: the overall mapping to iterators is then pretty close to 1:1.

svaarala:
The internal enumerator state object is not currently exposed to user code in any way. But it would be possible to make it inherit from a prototype object which provided a `.next()` method so that it would actually be iterator-like even with its weird internal state book-keeping.

fatcerberus:
Anyway, I think it's probably best to implement `Array.from()` in a separate pull (or a couple of them), since the implementation of all the ES6 semantics is a bit involved.  That constructor redirection behavior in particular is potentially useful and I'm not sure I'd want to leave it out of even a first implementation.

The necessary plumbing for iterator support is also best tackled separately, I think.  For this pull, I'd like to keep it confined to only those Array methods which are easy to implement without touching internals.

fatcerberus:
Hm, so it looks like `.find()` and `.findIndex()` might be able to piggyback on the shared indexOf helper too.  That function is getting quite the workout ;-)

svaarala:
That's true :) It's an OK approach as the fallback implementation because it minimizes footprint and reduces code repetition. But there should later be separate fast paths for dense array cases which don't need to be generic because they won't be enabled for footprint conscious targets.

fatcerberus:
Hm, would it be possible to add a new enumeration flag, something like `DUK_ENUM_ARRAY_LIKE`, which would cause `duk_enum()`/`duk_next()` to use array-like behavior, i.e. read `.length` and then enumerate all keys in [0,length) regardless of existence (non-present = undefined)?

svaarala:
It would - but it wouldn't work in the call sites because creating the enumerator would trigger the potential side effects of reading the properties too early.

fatcerberus:
Ah, good point.  Still maybe useful as a future addition to the C API though.

The idea pile doesn't stop from getting taller :)

svaarala:
For the C API an "array-like to array" conversion would maybe be most useful. Though I'm not sure where application writers would encounter "array like" values that were not arrays.

fatcerberus:
> Though I'm not sure where application writers would encounter "array like" values that were not arrays.

Maybe some JS code passed one into a Duktape/C function.  Lots of people confuse `arguments` with a real array, for example.

svaarala:
Yeah, `arguments` is one case that comes to mind. But can't really think of others.

For `arguments` specifically the useful API call would be to normalize it into an array instance, not necessarily enumerate it.

fatcerberus:
I'll close this pull for now.  I'm not happy with the mess this will make of the shared indexOf helper and it's bit late in the Duktape 2.0.x cycle for any kind of major refactoring.  I'll revisit ES6+ built-ins later, perhaps for Duktape 2.1.x.

PR #1119, opened 2016-11-28
Use same bitpacked format for heap strings and thread property string data

svaarala:
This reduces footprint, and is also required later on by the ES6 Symbol pull; the current string data has limitations for full 8-bit data which are removed here.


fatcerberus:
Looked at the diff, that bitpacked format is a pretty clever way to reduce footprint.

svaarala:
It's been in use since 1.0 for heap-wide strings (that get a stridx) but not for other strings that are part of the built-ins init data. As the number of built-ins grows, it's good to extend the bitpacking to strings without stridx too. The symbols pull needed support for the full 8-bit range which wasn't previously supported.

-400 bytes for the x64 default build. Not huge, but offsets some of the Symbol pull footprint (which was around +1.6kB or so).

PR #1124, opened 2016-12-01
Add a type summary document

svaarala:
Add a table summarizing external and internal types.


svaarala:
Github doesn't render this very well; a plain HTML version form Pandoc is scalable and more viewable:

```
$ pandoc -r rst -w html5 doc/types.rst > /tmp/types.html
```

Issue #1110, opened 2016-11-27
Improve Duktape.info() by using an object format


svaarala:
`Duktape.info()` is currently the only Duktape specific introspection primitive and it's only available to Ecmascript code. It accepts any value as an argument and returns a version specific array of values which describes the object: http://duktape.org/guide.html#builtin-duktape-info.

This is very footprint friendly because introspection doesn't involve a lot of separate bindings for different pieces of information. The array return format is also very footprint friendly, but quite awkward for calling code to deal with, especially because it's version specific. Calling code may need to check `Duktape.version` to figure out what the array values mean.

A few small changes would make the API more useful and easier to use:
- Change the return value from an array to an object. This makes the result self describing and makes it much easier to version - both for Duktape and the call site.
- Add an optional object argument to the call. If present, the object would provide options for introspection if that becomes useful at some point. For example, if the output for some introspection aspect creates a large output, a boolean can be used to request a partial/full version.
- Add a C binding with the same signature: argument object (or undefined/null) in, introspection result object out.

These changes would make introspection more easily usable and more extensible and hopefully prevent the need to create many partial introspection APIs.

See discussion in #969, e.g. https://github.com/svaarala/duktape/pull/969#issuecomment-263096373.

svaarala:
Tagged this for 2.0.0; even though it's tight: it'd be nice to make the format change with no change in the logical introspection result.

svaarala:
Example of a current result:

```
((o) Duktape [linenoise] 1.99.99 (v1.5.0-1025-g0ee5611-dirty)
duk> obj = { foo: 123, bar: 234 }
= {foo:123,bar:234}
duk> print(JSON.stringify(Duktape.info(obj), null, 4))
```

Prints out:

```js
[
    6,
    null,  // this is actually a pointer
    2,
    64,
    50,
    2,
    2,
    0,
    0
]
```

Example of what the above result array would look like as an object:

```js
{
    "type": 6,
    "ptr": null,  // this is actually a pointer
    "refCount": 2,
    "hdrSize": 64,
    "propSize": 50,
    "eSize": 2,
    "eNext": 2,
    "aSize": 0,
    "hSize": 0
}
```

svaarala:
#1125 implements an object result value. There's no object argument now (not even documented) so I'll leave this open but unassign from 2.0.0.

PR #1125, opened 2016-12-01
Convert Duktape.info() to return an object

svaarala:
- [x] Add a duk_get_info() API call
- [x] Naming: `duk_push_info()`? `duk_inspect_value()`?
- [x] Convert Duktape.info() to use duk_get_info()
- [x] Include previous data but also add thread state
- [x] Review key names
- [x] Config option for how deep inspection to support => Future work, #1156
- [x] Minimal testcase coverage
- [x] Move Duktape.info() keys list into API document
- [x] Reference API document from guide Duktape.info()
- [x] Check keys to match pull final result
- [x] Migration notes
- [x] Releases entry

Implements #1110.


svaarala:
@fatcerberus This reworked Duktape.info() now exposes the thread state:

```
duk> T = new Duktape.Thread(function foo() {})
= {}
duk> Duktape.info(T)
= {type:6,itag:9,hptr:(0x18fbcb0),refc:2,tstate:1,hbytes:816,pbytes:0,esize:0,enext:0,asize:0,hsize:0}
duk> Duktape.Thread.resume(T, 123)
= undefined
duk> Duktape.info(T)
= {type:6,itag:9,hptr:(0x18fbcb0),refc:2,tstate:5,hbytes:816,pbytes:0,esize:0,enext:0,asize:0,hsize:0}
```

The states are defined in `duk_hthread.h` and `tstate: 5` is `DUK_HTHREAD_STATE_TERMINATED`.

svaarala:
Not sure about the best API call name, right now it is `duk_get_info()`.

But `duk_inspect_value()` might be better: the name (inspect) would clearly convey that you're looking into internals, which might remind you that the output is version specific.

The C equivalent of `Duktape.act()` could be e.g. `duk_inspect_callstack_entry()` or similar.

svaarala:
This is around +300 bytes compared to the previous solution. That's quite much for an inspection binding that is only rarely needed, and illustrates a bit why I chose the raw array format initially. The thought behind that choice was that inspection is rarely used in practice, and when it is, the call site can promote the raw array into an object with useful property names if supporting multiple versions is necessary.

I still think the inspection output should be an object because that array was ultimately a bit too cryptic. If the array format was kept, a workable model might be to include a polyfill to convert the raw output into an object - and that polyfill would take Duktape version(s) into account when doing the conversion. So in practice applications would have an object-variant for the inspection result. But even this is awkard to maintain in the longer run because everything must be representable uniquely as an array which easily gets awkward.

So, this is maybe a good place to try to figure out how to make the generated C code smaller for writing multiple properties at once... :)

Another useful change would be to remove the Ecmascript `Duktape.info()` binding and only leave the C binding. This would have the effect of `duk_inspect_value()` being omitted during linking if it wasn't actually called by the application, thus removing the footprint impact in many cases without the need to edit built-in bindings to get rid of `Duktape.info()`.

svaarala:
With a format string + duk_int_t array based "set multiple properties" helper the x64 defsize impact is now +160 bytes. I also tried a vararg helper which was interestingly worse than individual `duk_put_prop_string()` calls. It's also important to put the value into an array or a struct rather than local variables, otherwise at least gcc will emit a lot of shuffling code which makes footprint much worse.

svaarala:
One reason I also want to move to the object format is that it'll then be easier to provide a lot of inspection results by default, but allow that to be scaled down to fewer properties for low memory targets. With properties it's simply a matter of setting fewer properties; with the array approach it was much more difficult.

svaarala:
I'll merge this as soon as the tests pass. It's not ideal, but the footprint impact is now within reason. Current output is something like:

```
duk> print(Duktape.enc('jx', Duktape.info({ foo: 123 }), null, 4))
{
    hptr: (0x8d0090),
    type: 6,
    itag: 9,
    refc: 1,
    hbytes: 64,
    class: 1,
    pbytes: 50,
    esize: 2,
    enext: 1,
    asize: 0,
    hsize: 0,
    variant: 0
}
```

I'll open a follow-up pull for whatever final tweaks are made, so that key naming etc can still be changed.

svaarala:
(Couple of documentation issues to be done still, just noticed.)

PR #1126, opened 2016-12-01
Add duk_push_bare_object() API call

svaarala:
Add a convenience API call `duk_push_bare_object()` which pushes an object without an internal prototype, same as `Object.create(null)`. Bare objects are useful for various tracking hashmap purposes.

Previously one would have to `duk_push_object()` and then `duk_set_prototype()`. This pull exposes an internal helper which pushed a bare object already.

- [x] Expose new API call
- [x] Testcases
- [x] API documentation
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1882,6 +1882,9 @@ Planned
   duk_components_to_time()) to allow C code to conveniently work with the
   same time provider as seen by Ecmascript code (GH-771)
 
+* Add duk_push_bare_object() API call which pushes an object without an
+  internal prototype, equivalent to Object.create(null) (GH-1126)
+
 * Add ability to perform an indirect debugger Eval with non-empty callstack by
   sending null for the callstack level (GH-747)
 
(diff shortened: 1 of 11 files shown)


fatcerberus:
I sometimes use stashed objects for tracking purposes - for example `RequireScript()` is like a one-time #include - so this will be useful.

I never thought to set the prototype to null though... there might be undiscovered bugs related to that, I'll need to check that now.

svaarala:
Mainly the bugs may be some code assuming that a `toString` property has "already been seen" or similar.

It seems that the module loaders may actually have this bug :) I'll add a pull to fix them.

svaarala:
It seems they're semi-broken. For example the Node.js loaded `require('toString')` won't trigger module loading (incorrect) but will return undefined because the cache object lacks an `id`.

fatcerberus:
I have to check my module loading code later then.  The Node.js loader extra was adapted from minisphere's loader so it likely has the same bug.

svaarala:
This is a good example why I wanted to have this in the API (also because it was already used internally so the footprint impact is 0): having it around makes it more likely applications will use a bare object for caches etc.

fatcerberus:
Somewhat related, I just noticed the documentation for `duk_set_prototype()`:

> Set the internal prototype of the value at index to the value at stack top (which must be an object or `undefined`). If the target value is not an object or the prototype value is not an object or `undefined`, an error is thrown.

Shouldn't that be `null`, not `undefined`?

svaarala:
> Shouldn't that be null, not undefined?

There's of course no specification requirement for the C API in this regard, and right now it is `undefined` or object. I think we've had this discussion a few times ;)

But in this case I agree it should accept `null` to mirror `Object.create()`. I'll add a pull to do so.

fatcerberus:
> There's of course no specification requirement for the C API in this regard, and right now it is undefined or object. I think we've had this discussion a few times ;)

I know, I just wanted to make sure that was the actual behavior and not a documentation mistake.  I actually wasn't making a value judgment as to "correct" behavior this time :)

svaarala:
> I actually wasn't making a value judgment as to "correct" behavior this time :)

:) It's still maybe worth changing ... :)

PR #1128, opened 2016-12-02
Add C API equivalent for Duktape.act()

### LINKED ISSUE #777: Duktape C API call to inspect callstack
Add an equivalent of `Duktape.act()` to the C API so that C code can inspect callstack entries. This would be useful e.g. in the logging extra which now falls back to using an Ecmascript helper to automatically figure a logger name.

The simplest API binding would be `duk_push_callstack_entry(ctx, level)` which would push the same object `Duktape.act()` pushes. This is not very efficient but extensible and low footprint. A more performant solution would either represent the callstack entry as a C struct; the API call could also identify the field we need and only return that. But as a baseline a `Duktape.act()` equivalent makes most sense to me.

svaarala:
- [x] Rebase after #1125
- [x] Add `duk_inspect_callstack_entry()`
- [x] API documentation
- [x] Website changes
- [x] Testcase coverage
- [x] Releases entry

Fixes #777.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1898,6 +1898,12 @@ Planned
 * Add DUK_GC_COMPACT flag to duk_gc() to force object property table
   compaction (GH-778, GH-1129)
 
+* Add duk_inspect_value() to provide internal information about an argument
+  value; the output matches Duktape.info() (GH-1125)
+
+* Add duk_inspect_callstack_entry() to provide internal information about a
+  callstack entry; the output matches Duktape.act() (GH-1128)
+
 * Add ability to perform an indirect debugger Eval with non-empty callstack by
   sending null for the callstack level (GH-747)
 
--- a/src-input/duk_api_inspect.c
+++ b/src-input/duk_api_inspect.c
@@ -182,3 +182,49 @@ DUK_EXTERNAL void duk_inspect_value(duk_context *ctx, duk_idx_t idx) {
 	    "bcbytes" "\x00" "dbytes" "\x00" "tstate" "\x00" "variant" "\x00" "\x00",
 	    (duk_int_t *) &vals);
 }
+
+DUK_EXTERNAL void duk_inspect_callstack_entry(duk_context *ctx, duk_int_t level) {
+	duk_hthread *thr = (duk_hthread *) ctx;
+	duk_activation *act;
+	duk_uint_fast32_t pc;
+	duk_uint_fast32_t line;
+
+	DUK_ASSERT_CTX_VALID(ctx);
+
+	/* -1             = top callstack entry, callstack[callstack_top - 1]
+	 * -callstack_top = bottom callstack entry, callstack[0]
+	 */
+	if (level >= 0 || -level > (duk_int_t) thr->callstack_top) {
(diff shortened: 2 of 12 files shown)


svaarala:
One naming idea:
- duk_inspect_callstack(ctx, level) and correspondingly duk_inspect_value(ctx, idx) for the Duktape.info() equivalent

fatcerberus:
Agree, "inspect" makes it clear that you're using some heavy introspection, rather than "get" which sounds more typical.

svaarala:
`duk_inspect_value()` and `duk_inspect_callstack_entry()` are the API call names now. Pretty verbose, but on the other hand these are very rarely called so it's good their names are descriptive.

fatcerberus:
Good to have this merged, I took advantage of it right away.  There were a few places in minisphere where I did stuff like this (this being written early in development before I thought about `duk_eval_string()` or even `duk_get_global_string()`):

```c
duk_push_global_object(ctx);
duk_get_prop_string(ctx, -1, "Duktape");
duk_get_prop_string(ctx, -1, "act");
duk_push_int(ctx, -4);
duk_call(ctx, 1);
// interpret callstack entry data
```

It was nice to replace all that with a single call, with the added bonus that I could choose to leave out the `Duktape` built-in in the future and callstack inspection would still work.  Or if some game script happens to modify it destructively.  That's important since I rely on it for error blaming, and error handling needs to be robust for obvious reasons.

fatcerberus:
One enhancement, if I could suggest, would be for `duk_inspect_callstack_entry()` to return a boolean indicating whether the callstack entry could be successfully retrieved while still leaving `undefined` on the valstack in that case, similar to the semantics of `duk_get_prop_xxx()`.  For example my blaming code does this:

```c
	dukrub_inspect_callstack_entry(ctx, -1 + blame_offset);
	if (!duk_is_object(g_duk, -1)) {
		dukrub_inspect_callstack_entry(ctx, -1);
		duk_replace(ctx, -2);
	}
```

It would be nice if I could move the `duk_inspect_callstack_entry()` call itself into the if-condition, saving a line of code.

svaarala:
That sounds good - maybe `duk_inspect_value()` could have a similar return value.

fatcerberus:
@svaarala Just wanted to mention, I've been using `Duktape.act()` for a long time in this handy "blamed" `duk_error()` variant:
https://github.com/fatcerberus/minisphere/blob/master/src/engine/api.c#L293-L327

It's really useful for throwing argument validation errors from Duktape/C functions, since I can blame the error on the Ecmascript activation just below.  Now that callstack inspection is a first-class citizen in Duktape, maybe this "blamed throw" call is something to consider for the public Duktape API?

svaarala:
Letting the caller select the "blamed" function for file/line assignment is one possibility - but I think the more robust approach is to allow functions to indicate if they're "infrastructure" (not blamed) or application (blamed).

Often a function can be called both by an application but also by another infrastructure function, so that it's difficult to know what level in the callstack to blame. For example, a target function may be called directly or via `.forEach()` and it would be incorrect to blame the `.forEach()` for the error.

PR #1130, opened 2016-12-02
Add duk_def_prop() ToString() key coercion

### LINKED ISSUE #836: duk_def_prop() doesn't allow index as key
At least, the last time I tried to use an index instead of a string for the key, it didn't work.  If it's indeed the case that numeric indices are not allowed as keys for `duk_def_prop()`, would it be difficult to change it so they are?

svaarala:
Add a ToString() coercion for duk_def_prop() key. Previously the key was required to be a string and e.g. a number wasn't accepted. Fixes #836.

- [x] Change coercion (note that will need to be ToPropertyKey() in the future)
- [x] Testcase update
- [x] Migration notes
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1803,6 +1803,9 @@ Planned
   duk_del_prop_lstring(), duk_has_prop_lstring(), duk_get_global_lstring(),
   duk_put_global_lstring() (GH-946, GH-953)
 
+* Add a ToString (later, ToPropertyKey) coercion to duk_def_prop() key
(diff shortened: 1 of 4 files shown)


fatcerberus:
Hm, I guess it will need to be changed to accept symbols later, since there's no symbol support in master yet.

svaarala:
Yes, in general all the property access functions need to use ToPropertyKey() rather than ToString() for coercion. I made a note in the symbol pull to check that before merge.

PR #1132, opened 2016-12-02
Add explicit lexenv/varenv fields to duk_hcompfunc

### LINKED ISSUE #1144: Assert failure when out-of-memory during duk_hcompfunc creation
When duk_hcompfunc creation (compilation or closure creation) fails due to an out-of-memory error, the `duk_hcompfunc`'s data field may remain NULL. There are a few asserts requiring that it is non-NULL and memory unsafe behavior may be possible if the pointer is in fact NULL.

This can be reproduced to some degree with ajduk:

```
$ python util/runtest.py  --duk ./ajduk tests/ecmascript/test-bi-nodejs-buffer-proto-varint.js --clip-columns 200
FAILURE test-bi-nodejs-buffer-proto-varint.js [593 diff lines] [errors: expect-mismatch,returncode-nonzero,stderr-nonempty] [0.0 sec]
- Test wrote to stderr:
    *** FATAL ERROR: assertion failed: DUK_HCOMPFUNC_GET_DATA(thr->heap, f) != NULL (duk_heap_refcount.c:128)
- Diff to expected result:
    @@ -1,590 +0,0 @@
    -variable size int test
    -16 bytes: 00000000000000000000000000000000
    -16 bytes: 41414141414141414141414141414141
    -AAAAAAAAAAAAAAAA
    -[object Uint8Array]
    -16
    -16 bytes: 55555555555555555555555555555555
    -16 bytes: 5555555555555555aaaa555555555555
    -4 111111fe111111111111111111111111
    [... 581 more lines]
```

The easy (and robust) fix is to tolerate NULL data pointer. It only really needs to be supported in refcount (and possibly mark-and-sweep) code because it only happens when out-of-memory happens and the value stack needs to be unwound. A better fix would be to structure the creation code so that when the `duk_hcompfunc` is created, the data buffer already exists so that setting it to the `duk_hcompfunc` always succeeds.

Tagged 'security' because this may lead to memory unsafe behavior.

svaarala:
Scope pointers are currently internal properties of Ecmascript functions. Because they're nearly always needed, being explicit properties has unnecessary property overhead for both reading and writing them, and for RAM footprint. Add explicit lexenv/varenv fields into `duk_hcompfunc`.

- [x] Add the fields and make related call handling etc changes
- [x] Debugger changes (artificial properties)
- [x] Test artificial properties manually
- [x] Internal debug print macro fixes
- [x] Compress pointers
- [x] Fix duk_hcompfunc data != NULL assumptions #1144
- [x] Performance test => negligible change
- [x] Releases entry

Fixes #1144.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2008,6 +2008,11 @@ Planned
 
 * Add an fmod() self test (GH-1108)
 
+* Fix duk_hcompfunc 'data' field != NULL assumptions which might lead to
+  memory unsafe behavior if Duktape ran out of memory when creating a
+  duk_hcompfunc during compilation or function instantiation (GH-1144,
+  GH-1132)
+
 * Fix a few bugs in object property handling (delete property and
   Object.defineProperty()) where an object property table resize triggered
   by a finalizer of a previous value could cause memory unsafe behavior
@@ -2117,7 +2122,7 @@ Planned
   GH-973, GH-1042); minor RegExp compile/execute optimizations (GH-974,
   GH-1033); minor IEEE double handling optimizations (GH-1051); precomputed
   duk_hstring array index (GH-1056); duk_get_{type,type_mask}() optimization
-  (GH-1077)
+  (GH-1077); explicit lexenv/varenv fields in duk_hcompfunc struct (GH-1132)
 
 * Miscellaneous footprint improvements: RegExp compiler/executor (GH-977);
   internal duk_dup() variants (GH-990); allow stripping of (almost) all
@@ -2125,7 +2130,8 @@ Planned
   helper and use duk_def_prop() instead (GH-1010); minor IEEE double handling
   optimizations (GH-1051); precomputed duk_hstring array index (GH-1056);
   internal value stack access improvements (GH-1058); shared bitpacked string
-  format for heap and thread initialization data (GH-1119)
+  format for heap and thread initialization data (GH-1119); explicit
+  lexenv/varenv fields in duk_hcompfunc struct (GH-1132)
 
 * Internal change: rework shared internal string handling so that shared
   strings are plain string constants used in macro values, rather than
--- a/debugger/duk_debug_proxy.js
+++ b/debugger/duk_debug_proxy.js
@@ -937,7 +937,7 @@ TargetConnHandler.prototype.trialParseDvalue = function trialParseDvalue() {
                 if (avail >= 2 + len) {
                     v = new Uint8Array(len);
                     v.set(buf.subarray(2, 2 + len));
-                    v = { type: 'heapptr', pointer: Duktape.enc('hex', plainof(v)) };
+                    v = { type: 'heapptr', pointer: Duktape.enc('hex', plainOf(v)) };
                     consume(2 + len);
                 }
             }
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -2433,6 +2433,10 @@ The following list describes artificial keys included in Duktape 1.5.0, see
 +---------------------------------+---------------------------+---------------------------------------------------------+
 | (not present yet)               | ``duk_hcompfunc``         | Ecmascript function data area, including bytecode.      |
 +---------------------------------+---------------------------+---------------------------------------------------------+
+| ``lex_env``                     | ``duk_hcompfunc``         | Function lexical environment.                           |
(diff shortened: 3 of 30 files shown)


svaarala:
Similar change should be considered for other function fields, e.g. `_Formals`.

svaarala:
Test Eval for a function, and look up `lex_env` and `var_env` artificial properties:
```
{"request":"Eval","args":[null,"(function foo(a,b,c) {})"]}                          
{"reply":true,"args":[0,{"type":"object","class":6,"pointer":"00000000011028c0"}]}
{"request":"GetHeapObjInfo","args":[{"type":"object","class":6,"pointer":"00000000011028c0"}]}
{"reply":true,"args":[0,"heapptr",{"type":"heapptr","pointer":"00000000011028c0"},0,"heaphdr_flags",806095232,0,"heaphdr_type",1,0,"refcount",2,0,"extensible",true,0,"constructable",true,0,"boundfunc",false,0,"compfunc",true,0,"natfunc",false,0,"bufobj",false,0,"thread",false,0,"array_part",false,0,"strict",false,0,"notail",false,0,"newenv",true,0,"namebinding",true,0,"createargs",false,0,"envrecclosed",false,0,"exotic_array",false,0,"exotic_stringobj",false,0,"exotic_arguments",false,0,"exotic_dukfunc",false,0,"exotic_proxyobj",false,0,"class_number",6,0,"class_name","Function",0,"prototype",{"type":"object","class":6,"pointer":"00000000010e0a20"},0,"props",{"type":"pointer","pointer":"0000000001102e40"},0,"e_size",7,0,"e_next",7,0,"a_size",0,0,"h_size",0,0,"nregs",3,0,"nargs",3,0,"lex_env",{"type":"object","class":15,"pointer":"0000000001102a30"},0,"var_env",{"type":"object","class":15,"pointer":"0000000001102a30"},0,"start_line",1,0,"end_line",1,0,"data",{"type":"heapptr","pointer":"0000000001101f30"}]}
{"request":"GetHeapObjInfo","args":[{"type":"object","class":6,"pointer":"0000000001102a30"}]}
{"reply":true,"args":[0,"heapptr",{"type":"heapptr","pointer":"0000000001102a30"},0,"heaphdr_flags",2013266048,0,"heaphdr_type",1,0,"refcount",2,0,"extensible",true,0,"constructable",false,0,"boundfunc",false,0,"compfunc",false,0,"natfunc",false,0,"bufobj",false,0,"thread",false,0,"array_part",false,0,"strict",false,0,"notail",false,0,"newenv",false,0,"namebinding",false,0,"createargs",false,0,"envrecclosed",false,0,"exotic_array",false,0,"exotic_stringobj",false,0,"exotic_arguments",false,0,"exotic_dukfunc",false,0,"exotic_proxyobj",false,0,"class_number",15,0,"class_name","DecEnv",0,"prototype",{"type":"object","class":14,"pointer":"00000000010e0800"},0,"props",{"type":"pointer","pointer":"0000000001102ac0"},0,"e_size",2,0,"e_next",1,0,"a_size",0,0,"h_size",0]}
```

svaarala:
x86 minsize is about -120 bytes which is nice, considering no functionality is lost.

The difference in `duk_hcompfunc` size is visible in Duktape.info(). This is from master (ajduk, x86):

```
= [6,(0xf7542350),3,44,91,7,7,0,0,4]
```

Header size (`duk_hcompfunc`) is 44 bytes, property table is 91 bytes (7x13), and bytecode + constants is 4 bytes (just a RETURN). Total: 139 bytes.

After the changes here:

```
duk> Duktape.info(Duktape.compact(function foo() {}))
= [6,(0xf7489948),3,52,78,6,6,0,0,4]
```

Header size (`duk_hcompfunc`) is 52 bytes, property table is 78 bytes (6x13), and bytecode + constants is 4 bytes (just a RETURN). Total: 134 bytes, with pointer compression the total would be 130 bytes. Omitting `_Formals` reduces 13 more bytes; compressing property table keys would be 10 more bytes (for 5 remaining properties), so total would be 107 bytes.

svaarala:
Here's the result with pointer compression applied to lexenv/varenv references:

```
duk> Duktape.info(Duktape.compact(function foo() {}))
= [6,(0xf74fa350),3,48,78,6,6,0,0,4]
```

Total is 48 + 78 + 4 = 130 bytes as expected.

PR #1133, opened 2016-12-02
Accept 'null' for duk_set_prototype()

svaarala:
Accept also `null` type for `duk_set_prototype()` to better align with `Object.create()`.

- [ ] Code change
- [ ] Testcases
- [ ] API documentation
- [ ] Releases entry


svaarala:
Related, `duk_get_prototype()` now returns an object or `undefined` to match the `duk_set_prototype()` behavior. That should also be changed to "null".

If that's done, this change is no longer a compatible change (for success path code anyway) so maybe `duk_set_prototype()` should no longer accept undefined either.

PR #1134, opened 2016-12-02
Allow bound function as initial Duktape.Thread function

svaarala:
- [x] Relax the requirement
- [x] Add test coverage for bound and native targets
- [x] Documentation fixes
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1905,6 +1905,9 @@ Planned
 * Remove duk_{get,put,has,del}_var() calls from API header; they were not
   fully implemented and not part of the documented public API (GH-762)
 
+* Allow a bound Ecmascript function as an argument to new Duktape.Thread()
+  (GH-1134)
+
 * Minor changes to error messages for errors thrown by Duktape internals
   (GH-827, GH-839, GH-840, GH-1016)
 
--- a/doc/testcase-known-issues.yaml
+++ b/doc/testcase-known-issues.yaml
@@ -57,10 +57,6 @@
   knownissue: "'\\u0000' should ToNumber() coerce to NaN, but now coerces to zero like an empty string"
   # Moved
 -
-  test: "test-dev-bound-thread-start-func.js"
-  knownissue: "initial function of a new coroutine cannot be bound"
-  # Moved
--
   test: "test-dev-func-cons-args.js"
   knownissue: "corner cases for 'new Function()' when arguments and code are given as strings"
   # Moved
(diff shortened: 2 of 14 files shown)


svaarala:
The previous limitation is essentially type checking to ensure that once we longjmp, we'll be able to complete the call handling and resume the target thread using the bytecode executor. That works if the target is an Ecmascript function; if it's not, the call setup (for an Ecmascript-to-Ecmascript call) fails and there's an internal error.

Just removing the type check in `Duktape.Thread.resume()` is sufficient to remove the limitation. So it may be enough to just remove the type check, and make the call handling setup error after the longjmp (in executor code) to provide a useful error.

svaarala:
This should now be functional, and I've updated the tests (seems I introduced an overlapping test, but doesn't matter). I still need to figure out if there's any documentation that's out of sync.

svaarala:
Oops, the longjmp resume handler cannot actually deal with the call setup error very well because that error causes a bytecode executor exit which is not the right thing to do if there's an active try-catch for example. Maybe that can be fixed - but if not, the resume call site must pre-validate the argument by resolving the bound function chain, etc.

@fatcerberus But this should work for your testing purposes in its current state.

fatcerberus:
"Invalid state" seems kind of vague -- but I'm not really sure what would be a better message.

svaarala:
Yeah, note that the error happens on resume, not thread creation. If the error was thrown in thread creation (which would be more logical) "invalid args" would be good enough IMO. For the resume() case the thread already exists but is in an invalid state.

svaarala:
@fatcerberus I'll merge this next (once the tests clear); it's not ideal but better than before and I don't want to rework the coroutine stuff too much if generators are hopefully around the corner. But this change should allow you to go forward with your generator experiment better (from what I understood).

svaarala:
This is almost +100 bytes for the x86 defsize build, for nothing except an argument validity check :(

The cost comes from needing a helper to look up the non-bound target function which is relatively awkward now and not needed elsewhere. It will be offset if bound function handling is improved.

fatcerberus:
Thanks for the heads-up, I'll try test out some generator stuff tonight :)

fatcerberus:
Now that I think of it I could have used a bind()-alike implemented in JS to get around this limitation... figures I didn't think of that until now...

svaarala:
:) No worries, this was a limitation that was supposed to be removed anyway. I'm not sure how useful this is for practical code, maybe forcing a "this" binding would make sense in some situations.

svaarala:
But in any case it's surprising for the caller if bound functions are not allowed, because they're intended to behave like normal (Ecmascript) functions just about everywhere.

fatcerberus:
The bound function works perfectly. :)

PR #1135, opened 2016-12-03
Fix bound function internal prototype handling to match ES6

svaarala:
In ES6 (and ES7) internal prototype should be copied from the current internal prototype of the target, at the time of the `.bind()` call:
- http://www.ecma-international.org/ecma-262/6.0/#sec-function.prototype.bind
- http://www.ecma-international.org/ecma-262/6.0/#sec-boundfunctioncreate
- Steps 2 and 8

In ES5 Function.prototype was used for bound functions:
- https://www.ecma-international.org/ecma-262/5.1/#sec-15.3.4.5
- Step 11: "Set the [[Prototype]] internal property of F to the standard built-in Function prototype object as specified in 15.3.3.1."

Tasks:
- [x] Change behavior
- [ ] Fix testcases
- [x] Migration note
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1996,6 +1996,10 @@ Planned
   for a target function with name "foo", bound function name is "bound foo"
   (GH-1113)
 
+* Change bound function internal prototype handling to match ES6 requirements;
+  bound function internal prototype is copied from the target function
+  instead of always being Function.prototype (GH-1135)
+
 * Add a fastint check for duk_put_number_list() values (GH-1086)
 
 * Remove an unintended fastint downgrade check for unary minus executor
--- a/doc/release-notes-v2-0.rst
+++ b/doc/release-notes-v2-0.rst
@@ -1058,6 +1058,10 @@ Other incompatible changes
   "enumerate" trap has been obsoleted.  Key enumerability is also now checked
   when "ownKeys" trap is used in Object.keys() and for-in.
 
+* Bound function internal prototype is copied from the target function to match
+  ES6 requirements; in ES5 (and Duktape 1.x) bound function internal prototype
+  is always set to Function.prototype.
+
 Known issues
 ============
 
--- a/src-input/duk_bi_function.c
+++ b/src-input/duk_bi_function.c
@@ -302,6 +302,7 @@ DUK_INTERNAL duk_ret_t duk_bi_function_prototype_call(duk_context *ctx) {
  * merges argument lists etc here.
  */
 DUK_INTERNAL duk_ret_t duk_bi_function_prototype_bind(duk_context *ctx) {
+	duk_hthread *thr = (duk_hthread *) ctx;
 	duk_hobject *h_bound;
 	duk_hobject *h_target;
 	duk_idx_t nargs;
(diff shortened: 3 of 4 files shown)


fatcerberus:
I wonder why they even went to the trouble of changing this in ES6.  Is it common in real-world code to change the internal prototype of a function to something other than `Function.prototype`?

svaarala:
I haven't encountered code doing that in practice. Just hypothetically, maybe if you implement some kind of RPC framework, RPC-backed methods might inherit from a different prototype.

fatcerberus:
Huh, so I was skimming through the specification and I found the section on `Function.prototype.toString()`.  Unless I'm misreading the spec, it seems to require that either `.toString()` returns the actual function source code for an Ecmascript function, or else to return something that will cause a SyntaxError:
http://www.ecma-international.org/ecma-262/6.0/#sec-function.prototype.tostring

Specifically the last two bullets:
> * If the object was defined using ECMAScript code and the returned string representation is not in the form of a MethodDefinition or GeneratorMethod then the representation must be such that if the string is evaluated, using eval in a lexical context that is equivalent to the lexical context used to create the original object, it will result in a new functionally equivalent object. In that case the returned source code must not mention freely any variables that were not mentioned freely by the original function’s source code, even if these “extra” names were originally in scope.
> * If the implementation cannot produce a source code string that meets these criteria then it must return a string for which eval will throw a SyntaxError exception.

Duktape's current behavior matches neither: `eval()`ing the string will neither produce a "functionality equivalent object" (in fact it will produce a nop function), nor will it cause a syntax error.

Of course ES5.1 is naturally much less specific about this :)
http://www.ecma-international.org/ecma-262/5.1/#sec-15.3.4.2
> An implementation-dependent representation of the function is returned. This representation has the syntax of a FunctionDeclaration. Note in particular that the use and placement of white space, line terminators, and semicolons within the representation String is implementation-dependent.

svaarala:
Yeah that seems changed - ES5.1 specifically requires that the .toString() result *doesn't* cause a SyntaxError :)

fatcerberus:
I guess there's so much code out there using .toString() to get function source code that throwing a syntax error on eval is preferable - otherwise the code may break in more subtle ways.

svaarala:
Code that relies on source code being available and working perfectly when parsed is quite weird in my opinion. There's never any guarantee that a particular function has been written in Ecmascript for instance.

But I see what you mean - maybe that's the reason for the change.

fatcerberus:
I agree - however, it does seem to be a common technique.  See this SO question, the bottom answer claims that at least AngularJS relies on `Function.prototype.toString()` returning actual source (I've yet to verify this claim but found it interesting):
http://stackoverflow.com/questions/16937940/in-what-javascript-engines-does-function-prototype-tostring-not-return-the-sourc

fatcerberus:
This is also interesting:
http://www.2ality.com/2016/08/function-prototype-tostring.html

svaarala:
So I think the best minimal change is to return e.g.:

```
function foo() { [ecmascript code] }
```

which won't parse, and will be similar to the suggested `function foo() { [native code] }`.

fatcerberus:
That has the nice property that it matches the syntax used by V8:
```
C:\Users\fatce>node
> Object.prototype.toString.toString()
'function toString() { [native code] }'
```

fatcerberus:
By the way, if function `.toString()` behavior is changed, be careful not to reintroduce the bug fixed by #554.

svaarala:
I know, although that's not really a bug in Duktape but rather a non-compatible expectation in Emscripten. There's no alternative though: either follow ES6 and potentially break Emscripten compatibility (not sure if that's the case - hopefully not), or don't follow ES6.

PR #1141, opened 2016-12-03
Prune duk_hcompfunc _Formals when possible

svaarala:
- [x] Rebase once #1132 is merged
- [x] Add proper _Formals omission condition to the compiler
- [x] Use _Formals, fallback to _Length, and finally fallback to assuming .length == nargs (which is almost always the case)
- [x] Because nargs == _Formals.length almost always (maybe always now), drop _Length and keep _Formals if that were not the case? (Reduces footprint and eliminates one unnecessary check.)
- [x] Check case where function accesses e.g. `Math` using slow path
- [x] Downgrade debug logs
- [x] Testcase coverage
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2131,7 +2131,8 @@ Planned
   optimizations (GH-1051); precomputed duk_hstring array index (GH-1056);
   internal value stack access improvements (GH-1058); shared bitpacked string
   format for heap and thread initialization data (GH-1119); explicit
-  lexenv/varenv fields in duk_hcompfunc struct (GH-1132)
+  lexenv/varenv fields in duk_hcompfunc struct (GH-1132); omit duk_hcompfunc
+  _Formals array when it is safe to do so (GH-1141)
 
 * Internal change: rework shared internal string handling so that shared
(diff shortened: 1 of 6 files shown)


svaarala:
Rebased on top of #1132 to test impact on function size (compiled ajduk without debugger support; otherwise _Formals is not omitted):

```
duk> Duktape.info(Duktape.compact(function foo() {}))
= [6,(0xf7496544),3,40,65,5,5,0,0,4]
```

Header size is 40, property table size is 65 (5x13), bytecode+consts is 4 bytes: total 109 bytes. The extra 8 byte reduction comes from disabling debugger support which drops function start/end line, so size comparable to examples in #1132 would be 117 bytes.

svaarala:
Concrete impact for ajduk-rom, running dist-files/mandel.js; align-by-1, debugger support disabled.

From master, heap dump after test has executed (mandelbrot function still in global object):

```
======= heap hwm = 24064 use = 10972 waste = 764 ======
```

From this branch:

```
======= heap hwm = 23524 use = 10440 waste = 732 ======
```

PR #1146, opened 2016-12-04
Allow dropping of _Varmap when only safe slow path variable reads are present

svaarala:
Currently any slow path variable read in a function (among with other conditions) causes a function's `_Varmap` to be kept. This is necessary in general so that a slow path variable access matches variables declared in a function; in the worst case the declared variables set is dynamic because eval() may introduce new variables.

This can be relaxed so that the _Varmap can be dropped even when there are slow path accesses, provided that the accesses are guaranteed not to match the function's statically declared variables (provided that there are no inner functions or eval()).

For example, if a function accesses `Math.PI` but: (1) doesn't declare `Math`, (2) doesn't call direct eval, and (3) doesn't have inner functions, it's safe to drop _Varmap.

- [x] Add separate compiler tracking state for "slow path, not own static variable" and "slow path, maybe own static variable"
- [x] Add compiler tracking state for "contains one or more try-catch statements" => Existing "catch_depth" is enough.
- [x] Change _Varmap condition
- [x] Change _Formals condition to allow dropping of formals when _Formals.length is zero, as long as _Formals.length == nargs
- [x] Use `duk_uint8_t`s for compiler status flags, reduces code and RAM footprint slightly
- [x] Reduce bytecode buffer preallocation for targets that prefer size (unrelated but implemented at the same time)
- [x] Testcase coverage
- [x] Add test for inner/outer function, with inner function dropping _Varmap but accessing outer function variables
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2146,7 +2146,9 @@ Planned
   internal value stack access improvements (GH-1058); shared bitpacked string
   format for heap and thread initialization data (GH-1119); explicit
   lexenv/varenv fields in duk_hcompfunc struct (GH-1132); omit duk_hcompfunc
-  _Formals array when it is safe to do so (GH-1141)
+  _Formals array when it is safe to do so (GH-1141); omit duk_hcompfunc
+  _Varmap in more cases when it is safe to do so (GH-1146); reduce initial
+  bytecode allocation in Ecmascript compiler for low memory targets (GH-1146)
 
 * Internal change: rework shared internal string handling so that shared
   strings are plain string constants used in macro values, rather than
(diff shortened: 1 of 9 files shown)


svaarala:
Here's a concrete indication of the change.

From master:

```
duk> function foo(a,b) { print(Math.PI); }
= undefined
duk> Duktape.info(foo)
= [6,(0x1850790),3,112,150,6,6,0,0,68]
```

There are 6 properties total, one of them is _Varmap.  From this branch:

```
duk> function foo(a,b) { print(Math.PI); }
= undefined
duk> Duktape.info(foo)
= 6,0x1286fd0,3,112,125,5,5,0,0,68
```

_Varmap is omitted in this (safe) case.

Unsafe cases (to be tested):
- With statement
- Try-catch (caught variable uses slow path access)
- Inner function
- Eval

PR #1150, opened 2016-12-05
Bump RAM built-ins initdata bits for normal/function property count

### LINKED ISSUE #846: Bump DUK__NUM_FUNC_PROP_BITS for RAM built-ins by 1 bit
The current limit, 64, is easily reached when adding custom built-ins via genbuiltins.py YAML metadata files. Make the bit count dynamic (chosen by genbuiltins.py) or bump it by 1 bit.

svaarala:
Previous limitation is encountered with custom bindings. Longer term solution is to add a shared, efficient varint encoding to the bit packed data formats and use it consistently for all count/length fields.

Fixes #846.


svaarala:
Oops, there's a FIXME in the releases file. I'll fix it separately.

PR #1152, opened 2016-12-05
Use varuint encoding for more initdata fields

svaarala:
Remove hard limits on stridx, bidx, natidx, and class number.


svaarala:
The main point here is to reduce fixed bit limits for e.g. built-in objects or built-in strings, but footprint also improves by around 100 bytes.

PR #1153, opened 2016-12-06
Add config options for function .name and .fileName properties

svaarala:
Add explicit options to control .name and .fileName property presence. These are hacked away already by some applications, so it's nice to support making them optional in master. Dropping .name leads to some ES6 compliance issues (not ES5); dropping either affects e.g. tracebacks.

- [x] Rebase after #1146 is merged
- [x] Add config options
- [x] Add ifdefs
- [x] Fix function declaration handling; now expects .name to be passed via function template
- [x] Cleaner handling for function declaration? If not, add asserts for stack top and type
- [x] Remove .fileName (which is non-standard) from low_memory.yaml or at least low_memory_strip.yaml
- [x] Make function `.name` property configurable to match http://www.ecma-international.org/ecma-262/7.0/
- [x] Make function `.length` property configurable to match http://www.ecma-international.org/ecma-262/7.0/
- [x] Make function `.fileName` non-writable (writable before) to match `.name` behavior
- [x] Make Error `.fileName` and `.lineNumber` non-writable to match `.name` behavior
- [x] Test coverage for `.name`, `.length`, `.fileName`, and `.lineNumber` attribute changes
- [x] Releases entry


svaarala:
With both .name and .fileName disabled (and no debugger support):

```
duk> foo = function(){}
= {_func:true}
duk> Duktape.info(foo)
= [6,(0x15ac060),3,112,75,3,3,0,0,4]
duk> Object.getOwnPropertyNames(foo)
= ["length","prototype"]
```

The only internal property is `_Pc2line` which can also be dropped when tracebacks are not enabled (typical for very low memory environments).

Only remaining properties are `.length` (can be virtualized) and `.prototype` (can be spawned on access).

PR #1158, opened 2016-12-07
Use packed args for internal duk_{get,put,del,has,xdef}_prop_stridx() call sites

svaarala:
Goal is to reduce footprint for call sites that occur very often. Almost all call sites use a small negative or positive value stack index and a stridx; these can be packed into a 32-bit argument which removes one argument load.

- [x] Check footprint potential
- [x] Check all call sites and ensure arguments are in range (if e.g. value stack index is a variable, may not be able to use internal packed args helper)
- [x] Rename callsites to highlight they expect packed arguments?
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2216,7 +2216,8 @@ Planned
   lexenv/varenv fields in duk_hcompfunc struct (GH-1132); omit duk_hcompfunc
   _Formals array when it is safe to do so (GH-1141); omit duk_hcompfunc
   _Varmap in more cases when it is safe to do so (GH-1146); reduce initial
-  bytecode allocation in Ecmascript compiler for low memory targets (GH-1146)
+  bytecode allocation in Ecmascript compiler for low memory targets (GH-1146);
+  packed arguments for some internal helper calls (GH-1158)
 
 * Internal change: rework shared internal string handling so that shared
   strings are plain string constants used in macro values, rather than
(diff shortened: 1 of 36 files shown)


svaarala:
Footprint potential is about -800 bytes for x64 defsize build which is pretty good considering there's no functional change, just packing arguments for internal helper calls.

svaarala:
For x86 minsize potential is around -1000 bytes.

svaarala:
To illustrate the change in output, `duk_put_prop_stridx(ctx, -3, DUK_STRIDX_SOMETHING)` compiles roughly to this in master (the arguments don't match but just for illustration):

```
  40fb00:       ba 6b 00 00 00          mov    $0x6b,%edx
  40fb05:       be fe ff ff ff          mov    $0xfffffffe,%esi
  40fb0a:       48 89 ef                mov    %rbp,%rdi
  40fb0d:       e8 2e 88 01 00          callq  428340 <duk_xdef_prop_stridx>
```

With the change here that becomes something like:

```
  42620f:       be 5b 00 fd ff          mov    $0xfffd005b,%esi
  426214:       48 89 df                mov    %rbx,%rdi
  426217:       e8 a1 de fe ff          callq  4140bd <duk_put_prop_stridx_short_raw>
```

There are ~100 call sites like these so the footprint savings add up quickly. Using one less argument has a direct benefit but it also reduces register pressure and shuffling. That's probably why the effect is most pronounced on x86 which is register poor compared to x64 or x32.

PR #1160, opened 2016-12-08
Add boolean return value for inspect API calls

svaarala:
See comment: https://github.com/svaarala/duktape/pull/1128#issuecomment-265661105.

- [x] API change
- [ ] Testcase update
- [ ] Documentation
- [ ] Releases (issue number)

### DIFF
--- a/src-input/duk_api_inspect.c
+++ b/src-input/duk_api_inspect.c
@@ -54,7 +54,7 @@ DUK_LOCAL void duk__inspect_multiple_uint(duk_context *ctx, const char *fmt, duk
 #define DUK__IDX_TSTATE   12
 #define DUK__IDX_VARIANT  13
 
-DUK_EXTERNAL void duk_inspect_value(duk_context *ctx, duk_idx_t idx) {
+DUK_EXTERNAL duk_bool_t duk_inspect_value(duk_context *ctx, duk_idx_t idx) {
 	duk_hthread *thr = (duk_hthread *) ctx;
 	duk_tval *tv;
 	duk_heaphdr *h;
@@ -181,9 +181,11 @@ DUK_EXTERNAL void duk_inspect_value(duk_context *ctx, duk_idx_t idx) {
 	    "pbytes" "\x00" "esize" "\x00" "enext" "\x00" "asize" "\x00" "hsize" "\x00"
 	    "bcbytes" "\x00" "dbytes" "\x00" "tstate" "\x00" "variant" "\x00" "\x00",
 	    (duk_int_t *) &vals);
+
+	return vals[DUK__IDX_TYPE] != DUK_TYPE_NONE;  /* 1 if value present, 0 if not */
 }
 
-DUK_EXTERNAL void duk_inspect_callstack_entry(duk_context *ctx, duk_int_t level) {
+DUK_EXTERNAL duk_bool_t duk_inspect_callstack_entry(duk_context *ctx, duk_int_t level) {
 	duk_hthread *thr = (duk_hthread *) ctx;
 	duk_activation *act;
 	duk_uint_fast32_t pc;
@@ -196,7 +198,7 @@ DUK_EXTERNAL void duk_inspect_callstack_entry(duk_context *ctx, duk_int_t level)
 	 */
 	if (level >= 0 || -level > (duk_int_t) thr->callstack_top) {
 		duk_push_undefined(ctx);
-		return;
+		return 0;
 	}
 	duk_push_bare_object(ctx);
 	DUK_ASSERT(level >= -((duk_int_t) thr->callstack_top) && level <= -1);
@@ -227,4 +229,6 @@ DUK_EXTERNAL void duk_inspect_callstack_entry(duk_context *ctx, duk_int_t level)
 	 * Duktape relies on them having consistent data, and this consistency
 	 * is only asserted for, not checked for.
 	 */
+
+	return 1;
 }
--- a/src-input/duk_api_public.h.in
+++ b/src-input/duk_api_public.h.in
@@ -776,8 +776,8 @@ DUK_EXTERNAL_DECL duk_bool_t duk_put_global_lstring(duk_context *ctx, const char
  *  Inspection
  */
 
-DUK_EXTERNAL_DECL void duk_inspect_value(duk_context *ctx, duk_idx_t idx);
-DUK_EXTERNAL_DECL void duk_inspect_callstack_entry(duk_context *ctx, duk_int_t level);
+DUK_EXTERNAL_DECL duk_bool_t duk_inspect_value(duk_context *ctx, duk_idx_t idx);
+DUK_EXTERNAL_DECL duk_bool_t duk_inspect_callstack_entry(duk_context *ctx, duk_int_t level);
 
 /*
  *  Object prototype
--- a/src-input/duk_bi_duktape.c
+++ b/src-input/duk_bi_duktape.c
@@ -14,7 +14,7 @@
 #if defined(DUK_USE_DUKTAPE_BUILTIN)
 
 DUK_INTERNAL duk_ret_t duk_bi_duktape_object_info(duk_context *ctx) {
-	duk_inspect_value(ctx, -1);
+	(void) duk_inspect_value(ctx, -1);
 	return 1;
 }
 
@@ -22,7 +22,7 @@ DUK_INTERNAL duk_ret_t duk_bi_duktape_object_act(duk_context *ctx) {
 	duk_int_t level;
 
 	level = duk_to_int(ctx, 0);
-	duk_inspect_callstack_entry(ctx, level);
+	(void) duk_inspect_callstack_entry(ctx, level);
 	return 1;
 }
 


svaarala:
I'm a bit on the fence if this change actually makes sense. The calls map 1:1 to `Duktape.info()` and `Duktape.act()` and the return value from those Ecmascript functions is exactly the value pushed on the value stack. So in essence the C API calls would then have two return values: one on the value stack, and another as a direct boolean return value. These are also called very rarely so they don't need to be very convenient; and adding footprint for making these rare calls more convenient seems like footprint better used elsewhere.

But I'll add the return values to the pull and see what the result would be.

svaarala:
For x64 defsize the footprint impact is +48 bytes. Seems tiny but also the change (and improvement in API usefulness) is tiny; this would be one third of the object coercion changes footprint impact. Hmm...

fatcerberus:
I 'm assuming the extra footprint is the cost of (at least) one extra instruction to load a 0 or 1 for each return.  Hm.

svaarala:
There are multiple exit paths and multiple loads, and generally functions returning a value will cause register allocation changes in the call site.

But I guess that's ultimately secondary to API consistency. On that front, e.g. `duk_get_finalizer()` will just push the current finalizer to the value stack exactly like `Duktape.fin()`, and doesn't provide a return value to indicate if a finalizer existed or not. While there are examples to the contrary (like the property functions you commented on earlier), these "rare" API calls like finalizer access, prototype access, etc are generally as minimal as possible.

So, don't know :)

fatcerberus:
I'll live if you choose not to merge the change, I have maybe four call sites total for these so it's pretty rare in practice as you say.

fatcerberus:
Thinking about `duk_get_prop_xxx()` a bit, I guess the difference there is that a property get might either *succeed* and push `undefined` (because that's the value of the property), or *fail* and push `undefined` (because the property doesn't exist).  So the only way to distinguish the two cases is by the return value.  `duk_inspect_xxx()` will only push `undefined` in the failure case, so there's no ambiguity.

svaarala:
That's true. Btw, even for `duk_get_prop_xxx()` and `foo.prop` there is a difference in handling non-existent properties. Ecmascript code generally doesn't care about the difference of the property not existing or existing with the value `undefined`. The end result is the same unless you explicitly check for the existence. So if C code mirrors Ecmascript code, it "shouldn't" use the return value but rather act on the pushed value.

So... If the return value is thought of as a shorthand for "read back `undefined`", it should actually be `false` when the pushed value is `undefined` (regardless of property existence) and `true` otherwise.

I'm not suggesting that change though, but it's an interesting question how closely the C API should mirror typical Ecmascript code behavior.

fatcerberus:
Ecmascript *does* have the `in` operator though so it's essentially shorthand for `"prop" in foo` followed by a property read--which is quite convenient and something that I often do in JS as well as C.

svaarala:
Sure, but my point was about the common idiom. How often do you see something like:

    x = foo.prop;
    if (x !== undefined) {
        // ...
    }

or:

    x = foo.prop;
    if (x != null) {
        // ...
    }

compared to:

    if ('prop' in foo) {
        x = foo.prop
        // ...
    }

Very rarely will application code actually bother to check for existence in my experience.

(And to be really technical, doing an "in" check and then reading the property is not 100% same as `duk_get_prop_xxx()` because there are no guarantees for the property not to disappear in the middle, e.g. with a Proxy target.)

fatcerberus:
I don't know about other real world code, but I know that I personally do existence checks a lot because if I forget to initialize a variable somewhere I don't want it to just silently be coerced to a default value later - I would prefer it to fail noisily.  Thus  I make a distinction between "property not present" vs. "present but set to undefined".

svaarala:
I'm not saying that's good or bad, but code bases I've worked with almost never do that. Maybe some other code bases do, shrug :)

svaarala:
Also based on previous discussions re: null and undefined, why not just:

    x = foo.prop;
    if (x !== void 0) {
    } else {
        // aiee, property is missing
    }

If "null" is used exclusively for a present-but-not-specified value (and "undefined" for a missing value, as I understood you prefer), "undefined" is actually never intentionally used for a property that's present but doesn't have a value?

fatcerberus:
For public-facing APIs I can't guarantee someone didn't implicitly pass me an `undefined` due to an accidentally uninitialized variable, for instance.  The main thing I try to do is avoid confusing "accidental" with "deliberate".  If I get a `null` that indicates intent, versus undefined which might be intended or not.  Since JS has two null types, only one of which is implicit, might as well take advantage of it :)

API design is really my passion, even though most of my Duktape work has been with internals.  So I guess I tend to think about these things differently than most.  I rarely view code as being a closed system, for example.

svaarala:
By the way, in Lua the equivalent of undefined/null is a single `nil` type which is more like "undefined" than "null". In particular, you delete properties by saying:

    x.foo = nil

which would be roughly the same as:

    delete x.foo;

If Ecmascript treated "undefined" as a missing property, it would actually make sense to delete properties by saying:

    x.foo = undefined;

That would actually make the role of "undefined" a bit more useful. Now there are essentially three null-like values: null-and-present, undefined-and-present, undefined-and-not-present.

svaarala:
> If I get a null that indicates intent, versus undefined which might be intended or not. Since JS has two null types, only one of which is implicit, might as well take advantage of it :)

But doesn't that then mean that you would want to provide that warning log entry for *both* a missing property and an "undefined" property -- because "undefined" wouldn't necessarily signal intent?

fatcerberus:
True, and now that I think about it, probably.  In that regard Lua's behavior makes more sense - nil properties aren't actually stored so there's no need for a separate existence check.

svaarala:
FWIW I'm not really sure what idioms are best, I'm just commenting on what I typically see. For example, one often sees:

    var timeout = timeout || 60;  // default to 60

which is error prone because a 0 would be treated as false and be converted to the default value. What I usually use for that is:

    var = (typeof timeout !== 'undefined') ? timeout : 60;

This doesn't (and cannot) differentiate between an actual "undefined" and a missing argument. One could make that distinction via the "arguments" object, but it's way too verbose for my taste. And I don't usually care about missing vs. undefined in this context.

Similarly for properties I typically use `typeof foo.prop !== 'undefined'`. I guess `"prop" in foo` would be be better, but I'm used to typeof because it also matches how I check arguments.

But again, I don't know if these are common or good idioms.

fatcerberus:
Wait, I know what I was referring to now.  

Suppose I have a function that takes an object with options.  If any options aren't present, the function uses defaults for them.  If they ARE present but set to undefined, then I warn: this was likely caused by an uninitialized variable.  I can't make that distinction with only a value comparison.

svaarala:
If one wants to treat a missing property and "undefined" the same way, it would actually be simpler if `duk_get_prop_xxx()` return value would be exactly the same as `!duk_is_undefined(ctx, -1)` after the property read. This allows:

    if (duk_get_prop_string(ctx, -3, "prop")) {
        /* obj.prop is present and not undefined */
    } else {
        printf("aiee: obj.prop is missing or undefined\n");
    }

Right now, if one wants to treat missing and undefined the same way, one must check for `duk_is_undefined()` explicitly.

It would also be possible to take the `DUK_TYPE_NONE` into actual use to signify a missing value for property reads. IOW, `duk_get_prop_xxx()` would push a "none" if the property doesn't exist. The downside of that is that `DUK_TYPE_NONE` is then a concrete value that can exist on the value stack and needs to be handled everywhere (probably defaulting to whatever behavior "undefined" has), and also that the C API return type won't match an Ecmascript property read type.

svaarala:
>  If they ARE present but set to undefined, then I warn: this was likely caused by an uninitialized variable. I can't make that distinction with only a value comparison.

That will then warn for this idiom:

```js
var timeout;
if (something) {
    timeout = 60;
}
doSomething({ timeout: timeout, ... });
```

The variable is intentionally uninitialized, and only set if there's a reason to do so. The fix would be to set `var timeout = null` but I wouldn't naturally do so because "undefined" is a valid value :-) But I do get your point though.

For comparison, in Lua that `timeout: timeout` would be a missing property if `timeout` was nil. That works out in these cases well too. The problem with Lua is actually the opposite: there's no way to signal an intentional null because a `nil` is exactly the same as absence. That's sometimes very awkward because there's no way to e.g. set a value to null and prevent inheritance of a non-null value.

fatcerberus:
Right, I would personally initialize `timeout` to `null` in that case as a clear signal to readers that it may be accessed without being assigned.  So the warning is desirable to me in that case :)

svaarala:
So with that sidetrack I can maybe reword my point a bit more clearly: I'm not sure what the best idiom (or even the most common one) for Ecmascript code is. So I'm also not really sure what the best C API idiom - it should ideally dovetail with the best/preferable Ecmascript convention.

In general one approach that may not be ideal but is at least conceptually simple is for C APIs to mirror their Ecmascript counterparts. For property reads that would mean that only the final returned value would be considered, without divulging whether the property was present or not.

That presence information is, by the way, not really always even available. For example, proxies return `undefined` for both missing properties and properties with an undefined value. You can't determine the presence based on the property read alone. You can of course do an "in" check but that's different from a side effect perspective -- and it would be *really* surprising if the C API property reads triggered a "has" trap for proxies because the Ecmascript property reads don't.

svaarala:
So as a concrete example:

```
duk> target = {}; P = new Proxy(target, { get: function () { return void 0; } });
= {}
duk> typeof P.foo
= "undefined"
```

If I gave "P" as an argument to a Duktape/C function, every `duk_get_prop_xxx()` read would return 1 (i.e. "property present"). So with proxies the distinction between not present and undefined already breaks down.

fatcerberus:
Huh.  That's interesting but I guess it makes sense from the perspective of Ecmascript semantics.

Ultimately I guess it's because as a C coder first, because `undefined` is the implicit value for anything that hasn't been assigned yet, it's natural for me to treat it as garbage.

svaarala:
Right, but that's really my main point: is it useful for the C API to provide information about the presence or non-presence of the property if (a) that's not generally used by script code (up to debate of course) and (b) that information is not always available or reliable?

In other words, shouldn't non-present and "undefined" just be treated the same way?

svaarala:
And that question is actually quite opposite to the question for the inspect API calls. For the inspect calls the "non-presence" is very well defined: for inspecting a value, an invalid index is "not present", for inspecting a callstack entry, and "out of bounds" index is also clearly "not present".

fatcerberus:
Hmm, going outside of my own little bubble for a bit, I guess that seems like the most reasonable default given typical idioms.  I would be a bit annoyed having to change the call sites in minisphere but I'm just one single user after all ;)

svaarala:
Sorry, this sidetrack is a bit long already, but that Proxy example actually brings up a related point: the current `duk_get_prop_xxx()` return value was originally added before Proxies were supported (or even fully specified). The return value makes sense for a concrete property lookup.

However, with proxies that return value is essentially bogus. And because Proxies should be relatively transparent compared to a plain object, this actually works quite badly for C code now. Suppose I called a C function like this:

```js
myNativeFunction(new Proxy({
    host: 'foo.bar', port: 8080, timeout: 30
}, { /* ... traps ... */ }));
```

If that C function now checked for presence of options in the call argument, all options would read back as "present with value X" or "present with value undefined". So the C call could never determine that a parameter was missing, unless it used `duk_has_prop_xxx()` explicitly.

So... the best practice for dealing with objects in C code would in fact be to generally ignore the `duk_get_prop_xxx()` return value, or risk unintended behavior with Proxy arguments. Given that, it might make more sense to remove the return value or change its semantics (= true if not undefined, false if undefined, regardless of presence).

svaarala:
Added #1170 to mention the Proxy behavior in documentation. There's no similar problem for any of the other property operations or Proxy traps; the return values map cleanly 1:1 otherwise.

fatcerberus:
Sorry, I always managed to sidetrack Duktape pulls with tangential stuff :)

svaarala:
Well, the conclusion for `duk_get_prop_xxx()` is actually useful at least for documentation.

svaarala:
I'll close this as unmerged; these are very rarely called and the added convenience doesn't seem worth it. Also the Ecmascript variants of these bindings don't provide a "present or not" indication; that's apparent from the return value.

PR #1166, opened 2016-12-09
Add internal duk_remove_m2() helper for footprint

svaarala:
Add a helper for `duk_remove(ctx, -2)` which occurs enough many times internally to benefit from a helper. Footprint saved is around 100 bytes.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2217,7 +2217,8 @@ Planned
   _Formals array when it is safe to do so (GH-1141); omit duk_hcompfunc
   _Varmap in more cases when it is safe to do so (GH-1146); reduce initial
   bytecode allocation in Ecmascript compiler for low memory targets (GH-1146);
-  packed arguments for some internal helper calls (GH-1158)
+  packed arguments for some internal helper calls (GH-1158); misc internal
+  helpers to reduce call site size (GH-1166)
 
 * Internal change: rework shared internal string handling so that shared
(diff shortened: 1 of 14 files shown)


svaarala:
x64 defsize is -150 bytes.

fatcerberus:
This is a much more common idiom than one would intuitively guess.  I find myself doing `duk_remove(ctx, -2)` a lot too.

svaarala:
The other values in the code base are: 0, -3, -4, and a variable `idx_args`.

The -2 is related to the idiom of replacing the stack top with another value:

```
if (!duk_is_string(ctx, -1)) {
    duk_push_string(ctx, "replacement");
    duk_remove(ctx, -2);
}
```

Maybe there's potential for same aptly named public API function?

svaarala:
Sometimes footprint work is unintuitive: the change to use packed arguments for internal property helper calls and the `duk_remove(ctx, -2)` shorthand together account for close to 1kB of footprint improvement:

duk-graph-size-x64

What those changes basically do is remove one argument from some commonly used calls affecting maybe 150-200 call sites. That's enough to gain about 1kB, much more easily than by e.g. optimizing specific built-in objects.

So I'll hunt for common call patterns a bit in the internals, and there may be opportunities for public API calls there too. Code policy check should also point out that a certain idiom (say, `duk_remove(ctx, -2)`) has a footprint-preferred variant because they're somewhat unintuitive and hard to remember.

PR #1169, opened 2016-12-10
Prefer #if defined() over #ifdef for consistency

svaarala:
Add code policy check to prefer `#if defined()` and `#if !defined()` over `#ifdef` and `#ifndef` for consistency and (at least to me) better readability.


fatcerberus:
I'm inclined to agree re: readability.  One thing I still prefer the abbreviated syntax for though is include guards.  #ifndef-#define-#endif is immediately recognizable as an idiom.  The long form clouds it a bit.

svaarala:
Yep, competing concerns here: consistency and easy automatic checking vs. mixed use with different idioms for different cases.

PR #1173, opened 2016-12-11
Misc footprint optimization by reducing call site argument count

svaarala:


fatcerberus:
The more of these I see, I'm a bit surprised that a size-optimizing compiler isn't able to figure some of these out on its own.

svaarala:
Size optimization doesn't seem to be a major concern for most compilers, as opposed to speed optimization. GCC apparently does support optimizing for constant call arguments, but it doesn't seem to work very well at least with `-Os`.

svaarala:
By the way, very early on I went through around 5 different GCC major/minor versions and compared `-Os` output. There was over 10% difference between the extremes, with sometimes newer versions producing much larger code. So if I had to guess, it's not a high priority and not necessarily a regression for output to increase in size significantly. I would assume a 10% performance loss would be considered a significant regression and perhaps be at least close to a release blocker.

fatcerberus:
Makes sense, most projects aren't as concerned with footprint as Duktape is. :)

svaarala:
But then again, gcc and clang are both used on many low memory targets, so it's not necessarily always a good match.

PR #1178, opened 2016-12-12
Align RegExp.prototype and instance properties with ES6

svaarala:
- [x] RegExp.prototype is no longer a RegExp instance
- [x] .source, .global, .ignoreCase, .multiline are inherited getters instead of own properties
- [x] .sticky, .unicode are new in ES6 (return false for now)
- [x] Change .sticky and .unicode to be absent until implemented (otherwise may interfere with feature detection)
- [x] .flags is new in ES6, implement
- [x] RegExp constructor looks up .flags rather than .global, .ignoreCase, .multiline when argument is a RegExp instance
- [x] RegExp constructor allows a RegExp instance and an (overriding) flags argument
- [x] RegExp constructor non-constructor call handling (maybe future work) #1179
- [x] RegExp .flags implementation future work (should do explicit property accesses which matters if getters are replaced) #1179
- [x] RegExp.prototype draft ES8 fixes
- [x] Testcase coverage
- [x] Test coverage for draft ES8 fixes
- [x] The .flags getter should read other flag fields explicitly using property reads. This matters if the getters have been replaced. Easy to demonstrate by replacing the getters with logging variants.
- [x] For some reason V8 and Firefox won't TypeError for e.g. `RegExp.prototype.source`. This seems incorrect because `.source` should TypeError if the `this` binding doesn't have RegExp internal slots. With RegExp.prototype now a plain object (and explicitly without RegExp internal slots) a TypeError should result. This should be double checked because it's quite possible V8 and Firefox are correct. => Draft ES2017 behavior.
- [x] The .toString() method is generic and should work on any object; however, it reads `.source` and `.flags` which should (again) cause a TypeError when applied to RegExp.prototype. As a result, `String(RegExp.prototype)` should be a TypeError as far as I can see. => Draft ES2017 behavior.
- [x] 2.0 migration notes
- [x] Releases entry
- [x] Migration notes and releases changes for draft ES8 fixes

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1945,6 +1945,13 @@ Planned
   trap; because "getOwnPropertyDescriptor" trap is not yet supported, the
   check is always made against the target object (GH-1115)
 
+* Align RegExp.prototype behavior more closely with ES6: .source, .global,
+  .ignoreCase, .multiline are now inherited getters; .flags, .sticky, and
+  .unicode have been added (they are inherited getters too); constructor
+  behavior has been revised for ES6 behavior; however, leniency to allow
+  e.g. RegExp.prototype.source (from ES2017 draft) is supported for real
+  world code compatibility (GH-1178)
+
 * Remove no longer needed platform wrappers in duk_config.h: DUK_ABORT(),
   DUK_EXIT(), DUK_PRINTF(), DUK_FPRINTF(), DUK_FOPEN(), DUK_FCLOSE(),
   DUK_FREAD(), DUK_FWRITE(), DUK_FSEEK(), DUK_FTELL(), DUK_FFLUSH(),
--- a/doc/release-notes-v2-0.rst
+++ b/doc/release-notes-v2-0.rst
@@ -10,6 +10,9 @@ Main changes in this release (see RELEASES.rst for full details):
 * Improve buffer bindings: plain buffers now behave like ArrayBuffers,
   and Duktape.Buffer has been removed with ArrayBuffer taking its place.
 
+* Many built-in behaviors have been aligned with ES6 or ES7, so there are
+  small behavioral changes throughout.
+
 * FIXME
 
 The release has API incompatible changes, see upgrading notes below.
@@ -1086,6 +1089,11 @@ Other incompatible changes
   follow more lenient ES6 coercion semantics: non-object arguments are either
   coerced to objects or treated like non-extensible objects with no own
   properties.
+
+* RegExp.prototype follows ES6 behavior more closely: it is no longer a RegExp
+  instance, .source, .global, .ignoreCase, and .multiline are now inherited
+  getters, etc.  However, leniency to allow e.g. RegExp.prototype.source (from
+  ES2017 draft) is supported for real world code compatibility.
 
 * Duktape.info() now returns an object rather than an array.  The object has
   named properties like ``.type`` and ``.enext`` for the internal fields which
--- a/src-input/builtins.yaml
+++ b/src-input/builtins.yaml
@@ -1774,7 +1774,7 @@ objects:
         attributes: ""
 
   - id: bi_regexp_prototype
-    class: RegExp
+    class: Object  # Object in ES6; RegExp in ES5
     internal_prototype: bi_object_prototype
     bidx: true
     present_if: DUK_USE_REGEXP_SUPPORT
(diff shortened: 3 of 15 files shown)


svaarala:
What I don't understand is why both Firefox and V8 accept this:

```
> RegExp.prototype.source
'(?:)'
```

The `.source` getter should TypeError if the target isn't an object (not triggered here) or doesn't have RegExp internal slots (should trigger here because RegExp is a plain object and is explicitly described not to have internal slots).

Specifically:
- http://www.ecma-international.org/ecma-262/7.0/#sec-properties-of-the-regexp-prototype-object: "The RegExp prototype object is an ordinary object. It is not a RegExp instance and does not have a [[RegExpMatcher]] internal slot or any of the other internal slots of RegExp instance objects."
- http://www.ecma-international.org/ecma-262/7.0/#sec-get-regexp.prototype.source: "If R does not have an [[OriginalSource]] internal slot, throw a TypeError exception."

This is noted in #1179 for figuring out as future work.

svaarala:
To clarify, the code in this pull will TypeError for `RegExp.prototype.source` etc.

svaarala:
As a related issue, `RegExp.prototype.toString()` is generic and allows any object argument. However, it constructs the result by reading `.source` and `.flags`, and when applied to `RegExp.prototype` itself `.source` should cause a TypeError as far as I can see.

As a result, `String(RegExp.prototype)` is TypeError at least for now.

svaarala:
Weird... Clicking on task items no longer causes the item to become checked (but I can edit the text manually). Oh well :)

fatcerberus:
I'm not able to check the boxes on mobile, but it still seems to work on the full desktop site.  Huh.

svaarala:
Some of the ES6 changes seem a bit odd (or maybe I'm misinterpreting them) but overall the changes are for the better. I actually had a low memory note that `.source` etc should be getters to reduce property count on instance objects. Also the flags handling in constructor is now simplified. So overall both code footprint and RAM footprint are reduced (slightly).

fatcerberus:
Regarding `RegExp.prototype.toString()`, I found this:
https://bugs.chromium.org/p/v8/issues/detail?id=4617

Apparently the ES6 behavior broke the Web so they patched V8 to return (some of) the ES5 behavior.  I guess SpiderMonkey followed suit.

svaarala:
Thanks, that's a good find. So, it's "real world" behavior that Duktape may need to imitate for similar reasons.

svaarala:
I also do find it quite annoying that `String(RegExp.prototype)` is a TypeError.

fatcerberus:
ES8 will change the RegExp accessors to return `undefined` when called on `RegExp.prototype` instead of throwing, see this TC39 pull:
https://github.com/tc39/ecma262/pull/263

The spec change that was ultimately done:
https://github.com/tc39/ecma262/commit/1a2ca97c1949ce0d0ee64dbc40da01574f0b1cd2

Note that calling them as methods of anything *other than* `RegExp.prototype` that's not a regexp is still a TypeError.

fatcerberus:
Just to check, that `undefined` clause is still present in the latest ES8 draft:
https://tc39.github.io/ecma262/#sec-get-regexp.prototype.sticky

svaarala:
That behavior sounds best for now I think. It may still change in finalization (so technically Duktape won't be tracking the latest *released* specification... :) but that's not a big deal.

Also the current behavior in this branch to return false for `.sticky` interferes with the feature detection described in the links you provided. So both `.sticky` and `.unicode` should be absent until implemented.

svaarala:
Thanks for the detective work BTW :)

fatcerberus:
No problem, always happy to help. :)

Re: Implementing ES-draft features: That's actually how the specification process is meant to work IIRC - engines implement proposals to see what works, and only if enough engines implement a feature (and keep it), it makes it into the final version.

To be clear I'm *not* implying that Duktape should do that - we don't even implement all of ES6 yet.  Just pointing out that I don't think anyone will mind if we take the occasional bit from a draft specification. :)

svaarala:
Yeah I agree. Though I think it mostly really makes sense when the draft is fixing something that has actual meaning in the current specification version. In other words, I'd like to avoid switching trivia back and forth as drafts change behavior. Sometimes that's not possible I agree.

svaarala:
Now this pull adopts the draft ES2017 behavior for `.source`, `.global`, etc. Also `.sticky` and `.unicode` were removed so that their presence won't interfere with any application feature detection.

PR #1181, opened 2016-12-12
Guide versioning note on ES-latest tracking

svaarala:
Because Duktape tracks the latest Ecmascript specification, semantics corrections to match newer specification version can also be made in minor releases. The semantic changes are usually relatively small; if they weren't, they'd also be an issue for web Javascript.

### DIFF
--- a/website/guide/intro.html
+++ b/website/guide/intro.html
@@ -47,11 +47,17 @@ where appropriate:</p>
 <li><a href="http://www.ecma-international.org/ecma-262/5.1/">ECMAScript&#x00ae; Language Specification 5.1 Edition</a></li>
 </ul>
 
-<p>Some features from Ecmascript 2015 (E6) and Ecmascript 2016 (E7) are implemented:</p>
+<p>Duktape tracks the latest Ecmascript specification for semantics and
+built-ins (however support for ES6 and later is still incomplete), see:</p>
 <ul>
 <li><a href="http://www.ecma-international.org/ecma-262/6.0/">ECMAScript&#x00ae; 2015 Language Specification</a></li>
 <li><a href="http://www.ecma-international.org/ecma-262/7.0/">ECMAScript&#x00ae; 2016 Language Specification</a></li>
 </ul>
+
+<p>In some specific cases Duktape may follow a specification draft,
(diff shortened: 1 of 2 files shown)


svaarala:
@fatcerberus How would you find the changed description re: versioning guarantees?

What I'm trying to do is explain up front that semantic fixes are not under strict versioning guarantees as far as the minor version is concerned, but patch versions won't change semantics unless that's really necessary to fix a bug.

Hopefully a compromise like this would allow minor releases to still make sense. Because if not, almost every release would be major :-o

fatcerberus:
That sounds good. It's about what I'd expect for a scripting engine using semantic versioning.  Perhaps a bit inconvenient for those using Duktape in their backend (rather than the frontend, like minisphere), but like you said if the changes were a significant compatibility issue they'd also be an issue for the web, so there shouldn't be a problem in practice.

I do like the strict guarantees for the Duktape API by the way - it ensures I can upgrade to a new minor release and get new features without having to change a bunch of call sites.  I personally don't mind Ecmascript changes as much because it's easier to change a JS script to align compared to C code (there's no compile step for the former).

svaarala:
Ok, good to hear. From a practical viewpoint I think the best way to proceed is that:
- If a later specification adjusts semantics in a way that's unlikely to impact call sites, then the change can be made in a minor version. For example, the Object constructor coercion changes are unlikely to break code because they only affect cases that would have thrown errors in the past. While it's possible to rely on that, it's not very likely.
- If a later specification adjusts semantics in a way that is likely to impact call sites, e.g. changes success path behavior or affects a known idiom, it's probably preferable to make the change in a major version.

This is far from clear cut or ideal, but is pragmatic in the sense that minor version upgrades should be unlikely to break call sites, while major version upgrades don't have similar expectations.

fatcerberus:
Note that from what I've seen, Ecmascript itself sometimes specifies an error throw as a placeholder exactly so that implementing meaningful behavior later *doesn't* create a compatibility issue.  For example this note in ES7 for `String.prototype.includes()`:

> Throwing an exception if the first argument is a RegExp is specified in order to allow future editions to define extensions that allow such argument values.

svaarala:
Right, but it's still possible for code to depend on the error throw. One can rely on error throwing (on purpose or by accident) for validation for example.

PR #1182, opened 2016-12-13
Change NativeError (TypeError, etc) constructor to inherit from Error constructor

svaarala:
In ES5 e.g. TypeError would inherit from Function.prototype. In ES6 it inherits from the Error constructor (which inherits from Function.prototype). See http://www.ecma-international.org/ecma-262/6.0/#sec-properties-of-the-nativeerror-constructors.

- [x] Change inheritance
- [x] Testcase coverage and fixes
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2088,6 +2088,10 @@ Planned
   behavior; this only matters when tracebacks are disabled and concrete
   error instance properties are in use (GH-1153)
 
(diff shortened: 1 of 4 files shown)


fatcerberus:
This seems like a very odd semantic change: Generally constructors don't inherit from anything (other than Function.prototype).

Unless... maybe this is the way classes work in ES6 in general?

PR #1183, opened 2016-12-13
Align anonymous function handling against ES6/ES7

svaarala:
Currently an empty string `.name` is set on the function instance. Check ES6/ES7 behavior.

- [x] Change anonymous function `.name` behavior to match ES6
- [x] Change `.name` for Function constructor result
- [x] Testcase for anonymous function name in ES6
- [x] Testcase for new Function() .name
- [x] Migration notes
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2090,6 +2090,13 @@ Planned
   non-writable and non-enumerable) to match ES6 requirements; also change
   .fileName to follow the same attribute convention (GH-1153, GH-1177)
 
+* Remove anonymous function own .name property to match ES6 requirements;
+  anonymous functions inherit an empty string as their name from
+  Function.prototype.name (GH-1183)
+
+* Change functions created using new Function() to have the .name
+  "anonymous" to match ES6 requirements (GH-1183)
+
 * Make Error instance .fileName and .lineNumber properties configurable
   but non-writable and non-enumerable to match function instance property
   behavior; this only matters when tracebacks are disabled and concrete
--- a/src-input/duk_api_bytecode.c
+++ b/src-input/duk_api_bytecode.c
@@ -429,12 +429,13 @@ static duk_uint8_t *duk__load_func(duk_context *ctx, duk_uint8_t *p, duk_uint8_t
 	/* Push function object, init flags etc.  This must match
 	 * duk_js_push_closure() quite carefully.
 	 */
-	h_fun = duk_push_compiledfunction(ctx);
+	h_fun = duk_push_hcompfunc(ctx);
 	DUK_ASSERT(h_fun != NULL);
 	DUK_ASSERT(DUK_HOBJECT_IS_COMPFUNC((duk_hobject *) h_fun));
 	DUK_ASSERT(DUK_HCOMPFUNC_GET_DATA(thr->heap, h_fun) == NULL);
 	DUK_ASSERT(DUK_HCOMPFUNC_GET_FUNCS(thr->heap, h_fun) == NULL);
 	DUK_ASSERT(DUK_HCOMPFUNC_GET_BYTECODE(thr->heap, h_fun) == NULL);
+	DUK_ASSERT(DUK_HOBJECT_GET_PROTOTYPE(thr->heap, (duk_hobject *) h_fun) == thr->builtins[DUK_BIDX_FUNCTION_PROTOTYPE]);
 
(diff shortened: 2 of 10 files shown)


svaarala:
V8 behavior seems a bit odd:

```
> var anon = (1, 2, function () {});  // avoid automatic name
undefined
> Object.hasOwnProperty(anon, 'name');
false
> JSON.stringify(anon.name);
'""'
> Object.getPrototypeOf(anon) === Function.prototype
true
> JSON.stringify(Function.prototype.name);
'""'
```

So far so good: `anon` inherits from Function.prototype, doesn't have an own `.name` property, and the property lookup comes out as the empty string (same as for Function.prototype).

But then:

```
> Object.defineProperty(Function.prototype, 'name', { value: 'aiee' });
[Function: aiee]
> Function.prototype.name
'aiee'
> JSON.stringify(anon.name);
'""'
```

Setting the `Function.prototype.name` succeeds - but `anon.name` is still empty, even if it has no such own property and inherits directly from `Function.prototype`.

svaarala:
It's get a bit more weird:

```
> Object.hasOwnProperty(anon, 'name');
false
> Object.getOwnPropertyDescriptor(anon, 'name')
{ value: '',
  writable: false,
  enumerable: false,
  configurable: true }
```

Maybe there's a typo somewhere :)

svaarala:
If my test above is correct, this is probably V8 somehow "caching" the name so that a change to `Function.prototype` isn't effective, but the function instance still pretends it inherits the value from Function.prototype. But then, it returns a descriptor for Object.getOwnPropertyDescriptor() so maybe this approach is "leaking" a bit. Or maybe there's a bug in my test.

svaarala:
Deleted a few comments on weirdness I though I found in V8, but it actually seems anonymous function `.name` is a concrete own property in V8:

```
> var anon = (1, 2, function () {});
undefined
> anon.hasOwnProperty('name')
true
```

So, need to check if this is ES6/ES7 behavior. If so, current behavior is OK.

fatcerberus:
ES6 has this to say, right under the section for `.name`:

> Anonymous functions objects that do not have a contextual name associated with them by this specification do not have a name own property but inherit the name property of %FunctionPrototype%.

Either this was changed in a later edition, or V8 is out of spec here.

svaarala:
That's how I quickly read it too, I'll re-read the relevant parts tonight. Function `.name` is operated from quite a few places.

svaarala:
Firefox also has an own `.name` property for anonymous functions. Time to check the ES2017 draft.

svaarala:
There's also this:

```
> x = new Function('')
[Function: anonymous]
> x.name
'anonymous'
```

fatcerberus:
That last one is backed up by ES6, at least:
ES6 Section 19.2.1.1.1, step 29:
Perform SetFunctionName(F, "anonymous").

svaarala:
Yeah, that `new Function(...)` behavior is quite clearly specified in ES6 (also checked draft ES2017).

Just wanted to note it has been explicitly defined; it didn't use to be, and Duktape currently creates anonymous functions (with empty string name) in this case.

fatcerberus:
At work now, so just putting this here:
http://www.ecma-international.org/ecma-262/6.0/#sec-function-definitions-runtime-semantics-evaluation

svaarala:
That's the place where the actual steps indicating an anonymous function is supposed to miss a `.name` is apparent: for the case with a BindingIdentifier the SetFunctionName() specification function is called. For the anonymous case it's not called so it should remain unset.

The `.length` property is in both cases handled via FunctionCreate() -> FunctionInitialize() -> step 3 sets `.length` with attribute C (configurable, other flags unset).

Behavior in this pull should now match that outcome.

svaarala:
@fatcerberus I'll merge this tomorrow unless you come up with shortcomings :)

fatcerberus:
I checked the ES8 draft to be sure, and it has the same note as ES6:
> Anonymous functions objects that do not have a contextual name associated with them by this specification do not have a name own property but inherit the name property of %FunctionPrototype%.

Runtime semantics for anonymous function expressions likewise do not seem to invoke SetFunctionName() (but those with a binding name of course do):
https://tc39.github.io/ecma262/#sec-function-definitions-runtime-semantics-evaluation

So it looks like the behavior in this pull is correct w.r.t. Ecmascript requirements.  So no complaints from me, merge away. :)

PR #1184, opened 2016-12-13
Add internal duk_push_hstring_empty() helper

svaarala:
More readable call sites, a tiny impact on footprint.

- [ ] Rebase after #1178
- [ ] Releases entry


svaarala:
Oops, wrong merge. Need to fix in #1178.

PR #1188, opened 2016-12-14
Fix object literal getter/setter .prototype and constructability to match ES6

svaarala:
- [x] Getter/setter in object literal should have no `.prototype` property
- [x] Getter/setter in object literal should not be constructable
- [x] Simplify template -> instance function flag copying
- [x] Testcase coverage
- [x] Releases entry

There are still ES6 issues left; e.g. getter/setter `.name` doesn't match ES6 requirements (ES5 had no requirements) for example.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2106,6 +2106,9 @@ Planned
   from the Error constructor rather than Function.prototype directly as
   required by ES6 (GH-1182)
 
+* Change object literal getter/setter to match ES6 requirements: no automatic
+  .prototype property, and the functions are non-constructable (GH-1188)
+
 * Add a fastint check for duk_put_number_list() values (GH-1086)
 
 * Remove an unintended fastint downgrade check for unary minus executor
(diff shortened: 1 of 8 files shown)


fatcerberus:
@svaarala How difficult would it be to implement ES6 method syntax?  e.g.

```js
obj = {
    func1() { console.log("func1"); },
    func2() { console.log("func2"); }
}
```

If it's not too much work I might look into implementing it myself.

svaarala:
Detecting the case should be a matter of finding an identifier and an opening parenthesis so no difficult long lookups or such. Then one would compile a function similar to the get/set case. It may be that the function parsing helpers in duk_js_compiler.c would need some adjustment for this case. Finally, if there are any semantics differences to the longhand, those need to be handled. The function objects generally don't remember their context after compilation, so any behavioral stuff must be encoded by their flags and properties.

svaarala:
@fatcerberus I'll do a quick test on that method syntax shortly.

svaarala:
The main problem is indeed that the function parser expects a lexing state is one step too forwards when parsing the function property shorthand. That should be fixable. It's probably good to add the plain variable property case too at the same time.

svaarala:
The method definition shorthand prototype will be in #1190 (still a work in progress).

svaarala:
Ok, this is ready to be merged. I'll do that in a few hours and then see if I can complete #1190.

PR #1190, opened 2016-12-14
Implement ES6 object literal identifier reference, method definition, remove dup check

svaarala:
- [x] Remove duplicate key check
- [x] Add identifier reference property shorthand
- [x] Add method definition shorthand
- [x] Fix method definition shorthand .prototype, constructability, and .name visibility
- [x] Rewrite parts of object literal parsing to reduce duplication and improve structure
- [x] Resolve FIXMEs
- [x] Make ES6 shorthands conditional to `DUK_USE_ES6` for now (no separate config option)
- [x] Testcase for duplicate key handling (side effects, overwriting behavior etc)
- [x] Testcase for identifier reference shorthand
- [x] Testcase for method definition shorthand (cover function .name, name binding)
- [x] Testcase for object literal shadowed by a non-writable and non-configurable ancestor
- [x] Testcase for computed name in a method definition shorthand
- [x] Testcase coverage for string and number in method definition shorthand
- [x] Testcase fixes for removal of duplicate key check
- [x] Cover "this" binding for getter/setter and method shorthand
- [x] Releases entry


svaarala:
@fatcerberus I'll experiment with implementing the method definition shorthand here. That's not yet implemented, but key duplicate check and identifier reference shorthand (in a very hacky way) is implemented. I'll probably rewrite parts of the object literal parsing function: it contains unnecessary duplication and should be possible to simplify a bit now.

fatcerberus:
Thanks, you didn't have to go to all this trouble but I appreciate it :)

svaarala:
No problem, with these changes ES6 object literal changes should be close to complete. It also reduces footprint (at least at present) because the complicated duplicate key logic can be eliminated. The duplicate key logic in ES5 is a bit complicated because it needs to reject duplicates, but allow a setter and a getter to be paired in separate steps.

fatcerberus:
The identifier reference shorthand is a bit strange, I actually had to read the testcase to understand it and even then it took a few seconds to sink in what was happening.

svaarala:
I find it a bit difficult to read (though perhaps that's just because I'm not used to it). I guess it could be useful in e.g. this idiom:

```js
var host = 'foo.bar.com'
var port = 8080;
var timeout = typeof timeout === 'number' ? timeout : 60;
net.connect({ host, port, timeout });
```

fatcerberus:
@svaarala Let me know when this is stable for testing (no rush), the current branch is causing segfaults for me.

svaarala:
I accidentally deleted a critical line in a rebase which caused all tests to fail. But that's now fixed and repushed. It's not necessarily fully stable now but should work enough to do simple testing. Maybe you pulled that broken version?

The main area of work is to write a proper implementation of the compiler parts. I wanted to create an initial hacky version quickly so that I could add the test coverage etc first.

svaarala:
Reworked the object literal parsing structure a bit. The two level loop structure is now a single level and shared helpers reduce code footprint (and reduce repetition) a bit.

svaarala:
The reworked structure which includes the new syntax is about 100 bytes smaller than previous, which is nice :)

fatcerberus:
In that case I'll help out with testing :)

svaarala:
@fatcerberus Let me know what happens if you have time to try this out. I'm still not finished and there may well be some fallout (there's a bunch of FIXMEs I need to think through). The method property syntax is not yet fully correct; the methods have a name binding, .prototype, and they're constructable (they should have none of these).

svaarala:
So to be clear - it'd be good to smoke test this. But detailed testing doesn't yet make much sense because some functionality is still incomplete. It won't be difficult to finish, it's mainly about some annoying plumbing/flags that need to be carried around the compiler.

fatcerberus:
Interesting, according to MDN computed keys work as method names:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions

V8 agrees:
```
D:\src\duktape\ms>node
> var key = "foo";
undefined
> var obj = { [key]() { console.log("called!"); } }
undefined
> obj
{ foo: [Function: foo] }
> obj.foo()
called!
undefined
>
```

svaarala:
I mean this:

```js
> obj = { foo() { console.log(typeof foo); } };
{ foo: [Function: foo] }
> obj.foo();
undefined
```

svaarala:
Re: that key example, that is supported by ES6:
- The MethodDefinition case is covered in http://www.ecma-international.org/ecma-262/6.0/#sec-method-definitions
- The syntax for it uses PropertyName production for the name slot
- PropertyName is defined in http://www.ecma-international.org/ecma-262/6.0/#sec-object-initializer
- The definition is LiteralPropertyName or ComputedPropertyName

This isn't implemented in the pull and I probably won't add it here because the name is dynamic and the compiler doesn't handle such cases now.

svaarala:
Added a (failing, known issue) testcase for the computed name in method definition case.

fatcerberus:
@svaarala I tested this out by changing a few function properties to methods in my game code, seemed to work okay.  I haven't tested the identifier shorthand yet, I'm sure I can find something that will benefit from it though.

svaarala:
Well here's a use case for computed method name:

```
$ nodejs
> obj = { [Symbol.for('aiee')] (a,b) { console.log(a,b); } };
{}
> obj[Symbol.for('aiee')](2,3)
2 3
```

That's actually something that would be useful in practice.

svaarala:
Ok, I think I now have the necessary flag plumbing done so this should be functionally complete. I also reworked the code to use a few more helpers. I'll read through the diff before merging but let's see what the tests think.

svaarala:
There are a few test coverage items also still to fix before merging. RELEASES entry also had a typo (fixed but unpushed).

svaarala:
Final footprint tuning: made the ES6 object literals conditional to `DUK_USE_ES6` (didn't want to add a specific define for it... maybe later) because they're not needed in ES5-ish low memory targets.

fatcerberus:
I assume the duplicate property check is removed even in the ES5-ish build?

svaarala:
Right, with the ES5-ish approach it makes no sense to include the check.

svaarala:
Effect on x86 minsize is around -350 bytes which is great :)

svaarala:
Had a bug where number and string wasn't accepted for method definition; now fixed:

```
duk> var obj = { 1(a,b) { print(a,b); } }; obj['1']('foo', 'bar');
foo bar
= undefined
duk> var obj = { "mystring"(a,b) { print(a,b); } }; obj.mystring('foo', 'bar');
foo bar
= undefined
```

fatcerberus:
@svaarala Thanks again for the work on this, this will come in handy. :)

svaarala:
That's good to hear :) This was a nice win-win because the footprint is smaller (also RAM because the duplicate key map isn't needed) and functionality is improved. There's probably a few easy kilobytes in the compiler code waiting to be optimized ;)

fatcerberus:
Found a bug :-(  It looks like Duktape is mixing up accessors with plain properties in some cases.

```
	var obj = {
		getMusicFilename: function() {},
		load: function() {},
		fadeIn: function() {},
		fadeOut: function() {},
		update: function() {},
		musicPath: "~/music/",
		play: function() {},
		playSound: function() {},
		forceFinish: function() {},
		setMaxMusicVolume: function() {},
		setMaxSoundVolume: function() {},
		setVolume: function() {},
		stop: function() {},
		save: function() {},
		get done() {},
		get volume() {},
		get svolume() {},
		get mute() {},
		set mute(value) {}
	}
	obj.setMaxSoundVolume();
```

Result: `TypeError: undefined not callable`.

What the object looks like to SSJ Blue:
Imgur

svaarala:
Thanks, I think I can reproduce this. Seems odd how that could happen so I'm sure the cause will be equally interesting :-)

svaarala:
An assert is triggered:

```
*** FATAL ERROR: assertion failed: comp_ctx->curr_func.is_setget == ((flags & DUK__FUNC_FLAG_GETSET) != 0) (duk_js_compiler.c:7434)
```

I should be able to fix this with the repro case - thanks!

fatcerberus:
What's interesting to me is that a data property was interpreted as an accessor *before* any getter/setter is even encountered.

fatcerberus:
@svaarala Let me know what you find out, I'm interested in what caused this bug too. ;)

svaarala:
The core issue is seen in this bytecode snippet:

```
00032 080201b5   MPUTOBJ    r1, r2, 8
00033 00100a02   LDCONST    r10, c16
00034 000d039c   CLOSURE    r3, 13
00035 000201b8   INITGET    r1, r2
```

A previous run of plain key/value pairs is finished with an MPUTOBJ when a getter is encountered. The getter is loaded with an INITGET opcode. However, the LDCONST for the getter name (c16 here) is loaded to r10 instead of r2 because of a bug in temp reg management.

This should be easy to fix - interesting that it wasn't caught by anything else.

svaarala:
> What's interesting to me is that a data property was interpreted as an accessor before any getter/setter is even encountered.

That's caused by the incorrect tempreg number in the bytecode snippet above. The INITGET opcode will be using the right target value (the getter function) but a key left over in r2 from a previous MPUTOBJ. So at first setMaxSoundVolume() is initialized correctly, but it gets overwritten by the getter because of the incorrect key from r2. The leftover key is loaded by:

```
00024 000c0202   LDCONST    r2, c12
```

And:

```
; c12 "setMaxSoundVolume"
```

So mystery solved :)

svaarala:
Fix is in #1193.

fatcerberus:
Thanks, that explains why `done` didn't make it into the object: its getter got assigned to `setMaxSoundVolume` instead.

fatcerberus:
@svaarala Thanks for implementing this so quickly, I'm really loving the method shorthand :-)

svaarala:
You're quite welcome. I'm looking forward to the identifier shorthand and destructuring which will allow stuff like:

```
```js
function connect({ host, port }) {
    // ...
}

var host = 'localhost';
connect({ host, port: 8080 });
```

which is great for named parameter handling.

PR #1191, opened 2016-12-15
Add ES6 %TypedArray% intrinsic object for typed array constructor functions

### LINKED ISSUE #921: Change built-in constructor inheritance chains to match ES6
For example, `TypeError` _constructor_ inherits (internal prototype) from `Error` constructor. In ES5.1 `TypeError` inherits from `Function.prototype`. See also #918.

svaarala:
See: http://www.ecma-international.org/ecma-262/7.0/#sec-properties-of-the-typedarray-constructors.

Also fixes some builtins.yaml property attributes which didn't have an effect on the binary as they're defaulted by the RAM built-ins init code.

Fixes #921.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2115,6 +2115,9 @@ Planned
 * Allow duplicate property names in object literals as required by ES6
   (GH-1190)
 
+* Change typed array constructor chain to match ES6, e.g. Uint8Array
+  constructor inherits from intrinsic %TypedArray% constructor (GH-1191)
+
 * Add a fastint check for duk_put_number_list() values (GH-1086)
 
 * Remove an unintended fastint downgrade check for unary minus executor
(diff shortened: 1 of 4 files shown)


svaarala:
```
> Uint8Array.__proto__
[Function: TypedArray]
> new Uint8Array.__proto__()
TypeError: Abstract class TypedArray not directly constructable
> Uint8Array.__proto__()
TypeError: Class constructor TypedArray cannot be invoked without 'new'
```

This is described in ES2016 as:

> The %TypedArray% intrinsic function object acts as the abstract superclass of the various TypedArray constructors. Because it is an abstract class constructor it will throw an error when invoked. The TypeArray constructors do not perform a super call to it.

So these changes are indeed part of modeling things more as classes. But no super call is made in this case.

PR #1193, opened 2016-12-16
Fix temporary register handling for object literal getters/setters

svaarala:
See bug description in https://github.com/svaarala/duktape/pull/1190#issuecomment-267533606. This was introduced by #1190 and is not present in the 1.x release.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1904,7 +1904,7 @@ Planned
   ({ [1+2]: 'three' }), identifier shorthand ({ foo, bar }), and method
   definition shorthand ({ func(a,b) { return a+b; } }); however, computed
   name for method definition ({ ['foo' + 'bar'](a,b) { ... } }) is not
-  yet supported (GH-985, GH-1190)
+  yet supported (GH-985, GH-1190, GH-1193)
 
 * Add support for ES7 exponentiation and exponentiation assignment operators,
   e.g. "2 ** 10" evaluates to 1024, avoiding the cost of an Ecmascript call to
--- a/src-input/duk_js_compiler.c
+++ b/src-input/duk_js_compiler.c
@@ -2951,8 +2951,8 @@ DUK_LOCAL void duk__objlit_flush_keys(duk_compiler_ctx *comp_ctx, duk__objlit_st
 		                st->temp_start,
 		                st->num_pairs * 2);
 		st->num_pairs = 0;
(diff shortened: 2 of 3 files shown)


svaarala:
@fatcerberus If you have time, it'd be good to test this fixes your Minisphere issue and that no other issues surface.

fatcerberus:
I can't test until I get home tonight but I'll be sure to let you know.

svaarala:
Sure, no hurry :)

svaarala:
@fatcerberus The reverted assert (and associated assignment change) is pushed and assertion test runs pass so this should be ready to merge, assuming there are no regressions left.

fatcerberus:
Thanks, I'll test it very shortly :)

svaarala:
Good catch by the way. This was quite an interesting bug in that it wasn't detected by other regression tests which seems intuitively unlikely given how it should manifest every time a setter/getter is created after a preceding data property.

fatcerberus:
Yep, one of the benefits of maintaining an engine designed for backward compatibility with one from the late 90s-early 2000s is that I have plenty of code to test against.

fatcerberus:
Alright, I tested a bunch of different games, including the one that originally triggered the bug, and everything seems to be in working order now.

fatcerberus:
What I still think is odd is the _way_ the bug manifests.  Knowing the cause, I'd expect the immediately preceding property (i.e. just before the getter) to be overwritten, but at least in my particular case, there were about 3 other properties in between.

svaarala:
I added some bytecode examples to #1190 which explain why that happens. But to recap with a concrete example:

    { foo: 123, bar: 234 get quux() }

The correct bytecode sequence would be something like:

    NEWOBJ r1
    LDCONST r2, "foo"
    LDINT r3, 123  ; foo: 123
    LDCONST r4, "bar"
    LDINT r5, 234  ; bar: 234
    MPUTOBJ r1, r2, 4  ; set foo and bar plain properties
    LDCONST r2, "quux"
    CLOSURE r3, quux()  ; quux: getter quux()
    INITSET r1, r2
    ...

The bug was the the `LDCONST r2, "quux"` would go to the wrong register, say r6 here. Register r2 would then hold the old value, "foo" here.

So, INITSET would operate on "foo" and the quux() closure value, setting `.foo = quux`.

Depending on how many proprties are in the MPUTOBJ set, the older the incorrect getter key used would be.

PR #1197, opened 2016-12-18
Make plain buffers mimic Uint8Array instead of ArrayBuffer

### LINKED ISSUE #870: Remove ArrayBuffer number indices
Standard ArrayBuffer values don't have virtual index properties like `arrayBuf[3] = 0x98`, but Duktape currently provides them.

For consistency these indices would also need to be removed from plain buffers (which mimic ArrayBuffer instances in 2.x). However, this would make plain buffer handling a little bit awkward: the main goal of using plain buffers over ArrayBuffer and typed arrays (Uint8Array etc) is to save memory; without index properties one is forced to create e.g. a `Uint8Array` over the plain buffer to get access to the bytes. This is counterproductive to saving memory.

Options:
1. Keep virtual number indices for both ArrayBuffers and plain buffers.
2. Keep virtual number indices for plain buffers but not ArrayBuffers. This would be more in line with standard ArrayBuffer behavior but rather confusing if plain buffers had indices and ArrayBuffers didn't.
3. Remove the indices from both plain buffers and ArrayBuffers. Most in line with E6 but basically means memory efficient buffer handling is only possible in C.
4. Change plain buffers to mimic Uint8Arrays but without a `.buffer` property. Remove virtual indices and other non-standard properties from ArrayBuffer.

svaarala:
Plain buffers in master (unreleased) mimic ArrayBuffer. Change them to mimic Uint8Arrays so that it's more consistent with ES6 for them to have e.g. index properties. With this in place, it's logical to remove the ArrayBuffer custom index properties and other custom properties, #870.

Tasks:
- [x] Change plain buffer behavior to mimic Uint8Array: inheritance, virtual properties
- [x] Change the built-ins so that plain buffers work with bufferobject support disabled (the low memory configuration)
- [x] Figure out a place for `ArrayBuffer.allocPlain()` and `ArrayBuffer.plainOf()` -> for now, move to Uint8Array constructor
- [x] Decide on object coercion behavior, `.buffer` is a bit of an issue in promoting to Uint8Array object -> now coerces to Uint8Array with a `.buffer` property pointing to an actual ArrayBuffer sharing the plain buffer as its backing buffer
- [x] Make DataView argument validation strict: reject anything but ArrayBuffers (including plain buffers)
- [x] Virtual autospawning .buffer accessor for plain buffers
- [x] Make .byteLength, .byteOffset, .BYTES_PER_ELEMENT, and .buffer inherited getters (as in ES6); .length remains a virtual own property
- [x] Remove non-standard ArrayBuffer properties
- [x] Remove non-standard DataView properties
- [x] `Duktape.dec()` return type => left as plain buffer for now, until Uint8Array is more memory friendly
- [x] Change Node.js Buffer constructor to special case ArrayBuffer argument (don't create a copy) to match newer Node.js behavior
- [x] Change low_memory.yaml example to disable bufferobject support as the lowmemory example is an ES5-ish configuration
- [x] Check configuration vs. symbols pulled in
- [x] Test duk-size build with Duktape.dec() enabled manually
- [x] Testcase coverage
- [x] Testcase fixes
- [x] Grep through testcases for ArrayBuffer / plain buffer refs
- [x] Documentation fixes
- [x] 2.0 migration note fixes
- [x] Releases entry

Fixes #870.


svaarala:
I'm somewhat inclined (at least right now :) to remove the Ecmascript bindings for creating plain buffers. They're a non-standard feature which is useful in low memory targets; requiring these environments to create plain buffers from C code (which is also often necessary to create e.g. dynamic or external buffers) seems reasonable. In similar vein there's no support for creating lightfuncs from Ecmascript code.

Actually, if anything, a useful feature would be to allow `new Uint8Array(N)` to create a *plain buffer* when bufferobject support is disabled. That would take care of `ArrayBuffer.allocPlain()`, and `ArrayBuffer.plainOf()` is meaningless when bufferobject support is disabled. Ecmascript code could still create useful buffers using `new Uint8Array(N)` which is syntactically standard.

The downside of this arrangement is that there's no Ecmascript facility to create plain buffers when bufferobject support is enabled. But as said above, maybe that's not too unreasonable.

svaarala:
Another related change is `Duktape.dec('hex')` and `Duktape.dec('base64')`. Currently these produce plain buffers which may be a bit odd if full bufferobject support is enabled. In a standard environment one would expect them to return ArrayBuffers or Uint8Arrays.

fatcerberus:
Hm, one issue with the Uint8Array constructor being co-opted like that is that one would then be using constructor syntax (`new Ctor(...)`) to ultimately end up with a non-extensible primitive (plain buffers are not really objects, they only pretend to be), which might be a bit of a headscratcher.

Then again, maybe that's no more surprising than lightfuncs mimicking Function objects but not being extensible.

svaarala:
@fatcerberus I had forgotten about disabling bufferobject support in the lowmem configuration. The configuration is intended to be an ES5-ish functionality subset so they should be disabled in that example. That explains the -10kB footprint drop in the measurements :)

svaarala:
The x86 minsize configuration + non-fancy `duk` command build to about 130kB now - a plain "eval file" command builds to around 120-125kB now I think. That's still with ES5 functionality intact (though non-verbose errors etc). So my earlier goal of a 128kB minimal ES5 build should now be achieved ;)

fatcerberus:
Bufferobject support amounts to ~10kB?  Wow, I would have guessed maybe 5, since the entire debugger feature is ~15kB (according to documentation).

svaarala:
C code is surprisingly large, people often underestimate the size of functions. Btw, those estimates in config option documentation, website, etc are very rough ballparks which are not diligently updates as features evolve (it'd be pretty time consuming and difficult to do so). Maybe at some point feature size could be measured by the build automatically (e.g. by flipping a config option and comparing the binary size).

svaarala:
> Hm, one issue with the Uint8Array constructor being co-opted like that is that one would then be using constructor syntax (new Ctor(...)) to ultimately end up with a non-extensible primitive (plain buffers are not really objects, they only pretend to be), which might be a bit of a headscratcher.

That's true - though which is better: that headscratcher or a custom binding? Tradeoffs... :)

If the Uint8Array constructor approach is taken, one must be careful to see how the constructor handling deals with the result: the default instance will be thrown away in this case but the return value is not technically an object and the "replacement object" semantics only kick in if the result is an object. I'm not sure if this works out of the box, and it can certainly be made to work.

fatcerberus:
> Btw, those estimates in config option documentation, website, etc are very rough ballparks which are not diligently updates as features evolve

Yes, I know - that's why I mentioned the debugger in particular, since I know for a fact it was updated from 10kB to 15kB after AppRequest was added (and it hasn't really changed since then). :)

svaarala:
I'm not sure what you're getting at - but just checked the debugger footprint impact on x64, and it's between 15-16kB so that estimate should be pretty close.

fatcerberus:
The footprint estimate in `debugger.rst` used to say 10-15kB.  When 1.5.0 was released that was changed to say 15-20, and the debugger hasn't changed significantly since then.  I was responding to your disclaimer about the estimates not being kept up to date.  I knew that particular one was. :)

svaarala:
Ah, I see.

I think I'll replace that `ArrayBuffer.allocPlain()` with a barebones `new Uint8Array()` call. To restate my reasoning:
- If bufferobject support is disabled, the script code will in any case need to deal with them missing in some fashion.
- Using `new Uint8Array(10)` in that case has the benefit that there's no custom binding name. Further, code written to do that will in most cases upgrade gracefully to use an actual Uint8Array if one is available.
- The downside of this solution in general is that there's no way to create plain buffers from Ecmascript code when bufferobject support is enabled. But maybe that's the right direction overall, trying to minimize custom bindings and let C code expose them if they make sense.
- And as said above, even with `ArrayBuffer.allocPlain()` there's only the facility to create plain fixed buffers - not dynamic or external buffers - so the mechanism is somewhat limited as is.
- The change to return ArrayBuffers or Uint8Arrays from `Duktape.dec()` would dovetail with this change.

fatcerberus:
Ultimately I do think that change makes sense.  If the plain buffer is a Uint8Array from Ecmascript perspective, it naturally follows that you'd create one using the Uint8Array constructor.  The non-extensibility is really a non-issue since lightfuncs have the same behavior despite looking like Function objects.

This maybe reduces the burden for documentation too, since there's one less buffer type.  It's conceptually just a "lightweight Uint8Array" rather than a completely separate type.

svaarala:
> This maybe reduces the burden for documentation too, since there's one less buffer type. It's conceptually just a "lightweight Uint8Array" rather than a completely separate type.

That's actually already the case in master - plain buffers are lightweight ArrayBuffers so with both approaches there's less to document and fewer surprises w.r.t. typing.

I wonder what's the trend in general - are new APIs returning ArrayBuffers or Uint8Arrays? TextEncoder() returns Uint8Arrays for example, but there are also APIs that return ArrayBuffers.

fatcerberus:
UTF-8 is specifically defined as having 8-bit "code units", so returning a byte array there makes sense.  In minisphere, `Socket#read()` and `FileStream#read()` both return ArrayBuffers, because the type of the data is unknown.

svaarala:
To get `new Uint8Array(N)` returning a plain buffer working `duk_new()` needs a revision so that plain buffers (and actually lightfuncs too) are allowed as return values that will replace the default instance. With that change the approach works.

svaarala:
Having tried it out, I'm having some doubts about supporting `new Uint8Array(N)` with bufferobjects disabled. It pulls in Uint8Array, %TypedArray%, Uint8Array.prototype, and %TypedArrayPrototype% just to support that one call.

Also, on a longer timescale I'd like to eliminate the plain buffer type as a distinct API / internal type and focus on optimizing the memory efficiency of standard buffer types. For example, it'd be more useful to support memory optimized layouts for common cases like Uint8Array without a slice offset, or to virtualize the `.buffer` prototype of a Uint8Array so that the backing ArrayBuffer is created on-the-fly only if it's actually accessed. Optimizations like this can be complicated to implement, but they're not visible in the API which makes the external behavior easier to understand and deal with.

To support that longer term goal a useful intermediate step would be to:
- Minimize the visibility of plain buffers to Ecmascript code so that scripts would be encouraged to work with the standard types only.
- Support plain buffers only from C code. Memory conscious environments can continue doing so.
- When bufferobject support is disabled, allow C code to create plain buffers which would mimic Uint8Arrays. They'd inherit from a hidden Uint8Array.prototype (perhaps pulling in %TypedArrayPrototype% or not) so that you could add some inherited methods if you wanted to.
- Duktape.dec() would create standard Uint8Arrays with bufferobjects enabled, and plain buffers (mimicing Uint8Arrays) with bufferobjects disabled.

With this done, the next (incompatible) step would be to eliminate the plain buffer type from the C API, and support resizable/external buffers some other way, e.g. as variants of ArrayBuffer which have special abilities from C point of view but appear as standard ArrayBuffers to Ecmascript code. (The buffer push calls can remain but they'd create objects of different internal type.)

Hmm...

fatcerberus:
You know, there's still the option I mentioned in #865, where when bufferobjects are disabled then plain buffers would work like before, i.e. they don't mimic any standard buffer type.  That way you wouldn't have to pull in any constructors. :)

svaarala:
So I guess you mean they would inherit from Object.prototype as far as property lookups go - with probably virtual indices, `.length`, `.byteLength`, etc virtual own properties (without them a plain buffer would be quite useless to script code)?

svaarala:
With this pull the final buffer semantics for 2.0.0 release will be in master. I may need to use one more pull to tweak DataView virtual properties as there are a few extra properties DataView now provides over ES6. That can also be dealt with as a compliance bug in 2.1.0 (the buffers howto should caution against using the properties) however.

After that, the next big step is to rethink the buffers for the 3.0.0 release, hopefully eliminating the plain/object buffer type distinction at least as far as scripts and C API are concerned. I still want there to be a memory efficient buffer type, but there are several approaches to provide that with minimal memory usage (perhaps not as efficiently as the current plain buffers, but close) and without calling code being aware of it except during creation.

fatcerberus:
Yes, that's what I meant - just avoid pulling in the Uint8Array binding entirely and make them plain byte arrays if bufferobjects are compiled out.  It almost defeats the purpose of disabling bufferobjects if you still have to pull in half of the standard buffer bindings.

svaarala:
It's nowhere near half though. It just pulls in the almost empty objects. All the methods are absent.

fatcerberus:
Ah, that might be okay then.  Just wanted to throw my two cents in before this got merged, don't mind me :)

svaarala:
Note that I'm leaning towards that alternative too - but it's just that there isn't a dramatic difference. Out of maybe ~20 objects 4 are pulled in, and they're almost empty. Pulling in Uint8Array.prototype is pretty reasonable, but pulling in the Uint8Array constructor is a bit borderline and only relevant if you want to construct plain buffers from script code.

But, maybe it would be OK not being able to construct plain buffers from script code when bufferobject support was disabled. One would then always have to rely on C bindings to construct buffers as there would be no standard binding.

Or... the binding could be in some other object (instead of Uint8Array constructor or constructor property) but that's a bit weird too. For example, if there was a `Duktape.allocPlain()` it would be easy to provide even with bufferobject support disabled.

In this pull the custom bindings provided are: Uint8Array.allocPlain() and Uint8Array.plainOf().

svaarala:
Addition to above: not being able to construct plain buffers with bufferobject support disabled would mean less functionality than in Duktape 1.x in that situation. With Duktape 1.x the Duktape.Buffer binding was available even when bufferobject support was disabled. This is why I'm a bit on the fence removing that ability. It only affects the low memory targets though.

svaarala:
@fatcerberus One practical downside with mimicing Uint8Array as opposed to ArrayBuffer is that plain buffers are then interpreted already as views, so that when they're used as view constructor arguments they are interpreted as initializers which causes a new buffer to be created. For example:

```js
// Want a Uint16Array.
var plain = something;  // 0x11 0x22 0x33 0x44
var u16 = new Uint16Array(plain);  // -> 0x11 0x00 0x22 0x00 0x33 0x00 0x44 0x00
```

So above a 4-byte plain buffer is interpreted as the initializer [ 0x11, 0x22, 0x33, 0x44 ] which causes a 4-element (8-byte) Uint16Array to be created. When plain buffers mimicked ArrayBuffers the Uint16Array view would be over the existing plain buffer without making a copy.

To create a view on top of the existing plain buffer one now needs:

```js
var plain = something;
var u16 = new Uint16Array(Object(plain).buffer);
```

The Object coercion is necessary because a plain buffer doesn't have a `.buffer` property at present whereas the object coerced version does. I'm considering adding an inherit `.buffer` accessor which would automatically spawn the ArrayBuffer on access. This would allow:

```js
var plain = something;
var u16 = new Uint16Array(plain.buffer);
```

which would be basically standard code. However, the downside is that each access would create a new ArrayBuffer because there would be no place to store the autospawned ArrayBuffer. This approach would actually dovetail better with ES6 where `.buffer` and other typed array properties are inherited accessors in any case.

In this instance ArrayBuffer mimicking is better, but with that approach one either needs non-standard virtual index properties or plain buffers become quite useless. But overall Uint8Array mimicking still seems like the better appoach.

fatcerberus:
It might be possible to memoize the generated ArrayBuffer somehow (that might be a challenge, since you can't add real properties to the plain buffer).  I do this in minisphere for, e.g. `Font.Default` - the first access creates a Font object, subsequent accesses return a cached one.  In this way `Font.Default === Font.Default` even though it's created on demand.  Lazy evaluation, basically.

Not sure whether that's worth the additional footprint in this case though.

fatcerberus:
P.S. Merry Christmas! 🎅🎄

svaarala:
Right, but that's really the issue: where to store it? I could add a slot into `duk_hbuffer` but that is a bit too much of a change right now. Storing it outside somewhere would be very odd on the other hand.

So I think that autospawned .buffer is a reasonable compromise for now. It's only needed (in practical terms) when you want to create a view over the plain buffer (which is already a view). Creating a fresh ArrayBuffer is not ideal if one creates a dozen views over the plain buffer (and reads .buffer every time) but I think that's going to be relatively rare.

Merry Christmas to you too :-)

svaarala:
Ok, made `.byteLength`, `.byteOffset`, and `.buffer` inherited properties, and added the typed array ArrayBuffer reference as an actual struct member of `duk_hbufobj` to allow the concrete property to be missing in most cases. Plain buffers also inherit the `.buffer` accessor; the getter returns an ArrayBuffer when the argument is a plain buffer. So this now works:

```
duk> x = Duktape.dec('hex', 'deadbeef')
= |deadbeef|
duk> y = new Uint32Array(x.buffer)
= |deadbeef|
duk> y[0] = 0x12345678;
= 305419896
duk> x
= |78563412|
duk> y
= |78563412|
```

fatcerberus:
I noticed in the commit logs:
> Remove DataView non-standard virtual properties

Did DataView have the virtual indices too?

svaarala:
Yes, all `duk_hbufobj` instances had all of the virtual properties (indices, .length, .byteLength, .byteOffset, .BYTES_PER_ELEMENT).

svaarala:
I made all `duk_hbufobj` have the same property set for simplicity in Duktape 1.x. It's also not (at least technically) non-compliant because the specification doesn't generally forbid additional properties, it just mandates what must be present. Having the properties around also allows more memory conscious programming because e.g. a DataView or an ArrayBuffer doubles as a Uint8Array.

The downside is of course that application code may accidentally rely on a non-standard property. So at least in this instance removing the non-standard properties seems better overall, as it reduces the need to document the variances, caution against relying on them if portability to other engines matter, etc.

svaarala:
As a side note re: the "technically" above, it is really a bit of a grey area in practice. The interpretation that no custom properties are allowed would make every engine or at least every execution environment non-compliant, so this clearly can't be a useful way to interpret the requirements.

Also the interpretation that any additional properties are allowed can also result in absurdity. For example, if I added a custom own `.BYTES_PER_ELEMENT = 1000` to Uint8Arrays, it wouldn't technically be a compliance violation because `.BYTES_PER_ELEMENT` is an inherited getter, and the specification doesn't prohibit a shadowing own property of the same name. The result is compliant but bad.

So I think the most useful interpretation is that custom properties are OK as long as it's likely they won't interfere with practical code. For example, an additional `.BYTES_PER_ELEMENT` virtual property that had the same value as the inherited one would be OK IMO. Virtual index properties for ArrayBuffer would be OK in this sense, but it can lead to user code accidentally relying on the behavior.

fatcerberus:
The spec actually does say it's okay for an engine to provide custom stuff (even new syntax!):

> A conforming implementation of ECMAScript may provide additional types, values, objects, properties, and functions beyond those described in this specification. In particular, a conforming implementation of ECMAScript may provide properties not described in this specification, and values for those properties, for objects that are described in this specification.

> A conforming implementation of ECMAScript may support program and regular expression syntax not described in this specification. In particular, a conforming implementation of ECMAScript may support program syntax that makes use of the “future reserved words” listed in subclause 11.6.2.2 of this specification.

> A conforming implementation of ECMAScript must not implement any extension that is listed as a Forbidden Extension in subclause 16.1.

svaarala:
Right, so it would be compliant to have `Object.prototype.toJSON()` set to a function return `"TROLLFACE"`. But the engine would be pretty useless if one did so... :)

So, a practical approach allows custom properties but it must be clear that the custom properties are not likely to break common idioms.

svaarala:
Interestingly the quote allows e.g. custom syntax and custom types, but doesn't explicitly allow modification of built-in function behavior for functions defined in the specification. It would be weird if you added a custom type but none of the built-in functions could support it (via custom behavior), even when it didn't impact code not using the custom type at all. But then, 16.2 Forbidden Extensions explicitly lists a set of functions with "The behaviour of the following methods must not be extended except as specified in ECMA-402", so maybe the intent is that any other functions are free game for custom arguments, custom behaviors, etc.

svaarala:
Footprint note: the changes here are around +300 bytes for x64 default build, and roughly +/- 0 for minsize or stripsize which is good.

svaarala:
Phew, this is finally about complete. I still need to review the releases entry and migration notes. The rest of the documentation changes are in the Wiki.

The pull makes the `.buffer` reference in typed arrays an inherited getter, with the actual reference stored in `duk_hbufobj->buf_prop`. This would allow spawning the ArrayBuffer instance on-the-fly only if it is accessed. In other words, if you did:

```js
var u8 = new UintArray(10);
```

but didn't explicitly read `.buffer`, the ArrayBuffer object wouldn't get created unnecessarily. Same would apply to the C API.

While this should be transparent, I'll use a separate pull and merge that after 2.0.0 release.

fatcerberus:
That lazy evaluation behavior seems useful, especially since much practical code will create the typed view directly without an intermediate ArrayBuffer.  One of the benefits of using a plain buffer as the concrete backing store, I suppose. :)

svaarala:
I had planned to virtualize `.buffer` in any case at some point, but now with it being an inherited getter in ES6 it just got a lot easier.

This approach could also be used without a backing plain buffer BTW (and that's been my plan for 3.0.0 buffer model): a typed array can hold the necessary buffer fields itself. If a buffer gets created, they get copied over to the spawned ArrayBuffer and the accesses then flow through the ArrayBuffer.

svaarala:
Ok, this is going in once the tests pass. Then there's only small trivia left for 2.0.0 release and I should be able to complete them in a few days.

PR #1202, opened 2016-12-19
Documentation updates for initial symbol support

svaarala:
Updates to:
- [x] Guide types section
- [x] Guide internal properties section; link to symbol howto
- [x] Guide references to internal keys/strings/properties in general
- [x] API call documentation, at least property API calls, enum call
- [x] Sandboxing document


svaarala:
@fatcerberus Now with the symbol merge in master, I'm trying to add that cautionary documentation in this pull soon :)

fatcerberus:
By the by, I would have considered the current symbol support experimental even if not for the API issues: none of the specification hooks like toStringTag etc. are implemented yet.  That in itself might be surprising, so I figured it was best to leave it as opt-in for now.

svaarala:
True, though those are not really part of the symbol feature but "users" of symbols.

svaarala:
Most of the minimally required text is in place. I'll grep through mentions of "internal properties" in the Guide and the API documentation.

`duk_push_string()` has a note that the result may appear as a symbol to Ecmascript code. Maybe a similar reference should be in e.g. `duk_is_string()` and `duk_require_string()`. Though I'd like to avoid adding the note in every possible place.

fatcerberus:
I think it's important to have a short note at least in `duk_require_string()` that it will not throw for symbols.  As has been discussed to death ;) many users may find that behavior surprising.

If a user needs to know whether a string is actually a symbol then point them to `symbols.rst` for the specific format.

svaarala:
I could add a small includable `<div class=note>` that can be added to at least these:
- duk_require_string
- duk_get_string
- duk_is_string
- duk_get_type
- duk_get_type_mask

The div could say something like "Note that Symbol values are visible in the C API as strings, e.g. duk_is_string() is true (this behavior is similar to Duktape 1.x internal strings). You can distinguish symbols from ordinary strings by looking at their initial byte, see symbols.rst."?

fatcerberus:
Nice and concise, I like it. :)

svaarala:
Reworded a bit to:

> Symbol values are visible in the C API as strings, e.g. duk_is_string() is true (this behavior is similar to Duktape 1.x internal strings). Symbols are still an experimental feature. For now, you can distinguish symbols from ordinary strings by looking at their initial byte, see symbols.rst.

fatcerberus:
Out of curiosity, how does `duk_to_string()` behave when the value is a symbol?  Is it a no-op?

svaarala:
`duk_to_string()` causes a ToString() coercion which is a TypeError for symbol values. That's worth noting in `duk_to_string()` documentation.

svaarala:
Added that `duk_to_string()` note. I'm ready to merge unless there is anything else to add.

fatcerberus:
Hm, so `duk_require_string()` will accept a symbol, but `duk_to_string()` will throw.  I agree on adding a note for that, it seems like that might surprise someone otherwise.

The thing that bugs me about that is that it's maybe inconsistent - duk_require_string() means "C string required", while duk_to_string() is an Ecmascript coercion.

svaarala:
The intent of `duk_to_string()` is to invoke the Ecmascript ToString() coercion. The other alternative standard coercion would be the `String(x)` coercion (called as a normal function) which *does* accept symbols and creates the `Symbol(foo)` string in response.

I don't think there are other standard alternatives to coerce an arbitrary value to a string - and anything other than standard coercions would be surprising in a different way.

fatcerberus:
Yeah, I agree there's not really any other sane thing for it to do.  Hopefully adding that note in the API docs should be enough to prevent surprises. :)

svaarala:
> Yeah, I agree there's not really any other sane thing for it to do. Hopefully adding that note in the API docs should be enough to prevent surprises. :)

Let's hope so :) Just to keep things in perspective, I don't want to spend a lot of time cementing the documentation for the experimental symbol support, because it's quite likely the solution will change slightly for 3.0.

Issue #765, opened 2016-05-09
Deprecate DUK_DEFPROP_HAVE_XXX?


fatcerberus:
In Duktape 1.3.0 (or was it 1.4.0?), Duktape added the convenience flags `DUK_DEFPROP_SET_XXX` and `DUK_DEFPROP_CLEAR_XXX`.  In light of these, the original `DUK_DEFPROP_HAVE_XXX` variants seem redundant.  If you want to leave one of the flags as-is, you can just not pass a set or clear flag for it.  The HAVE_XXX variants are thus not needed for the 3-way logic.

svaarala:
Good point. There might be some cases where the flags are e.g. computed where the current value vs. have flag (basically a mask) might be useful.

fatcerberus:
Actually one thing I would really like is convenience flags to set/clear multiple attributes:

```
DUK_DEFPROP_SET_EWC
DUK_DEFPROP_SET_EC
DUK_DEFPROP_CLEAR_EWC
etc.
```

It gets tedious typing out ENUMERABLE, CONFIGURABLE, WRITABLE over and over again and bloats any code that does a lot of `duk_def_prop()` calls, such as environment setup code.

Likewise, we could also have:
```
DUK_DEFPROP_VALUE
DUK_DEFPROP_GET
DUK_DEFPROP_SET
DUK_DEFPROP_GET_SET
```

Incidentally this is why I tend to avoid `Object.defineProperty()` when coding in JS - it's too verbose.

svaarala:
Internally there are indeed `DUK_PROPDESC_FLAGS_{W,E,C,WE,WC,WEC}` for similar reasons. Adding similar constants seems fine to me.

Would `DUK_DEFPROP_VALUE` be the same as `DUK_DEFPROP_HAVE_VALUE`?

I agree Object.defineProperty() is verbose - but what would be the alternative, considering properties may get new attributes over time and an implementation might have custom attributes? Linear argument lists are very awkward for extensibility for example.

Since it's also a "slow path" API I think using an object to describe the requested change is actually best for API stability. This is also why I'd like to switch `duk_def_prop()` over to using an object argument (or perhaps a flag field + object argument combination) and to make it easier to work with objects to compensate for the change.

In general IMO object arguments to pass key/value arguments are great when: (1) the call is a "slow path" one (object creation etc), (2) the API is long-lived and difficult to change, (3) the API is likely to gain new features which can't be easily foreseen.

[...]

PR #1208, opened 2016-12-19
Add convenience defines to duk_def_prop()

svaarala:
See #765.


svaarala:
There are some draft flags in the pull -- all of them are not necessarily useful additions to the API, but it's something for discussion.

PR #1218, opened 2016-12-21
Make default fatal behavior to call abort() via DUK_ABORT()

svaarala:
Current behavior in master is to cause an intentional segfault from a fatal error. Even with suggestions to define a fatal error handler users often still don't do that (at least when exploring things, which is quite understandable). The segfault then causes confusion regarding memory safety, etc.

So add back `DUK_ABORT()` to `duk_config.h` and use `abort()` by default for fatal errors. There's still no I/O (stderr) on fatal error however. But the "aborted" text is more easy to associate with a fatal error. The downside is stubbing out `DUK_ABORT()` on platforms where that binding doesn't exist (bare metal systems etc).

The main reason I don't want to print to stderr is that it pulls in an I/O dependency which is otherwise entirely avoided. When debug prints are enabled, fatal errors are debug logged though.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1801,9 +1801,10 @@ Planned
   call to match revised fatal error handler function signature (GH-781)
 
 * Incompatible change: default fatal error handler (similar to panic handler
-  in Duktape 1.x) segfaults and infinite loops without calling e.g. abort()
-  (which may not be available); this behavior can be overridden by defining
-  DUK_USE_FATAL_HANDLER() in duk_config.h (GH-781)
+  in Duktape 1.x) calls abort() without printing anything to stdout or stderr
+  (which avoids unnecessary platform dependencies); this behavior can be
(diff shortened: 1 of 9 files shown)


fatcerberus:
I have actually left the segfault behavior in minisphere on purpose, because the MSVC debugger picks up the segfault and lets me see the callstack at the time of the error.  I found that I tend to lose the stack trace with a more "well-behaved" fatal handler.  This works for me because the engine is coded in a way that a fatal error should never happen (barring extenuating circumstances such as out of memory).

fatcerberus:
I think the debugger catches `abort()` as well, so adding that back is fine too :)

svaarala:
That's why I personally prefer the intentional segfault too. But it causes confusion and a lot of unnecessary questions like "I do X and Duktape crashes, so there must be a bug in Duktape".

One API change I've been thinking about is making the fatal error handler mandatory. That would remove the confusion as to the source of the error (= uncaught error in many cases). But that's another small thing that you need to get started.

PR #1221, opened 2016-12-22
Add duk_is_buffer_data() API call

svaarala:
Add `duk_is_buffer_data()` API call for reliably checking whether a value stack entry is a plain buffer or any of the buffer types, in other words: "value stack entry has data". Before this addition one could use `duk_get_buffer_data() != NULL` but there are corner cases with zero-size buffers where that method doesn't work.

- [x] Add API call
- [x] Testcase coverage
- [x] Test -UDUK_USE_BUFFEROBJECT_SUPPORT manually
- [x] API docs
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -1963,6 +1963,9 @@ Planned
   DUK_FREAD(), DUK_FWRITE(), DUK_FSEEK(), DUK_FTELL(), DUK_FFLUSH(),
   DUK_FPUTC(), DUK_STDOUT, DUK_STDERR, DUK_STDIN, duk_file (GH-787, GH-761)
 
+* Add duk_is_buffer_data() API call to reliably test whether a value stack
+  entry is a plain buffer or any buffer object (GH-1221)
+
 * Add time functions to the C API (duk_get_now(), duk_time_to_components(),
   duk_components_to_time()) to allow C code to conveniently work with the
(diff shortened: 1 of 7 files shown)


fatcerberus:
I wonder if the API shouldn't be called `duk_has_buffer_data()` instead...

svaarala:
That was my instinct as well - but it would then be inconsistent with existing type check functions. It's a grey area because while it's checking for specific types in practice (plain buffer or any buffer object) the programmer intent is closer to "does this value have a read/write data buffer?". Those two questions are now 1:1.

svaarala:
One option was something like `duk_is_buffer_like()` to signify it's checking for a range of different buffer types (and "like" because views are not actually buffers if you follow ES terminology).

fatcerberus:
Ecmascript also has a concept of ArrayBuffer "neutering".  I don't think Duktape supports that now, but assuming it did at some point in the future, would we want this API to return false then?  If so then it would no longer be strictly a type check.

svaarala:
Duktape handles neutering (plain buffer reference is NULL) in a very rough manner so the data structure is there already. But there's no API to actually neuter buffers (Khronos doesn't provide it for example) yet.

But I think a neutered buffer should probably work similarly to an "uncovered" buffer where the apparent size is larger than the backing buffer. The safe thing is to say that data pointer is NULL and size is 0, but recognize it as a buffer.

The other option of considering a non-buffer for type checking would also be reasonable though.

PR #1225, opened 2016-12-27
Spawn typed array .buffer ArrayBuffer lazily on first access

svaarala:
- [x] Remove inline creation of the ArrayBuffer when a typed array is constructed directly without using an ArrayBuffer argument (applies to C API too)
- [x] Spawn ArrayBuffer on first `.buffer` getter access
- [x] Memory test to demonstrate difference
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2369,5 +2369,9 @@ Planned
 2.1.0 (XXXX-XX-XX)
 ------------------
 
+* Spawn the ArrayBuffer object backing a typed array lazily when its .buffer
+  property is first read, reducing memory usage in common cases where the view
+  is constructed directly without needing the ArrayBuffer object (GH-1225)
+
 3.0.0 (XXXX-XX-XX)
 ------------------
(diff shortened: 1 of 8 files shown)


svaarala:
Tested a bit with valgrind massif. Test case creates 100000 times `new Uint8Array(256)`, x64 defaults:
- From master: 50.91MB (533 bytes / uint8array)
- From this branch: 40.97MB (429 bytes / uint8array)

With the .buffer autospawned, the overhead beyond the 256 bytes comes from:
- A plain buffer heap header: 40 bytes on 64-bit target. This is a fixed buffer so the heap allocation is actually header size + 256 bytes.
- A Uint8Array object: 96 bytes on 64-bit target. With .buffer virtualized the object no longer has a concrete property table.
- Libc allocator overhead.

The plain buffer and the Uint8Array object come out 392 bytes on x64; there's roughly 30 bytes of unaccounted (libc) overhead.

svaarala:
For comparison, the above test modified to create 256-byte plain buffers uses 31.44MB on x64 default config, i.e. ~314 bytes/buffer.

The size of a plain fixed buffer is 40 + 256 = 296 bytes + some libc overhead. Actual pool allocators on low memory targets can be overhead free however. On a low memory target the header overhead is also smaller (~16 bytes).

PR #1231, opened 2016-12-28
Add TextEncoder/TextDecoder to Guide

svaarala:
Add TextEncoder/TextDecoder to "Duktape built-ins" section. The section now summarizes the top level bindings not part of the Ecmascript standard.

### DIFF
--- a/website/guide/duktapebuiltins.html
+++ b/website/guide/duktapebuiltins.html
@@ -1,7 +1,7 @@
 <h1 id="duktapebuiltins">Duktape built-ins</h1>
 
-<p>This section describes Duktape-specific built-in objects, methods, and
-values.</p>
+<p>This section summarizes Duktape-specific and non-Ecmascript built-in
+objects, methods, and values.</p>
 
 <h2>Additional global object properties</h2>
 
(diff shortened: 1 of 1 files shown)


svaarala:
@fatcerberus Any suggestions on what should be covered in the Guide re: the Encoding API?

So far the approach is:
- List them in the Duktape built-ins section. Reasoning: bindings beyond Ecmascript standard bindings should be listed explicitly because they're not obvious from anywhere else.
- The Guide lists them in the global object summary and provides separate sections for them.
- Each section provides a single use case demonstrating UTF-8 and surrogate pair handling, which are the main features as far as encoding/decoding is concerned.

I didn't want to provide a lengthy explanation in the guide; I'm trying to move into this direction in general, with longer explanations in the Wiki or behind external links.

Should there be a Wiki article for Encoding?

fatcerberus:
@svaarala Sorry, didn't get a notification for that mention, GitHub is slacking off apparently. :)

But what's here looks good.  I agree that the Guide shouldn't discuss bindings at length, it's not an API reference.  Just the single paragraph explanation + short code example works well as a primer, I think.

PR #1233, opened 2016-12-28
Fix some compiler warnings for 2.0.0 release

svaarala:
Fix a few Clang and MSVC warnings. Clang warning for `if (sizeof(x) < y) {...}` remains.

Also fix DUK_UNPREDICTABLE() placeholder.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,11 +29,11 @@ build_script:
 
   # Prep a few variants of headers and sources.
 
-  - cmd: python tools\configure.py --output-directory prep-nondll --source-directory src-input --config-metadata config
+  - cmd: python tools\configure.py --line-directives --output-directory prep-nondll --source-directory src-input --config-metadata config
   - cmd: dir prep-nondll
-  - cmd: python tools\configure.py --output-directory prep-dll --source-directory src-input --config-metadata config --dll
+  - cmd: python tools\configure.py --line-directives --output-directory prep-dll --source-directory src-input --config-metadata config --dll
   - cmd: dir prep-dll
-  - cmd: python tools\configure.py --output-directory prep-cpp --source-directory src-input --config-metadata config -DDUK_USE_CPP_EXCEPTIONS
+  - cmd: python tools\configure.py --line-directives --output-directory prep-cpp --source-directory src-input --config-metadata config -DDUK_USE_CPP_EXCEPTIONS
   - cmd: dir prep-cpp
 
   # --- Visual Studio 2015 ---
--- a/config/compilers/compiler_clang.h.in
+++ b/config/compilers/compiler_clang.h.in
@@ -19,7 +19,7 @@
 #define DUK_UNLIKELY(x)  __builtin_expect((x), 0)
 #if defined(__clang__) && defined(__has_builtin)
 #if __has_builtin(__builtin_unpredictable)
-#define DUK_UNPREDICTABLE()  do { __builtin_unpredictable(); } while (0)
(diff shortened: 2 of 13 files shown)


svaarala:
One difficult warning to avoid is unreachable code when comparing `sizeof(something)` at run time:

```c
if (sizeof(time_t) < 8) {
    ...
}
```

The check is difficult to remove if the type in question doesn't have a portable limit define. Adding the limit define to `duk_config.h` creates another issue to consider for portability for little gain (there are only a few sizeof checks in the code base). Suppressing the warning means compiler specific pragmas which seems overkill. Assigning the sizeof() to a volatile variable before comparison generates unnecessary code.

Not sure yet what approach (if any) to take.

svaarala:
MSVC x64 has this:

```
duk_bi_buffer.c(2938): warning C4267: 'function': conversion from 'size_t' to 'duk_uint_t', possible loss of data
```

The warning is (currently) harmless with buffer sizes limited to <4GB, so I'll add an assert and explicit cast.

svaarala:
Or maybe actually add an internal `duk_push_size_t()`.

fatcerberus:
Doesn't that fall under the same issue as assigning the value to a variable (= unnecessary extra code)?

svaarala:
This was about a different warning; about casting a size_t to a duk_uint_t, and there's no dead code there.

svaarala:
MSVC old versions:

```
duk_bi_date.c(631) : warning C4551: function call missing argument list
```

Caused by:

```
        DUK_UNREF(duk_bi_date_timeval_in_leeway_range);
        DUK_ASSERT(duk_bi_date_timeval_in_leeway_range(d));
```

MSVC doesn't like the "unref" macro which does `(void) func_name`. But this seems to be the case only with older MSVC versions.

svaarala:
LTO builds get this:

```
/tmp/ccScKgQG.o (symbol from plugin): warning: memset used with constant zero length parameter; this could be due to transposed parameters
```

This is difficult to pin down. It's quite possibly harmless but I need to check how to find the relevant source line.

svaarala:
Remaining clang warnings are:

```
duk_bi_date_unix.c:279:7: warning: will never be executed [-Wunreachable-code]
            (parts[DUK_DATE_IDX_YEAR] < 1970 || parts[DUK_DATE_IDX_YEAR] > 2037)) {
             ^~~~~
duk_heap_alloc.c:736:3: warning: will never be executed [-Wunreachable-code]
                fatal_func(heap_udata, "sizeof(duk_tval) not 8 or 16, cannot use DUK_USE_EXEC_REGCONST_OPTIMIZE option");
                ^~~~~~~~~~
2 warnings generated.
```

Both are `sizeof()` check related.

svaarala:
Ajduk build has:

```
duk_heap_alloc.c:1010:23: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
```

Caused by:

```
        res->rnd_state[1] ^= (duk_uint64_t) (void *) res;
```

This is harmless. But I'll try to see if I can incorporate the pointer into the random state without the warning.

fatcerberus:
As a developer, I have to say that I find the `sizeof` comparison warnings to be rather annoying.  The entire reason the operator exists is because you *don't know* the size of the entity in question, otherwise you would just hardcode it.  So while sizeof may be a constant from the compiler's perspective, from the point of view of the source code (and the programmer who wrote it), it very much isn't.

svaarala:
Yeah, and adding a size/limit define for the type involved is error prone: if the define is wrong, it's hard to catch. Asserting for correctness of the size/limit define leads to the original problem :)

PR #1234, opened 2016-12-28
Add DUK_UNPREDICTABLE() placeholder

svaarala:
`__builtin_unpredictable()` is available in newer Clang versions and may be useful in performance optimizing e.g. refcounting for Clang. This pull adds `DUK_UNPREDICTABLE(x)` but doesn't add any call sites yet.


fatcerberus:
Looking at this now, this seems incorrect:

```c
#define DUK_UNPREDICTABLE()  do { __builtin_unpredictable(); } while (0)
```

Shouldn't both the macro and builtin take an argument?

svaarala:
Oops, quite right. I'll fix it. Thanks!

svaarala:
The compiler fillin did handle the parameter correctly, but not the Clang version.

PR #1260, opened 2017-01-02
Add a 'global' binding

### LINKED ISSUE #1259: Add a `global` binding, like Node.js
Node.js has a `global` binding which by default refers to the global object.  This is quite useful as it is otherwise awkward (in the general case) to get at the global object in Ecmascript.  Consider adding such a binding (maybe optional) in Duktape.

Note that there is an active proposal to implement a Node.js-style `global` binding in Ecmascript proper, presently in Stage 3:
https://github.com/tc39/proposal-global

svaarala:
Fixes #1259.

- [x] Add the binding
- [x] Config option? => No option for now.
- [x] Polyfill
- [x] Test coverage
- [x] Website documentation update (non-standard built-ins)
- [x] Search repo for 'new Function("return this")()' occurrences and see if they should reference 'global'
- [x] Wiki post-ES5+ features - https://github.com/svaarala/duktape-wiki/pull/160
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2369,6 +2369,11 @@ Planned
 2.1.0 (XXXX-XX-XX)
 ------------------
 
+* Add a "global" property to the global object to provide easy access to the
+  global object itself without needing idioms like
+  "new Function('return this')()"; implemented based on
+  https://github.com/tc39/proposal-global (GH-1259, GH-1260)
+
 * Spawn the ArrayBuffer object backing a typed array lazily when its .buffer
   property is first read, reducing memory usage in common cases where the view
   is constructed directly without needing the ArrayBuffer object (GH-1225)
--- a/polyfills/global.js
+++ b/polyfills/global.js
@@ -1,0 +1,15 @@
+/*
+ *  Duktape 2.1.0 adds a 'global' binding.  Polyfill for earlier versions.
+ */
+
+if (typeof global === 'undefined') {
+    (function () {
+        var global = new Function('return this;')();
+        Object.defineProperty(global, 'global', {
+            value: global,
+            writable: true,
+            enumerable: false,
+            configurable: true
+        });
+    })();
+}
--- a/src-input/builtins.yaml
+++ b/src-input/builtins.yaml
@@ -168,6 +168,16 @@ objects:
     bidx: true
 
     properties:
+      # 'global' binding giving easy to access to the global object.
+      # Not yet standard, see https://github.com/tc39/proposal-global.
+      - key: "global"
+        value:
+          type: object
+          id: bi_global
+        attributes: "wc"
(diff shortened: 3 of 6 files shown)


svaarala:
```
$ ./duk
((o) Duktape [linenoise] 2.0.0 (v2.0.0-dirty)
duk> global
= {console:undefined,readFile:{_func:true},writeFile:{_func:true}}
duk> global.eval
= {_func:true}
```

fatcerberus:
(Candidate) specification text:
https://github.com/tc39/proposal-global/blob/master/spec.md

svaarala:
In my Node.js version (v6.7.0) the property is enumerable but the proposal has it non-enumerable which makes more sense.

In fact it'd make most sense and be closer to some of the other ES2015 attributes if it were non-writable but configurable. That prevents accidental writes but allows intentional modification.

But, I'll implement it according to the current proposal of course.

fatcerberus:
The writable and configurable attributes are a bit surprising to me too.  I'd expect the property to be completely immutable like the `undefined` binding, i.e. a pseudo-keyword.  I guess it might make sense for Ecmascript code to replace it sometimes though, for example in sandboxing scenarios and the like.

svaarala:
I suppose that the binding is writable to allow existing code like:

```js
var global = new Function('return this;')();
```

to work without changes.

svaarala:
@fatcerberus This is merge ready from my part.

fatcerberus:
@svaarala Thanks for the heads up, I'll test this out tonight and let you know if there are any issues.

fatcerberus:
I think something's wrong with GitHub's notification system, I haven't been getting emails for a lot of pulls, even when you @ mention me.  It's not going into my spam folder either, I checked that just in case.  Very annoying.

fatcerberus:
No issues I can see in my quick tests.  I checked the property attributes in SSJ, and verified that `global` indeed maps to the global object.  Hard to imagine what else could go wrong with such a simple binding, so this looks good to me too.

svaarala:
Ok, merged. Just wanted to make sure it actually did what you requested for in #1259.

fatcerberus:
Yep, just a named reference to the global object is all. :)

I'm surprised it took this long to standardize this, honestly.  Web JS has had `window` literally since the beginning.

svaarala:
No need for that because it's easy to `new Function("return this;")().foo = 'bar'` ;-)

Mouvedia:
In light of tc39/proposal-global#20, it should be removed.

svaarala:
@Mouvedia Thanks, luckily it's not yet in a release so there's time to remove it or make it disabled by default.

Note that most Duktape users don't write web browsers, so breakage on the web is not necessarily a major issue. However, it doesn't make sense to add the binding to Duktape built-ins if it's not adopted elsewhere.

svaarala:
Opened #1440.

PR #1264, opened 2017-01-02
Add API calls for duk_get_xxx() with explicit default

### LINKED ISSUE #1263: Provide a 'duk_get_**_or' API?
What do you think of replacing the `duk_get_**` API calls with one which allows the user to provide a default value which will be returned in the case that the stack index is invalid or it is an unexpected value type. Then, simply change the current `duk_get_**` API to macros which will in turn call new API with the default argument assigned to them in the current API.

For example, the `duk_get_number` implementation could be changed to:

``` C
DUK_EXTERNAL duk_double_t duk_get_number_or(duk_context *ctx, duk_idx_t index, duk_double_t defvalue ) {
	duk_double_union ret;
	duk_tval *tv;

	DUK_ASSERT_CTX_VALID(ctx);

	ret.d = defvalue;  /* Only line changed */
	tv = duk_get_tval(ctx, index);
	if (tv && DUK_TVAL_IS_NUMBER(tv)) {
		ret.d = DUK_TVAL_GET_NUMBER(tv);
	}

	/*
	 *  Number should already be in NaN-normalized form, but let's
	 *  normalize anyway.
	 */

	DUK_DBLUNION_NORMALIZE_NAN_CHECK(&ret);
	return ret.d;
}
```

Then the current `duk_get_number` could be changed to:
``` C
#define duk_get_number(cx, index) duk_get_number_or((cx), (index), DUK_DOUBLE_NAN)
```

I'm suggesting this because to often I have to check the value type at an index first before obtaining it for optional/default arguments in my native bindings. This would simplify that and avoid branching.

Example:
``` C
// Instead of this:
static duk_ret_t play( duk_context* cx )
{
    // JS Signature play( offset:number /*= 0.0*/, loops:number /*=2*/ ):void

    float offset = 0.0f;
    int   loops  = 2;

    if( duk_is_number( cx, 0 ) )
    	offset = (float)duk_get_number( cx, 0 );

    if( duk_is_number( cx, 1 ) )
    	loops = (int)duk_get_int( cx, 1 );

    return 0;
}

// Instead of this:
static duk_ret_t play( duk_context* cx )
{
    // JS Signature play( offset:number /*= 0.0*/, loops:number /*=2*/ ):void

    float offset = (float)duk_get_number_or( cx, 0, 0.0 );
    int   loops  = (int)duk_get_int_or( cx, 1, 2 );

    return 0;
}
```

`duk_get_***_or` is not necessarily the name I'm suggesting — that's another conversation— I'm rather focusing on the purpose :)

Thoughts?

svaarala:
Fixes #1263.

- [x] Add duk_opt_xxx() API calls to match duk_get_xxx() calls
- [x] Testcase coverage
- [ ] API documentation
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2369,6 +2369,11 @@ Planned
 2.1.0 (XXXX-XX-XX)
 ------------------
 
+* Add duk_opt_xxx() API calls which behave like duk_get_xxx() but allow a
+  custom default value to be given if the value has unexpected type or the
+  index is invalid; for example: "int port = duk_opt_int(ctx, -3, 80);"
+  (GH-1263, GH-1264)
+
 * Add a "global" property to the global object to provide easy access to the
   global object itself without needing idioms like
   "new Function('return this')()"; implemented based on
--- a/src-input/duk_api_codec.c
+++ b/src-input/duk_api_codec.c
@@ -18,7 +18,10 @@ DUK_LOCAL const duk_uint8_t *duk__prep_codec_arg(duk_context *ctx, duk_idx_t idx
 
 	DUK_ASSERT(duk_is_valid_index(ctx, idx));  /* checked by caller */
 
-	ptr = duk_get_buffer_data_raw(ctx, idx, out_len, 0 /*throw_flag*/, &isbuffer);
+	/* FIXME: with def_ptr set to a stack related pointer, isbuffer could
+	 * be removed?
+	 */
+	ptr = duk_get_buffer_data_raw(ctx, idx, out_len, NULL /*def_ptr*/, 0 /*def_len*/, 0 /*throw_flag*/, &isbuffer);
 	if (isbuffer) {
 		DUK_ASSERT(*out_len == 0 || ptr != NULL);
 		return (const duk_uint8_t *) ptr;
--- a/src-input/duk_api_internal.h
+++ b/src-input/duk_api_internal.h
@@ -101,7 +101,7 @@ DUK_INTERNAL_DECL duk_hthread *duk_get_hthread(duk_context *ctx, duk_idx_t idx);
 DUK_INTERNAL_DECL duk_hcompfunc *duk_get_hcompfunc(duk_context *ctx, duk_idx_t idx);
 DUK_INTERNAL_DECL duk_hnatfunc *duk_get_hnatfunc(duk_context *ctx, duk_idx_t idx);
 
-DUK_INTERNAL_DECL void *duk_get_buffer_data_raw(duk_context *ctx, duk_idx_t idx, duk_size_t *out_size, duk_bool_t throw_flag, duk_bool_t *out_found);
+DUK_INTERNAL_DECL void *duk_get_buffer_data_raw(duk_context *ctx, duk_idx_t idx, duk_size_t *out_size, void *def_ptr, duk_size_t def_len, duk_bool_t throw_flag, duk_bool_t *out_isbuffer);
 
 DUK_INTERNAL_DECL duk_hobject *duk_get_hobject_with_class(duk_context *ctx, duk_idx_t idx, duk_small_uint_t classnum);
 
--- a/src-input/duk_api_public.h.in
+++ b/src-input/duk_api_public.h.in
@@ -645,8 +645,23 @@ DUK_EXTERNAL_DECL void *duk_get_pointer(duk_context *ctx, duk_idx_t idx);
 DUK_EXTERNAL_DECL duk_c_function duk_get_c_function(duk_context *ctx, duk_idx_t idx);
 DUK_EXTERNAL_DECL duk_context *duk_get_context(duk_context *ctx, duk_idx_t idx);
 DUK_EXTERNAL_DECL void *duk_get_heapptr(duk_context *ctx, duk_idx_t idx);
-DUK_EXTERNAL_DECL duk_size_t duk_get_length(duk_context *ctx, duk_idx_t idx);
-DUK_EXTERNAL_DECL void duk_set_length(duk_context *ctx, duk_idx_t idx, duk_size_t len);
+
+/*
+ *  Opt operations: like get operations but with an explicit default value.
+ */
+
+DUK_EXTERNAL_DECL duk_bool_t duk_opt_boolean(duk_context *ctx, duk_idx_t idx, duk_bool_t def_value);
+DUK_EXTERNAL_DECL duk_double_t duk_opt_number(duk_context *ctx, duk_idx_t idx, duk_double_t def_value);
+DUK_EXTERNAL_DECL duk_int_t duk_opt_int(duk_context *ctx, duk_idx_t idx, duk_int_t def_value);
+DUK_EXTERNAL_DECL duk_uint_t duk_opt_uint(duk_context *ctx, duk_idx_t idx, duk_uint_t def_value);
+DUK_EXTERNAL_DECL const char *duk_opt_string(duk_context *ctx, duk_idx_t idx, const char *def_value);
+DUK_EXTERNAL_DECL const char *duk_opt_lstring(duk_context *ctx, duk_idx_t idx, duk_size_t *out_len, const char *def_ptr, duk_size_t def_len);
+DUK_EXTERNAL_DECL void *duk_opt_buffer(duk_context *ctx, duk_idx_t idx, duk_size_t *out_size, void *def_ptr, duk_size_t def_len);
+DUK_EXTERNAL_DECL void *duk_opt_buffer_data(duk_context *ctx, duk_idx_t idx, duk_size_t *out_size, void *def_ptr, duk_size_t def_len);
+DUK_EXTERNAL_DECL void *duk_opt_pointer(duk_context *ctx, duk_idx_t idx, void *def_value);
+DUK_EXTERNAL_DECL duk_c_function duk_opt_c_function(duk_context *ctx, duk_idx_t idx, duk_c_function def_value);
+DUK_EXTERNAL_DECL duk_context *duk_opt_context(duk_context *ctx, duk_idx_t idx, duk_context *def_value);
+DUK_EXTERNAL_DECL void *duk_opt_heapptr(duk_context *ctx, duk_idx_t idx, void *def_value);
 
 /*
  *  Require operations: no coercion, throw error if index or type
@@ -723,6 +738,14 @@ DUK_EXTERNAL_DECL void duk_to_primitive(duk_context *ctx, duk_idx_t idx, duk_int
 DUK_EXTERNAL_DECL const char *duk_safe_to_lstring(duk_context *ctx, duk_idx_t idx, duk_size_t *out_len);
 #define duk_safe_to_string(ctx,idx) \
 	duk_safe_to_lstring((ctx), (idx), NULL)
+
+/*
+ *  Value length
+ */
+
+DUK_EXTERNAL_DECL duk_size_t duk_get_length(duk_context *ctx, duk_idx_t idx);
+DUK_EXTERNAL_DECL void duk_set_length(duk_context *ctx, duk_idx_t idx, duk_size_t len);
+DUK_EXTERNAL_DECL duk_size_t duk_opt_length(duk_context *ctx, duk_idx_t idx, duk_size_t def_value);  /* FIXME: not sure if workable */
 
 /*
  *  Misc conversion
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -76,7 +76,7 @@ DUK_LOCAL const duk_uint_t duk__type_mask_from_tag[] = {
 
 DUK_LOCAL_DECL duk_heaphdr *duk__get_tagged_heaphdr_raw(duk_context *ctx, duk_idx_t idx, duk_uint_t tag);
 
-DUK_LOCAL duk_int_t duk__api_coerce_d2i(duk_context *ctx, duk_idx_t idx, duk_bool_t require) {
+DUK_LOCAL duk_int_t duk__api_coerce_d2i(duk_context *ctx, duk_idx_t idx, duk_int_t def_value, duk_bool_t require) {
 	duk_hthread *thr;
 	duk_tval *tv;
 	duk_small_int_t c;
@@ -136,10 +136,11 @@ DUK_LOCAL duk_int_t duk__api_coerce_d2i(duk_context *ctx, duk_idx_t idx, duk_boo
 		DUK_ERROR_REQUIRE_TYPE_INDEX(thr, idx, "number", DUK_STR_NOT_NUMBER);
 		/* not reachable */
 	}
-	return 0;
+
+	return def_value;
 }
 
-DUK_LOCAL duk_uint_t duk__api_coerce_d2ui(duk_context *ctx, duk_idx_t idx, duk_bool_t require) {
+DUK_LOCAL duk_uint_t duk__api_coerce_d2ui(duk_context *ctx, duk_idx_t idx, duk_uint_t def_value, duk_bool_t require) {
 	duk_hthread *thr;
 	duk_tval *tv;
 	duk_small_int_t c;
@@ -189,7 +190,8 @@ DUK_LOCAL duk_uint_t duk__api_coerce_d2ui(duk_context *ctx, duk_idx_t idx, duk_b
 		DUK_ERROR_REQUIRE_TYPE_INDEX(thr, idx, "number", DUK_STR_NOT_NUMBER);
 		/* not reachable */
 	}
-	return 0;
+
+	return def_value;
 }
 
 /*
@@ -1157,7 +1159,7 @@ DUK_EXTERNAL void duk_xcopymove_raw(duk_context *to_ctx, duk_context *from_ctx, 
 }
 
 /*
- *  Get/require
+ *  Get/opt/require
  */
 
 DUK_EXTERNAL void duk_require_undefined(duk_context *ctx, duk_idx_t idx) {
@@ -1186,8 +1188,8 @@ DUK_EXTERNAL void duk_require_null(duk_context *ctx, duk_idx_t idx) {
 	}
 }
 
-DUK_EXTERNAL duk_bool_t duk_get_boolean(duk_context *ctx, duk_idx_t idx) {
-	duk_bool_t ret = 0;  /* default: false */
+DUK_LOCAL DUK_ALWAYS_INLINE duk_bool_t duk__getopt_boolean(duk_context *ctx, duk_idx_t idx, duk_bool_t def_value) {
+	duk_bool_t ret;
 	duk_tval *tv;
 
 	DUK_ASSERT_CTX_VALID(ctx);
@@ -1196,10 +1198,25 @@ DUK_EXTERNAL duk_bool_t duk_get_boolean(duk_context *ctx, duk_idx_t idx) {
 	DUK_ASSERT(tv != NULL);
 	if (DUK_TVAL_IS_BOOLEAN(tv)) {
 		ret = DUK_TVAL_GET_BOOLEAN(tv);
+		DUK_ASSERT(ret == 0 || ret == 1);
+	} else {
+		ret = def_value;
+		/* Not guaranteed to be 0 or 1. */
 	}
 
-	DUK_ASSERT(ret == 0 || ret == 1);
 	return ret;
+}
+
+DUK_EXTERNAL duk_bool_t duk_get_boolean(duk_context *ctx, duk_idx_t idx) {
+	DUK_ASSERT_CTX_VALID(ctx);
+
+	return duk__getopt_boolean(ctx, idx, 0);  /* default: false */
+}
+
+DUK_EXTERNAL duk_bool_t duk_opt_boolean(duk_context *ctx, duk_idx_t idx, duk_bool_t def_value) {
+	DUK_ASSERT_CTX_VALID(ctx);
+
+	return duk__getopt_boolean(ctx, idx, def_value);
 }
 
 DUK_EXTERNAL duk_bool_t duk_require_boolean(duk_context *ctx, duk_idx_t idx) {
@@ -1219,17 +1236,26 @@ DUK_EXTERNAL duk_bool_t duk_require_boolean(duk_context *ctx, duk_idx_t idx) {
 	return ret;
 }
 
-DUK_EXTERNAL duk_double_t duk_get_number(duk_context *ctx, duk_idx_t idx) {
+DUK_LOCAL DUK_ALWAYS_INLINE duk_double_t duk__getopt_number(duk_context *ctx, duk_idx_t idx, duk_double_t def_value) {
 	duk_double_union ret;
 	duk_tval *tv;
 
 	DUK_ASSERT_CTX_VALID(ctx);
 
-	ret.d = DUK_DOUBLE_NAN;  /* default: NaN */
 	tv = duk_get_tval_or_unused(ctx, idx);
 	DUK_ASSERT(tv != NULL);
-	if (DUK_TVAL_IS_NUMBER(tv)) {
-		ret.d = DUK_TVAL_GET_NUMBER(tv);
+#if defined(DUK_USE_FASTINT)
+	if (DUK_TVAL_IS_FASTINT(tv)) {
+		ret.d = (duk_double_t) DUK_TVAL_GET_FASTINT(tv);  /* FIXME: cast trick */
+	}
+	else
+#endif
+	if (DUK_TVAL_IS_DOUBLE(tv)) {
+		ret.d = DUK_TVAL_GET_DOUBLE(tv);
+	} else {
+		ret.d = def_value;
+		/* FIXME: generates unnecessary code for duk_get_double()? */
+		DUK_DBLUNION_NORMALIZE_NAN_CHECK(&ret);  /* FIXME: test coverage */
 	}
 
 	/* When using packed duk_tval, number must be in NaN-normalized form
@@ -1238,6 +1264,14 @@ DUK_EXTERNAL duk_double_t duk_get_number(duk_context *ctx, duk_idx_t idx) {
 	 */
 	DUK_ASSERT(DUK_DBLUNION_IS_NORMALIZED(&ret));
 	return ret.d;
+}
+
+DUK_EXTERNAL duk_double_t duk_get_number(duk_context *ctx, duk_idx_t idx) {
+	return duk__getopt_number(ctx, idx, DUK_DOUBLE_NAN);  /* default: NaN */
+}
+
+DUK_EXTERNAL duk_double_t duk_opt_number(duk_context *ctx, duk_idx_t idx, duk_double_t def_value) {
+	return duk__getopt_number(ctx, idx, def_value);
 }
 
 DUK_EXTERNAL duk_double_t duk_require_number(duk_context *ctx, duk_idx_t idx) {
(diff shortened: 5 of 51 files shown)


svaarala:
The x64 footprint impact for a non-LTO (link time optimization) build is ~800 bytes. For a link time optimized default build the impact is practically zero (16 bytes now) because Duktape doesn't call these (yet) internally. This allows the new functions to be dropped unless an application calls them.

Grepping around a bit, there don't seem to be many internal call sites which can actually use these. Typically Ecmascript semantics require either coercion (say, ToNumber()) or type checking and rejecting, and neither use case matches these API calls.

svaarala:
@harold-b This should now be in working order if you want to test it out.

harold-b:
Great!
I'll try this out tomorrow. :)

svaarala:
Let me know how it goes :) There are a few missing API docs that I'll complete tomorrow, and a few FIXMEs in the code to resolve (mainly about refactoring the get/opt calls and their helpers).

harold-b:
Will do :+1:

fatcerberus:
For what it's worth I probably won't use these much.  For optional arguments and such, the semantics I usually prefer are:

* If undefined (and in some cases, null), use the default value.
* Else, if the value is the correct type, use it as-is
* Else, if coercion is allowed, try to coerce
* Otherwise, TypeError

The way I understand these APIs, they use the default value even in the case that the original value is the wrong type, which seems weird (to me at least).

svaarala:
> The way I understand these APIs, they use the default value even in the case that the original value is the wrong type, which seems weird (to me at least).

What would be the alternative to that if the call doesn't do coercion?

svaarala:
I guess one possible behavior would be to coerce but without changing the value on the value stack. In fact there are quite many different combinations one might come up with re: type accessors:
- Coercing the value stack entry vs. coercing the result but not the value stack entry vs. not coercing the value stack entry or the result.
- Coercing all types (except perhaps undefined, null, or both) vs. coercing only specific types (identified by a type mask; some internals do this).
- Using a default value for "missing" arguments (undefined, null, or both) and throwing for other unexpected types vs. using a default value for any unexpected type.

Unfortunately the API can't really provide all possible variants, so it's really a question of what combinations are most useful in practice.

@fatcerberus Given your validation use case, wouldn't that also mean `duk_get_string()` would not be really useful (in that case) because it doesn't distinguish between a missing value (= undefined) vs. an actual unexpected type (say, a buffer)?

svaarala:
@fatcerberus I do agree BTW that in some use cases one may want to:
- Treat undefined/null specially as a "missing value", and using a default in its place.
- Detect a certain allowed type (e.g. a string) as the expected value.
- Reject other types.

This might be useful in some internals too, though Ecmascript bindings are subtly different in how they type check and coerce.

If something like this seems more useful overall, I can change the behavior in the pull. But if I do that, the proposed naming seems misleading - any idea what an API call like this could be named? It is something like "require string but allow default for missing value" in semantics.

As the pull is now, the API calls are modelled after `duk_get_xxx()`: (1) side effect free, (2) doesn't coerce (except for integer truncation), (3) doesn't throw. With these semantics I think the behavior in this pull seems difficult to improve, but it'd of course be possible to do something different.

harold-b:
@fatcerberus  I agree there's occasions where you want to explicitly do all those checks, but the ability to do that is already in place. And as you showed, there's several checks/branches needed when one goes the explicit route. But there is no current option in place which would allow simply getting a value or a default if the value at stack is what you expected, and since this can map nicely underneath current duk_get_xxx is a win for performance aware applications like games. Especially for mobile platforms.

I also use typescript to which I defer type-checking statically at compile time, which allows me to ease off on the checks on the runtime and opt for a faster method :)

Obviously there's other places in which explicit checks are a must.

svaarala:
Re: naming, Lua's [luaL_optint](http://pgl.yoyo.org/luai/i/luaL_optint) has semantics close to what @fatcerberus described:
- If the argument is a number, return the value (cast to int).
- If the argument is nil (roughly ES undefined), return default value.
- Otherwise raise an error.

So, current naming vs. behavior in this pull would be a bit misleading for Lua users.

svaarala:
@harold-b Your originally proposed naming `duk_get_string_or()` would match the current behavior better. Something like `duk_get_string_with_default()` or `duk_get_string_default()` would be similar alternatives.

To be clear, because of `luaL_optXXX()` behavior I think either the naming or the behavior should be changed.

harold-b:
@svaarala I see what you're saying, and agree.
I hadn't really considered the other stuff mentioned, re:coercion, for this API, so I have nothing to contribute at the moment without giving some thought.

I have to go for tonight, I'll be back tomorrow to test it out and see how the discussion is going :)

svaarala:
Yeah, there's no hurry to merge, and it's good to go through the various alternatives especially when committing to new APIs.

fatcerberus:
> Given your validation use case, wouldn't that also mean `duk_get_string()` would not be really useful (in that case) because it doesn't distinguish between a missing value (= undefined) vs. an actual unexpected type (say, a buffer)?

Indeed, which is why I don't find myself using *that* one much either.  I find that I'll either go right to `duk_require_string()`, or (for the described use case above), `duk_is_undefined[_or_null]()` followed by `duk_require_string()`.

fatcerberus:
> What would be the alternative to that if the call doesn't do coercion?

Throw an exception, like what Lua does (note that in my validation algorithm above, the coercion step is optional, falling through to a TypeError).  The downside in this case is that then the call is no longer side-effect free... ;-)

svaarala:
> > What would be the alternative to that if the call doesn't do coercion?
> Throw an exception, like what Lua does (note that in my validation algorithm above, the coercion step is optional, falling through to a TypeError). The downside in this case is that then the call is no longer side-effect free... ;-)

Right, this pull began with essentially `duk_get_xxx()` with an explicit default (described in #1263) so throwing isn't really an answer to that.

Doing something else is of course an option but then this pull is no longer a solution for #1263 (which is OK too if some other problem is more worth solving :).

svaarala:
Re: coercion behavior, only coercing in place is workable consistently for all types. If the value is not coerced in place, one runs into reachability issues. For example, if a number on the value stack is coerced to a string result without coercing the value stack entry itself, the string wouldn't necessarily be reachable and thus the `const char *` would be dangling.

So in general coercion should be in place. There's some grey area in `duk_get_(u)int()` which won't change the basic type of the value stack entry but will do a double-to-integer coercion which doesn't have any such problems.

fatcerberus:
Yeah, sorry, I didn't mean to suggest that these calls should actually do any coercion, my validation workflow above was meant only to describe what is the general case for me.  The Lua behavior is what I'm looking for, but you're quite right that implementing that doesn't actually solve the issue in #1263.  Hmm...

svaarala:
Given that coercion limitation, the rough alternatives seem to be:
- Current behavior: sloppy "duk_get_xxx() or default".
- Lua-like behavior: accept undefined/missing (and maybe NULL) to return default value, otherwise require specific type or TypeError. This would be roughly "duk_require_xxx() + defaulting for undefined/missing/NULL".
- Coercion behavior: accept undefined/missing (and maybe NULL) to return default value, otherwise coerce value *in place* (to avoid the coercion issue noted above). This would be roughly "duk_to_xxx() + defaulting for undefined/missing/NULL".

Other options? :)

svaarala:
The common theme in the alternatives is defaulting for undefined/missing/null, with each alternative doing the same thing for undefined/missing/null. The alternatives are also equal for the specific type, so they only differ in what to do with the remaining types: use default vs. coerce vs. throw.

fatcerberus:
@svaarala Note that as we have discussed many times, I don't always want to include `null` in the defaulting cases.  Sometimes it's meaningful in its own right ;)

svaarala:
Yeah I know - all the alternatives would have sub-alternatives for how to treat undefined, null, and missing values (= e.g. invalid value stack index). But I didn't want to make the list 8x larger by expanding them in the list ;)

The best case would IMO be to settle on one useful behavior for the undefined/missing/null cases in the API and try to apply it consistently. It won't be 100% of course, as even some Ecmascript bindings have different behaviors.

Providing different variants with different undefined/null/missing behavior will be difficult to remember, difficult to name compactly, etc, so I would like to avoid that if at all possible.

svaarala:
> [re: null] Sometimes it's meaningful in its own right ;)

Hmm, so as a concrete example, if you were reading a string off the value stack:
- Missing/undefined value: presumably default used?
- Null: what would be used here if one distinguishes between undefined/null?
- String: value used as is
- Anything else: TypeError

fatcerberus:
> Providing different variants with different undefined/null/missing behavior will be difficult to remember, difficult to name compactly, etc, so I would like to avoid that if at all possible.

Yeah, I know how that goes.  API design is hard, but fun :)

fatcerberus:
> Hmm, so as a concrete example, if you were reading a string off the value stack:
> [list snipped]

Assuming that the differentiation between null and undefined is used consistently throughout the codebase, I would almost certainly treat `null` in that context as either:
1. Invalid: Throw a TypeError, -OR-
2. "Value omitted".  This is a bit difficult to explain, but I'll try.  This would be for cases where an argument is optional, but completely omitting the associated value has semantics different from when the value is present.  So that calling a function *without* that argument would use a default value (value considered present) is different from calling the function with the argument set to null (value considered not present).

svaarala:
Re: undefined/null, both undefined and null are considered "not object coercible". This maybe favors treating them the same for defaulting too, at least for a "duk_to_xxx() with defaulting" call which does coercion.

I would also quite strongly expect undefined, missing, and null all to trigger defaulting myself. Somewhat as a hunch I think most -- but definitely not all -- Ecmascript built-ins behave the same way. For example it's OK to say `JSON.stringify(value, null, 4)` and because `undefined` is not a keyword it's also quite common to use `null` when `undefined` would (otherwise) be more logical.

svaarala:
> [...] but completely omitting the associated value has semantics different from when the value is present. 

This case is pretty awkward to solve using a single API call, because you would need both the direct return value (say a `const char *`) but an additional boolean to distinguish the cases. Do you think a single API call could (or should) do that?

svaarala:
So ultimately what I'm getting at is: assuming there's a *single API call* that does defaulting, what should that API call do? I agree there are some argument checking cases that may need more complicated type checks (e.g. multiple type mask checks) but in general I don't think those can be crammed into a single API call or even a set of API calls because there are just too many possible combinations.

fatcerberus:
Not sure I follow - for this call in particular, the TypeError would be sufficient.  If I wanted to treat `null` specially (as in the second case) I would just do a `duk_is_null()` check first.  You're right though that in most cases Ecmascript programmers treat undefined and null equivalently.  I'm the oddball that actually takes advantage of the existence of two values and treats them differently :)

To answer your question:
> assuming there's a single API call that does defaulting, what should that API call do?

My vote goes to the Lua semantics, perhaps with coercing and non-coercing variants.

fatcerberus:
Concretely:

Non-coercing:
```c
if (duk_is_[null_or_]undefined(ctx, idx))
    return def_value;
return duk_require_xxx(ctx, idx);
```

Coercing:
```c
if (duk_is_[null_or_]undefined(ctx, idx))
    return def_value;
return duk_to_xxx(ctx, idx);
```

svaarala:
> > assuming there's a single API call that does defaulting, what should that API call do?
> My vote goes to the Lua semantics, perhaps with coercing and non-coercing variants.

I meant particularly w.r.t. undefined, null, and missing.

svaarala:
> Not sure I follow - for this call in particular, the TypeError would be sufficient.

I was referring to this:

> So that calling a function without that argument would use a default value (value considered present) is different from calling the function with the argument set to null (value considered not present).

If I understood that correctly, in this case your target function would behave differently for undefined/missing and null? So I was looking for a concrete example how that would work in practice, i.e. what would be returned for (1) the expected type, (2) undefined, and (3) null, when reading a string from the value stack.

svaarala:
> Concretely [...]

And for comparison, the sloppy behavior similar to duk_get_xxx():

```
if (duk_is_string(ctx, idx)) {
    return duk_get_string(ctx, idx);
}
return def_value;
```

svaarala:
@fatcerberus Btw note that `duk_is_undefined()` or `duk_is_null()` returns false for an invalid index. So I would actually suggest the following semantics for the main alternatives:

duk_get_xxx()-like:
```c
if (duk_is_string(ctx, idx)) {
    return duk_get_string(ctx, idx);
}
return def_value;
```

duk_require_xxx()-like:
```c
if (!duk_is_valid_index(ctx, idx) || duk_is_[null_or_]undefined(ctx, idx)) {
    return def_value;
}
return duk_require_xxx(ctx, idx);
```

duk_to_xxx()-like:
```c
if (!duk_is_valid_index(ctx, idx) || duk_is_[null_or_]undefined(ctx, idx)) {
    return def_value;
}
return duk_to_xxx(ctx, idx);
```

svaarala:
One useful data point would be to look at e.g. 10 internal call sites to see which defaulting behavior would be useful for them. The internal call sites don't represent application call sites because they have specific Ecmascript specification requirements, but it might still shed some light here.

svaarala:
Re: checking for "missing", the duk_is_xxx() calls narrowly check for their specific types and always return false for out-of-bounds value stack indices. But there is a type tag DUK_TYPE_NONE which represents an out of bounds value. So checking for missing, undefined, or null is possible with a single API call as:

```
/* Same as: !duk_is_valid_index(ctx, idx) || duk_is_[null_or_]undefined(ctx, idx) */
duk_check_type_mask(ctx, idx, DUK_TYPE_MASK_NONE |
                              DUK_TYPE_MASK_UNDEFINED |
                              DUK_TYPE_MASK_NULL);

**Comment by svaarala** [comment]
@harold-b Regarding performance, checking for sets of types is relatively fast internally. There's an internal primitive to look up a type mask which boils down to reading the type tag, and doing one array lookup to get the mask. With the mask at hand, checking for individual types or type sets can then be simply e.g. `type_mask & (DUK_TYPE_MASK_NULL | DUK_TYPE_MASK_UNDEFINED)` which compiles to a single AND instruction and a branch. The concrete implementation for the example algorithms above would be very different and wouldn't involve multiple calls.

**Comment by fatcerberus** [comment]
> If I understood that correctly, in this case your target function would behave differently for undefined/missing and null? So I was looking for a concrete example how that would work in practice, i.e. what would be returned for (1) the expected type, (2) undefined, and (3) null, when reading a string from the value stack.

For a "get with default" call in particular?  I don't expect anything more than a TypeError.  Anything beyond that requires an explicit branch because the `null` is a meaningful value in its own right - there's nothing to substitute for it.  In Ecmascript, the `null` would be stored as-is instead of a string.

As a concrete example of treating `null` specially, imagine a version of `Object.create()` where the prototype argument is optional, defaulting to `Object.prototype`.  If one does `Object.create(null)`, that creates a bare object, but `Object.create()` makes one inheriting from `Object.prototype`.  In Ecmascript, that could be:

```js
function createObj(proto)
{
    proto = proto !== undefined ? proto : Object.prototype;
    return Object.create(proto);
}
```

There's no real analogy in C because the language is strongly typed.  Therefore a "defaulting get" wouldn't be appropriate there.

svaarala:
> For a "get with default" call in particular? I don't expect anything more than a TypeError. Anything beyond that requires an explicit branch because the null is a meaningful value in its own right - there's nothing to substitute for it.

Ok, that makes sense. So this case is a bit out of scope re: a defaulting API call: regardless of the call semantics one would most likely check for `null` explicitly. Handling of undefined/null/missing in a defaulting API call wouldn't then matter because null was already handled.

I wasn't sure about what you meant because it is possible to have a "tri-state" function that returns a value but also some additional information. One such call is actually used internally, to get a buffer data pointer but also return a boolean indicating if the value was a buffer or not. The boolean is needed because a buffer data pointer may be NULL even for a valid buffer argument. But naturally such calls are pretty awkward and to be avoided whenever possible ... :)

harold-b:
Oh boy, lots to catch up with... I'll get to.

harold-b:
It seems this has largely from the original intent.

> Regarding performance, checking for sets of types is relatively fast internally. There's an internal primitive to look up a type mask which boils down to reading the type tag, and doing one array lookup to get the mask. With the mask at hand, checking for individual types or type sets can then be simply e.g. type_mask & (DUK_TYPE_MASK_NULL | DUK_TYPE_MASK_UNDEFINED) which compiles to a single AND instruction and a branch. The concrete implementation for the example algorithms above would be very different and wouldn't involve multiple calls.

@svaarala  Yes I use DUK_TYPE_MASKS in tons of places but this is unrelated to the original intent of what I proposed in #1263. The idea was basically to have a call with the exact same behavior as `duk_get_xxx` calls except allowing you to provide the default value it returns. There by leveraging what is already there, as I showed in the post, I only changed the line of code that sets the default value to NaN on `duk_get_number`. Since it's always going to return a default value, I wanted to be able to leverage it for free and simply change the current API to a macro to use the defaulting method in place. This would eliminate verbosity with a simply call to an API that is already returning a default value.

I see that you guys have focused more on moving towards an API which is in line with the lua opt stuff you mentioned? I don't have much to say in that regard, I've had no issues in that case with simply using the type masks or `duk_is_xxx` functions, so I don't really have interest in other API calls similar to it, since the provided already gives me finely-grained control over how to handle values.

svaarala:
@harold-b @fatcerberus So how about we split this discussion into two separate issues?
- Issue #1: extending duk_get_xxx() with an explicit default value, which Harold brought up initially. This pull does exactly that, but I think the naming needs to be changed to avoid confusion with `luaL_optXXX()`.
- Issue #2: a stricter type checking mechanism with a controllable default value.

fatcerberus:
@svaarala Sounds good to me.

harold-b:
> Issue #1: extending duk_get_xxx() with an explicit default value, which Harold brought up initially. This pull does exactly that, but I think the naming needs to be changed to avoid confusion with luaL_optXXX().

But issue #1 is "**Fix g++ (with -std=c++11) compilation**"! ...just joking :stuck_out_tongue_closed_eyes: 

Yes this sounds good to me.

I'll be trying out this branch in a little bit, I just need to change the duktape callsites in my code that changed on 2.0.0.

svaarala:
@harold-b I edited out the hash marks from the comment :)

So, for this pull the remaining question is naming. Once there's a good idea of that, I'll update the code and documentation and this would be mergeable. I'll then open a separate pull for @fatcerberus's use case which I think is also worth an API call. It's important these two calls have naming that intuitively conveys the difference.

harold-b:
OK excellent, I'll read a bit more thoroughly what you and @fatcerberus where discussing so I can clearly understand its purpose, in the interest of discussing clear API naming for both.

fatcerberus:
Regarding that use case, a point in favor of an API for it is that I find myself doing this a lot:

```c
num_args = duk_get_top(ctx);
// ...
priority = num_args >= 2 ? duk_require_number(ctx, 1)
	: 0.0;
```

Which also requires me to declare the function vararg.  I can't replace the `num_args` check with `duk_is_undefined()` either because that returns false in case of an invalid index (= missing argument).  (I could declare the function with a set arg count to avoid that, but that leads to weird bugs later when I forget to update the count :)

svaarala:
@fatcerberus There is `duk_is_valid_index()` which allows you to avoid using duk_get_top() etc. I could also add a `duk_is_none()` for completeness - but it would essentially be `!duk_is_valid_index()`.

svaarala:
But, assuming the `duk_opt_xxx()` call accepted undefined/null/missing all the same way (= use default) but rejected other unexpected types, that would seem to solve that particular use case without needing to look at top etc? Of course assuming you'd be OK with accepting null to trigger default there.

fatcerberus:
Anyway, I'll step aside for now and let you and @harold-b hash out this pull.  I'll keep an eye out for the other one. :)

svaarala:
Added https://github.com/svaarala/duktape-wiki/pull/161 for summarizing the different behaviors. It might also make the discussion a bit easier to follow.

svaarala:
@fatcerberus Feedback on the naming would be appreciated so that the calls are not confused. Also feedback on how to illustrate the differences https://github.com/svaarala/duktape-wiki/blob/6fc95a3dda72d0483b6284e74d5b19f01b4c7d36/HowtoValueStackTypes.md would be very useful :)

harold-b:
OK, caught up better now with the discussion.

For naming, what we know:
Because [luaL_optint()](http://pgl.yoyo.org/luai/i/luaL_optint) has different behavior than that of `duk_get_xxx`, it would be confusing for users to use that naming for the #1263 call. So that one is out for this one.

In my opinion, since the behavior is meant to basically be `duk_get_xxx` with a user-specified default,
`duk_get_xxx_<something>` would probably make the most sense. Don't you guys think?
A `duk_get_xxx_with_default` is probably what we want, less the long name, or the original issue title of `duk_get_xxx_or` seem to fit.

Since the other call that @svaarala  and @fatcerberus  were discussing is liable to throw, it sounds more like it's in the `duk_require_` camp. It seems the behavior is similar rather similar as of what has been discussed.

Let me look at the table @svaarala just posted to provide more feedback.

harold-b:
Hijacking this thread real quick to ask about tools/configure.py...

I'm trying to migrate my branch to use 2.0.0 but I'm getting some issues when generating the config.

It's of course working fine from the v2.0.0 tag, but if I [rebase my changes](https://github.com/harold-b/duktape/commit/5edc585e0c8452d20af74280fff52d6ad77259e8#diff-df880d4c6b6e3b1ee1453743eba813eb) (which only affect duk_debugger.c) onto the v2.0.0 tag, I get an this error when using configure.py:

``` python
configure.py          INFO    Configuring Duktape version 2.0.0, commit f643bf5ac2e49cefd449345451c8e7d7639fbfef, describe v2.0.0-3-gf643bf5-dirty, branch musa/v4
configure.py          INFO      - source input directory: D:\Projects\Games\Musashi2\Source\3P\duktape\src-input
configure.py          INFO      - license file: D:\Projects\Games\Musashi2\Source\3P\duktape\LICENSE.txt
configure.py          INFO      - authors file: D:\Projects\Games\Musashi2\Source\3P\duktape\AUTHORS.rst
configure.py          INFO      - config metadata directory: D:\Projects\Games\Musashi2\Source\3P\duktape\config
configure.py          INFO      - output directory: D:\Projects\Games\Musashi2\Source\3P\duktape\tmp\configs
configure.py          INFO    Reusing output directory (already exists but contains only safe, known files)
genconfig.py          INFO    Creating duk_config.h: platform=any, architecture=any, compiler=any, dll mode
genconfig.py          WARNING Recommended config option DUK_USE_FATAL_HANDLER not provided
genbuiltins.py        INFO    Creating built-in initialization data: ram built-in support
genbuiltins.py        INFO    Removed 1 objects (0 disabled, 1 not needed by config), 6 properties (0 disabled, 6 not needed by config)
  File "D:\Projects\Games\Musashi2\Source\3P\duktape\tools\genbuiltins.py", line 3135, in <module>
    main()
  File "D:\Projects\Games\Musashi2\Source\3P\duktape\tools\genbuiltins.py", line 2983, in main
    ram_meta = load_metadata(opts, rom=False, build_info=build_info, active_opts=active_opts)
  File "D:\Projects\Games\Musashi2\Source\3P\duktape\tools\genbuiltins.py", line 1094, in load_metadata
    metadata_add_string_used_stridx(meta['strings'], json.loads(f.read()))
  File "D:\Projects\Games\Musashi2\Source\3P\duktape\tools\genbuiltins.py", line 900, in metadata_add_string_used_stridx
    raise Exception('source code needs define %s not provided by strings' % repr(k))
Exception: source code needs define u'DUK_STRIDX_INT_LEXENV' not provided by strings
  File "tools/configure.py", line 972, in <module>
    main()
  File "tools/configure.py", line 669, in main
    exec_print_stdout(cmd)
  File "tools/configure.py", line 60, in exec_print_stdout
    ret = exec_get_stdout(cmd, input=input, print_stdout=True)
  File "tools/configure.py", line 51, in exec_get_stdout
    raise Exception('command failed, return code %d: %r' % (proc.returncode, cmd))
Exception: command failed, return code 1: ['C:\\Python27\\python.exe', 'D:\\Projects\\Games\\Musashi2\\Source\\3P\\duktape\\tools\\genbuiltins.py', '--git-commit', 'f643bf5ac2e49cefd449345451c8e7d7639fbfef', '--git-branch', 'musa/v4', '--git-describe', 'v2.0.0-3-gf643bf5-dirty', '--duk-version', '20000', '--used-stridx-metadata', 'c:\\users\\harito\\appdata\\local\\temp\\tmp-duk-prepare-vkfpqe\\duk_used_stridx_bidx_defs.json.tmp', '--strings-metadata', 'D:\\Projects\\Games\\Musashi2\\Source\\3P\\duktape\\src-input\\strings.yaml', '--objects-metadata', 'D:\\Projects\\Games\\Musashi2\\Source\\3P\\duktape\\src-input\\builtins.yaml', '--active-options', 'c:\\users\\harito\\appdata\\local\\temp\\tmp-duk-prepare-vkfpqe\\duk_config_active_options.json', '--out-header', 'c:\\users\\harito\\appdata\\local\\temp\\tmp-duk-prepare-vkfpqe\\src\\duk_builtins.h', '--out-source', 'c:\\users\\harito\\appdata\\local\\temp\\tmp-duk-prepare-vkfpqe\\src\\duk_builtins.c', '--out-metadata-json', 'c:\\users\\harito\\appdata\\local\\temp\\tmp-duk-prepare-vkfpqe\\genbuiltins_metadata.json', '--ram-support']

```

My input is:
``` bash
py -2 tools/configure.py \
    --output-directory ./tmp/configs \
    --source-directory src-input \
    --config-metadata config \
    --dll \
    -DDUK_USE_FASTINT \
    -DDUK_USE_ASSERTIONS \
    -DDUK_USE_DEBUGGER_SUPPORT \
    -DDUK_USE_INTERRUPT_COUNTER \
    -DDUK_USE_DEBUGGER_INSPECT \
    -DDUK_USE_DEBUGGER_DUMPHEAP \
    -DDUK_USE_DEBUGGER_THROW_NOTIFY
```

This input works fine with the vanilla v2.0.0 tag. I also renamed my own option `MUDUK_USE_DEBUG_GET_CLOSURES` because that was causing an error as well claiming that `DUK_USE_DEBUG_GET_CLOSURES` was not a valid config option.

What am I messing up that makes it not find `DUK_STRIDX_INT_LEXENV`?

Any help with this would be appreciated.

svaarala:
Hmm, are you sure you're not mixing source code / config versions somehow? That DUK_STRIDX_INT_LEXENV string shouldn't occur in the repo at all (git grep INT_LEXENV is empty).

svaarala:
Ah, I just remembered you did custom work on scopes and that code probably accessed the internal `_Varenv` and `_Lexenv` properties. Those properties are no longer present in 2.0.0 and are replaced by actual struct fields in #1132, so you'd need to reimplement those parts of your custom changes.

harold-b:
Ah nice find.

It's in my changes in (this line)[https://github.com/harold-b/duktape/blob/5edc585e0c8452d20af74280fff52d6ad77259e8/src/duk_debugger.c#L3034]
which uses `DUK_HTHREAD_STRING_INT_LEXENV`.

I'll have to dig in deeper to re-implement these then. 
Thanks. :)

fatcerberus:
@harold-b Ah, I forgot you were the one who implemented the custom GetClosures debugger command. :)  Let me know when you get that working again, I'd like to implement something like it in SSJ.

harold-b:
@fatcerberus Sure thing :)

svaarala:
@harold-b Just a heads up - to improve performance and footprint of scope handling, it's quite likely scope objects will get their own data structures (either special heap objects or specialized `duk_hobject` subclasses like `duk_harray` etc) in the near future. This will affect scope walking once again.

harold-b:
@svaarala Got it. Thanks for letting me know.

harold-b:
Hey @svaarala, sorry for the delay, I tried to swap versions in the middle of the week but was unable to.
I managed to get my App ready for v2.0.0 this week so I tried out this branch and swapped some of my native methods with the `duk_opt_calls`, works perfectly! :) Thank you.

Do you think this branch is safe to run on a project? Or should I stay with v2.0.0 until v2.1.0 is released?

Also a quick Q with v2.0.0. I have a binder helper class to bind JS classes with their native constructors, methods and properties. In this binder, I gave the native constructor functions a "name" property so that the debug client could figure out which object it is, if it returned the default `toString` method. In v2.0.0 this caused a fatal error, can't recall what it said, I think non-configurable or something.

Is there any way to overcome this for v2.0.0?

svaarala:
@harold-b For this to get merged at least API function renaming would be needed, and quite likely the current names (duk_opt_boolean etc) would be used for calls with different semantics. So as soon as this naming issue is resolved, I can move to change the calls and merge this into master.

Re: .name, in ES6 function `.name` properties are not writable, but are configurable. That means you can't just write to them, but you can overwrite them using `Object.defineProperty()` or `duk_def_prop()`. This change was a bit annoying but it was needed to align with ES6.

harold-b:
>Re: .name, in ES6 function .name properties are not writable, but are configurable. That means you can't just write to them, but you can overwrite them using Object.defineProperty() or duk_def_prop()

Perfect, that works :+1: 

>For this to get merged at least API function renaming would be needed, and quite likely the current names (duk_opt_boolean etc) would be used for calls with different semantics. So as soon as this naming issue is resolved, I can move to change the calls and merge this into master.

But safety-wise, should I be OK to use it as is (I can always to a find and replace when it gets merged with their proper naming)?

Speaking of which any new thoughts on name ideas for this one?

svaarala:
The API calls here should be fully working. For naming the best proposal so far, in my view, is:
- `duk_get_xxx_default()`

I would prefer `duk_something_xxx()` if a good, intuitive name came up :)

harold-b:
I'm still with you on `duk_something_xxx()` being preferable, but I'm also still drawing blanks on that one :(

I'll try to brainstorm a bit more...

fatcerberus:
How about `duk_maybe_xxx()`?

svaarala:
It's hard to associate with the lenient nature of the calls; for example, how would duk_opt_xxx() and duk_maybe_xxx() be intuitively different?

If duk_get_xxx() didn't exist it would probably be best to add the default values to the API calls without introducing new API calls. This would be an option for 3.0.0 release but even so it doesn't seem like a good option.

fatcerberus:
It reads somewhat differently to me at least:
* "optional string" - value is optional, but if present it must be a string.
* "maybe string" - value may be a string, but caller doesn't know for sure.  If not, use default value.

svaarala:
Hmm, I guess so. To me it reads more like an assertion than reading a value though :-/

How about something close to get, like `duk_read_number()`? The "read" would refer to "reading from the value stack". But it's a bit weird because it can be confused with reading e.g. string contents.

`duk_peek_number()`? But that sounds to me like it wouldn't have a default. It does have the upside of conveying the "no side effects" policy of the calls.

`duk_lookup_number()`?

Or maybe `duk_grab_number()` ;-)

fatcerberus:
Regarding that null/undefined bikeshed: It's worth mentioning I think that when using ES6 default arguments, `null` is treated as "present" but `undefined` is not:

```js
function eat(what, where = "farm", howMany = 812) {
    console.log(`the pig ate ${howMany} ${what} at the ${where}!`);
}
eat("cows", undefined, 1000);  // the pig ate 1000 cows at the farm
eat("cows", null, 1000);  // the pig ate 1000 cows at the null
```

So perhaps `duk_opt_xxx()` *should* be a TypeError in case of `null`.

svaarala:
@fatcerberus That's a very good consideration IMO. It would definitely make sense the C API mostly intended for handling optional arguments would behave the same way as ES6 optional arguments.

harold-b:
> How about something close to get, like duk_read_number()? The "read" would refer to "reading from the value stack". But it's a bit weird because it can be confused with reading e.g. string contents.

So far duk_read_xxx so far is my top choice... Was thinking about it a few days ago and it was the only one that felt ok. I came to check the thread and you had already suggested it, so my :+1:  on that one.

svaarala:
I'll open a separate pull for the semantics of duk_require_xxx() with an explicit default for undefined and invalid index (which generally behaves like undefined). Still not sure about naming, and I'm maybe leaning more toward just using duk_get_xxx_default() and duk_require_xxx_default() simply because any of the shorter alternatives don't have an intuitively clear meaning.

fatcerberus:
No `duk_opt_xxx()` to align with ES6 optional parameters?

svaarala:
Yes, that's what I meant with duk_require_xxx() + defaulting for undefined and invalid index (= align with ES6 optional parameters). Just not sure about the naming still.

fatcerberus:
Yes, I was referring to the name :)

svaarala:
Oh OK, so I'm just not sure about the name, whether `duk_require_string_default()` would convey the semantics better (even if a lot more verbose), especially if it's paired with `duk_get_string_default()`.

These are actually quite similar operations, and could be implemented with a single internal function:

```c
const char *duk_xxx_string(duk_context *ctx, duk_idx_t idx, duk_uint_t mask, const char *def_ptr) {
    if (duk_check_type_mask(ctx, idx, mask)) {
        return def_ptr;
    }
    return duk_require_string(ctx, idx);
}
```

For `duk_get_string_default()` the mask includes everything but an actual string, i.e. anything else triggers the default. For `duk_require_string_default()` the mask is `DUK_TYPE_MASK_NONE | DUK_TYPE_MASK_UNDEFINED`.

This is of course an internal detail and doesn't affect public API naming.

svaarala:
@fatcerberus One detail in mimicking ES6 semantics is the handling of "unused" (non-existent) value stack entries. For the API to work well with vararg functions, it needs to treat invalid indices the same as `undefined`. This is in line with ES semantics for most functions, at least user functions (some built-ins make the difference).

But other C API calls don't generally treat invalid index and undefined the same. That's actually maybe something to change, maybe `duk_is_undefined()` should be true for out-of-bounds indices for example so that you could use `duk_is_undefined()` to check vararg function arguments without having to explicitly check whether the index is valid.

fatcerberus:
Hm, not sure - JS doesn't have a concept of a stack so that "non-present argument" is (as far as Ecmascript code is concerned) exactly the same thing as "undefined value argument".  To get that in C you declare the function with a set number of arguments.  Declaring it vararg is basically saying you don't care about the specific shape of the value stack - and in that case you should be prepared to deal with non-present values.  I'm not convinced that, in this instance at least, perfect alignment with Ecmascript is necessary.

svaarala:
Sure, I agree it's not necessary to align 100% -- and `duk_opt_xxx()` doesn't really have an Ecmascript equivalent anyway (for value defaulting yes, but not for type checking).

For duk_opt_xxx() the least surprising way would be to treat out-of-bounds values like undefined (at least to me) - basically if you omit arguments to a vararg function they are treated as if `undefined` was passed in -- which is generally the case for script functions unless they dig into `arguments` specifically which is rare.

svaarala:
The equivalent of a vararg function is roughly an Ecmascript function taking zero arguments which then inspects `arguments`. Argument default isn't possible in that case so there's no direct equivalent.

In any case, the basic options are to either TypeError or use-default for invalid indices. As a user I'd prefer the latter (matching duk_get_xxx() which also return default for invalid index), but the former could also be justified because the type check is more strict.

But there's really no direct equivalent because there's no "none" value in Ecmascript, at least in an easily usable form. You can determine the different by inspecting `arguments` but that's uncommon and not often done so it doesn't really guide design here IMO.

svaarala:
Actually, looking `duk_is_undefined()` for out-ot-bounds values, the equivalent in Ecmascript would be roughly:

```js
if (typeof arguments[N] === 'undefined') {
}
```

which would actually test true for missing values:

```
duk> function foo() { print(typeof arguments[3] === 'undefined'); }
= undefined
duk> foo(1);
true
= undefined
```

You only detect that the argument was "out of bounds" by comparing the index to `arguments.length` -- the C equivalent of which is to inspect `duk_get_top()` or maybe `duk_is_valid_index()`.

So here's one argument why duk_is_undefined() should be true for out-of-bounds indices :)

svaarala:
Closing as unmerged, #1458 adds the Lua/ES2015 like optional argument semantics, and a separate pull will be created for "get with default".

PR #1270, opened 2017-01-05
Fix String.prototype.repeat() copy_size handling when not preferring footprint

svaarala:
When DUK_USE_PREFER_SIZE is disabled, Duktape 2.0.0 String.prototype.repeat() mismanages the `copy_size` variable in the copy loop. This causes memory unsafe behavior and is visible both as valgrind warnings but also in actual repeat results. The error causes potentially uninitialized to be read during the copy, may corrupt the resulting string, but shouldn't cause the allocated result buffer to be overrun. Still tagged security because memory unsafe behavior is going on.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2366,6 +2366,9 @@ Planned
 1.7.0 (XXXX-XX-XX)
 ------------------
 
+2.0.1 (XXXX-XX-XX)
+------------------
+
 2.1.0 (XXXX-XX-XX)
 ------------------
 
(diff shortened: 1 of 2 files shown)


fatcerberus:
I wonder why this didn't show up in the testcases?

svaarala:
What happens in practice is a `memcpy()` with overlapping memory areas because the copy_size is larger than intended. That causes undefined behavior but in practice it works correctly if the copy starts from the beginning of the buffer (which is of course not guaranteed) and the already copied prefix is long enough. So it's not triggered on every repeat; I managed to produce it when working on property caching.

svaarala:
One triggering factor is GCC optimization level; with -Os (which `duk` uses) it's not triggered, but with -O2 it is.

fatcerberus:
minisphere Ubuntu packages are compiled with -O2 (because that's what pbuilder uses), so it's good that you caught this then :)

svaarala:
Yeah :) It would be nice to have automatic coverage for running the testcases with various option combinations; covering 3-4 profiles would be nice at least. That was one of the webhook testrunner goals that I didn't have time to complete. It also means a few more test clients are needed because running the tests will take a lot longer.

PR #1272, opened 2017-01-06
Fix exponentiation operator behavior on Linux gcc 4.7+ -O2

svaarala:
For some reason exponentation operator behaves incorrectly on Linux with gcc 4.8.4 -O2 (-Os used for commit regression tests is fine). Symptom example:

```
duk> 2 ** 10
= 1024
duk> x = 2
= 2
duk> x ** 10
= 1  // <-- ?
duk> Math.pow(2,10)
= 1024
```

Exp at compile time works while runtime results are incorrect. The mysterious thing is that in both cases the internal function used to compute the exponentiation is the same. Also other arithmetic operations, handled by the same arith helper, work.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2386,5 +2386,8 @@ Planned
 * Fix memory unsafe behavior in Duktape 2.0.0 String.prototype.repeat()
   (GH-1270)
 
+* Fix incorrect exponentation operator behavior which happened at least on
+  Linux gcc 4.8.4 with -O2 (not -Os) (GH-1272)
+
 3.0.0 (XXXX-XX-XX)
 ------------------
--- a/src-input/duk_js_executor.c
+++ b/src-input/duk_js_executor.c
@@ -276,7 +276,7 @@ DUK_LOCAL DUK__INLINE_PERF void duk__vm_arith_binary_op(duk_hthread *thr, duk_tv
 			break;
 		}
 		default: {
-			DUK_UNREACHABLE();
+			/* Possible with DUK_OP_EXP. */
 			goto skip_fastint;
 		}
 		}


svaarala:
This issue is also GCC version specific. Gcc 4.6.4 -O2:

```
duk> x = 2
= 2
duk> x ** 10
= 1024
```

Gcc 4.8.4 -O2:

```
duk> x = 2
= 2
duk> x ** 10
= 1
```

So it's most likely related to some optimization decision (aliasing assumptions, union assumptions, for example) that GCC 4.7+ make with -O2.

fatcerberus:
Maybe an optimizer bug?

svaarala:
Could be, but at the outset a Duktape bug would be more likely IMO.

svaarala:
`dukd` compiled with gcc-4.8.4 and forced -O2 exhibits the issue, and managed to get this debug dump:

```
duk> x = 2
= 2
duk> x ** 10
D0 duk_js_arith.c:69 (duk_js_arith_pow): POW: 2.000000 0.000000
= 1
```

Disabling fastint support but otherwise the same:

```
duk> x = 2
= 2
duk> x ** 10
D0 duk_js_arith.c:69 (duk_js_arith_pow): POW: 2.000000 10.000000
= 1024
```

So the issue may be fastint related. Indeed, adding back fastint support and ensuring that the exponent is *not* a fastint:

```
duk> z = 0.1
= 0.1
duk> z = 10
= 10
duk> x ** z
D0 duk_js_arith.c:69 (duk_js_arith_pow): POW: 2.000000 0.000000
= 1
duk> x = 2
= 2
duk> y = 10
= 10
duk> x ** y
D0 duk_js_arith.c:69 (duk_js_arith_pow): POW: 2.000000 0.000000
= 1
duk> y += 0.1; y -= 0.1  // <-- remove fastint status
= 10
duk> x ** y
D0 duk_js_arith.c:69 (duk_js_arith_pow): POW: 2.000000 10.000000
= 1024
```

So, the issue is most likely in converting a fastint into an IEEE double which does use union tricks. No idea why it would manifest here but not in other arithmetic though.

svaarala:
I think I found the issue. The executor arith helper has a fastpath for both arguments being fastints. That fastpath handles SUB, MUL, DIV, MOD (ADD has a separate helper) but not EXP. The case clause for the fastint check has a default clause with `DUK_UNREACHABLE()` which is incorrect because the default clause is reached with EXP. The case is concretely:

```c
default:
    DUK_UNREACHABLE():
    goto skip_fastint;
```

Because `DUK_UNREACHABLE()` compiles to `__builtin_unreachable()`, the compiler is justified in not making that code path work. Interestingly it only affects -O2 - but that's par for the course for this kind of bugs.

Removing the `DUK_UNREACHABLE()` fixes the issue.

svaarala:
Running the test set manually nothing else fails with gcc 4.8.4 -O2 after this fix.

fatcerberus:
This was my bad, I saw that the default case jumped to `skip_fastint` but somehow missed the `DUK_UNREACHABLE()` hint... :)

svaarala:
These are pretty hard to catch because they only manifest in only certain conditions (like compiler versions, register pressure, whatever). Here also if I added debug prints the problem went away because the debug prints affected register handling.

Realistically this could have been caught with an updater matrix compile test that would run testcases with different compiler versions, different compiler options, and different Duktape options. I had that tooling for the Duktape 1.x config model but not yet for the Duktape 2.x config. It also needs a bit more testing hardware, and I do have 4 dedicated physical hosts now for continuous testing.

PR #1281, opened 2017-01-09
Avoiding conditional directives that split parts of statements

RomeroMalaquias:
Hello, this change is only aesthetic, it's a suggestion to avoid the use of conditional directives that break statements i think that kind of concern improve the code maintainability do you agree?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -41,6 +41,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Brett Vickers (https://github.com/beevik)
 * Dominik Okwieka (https://github.com/okitec)
 * Remko Tronçon (https://el-tramo.be)
+* Romero Malaquias (rbsm@ic.ufal.br)
 
 Other contributions
 ===================
--- a/src-input/duk_js_compiler.c
+++ b/src-input/duk_js_compiler.c
@@ -6171,6 +6171,7 @@ DUK_LOCAL void duk__parse_stmt(duk_compiler_ctx *comp_ctx, duk_ivalue *res, duk_
 	duk_small_uint_t stmt_flags = 0;
 	duk_int_t label_id = -1;
 	duk_small_uint_t tok;
+	duk_bool_t test_func_decl;
 
 	DUK__RECURSION_INCREASE(comp_ctx, thr);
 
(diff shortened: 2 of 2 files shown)


svaarala:
Conditionals are sometimes also used to break if-else statements. It's sometimes rather difficult to avoid, e.g.:

```
#if defined(FEATURE_X)
    if (something) {
        /* something */
    } else 
#endif
    {
        /* common case */
    }
```

There's no style policy against that right now because it's sometimes awkward to avoid.

svaarala:
That said, I do agree that for if-conditions using an explicit variable (with the variable initialized in an #ifdef-chain) is cleaner. For very simple cases it's tempting not to add an explicit variable and it's probably not worth avoiding them as absolute policy. But for anything more than a trivial case it's easier to read an explicit variable. For example, the compiler conditionals for keeping various properties of function templates were converted to use explicit variables for readability, see e.g.:

- https://github.com/svaarala/duktape/blob/master/src-input/duk_js_compiler.c#L858-L874

RomeroMalaquias:
I see, i'm working in a research to find if the developers care about that kind of stuff.
There is a paper that suggests some approachs to avoid this use of conditionals, the title [A Catalogue of Refactorings to Remove Incomplete Annotations](http://www.dsc.ufcg.edu.br/~spg/uploads/medeiros-jucs-14.pdf).
I have a question, the test is failing because the change that i did?

svaarala:
Interesting - do you have a recommendation for dealing with the if-else blocks and #ifdefs?

Re: test failure, it's caused by a "codepolicycheck":

```
Code policy check

vim +6240  src-input/duk_js_compiler.c : trailing whitespace                     - test_source_elem = allow_source_elem;

vim +6240  src-input/duk_js_compiler.c : non-leading tab (idx 6239)              - test_source_elem = allow_source_elem;

*** Total: 2 problems
```

RomeroMalaquias:
The paper suggests something like:

```c
if (condition1) {
     //lines of code 1
}
#ifdef expression1
else if(condition2){
     //lines of code 2
}
#endif
```
could be changed to:
```c
bool test = condition1;
if (test) {
     //lines of code 1
}
#ifdef expression1
if(!test && condition2){
     //lines of code 2
}
#endif
```

RomeroMalaquias:
I will fix the commit

RomeroMalaquias:
Hello @svaarala , is there something more to do to improve this pull request?

svaarala:
I'll check it out later tonight.

RomeroMalaquias:
Thanks for your time

svaarala:
Hmm, I actually wrote some review comments that are not yet addressed. Also, please add yourself to AUTHORS.rst (the first author list, for code contributions) :)

svaarala:
Ah, I had forgotten to press "Submit review" so my comments were not visible.

svaarala:
Please squash the code commits (AUTHORS commit is better left separate).

The temporary name "test_source_elem" is actually a bit misleading -- maybe "test_func_decl" would be more readable given its purpose.

RomeroMalaquias:
Ok!

svaarala:
Thanks @RomeroMalaquias !

RomeroMalaquias:
Thanks for your time

PR #1282, opened 2017-01-10
Fix DUK_UNREF() warning for volatile

svaarala:
Reported by @markand:

```
duk_js_executor.c:5008:14: warning: conversion to void will not access object of type ‘volatile duk_small_int_t {aka 
                 volatile int}’
duk_config.h:2436:36: note: in definition of macro ‘DUK_UNREF’
#define DUK_UNREF(x)  do { (void) (x); } while (0)
```

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2424,6 +2424,8 @@ Planned
 * Fix incorrect exponentiation operator behavior which happened at least on
   Linux gcc 4.8.4 with -O2 (not -Os) (GH-1272)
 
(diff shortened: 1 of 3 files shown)


svaarala:
The volatile variable in 2.0.0 was used as a way of ensuring that the switch-case generated for an 8-bit opcode doesn't need a bounds check. The reworked version here (omitting the default clause) seems to avoid the bounds check too on GCC. Clang emits the bound check either way; I'm not sure how to fix that but it's not within the scope of this pull in any case.

PR #1286, opened 2017-01-13
Fix typedarray .set(undefined) assert failure and incorrect behavior

### LINKED ISSUE #1285: TypedArray.set() asserts if the first parameter is undefined.
Example:
``` javascript

var a = new Float32Array(2);
a.set( undefined ); // Causes assert crash in builds with DUK_USE_ASSERTIONS
```
Maybe I'm mistaken, but isn't the expected behavior here to throw if the parameter is undefined even in debug builds?

This is called from the protected variant `duk_pcall_method()`. See the offending call stack:

```
duk_default_fatal_handler(void * udata, const char * msg) Line 10537    C
duk_known_hobject(duk_hthread * ctx, int idx) Line 18310    C
duk_bi_typedarray_set(duk_hthread * ctx) Line 25449 C
duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 59505 C
duk_handle_call_unprotected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 59038  C
duk__js_execute_bytecode_inner(duk_hthread * entry_thread, unsigned int entry_callstack_top) Line 73087 C
duk_js_execute_bytecode(duk_hthread * exec_thr) Line 70899  C
duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 59453 C
duk_handle_call_protected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 58936    C
duk_pcall_method(duk_hthread * ctx, int nargs) Line 13382   C
Musashi::ScriptGameSys::InitScriptApp() Line 424    C++
```
I am running off commit f7e0d6f72195507382a4460e7f085b17e7280b1a from the #1264 PR.

svaarala:
Present in 2.0.0 but not in 1.x; incorrect behavior causes a segfault so tagged security.

Fixes #1285.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2390,6 +2390,9 @@ Planned
 * Improve duk_hstring array index handling performance when
   DUK_USE_HSTRING_ARRIDX is disabled (GH-1274)
 
+* Fix argument validation bug in typedarray .set() which would cause a segfault
+  for e.g. new Float64Array(2).set(undefined) (GH-1285, GH-1286)
+
 * Fix duk_hstring array index check integer overflow, which caused certain
   integer strings (such as '7394299990') to be incorrectly treated as array
   indices (GH-1273, GH-1276)
--- a/src-input/duk_bi_buffer.c
+++ b/src-input/duk_bi_buffer.c
@@ -1656,7 +1656,7 @@ DUK_INTERNAL duk_ret_t duk_bi_typedarray_set(duk_context *ctx) {
 	}
 
 	duk_hbufobj_promote_plain(ctx, 0);
-	h_obj = duk_known_hobject(ctx, 0);
+	h_obj = duk_require_hobject(ctx, 0);
 
 	/* XXX: V8 throws a TypeError for negative values.  Would it
 	 * be more useful to interpret negative offsets here from the
--- a/tests/ecmascript/test-bug-typedarray-set-undefined.js
+++ b/tests/ecmascript/test-bug-typedarray-set-undefined.js
@@ -1,0 +1,12 @@
+/*===
+TypeError
+still here
+===*/
+
+var x = new Float64Array(2)
+try {
+    x.set(undefined)
+} catch (e) {
+    print(e.name);
+}
+print('still here');


svaarala:
```
duktape@keppihevonen:~/duktape$ ./duk.O2.200 tests/ecmascript/test-bug-typedarray-set-undefined.js
Segmentation fault
duktape@keppihevonen:~/duktape$ ./duk tests/ecmascript/test-bug-typedarray-set-undefined.js
TypeError: object required, found undefined (stack index 0)
    at [anon] (duk_api_stack.c:1608) internal
    at set () native strict preventsyield
    at global (tests/ecmascript/test-bug-typedarray-set-undefined.js:6) preventsyield
error in executing file tests/ecmascript/test-bug-typedarray-set-undefined.js
```

fatcerberus:
Huh, I wonder how that `duk_known_hobject()` call got there...

svaarala:
Human error I guess? :) I converted some `duk_require_hobject()` calls to `duk_known_hobject()` when the value is known to be an object (e.g. by a prior coercion or check). Here that assumption was incorrect.

svaarala:
Some post-mortem notes:
- Core bug was an incorrect assumption when converting call sites to use `duk_known_hobject()` where (presumably) safe.
- The conversion function `duk_known_hobject()` has an assert which was correctly triggered.
- The bug was not caught by automatic test runs because no testcase exercised that case.

Even with the relatively high effort on test cases, the test coverage wasn't enough to catch this. There are exhaustive argument type tests for many built-in bindings but not for this specific call.

There's maybe some room for improvement here. Exhaustive argument type tests serve two purposes (at least):
- Documenting the expected behavior, e.g. TypeError throwing or coercion.
- Ensuring no assertions are triggered and no memory unsafe behavior happens.

The former needs manual test writing because the expected behavior is context dependent. It might be possible to develop some helpers or some shorthands to compactly run through the available types for each binding. Now the process is quite tedious.

The latter could be autogenerated based on some interface descriptions. For example, if there was metadata to say that a certain built-in function takes 3 arguments, and a shared init code which e.g. created the necessary object instances, it'd be straightforward to create test calls for a bunch of test argument values and simply test that no unsafe behavior occurs (ignoring the functional behavior otherwise) -- this is ultimately more important than functionally correct behavior. This kind of automatic testing would have caught this particular issue.

Instead of automatic generation simple test utilities might also be enough, e.g.:

```js
// in test-something.js
var something = new Something();
methodArgumentSafetyTest(something, 'method', 3 /*three args*/);
print('safety tests passed');
```

The helper would run a test value set through all the argument slot, in all combinations, and print nothing regardless of TypeErrors etc. The intent would simply be to trigger assertions or memory unsafe behavior.

PR #1291, opened 2017-01-15
Follow-ups to object property hash rework

svaarala:
- Keep probe step as 1?
- 16-bit hash part (works well with slot cache)?
- Drop hash parts on emergency GC?


svaarala:
Not merged, moved to future work.

PR #1294, opened 2017-01-15
Fix textencoder uninitialized variable

### LINKED ISSUE #1293: Bug in TextEncoder: use of uninitialized variable
Kept getting "invalid args" when calling TextEncoder#encode() with valid input (i.e. a string), so I decided to investigate.  Got this back from MSVC:
```
Run-Time Check Failure #3 - The variable 'final_len' is being used without being initialized.
```

Offending line of code is here:
https://github.com/svaarala/duktape/blob/v2.0.0/src-input/duk_bi_encoding.c#L430

Not sure what would cause this, but the variable must indeed be uninitialized, since MSVC reports its value as 14757395258967641292, or in hex `0xcccccccccccccccc`, which is the sentinel for an uninitialized stack variable.

`len` is zero, so the input seems to be an empty string.

svaarala:
Final_len was uninitialized if input string was empty. Doesn't cause issues on Linux gcc or regression test coverage, but did cause issues with MSVC.

Fixes #1293.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2418,6 +2418,9 @@ Planned
 * Fix memory unsafe behavior in Duktape 2.0.0 String.prototype.repeat()
   (GH-1270)
 
+* Fix memory unsafe behavior in new TextEncoder().encode('') (applied to an
+  empty string) which manifested at least on MSVC (GH-1293, GH-1294)
+
 * Fix incorrect exponentiation operator behavior which happened at least on
   Linux gcc 4.8.4 with -O2 (not -Os) (GH-1272)
 
--- a/src-input/duk_bi_encoding.c
+++ b/src-input/duk_bi_encoding.c
@@ -362,7 +362,6 @@ DUK_INTERNAL duk_ret_t duk_bi_textencoder_prototype_encode(duk_context *ctx) {
 	DUK_ASSERT_TOP(ctx, 1);
 	if (duk_is_undefined(ctx, 0)) {
 		len = 0;
-		final_len = len;
 	} else {
 		duk_hstring *h_input;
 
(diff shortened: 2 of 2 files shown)


svaarala:
@fatcerberus Since the original issue is not caught by automatic regression testing (compiler luck I guess), could you check if the fix here works in MSVC and your specific options?

fatcerberus:
Sure, I'll test it out in a bit.

fatcerberus:
Yep, that fixed it.  Thanks :)

What I can't figure out is that "invalid args" error.  If anything it should just create a huge unbacked buffer like you say in https://github.com/svaarala/duktape/issues/1293#issuecomment-272705932, but instead it's throwing an error.

fatcerberus:
Huh, this bug produces really weird symptoms.  On x86 the thrown error is different:

```
C:\src\spectacles-i>C:\src\minisphere\msw\cell -bd dist
Cell 5.0a1 Sphere packaging compiler (x86)
the JavaScript-powered compiler for Sphere v2
(c) 2015-2017 Fat Cerberus

evaluating Cellscript.js...
building targets...
    transpile script @/lib/map/index.js
        TypeError: buffer required, found [object Uint8Array] (stack index 1)
        @ [#/cell_modules/transpile.js:27]
writing game.json... OK.
```

svaarala:
Thanks for reporting the bug :) Always difficult to find if automatic regressions don't catch them. One more case where running the test set automatically on Windows too would be nice. Now that's a bit easier tooling-wise so maybe that coverage can be achieved in the coming months.

Just to be sure, was that x86 issue also fixed here?

fatcerberus:
```
Cell 5.0a1 Sphere packaging compiler (x86)
the JavaScript-powered compiler for Sphere v2
(c) 2015-2017 Fat Cerberus

evaluating Cellscript.js...
building targets...
    transpile script @/lib/map/index.js
writing game.json... OK.
```

Indeed it is. 😄

fatcerberus:
@svaarala I'd like to do a release soon (in the next week or so) with ES2015 support; what's the timeframe on Duktape 2.0.1?

svaarala:
I can do it anytime - I've just been waiting for any small fixes to crop up for a bit. But it's not a very heavy process to release a 2.0.2 if more fixes crop up.

fatcerberus:
Alright, I'll keep an eye out and let you know when I'm ready to release.  For now I can just build against master in-house.

svaarala:
Sounds good - it takes a few hours to do a release so couple of days notice is more than enough.

PR #1296, opened 2017-01-16
Document profile guided optimization

svaarala:
For performance sensitive environments using profile guided optimization (PGO) is very useful, so reference it from the website and performance sensitive environment documentation.

Also add `duk-pgo.O2` and `duk-perf-pgo.O2` test targets to the Makefile. The difference to plain GCC -O2 is almost 20% on `tests/perf/test-mandel.js`:

```
# -O2 without PGO.
$ time ./duk.O2 tests/perf/test-mandel.js
[...]

real    0m3.061s
user    0m3.059s
sys     0m0.000s

# -O2 with PGO.
$ time ./duk-pgo.O2 tests/perf/test-mandel.js
[...]

real    0m2.488s
user    0m2.487s
sys     0m0.000s
```

On Google V8 benchmark the plain -O2 result is around 305, while -O2 + PGO is around 376.

The training set for duk-pgo.O2 is quite dummy and doesn't involve e.g. regexps, so these results are probably possible to improve on by using a better profile generation test set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.o
 *.strip
 *.pyc
+/*.gcda
 /_*
 /duk.*
 /duk
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,13 @@ CONFIGOPTS_DEBUG_ROM=--rom-support --rom-auto-lightfunc --option-file util/maked
 CONFIGOPTS_EMDUK=-UDUK_USE_FASTINT -UDUK_USE_PACKED_TVAL
 CONFIGOPTS_DUKWEB=--option-file util/dukweb_base.yaml --fixup-file util/dukweb_fixup.h
 
+# Profile guided optimization test set.
+PGO_TEST_SET=\
+	tests/ecmascript/test-dev-mandel2-func.js \
+	tests/ecmascript/test-dev-totp.js \
+	tests/perf/test-fib.js \
+	tests/ecmascript/test-regexp-ipv6-regexp.js
+
 # Compiler setup for Linux.
 CC	= gcc
 GXX	= g++
@@ -193,6 +200,7 @@ checksetup:
 # which we don't want to delete by default with 'clean'.
 .PHONY:	clean
 clean:
+	@rm -f *.gcda
 	@rm -rf dist/
 	@rm -rf prep/
 	@rm -rf site/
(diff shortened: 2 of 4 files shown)


svaarala:
@fatcerberus It would be interesting to hear how MSVC PGO (https://msdn.microsoft.com/en-us/library/e7k32f4k.aspx) affects Minisphere -- in case you haven't tried that yet. It seems MSVC won't be getting much in the way of static hot/cold attributes and PGO is recommended instead.

svaarala:
The -O2+PGO build can also be smaller (this will obviously depend on the compiler):

```
 356973	   1468	    328	 358769	  57971	duk.O2.pgo
 380836	   1436	    216	 382488	  5d618	duk.O2.master
```

fatcerberus:
@svaarala I actually haven't tried PGO yet, mainly because I haven't felt the need.  minisphere isn't really CPU-bound at all; by default it even throttles to 60fps (this is configurable by game code of course) so you have ~16ms to get everything done between frames, which in almost all cases is more than enough.  I honestly don't think I would benefit much from a JIT engine.  Cell would of course, but I'm not going out of my way to embed V8 just for the build tool ;-)

svaarala:
Ok, maybe that transpilation pass then? That seemed to be a performance bottleneck in your case?

I don't know how complex using PGO is with MSVC but with GCC it was trivial: https://github.com/svaarala/duktape/blob/master/Makefile#L373-L386.

First pass with -fprofile-generate, then just run the program, and second pass with -fprofile-use.

svaarala:
To be clear, I'm interested in the impact of PGO with MSVC in general. But since you have an actively developed codebase and there was discussion about a Minisphere benchmark run, I thought it might be relatively easy for you to try that out.

I'll look into MSVC PGO at some point anyway (just for documenting the bare essentials) and I can test it then too.

fatcerberus:
I'm still investigating how to use PGO with MSVC, but here's the result of a profiling run on Cell (note: optimized with `/O1` "Minimize Size") while transpiling ES6 code with Babel:

Imgur

The profile looks quite a bit different with Babel compared to TypeScript.  With TypeScript, I found that I spent a large portion of my execution time in `duk__vm_arith_add()` (i.e. string concatenation), while here, that function is nowhere to be found.  Bytecode dispatch dominates of course, but property retrieval accounts for a decent chunk too.

fatcerberus:
No idea why I was optimizing Cell for size - switching to `/O2` vastly improved my transpilation speeds :)

svaarala:
How much is "vast"? :-)

Re: that string concatenation issue, how dominating is `duk__vm_arith_add()` in practice? Just ballpark-wise; in the pathological cases it can be something like 90+%.

fatcerberus:
Just eyeballing it, switching from size to speed optimization seemed like about a 2x difference.  I wasn't exaggerating :)

I'd have to find the post with the TypeScript profiling results again, but if I remember correctly, `duk__vm_arith_add()` was maybe about 30% of the execution time?  Nowhere near 90% though.

svaarala:
Ok, 30% sounds like a case where it's maybe doing some string concatenation using `res += x` but it's not one of the pathological cases like:

```js
var res = '';
for (var i = 0; i < 1e5; i++) {
    res += String.fromCharCode(i);
}
```

This takes 10 seconds on this x64 laptop, and interns ~4.5GB of string data. The equivalent:

```js
var res = [];
for (var i = 0; i < 1e5; i++) {
    res.push(String.fromCharCode(i));
}
res = res.join('');
```

takes 0.15 seconds and only interns ~200kB of string data.

Real code mostly falls in between (though that pathological case does happen in char-by-char accumulation). Building a result from snippets (say, XML tags) is a common case which is in between.

To be clear, I'm not saying the application code is at fault -- one might reasonably expect char-by-char concatenation to be faster than using a temporary array. But currently it does lead to pathological behavior (same as in Lua, incidentally, and where I've been bitten by it :-).

svaarala:
@fatcerberus If you want to see how much more performance you can achieve, you could try prepending:

    --option-file config/examples/performance_sensitive.yaml

to your configure.py run. That enables a bunch of options to improve performance. It's not a dramatic difference but can be 10-20% too.

fatcerberus:
Thanks, I'll try that.  I'm already using fastints but I see there's a bunch of other low-level optimizations in that file, so it might be worth a try.

PR #1306, opened 2017-01-18
Add closure microbenchmark (inner functions)

svaarala:
Add a microbenchmark for exercising a function with a lot of inner functions. This exercises closure creation and entry to a function with a lot of inner function, which doesn't currently perform very well.


svaarala:
Just quickly testing:
- 1.5.0: 2.4 seconds
- 2.0.0: 1.7 seconds

So there was some improvement in 2.0.0 release (I did optimize a few code paths in closure creation there), but closure creation would benefit from an actual rewrite.

PR #1312, opened 2017-01-19
Make symbol checks unlikely

svaarala:
Usually symbol checks are in code paths expecting a string; make these unlikely.  For code paths where a symbol is expected, make the check likely.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2532,8 +2532,8 @@ Planned
 * Miscellaneous compiler warning fixes (GH-1358)
 
 * Miscellaneous performance improvements: more likely/unlike attributes and
-  hot/cold function splits (GH-1308, GH-1309), integer refzero-free-running
-  flag (instead of a flag bit) (GH-1362)
+  hot/cold function splits (GH-1308, GH-1309, GH-1312), integer
+  refzero-free-running flag (instead of a flag bit) (GH-1362)
 
 * Miscellaneous footprint improvements: more compact duk_hobject allocation
   (GH-1357), explicit thr->callstack_curr field for current activation
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -1598,7 +1598,7 @@ DUK_INTERNAL duk_hstring *duk_get_hstring(duk_context *ctx, duk_idx_t idx) {
 
 DUK_INTERNAL duk_hstring *duk_get_hstring_notsymbol(duk_context *ctx, duk_idx_t idx) {
 	duk_hstring *res = (duk_hstring *) duk__get_tagged_heaphdr_raw(ctx, idx, DUK_TAG_STRING);
-	if (res && DUK_HSTRING_HAS_SYMBOL(res)) {
+	if (DUK_UNLIKELY(res && DUK_HSTRING_HAS_SYMBOL(res))) {
 		return NULL;
 	}
 	return res;
@@ -1920,7 +1920,7 @@ DUK_EXTERNAL duk_size_t duk_get_length(duk_context *ctx, duk_idx_t idx) {
 	case DUK_TAG_STRING: {
 		duk_hstring *h = DUK_TVAL_GET_STRING(tv);
 		DUK_ASSERT(h != NULL);
-		if (DUK_HSTRING_HAS_SYMBOL(h)) {
+		if (DUK_UNLIKELY(DUK_HSTRING_HAS_SYMBOL(h))) {
 			return 0;
 		}
 		return (duk_size_t) DUK_HSTRING_GET_CHARLEN(h);
@@ -2461,7 +2461,7 @@ DUK_INTERNAL void duk_push_class_string_tval(duk_context *ctx, duk_tval *tv) {
 		duk_hstring *h;
 		h = DUK_TVAL_GET_STRING(tv);
 		DUK_ASSERT(h != NULL);
-		if (DUK_HSTRING_HAS_SYMBOL(h)) {
+		if (DUK_UNLIKELY(DUK_HSTRING_HAS_SYMBOL(h))) {
 			stridx = DUK_STRIDX_UC_SYMBOL;
 		} else {
 			stridx = DUK_STRIDX_UC_STRING;
@@ -2699,7 +2699,7 @@ DUK_INTERNAL duk_hstring *duk_to_hstring_acceptsymbol(duk_context *ctx, duk_idx_
 	duk_hstring *ret;
 	DUK_ASSERT_CTX_VALID(ctx);
 	ret = duk_get_hstring(ctx, idx);
-	if (ret && DUK_HSTRING_HAS_SYMBOL(ret)) {
+	if (DUK_UNLIKELY(ret && DUK_HSTRING_HAS_SYMBOL(ret))) {
 		return ret;
 	}
 	return duk_to_hstring(ctx, idx);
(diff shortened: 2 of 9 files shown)


fatcerberus:
Maybe I'm wrong but this seems like an abuse of likely/unlikely?  When I researched these a while ago, I remember reading you're only supposed to tag a check as likely/unlikely when you're nearly 100% certain of the outcome--that is, when the alternative is an extreme edge case.  Validation doesn't seem to fit that criteria.

You'd know better of course, since you're the one running the performance benchmarks. :)

svaarala:
> [...] you're only supposed to tag a check as likely/unlikely when you're nearly 100% certain of the outcome [...]

Hmm, can you provide a link for that? This may be compiler specific but I don't think that's the case with e.g. GCC; the hints seem to provide information about "likely" and "unlikely", not "nearly 100%" or "nearly 0%". GCC says this for __builtin_expect():

> You may use __builtin_expect to provide the compiler with branch prediction information. [...] The return value is the value of exp, which should be an integral expression. The semantics of the built-in are that it is expected that exp == c. 

With profiling these hints are ignored and the profile measurements are used instead. The practical effect here, for example, is that the string path is made the branch predicted outcome and the symbol path is moved out of the way to improve code locality. At least to my inspection the asm output looks better with the hints in place.

It's certainly easy to write a compiler where likely/unlikely would behave differently. But because they're macro based, it's easy to change what they do. The current Duktape macros indicate, as they're named, that the case is likely (say 80+%), not that it's virtually certain.

It would actually be nice to have separate hints for: almost 100% certain vs. 80+% certain. The former would be used for error paths where the error path could very well be located far away from the main code. The latter would be useful for e.g. the string/symbol case where strings are much more prevalent (except in symbol microbenchmarks) and favoring strings is better, but symbols still do occur in "normal" code and it makes sense to locate the symbol code nearby but still outside of the main path (which is the intended effect here).

svaarala:
So one improvement for the Duktape macros might to provide e.g. `DUK_LIKELY()` and `DUK_VIRTUALLY_CERTAIN()` (with a shorter name in practice of course). The latter could be used for error paths and the like, which are rarely taken. The former would be useful for denoting ordinary checks where the likely outcome is relatively easy to know statically, without trying to imply it's an "abnormal" case.

The macros could then be targeted at the compiler as appropriate. If the 80+% case worked badly with likely/unlikely attributes, that could then be made a no-op, but the likely/unlikely attributes could still be kept for the error handling paths (which are in the "virtually certain") category.

But unless there's a concrete case where these two cases actually perform very differently, I'd hesitate to add this distinction because without a concrete benefit it'd be an academic exercise.

fatcerberus:
It's been a while, so my memory is fuzzy but I think I might be remembering this Stack Overflow question:
http://stackoverflow.com/questions/7346929/why-do-we-use-builtin-expect-when-a-straightforward-way-is-to-use-if-else

In particular the second answer.

SO ansewrs are of course not authoritative though :-)

svaarala:
@fatcerberus This string/symbol case specifically is a good example of a case where I think branch hints are useful. Typically compilers have good heuristics for loop and such, but in this case there's no reasonable heuristic to allow the compiler to figure out that it's OK to make symbols perform 100% worse if strings perform 10% better (just for that check, not at the global level of course). For me that trade-off would be a no-brainer because strings are so much more prevalent.

Compiler heuristics vary, and apparently it's possible to ensure a good outcome by simply swapping the if-else cases with one another (one being "favored" by default). But I very much prefer an explicit hint, because it communicates the intent better when reading the source.

fatcerberus:
Fair enough.

By the way, also note the advisory at the bottom here (I'd hope Linux kernel developers would know what they're talking about :):
> You should use it only in cases when the likeliest branch is very very very likely, or when the unlikeliest branch is very very very unlikely.

svaarala:
> In particular the second answer.

I don't read that as saying that you should only use branch hints for the nearly 100% cases though? :)

The hints do in fact also improve cold path code, because if the expectation is correct, the rare path doesn't get fetched.

fatcerberus:
> I don't read that as saying that you should only use branch hints for the nearly 100% cases though? :)

Like I said, chalk it up to a fuzzy memory.  I guess my recollection got distorted since then :)

svaarala:
Whenever I add hints I try to look at the generated asm to see if it's improved. This is of course not a 100% approach because the results can vary between compilers, compiler versions, architectures, etc.

So really the only good way would be for there to be different hints for different likelihoods, and those would then be mapped to the compiler/version/architecture as appropriate. For example, if "likely" causes a very heavyweight outcome (emitting explicit branch prediction hints that are expensive if they're wrong) then the hints would only be enabled for virtually certain cases.

This would also improve the code readability a bit if the hint macros would convey the likelihood more explicitly (as said above, error paths and normal but less common paths are different conceptually).

But, on the whole, I'm not sure if this is worth the effort. Performance conscious environments should use PGO anyway, and e.g. MSVC seems to take the profiling approach only (?).

fatcerberus:
> ...it's OK to make symbols perform 100% worse if strings perform 10% better

Definitely agree here.  Considering how branch hints seem to work based on my research (avoiding a jump for the likely case), I imagine this actually is an improvement.  Since the unlikely path is an error path anyway, performance there isn't as important

> But, on the whole, I'm not sure if this is worth the effort. Performance conscious environments should use PGO anyway, and e.g. MSVC seems to take the profiling approach only (?).

Speaking of which, I still haven't figured out how to use PGO in MSVC.  MS didn't make it easy to find :-/

svaarala:
One useful data point here may be that the difference between a PGO and non-PGO -O2 build was about 70 points. By adding a set of likely/unlikely hints I managed to reduce the gap to 60 points, i.e. gain 10 points on the score.

That's still far from how well PGO works, so I can only really recommend PGO if one cares about performance a great deal. Getting close to the PGO result with static hints would probably be very difficult. Still, it's nice if the default non-PGO build behaves reasonably well :)

svaarala:
> [...]  Considering how branch hints seem to work based on my research (avoiding a jump for the likely case), I imagine this actually is an improvement. Since the unlikely path is an error path anyway, performance there isn't as important.

Hmm, I was actually referring to paths like:

```c
if (DUK_HSTRING_HAS_SYMBOL(h)) {
    return "symbol";
} else {
    return "string";
}
```

Here both cases are non-error-paths, but I'd still take a 10% improvement in the string path if it made the symbol path 100% worse (just locally for this check), because strings outnumber symbols by a huge margin in typical programs.

For error checking and argument validation and such, the hints are a no-brainer IMO, and that's one typical idiom in the Linux kernel too AFAIK (it's been years since I wrote any kernel code though :-).

fatcerberus:
Ah, I see.  I think I ultimately agree there too - one doesn't typically apply `typeof` to a symbol, in my experience symbols are "special" enough that you usually already know if you have one.

svaarala:
> Ah, I see. I think I ultimately agree there too - one doesn't typically apply typeof to a symbol, in my experience symbols are "special" enough that you usually already know if you have one.

Right, but I meant just generically all the "normal" paths where strings and symbols are differentiated like that. They're not error handling or validation as such, but just different behaviors in string coercion, typeof, ToPropertyKey() coercion, etc.

svaarala:
So these paths are the ones where I would like to statically annotate in code that "this path is more important" or "this path happens in most cases, but the other path is normal too". But the distinction from almost-never-executed error paths is probably too fine one for practical compiler attributes (or ISAs for that matter :).

PR #1317, opened 2017-01-21
Expand duk_push_heapptr() pointer asserts

svaarala:
- Change the finalize_list check, see discussion below
- Add a refzero_list check, see discussion below
- Check that a string is present in the stringtable
- Check that a non-string is present in heap_allocated (or finalize_list/refzero_list)

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2408,6 +2408,11 @@ Planned
 * Improve duk_hstring array index handling performance when
   DUK_USE_HSTRING_ARRIDX is disabled (GH-1274)
 
+* Improve duk_push_heapptr() assert validation to include checks that the
+  pointer is only allowed in finalize_list or refzero_list if currently being
+  finalized, and must otherwise be in either the string table (for strings)
+  or heap_allocated (non-strings) (GH-1317)
+
 * Fix a few incorrect asserts related to reference count triggered finalizer
   execution; the functionality itself was correct in these cases but a few
   asserts were too strict (GH-1318)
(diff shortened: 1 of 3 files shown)


svaarala:
With current finalization code the asserts in the pull may be too strict now:
- When an object is finalized by refcounting, it is still in refzero_list.
- When an object is finalized by mark-and-sweep, it is still in finalize_list.

In these cases the argument pointer is OK for duk_push_heapptr() in practice. But, only for the object being finalized, not other objects in these lists.

It's technically an API violation however, because the heapptr has already become unreachable and the API requirement for duk_get_heapptr() -> duk_push_heapptr() is that the object is reachable 100% of the time between the two. However, if calling code is 100% certain that duk_get_heapptr() of the finalizer argument matches the duk_push_heapptr() argument (without doing the duk_get_heapptr() explicitly), the push works.

svaarala:
Because there's a corner case where the duk_push_heapptr() is legitimately (well, sort of, see above) in finalize_list, this pull:
- Doesn't assert against that case happening.
- If the object is found in finalize_list, the other lists are not scanned because the object won't be in heap_allocated or any other list.

This could be improved a bit by *also* scanning the other lists and ensure it's not *also* there (which would indicate corrupted linked lists).

svaarala:
Ok, rewrote the asserts a bit so that the asserts also ensure:
- An object doesn't appear twice in a list (looping, corrupted linked list).
- An object doesn't appear in two lists (corrupted linked list).

svaarala:
Hmm, it occurred to me that I could make the assertions a bit stronger by checking that:
- If the ptr argument is in refzero_list, allow it if the object also has the DUK_HOBJECT_FLAG_FINALIZED set. This indicates that the finalizer is running for that object.
- If the ptr argument is in finalize_list, same check.

In both cases, once the finalizer returns the object is removed from the list (refzero_list or finalize_list) and is either freed or re-queued into heap_allocated with the FINALIZED flag cleared. Because refzero/mark-and-sweep finalizer nesting is now allowed, at most two objects (one per list) should have FINALIZED set.

PR #1319, opened 2017-01-21
Clarify finalizer rescue behavior in Guide

svaarala:
When mark-and-sweep runs finalizers, it may involve a set of finalizable objects in an arbitrary graph. Duktape calls finalizers for each object in the finalizable set in an unspecified order, and rescuing an object X may rescue other objects -- even if the finalizers for those objects has already executed.

Clarify the Guide summary which implied that only the object's own finalizer could rescue an object (which is the typical case but not always true).

### DIFF
--- a/website/guide/finalization.html
+++ b/website/guide/finalization.html
@@ -60,12 +60,17 @@ finalizer are also silently ignored.</p>
 <li>Finalizers are also executed for all remaining objects, regardless of
     their reachability status, when a Duktape heap is destroyed.</li>
 <li>A finalizer is called exactly once, at the latest when the heap is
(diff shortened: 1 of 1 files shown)


svaarala:
@fatcerberus Feedback and concrete wording suggestions welcome :-)

It's difficult to summarize a complex issue while remaining short, and 100% exact ;)

fatcerberus:
I'm at work now, but I'll look at this later.

svaarala:
Sure, no hurry :)

svaarala:
@fatcerberus I'll merge this now - if you have suggestions for improvements, I'll open a separate pull for that.

PR #1320, opened 2017-01-22
Fix ROM object incref with slow refcount default

svaarala:
With slow refcount default, duk_heaphdr_incref() didn't include a "readonly" check as intended.

Reported by @nkolban.


svaarala:
Repro: ajduk-rom, but DUK_USE_FAST_REFCOUNT_DEFAULT: false.

PR #1323, opened 2017-01-22
Minor improvements to heap queueing code

svaarala:
- [x] Reduce pointer compression calls when removing an object back to heap_allocated
- [x] Simplify heap_allocated vs. prev check in mark-and-sweep sweep phase (both are NULL or not NULL simultaneously)
- [x] Add a few more NULL asserts for queue head position cases
- [x] Manual assertion test case run
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2417,6 +2417,9 @@ Planned
   finalized, and must otherwise be in either the string table (for strings)
   or heap_allocated (non-strings) (GH-1317)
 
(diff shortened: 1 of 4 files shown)


svaarala:
The improvement with pointer compression is from 9 enc16/dec16 operations to 4 operations per each `duk_heap_remove_any_from_heap_allocated()` call. Without pointer compression the minor change has essentially no effect.

PR #1324, opened 2017-01-22
Add String.prototype.{startsWith,endsWith}

svaarala:
- [x] Add String.prototype.startsWith and String.prototype.endsWith
- [x] Add testcase coverage
- [x] Releases entry


fatcerberus:
Oh goody, I just started using these in Cell (thanks to a *very* comprehensive ES6 polyfill), but it's great to have them available natively.

PR #1326, opened 2017-01-22
Add string.prototype.includes()

svaarala:
Add support for String.prototype.includes():

- [x] Rebase after #1324 is merged
- [x] Add the binding
- [x] Testcase coverage
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2395,7 +2395,8 @@ Planned
   which uses a 2^N size and a bit mask instead of a prime size and a MOD;
   use a hash table more eagerly than before (GH-1284)
 
-* Add ES2015 String.prototype.{startsWith,endsWith} (GH-1324)
+* Add ES2015 String.prototype.{startsWith,endsWith,includes}() (GH-1324,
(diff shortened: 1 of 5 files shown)


fatcerberus:
Can this share code with `String.prototype.indexOf()`?

svaarala:
I didn't do any real implementation work yet, but it should be able to do that. The argument check / coercion behavior is slightly different, but a shared helper for the main scanning operation could (and should) be shared. I'll finish startsWith() and endsWith() first and then look into this one.

svaarala:
The return value convention also differs (1/-1 vs. true/false).

svaarala:
That search helper doesn't also perform well now (it's size optimized), and I'll maybe improve it in a separate pull. Basically the forward/backward scan modes should be outside the loop (which is pretty obvious for performance), and the `.includes()` variant doesn't need to track the character position at all.

svaarala:
Also the pointer limit checks can be optimized differently depending on scan direction. But, separate pull.

PR #1334, opened 2017-01-27
Fix 'duk' bytecode loading

### LINKED ISSUE #1333: TypeError: cannot string coerce Symbol
There is an error in version 2.0 of compiled code.
I run an example

```
    examples\cmdline\cmdline.exe -c mandel.bin  mandel.js
    examples\cmdline\cmdline.exe  mandel.bin    
```

And there is an error:

```
TypeError: cannot string coerce Symbol
    at [anon] (duk_api_stack.c:2588) internal
error in executing file mandel.bin
```

In version 1.6 all worked without errors.

svaarala:
Bytecode loading in 'duk' created a buffer via a string intermediary which caused a ToString() conversion; the bytecode prefix gets interpreted as a symbol value preventing the conversion. Fix by pushing a buffer value directly.

Fixes #1333.

### DIFF
--- a/examples/cmdline/duk_cmdline.c
+++ b/examples/cmdline/duk_cmdline.c
@@ -239,8 +239,9 @@ static duk_ret_t wrapped_compile_execute(duk_context *ctx, void *udata) {
 
 	if (src_data != NULL && src_len >= 2 && src_data[0] == (char) 0xff) {
 		/* Bytecode. */
-		duk_push_lstring(ctx, src_data, src_len);
-		duk_to_buffer(ctx, -1, NULL);
+		void *buf;
+		buf = duk_push_fixed_buffer(ctx, src_len);
+		memcpy(buf, (const void *) src_data, src_len);
 		duk_load_function(ctx);
 	} else {
 		/* Source code. */


examples/cmdline/duk_cmdline.c
```diff
@@ -239,8 +239,9 @@ static duk_ret_t wrapped_compile_execute(duk_context *ctx, void *udata) {
 
 	if (src_data != NULL && src_len >= 2 && src_data[0] == (char) 0xff) {
 		/* Bytecode. */
-		duk_push_lstring(ctx, src_data, src_len);
-		duk_to_buffer(ctx, -1, NULL);
+		void *buf;
+		buf = duk_push_fixed_buffer(ctx, src_len);
+		memcpy(buf, (const void *) src_data, src_len);
```
fatcerberus: For the future, maybe it would be nice to have a convenience API to do this in one step.

svaarala: Yep, there's an internal use for such a call (though there weren't enough call sites to justify it based on footprint alone). Also such a call avoids unnecessary zeroing.

PR #1341, opened 2017-01-30
Avoid log2(), log10(), cbrt(), trunc() on Android

### LINKED ISSUE #1325: Duktape shouldn't use `log2` on Android even with C99 detected
After updating duktape to 2.0, we started getting build failures on Android:
`duk_bi_math.c:120: error: undefined reference to 'log2'`

Our projects are built with `-std=c99`, but googling indicates that Android doesn't support log2 regardless of the chosen C standard:
http://stackoverflow.com/questions/11080118/does-android-support-log2

As a quick hack, I updated the `#if` around the `log2` stuff to:
`#if (defined(DUK_F_C99) || defined(DUK_F_CPP11) || (defined(_MSC_VER) && (_MSC_VER >= 1800))) && !(defined(ANDROID)||defined(__ANDROID__))`

Everything works after that.

svaarala:
Fixes #1325.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2502,6 +2502,8 @@ Planned
 
 * Fix 'duk' command line bytecode load error (GH-1333, GH-1334)
 
+* Avoid log2(), log10(), cbrt(), and trunc() on Android (GH-1325, GH-1341)
+
 * Compiler warning fix for using DUK_UNREF() on a volatile argument (GH-1282)
 
 * Add DUK_HOT() and DUK_COLD() macros, and use them for a few internal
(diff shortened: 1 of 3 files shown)


svaarala:
@HouQiming I'd appreciate if you could test I got the fix right :)

fatcerberus:
Good thing I implemented all those workarounds. :)

svaarala:
That's the good part of supporting C89: it provides a baseline for environments which are C99 but not all the way.

HouQiming:
Just saw this. I'm on vacation right now and don't have an Android machine
handy. I'll test it once I get back to work.

Thanks for the fix!

svaarala:
@HouQiming Sounds good, have a good vacation :)

HouQiming:
Just tested it. It works :)

PR #1347, opened 2017-02-03
Fix recursive mark-and-sweep entry

svaarala:
- [x] Repro testcase
- [x] Fix recursive mark-and-sweep entry from refzero triggered voluntary GC, see https://github.com/svaarala/duktape/issues/1311#issuecomment-277299860
- [x] Reject recursive mark-and-sweep explicitly rather than just asserting for it
- [x] Schedule fixes for 1.7.0 and 2.0.2
- [x] Releases entry

The latter change is necessary so that, for example, if a native finalizer calls duk_gc(), that won't cause recursive entry into mark-and-sweep.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2460,6 +2460,15 @@ Planned
 * Improve lexical scope handling performance by adding internal duk_hdecenv
   and duk_hobjenv structures (previously generic objects were used) (GH-1310)
 
+* Fix a garbage collection bug where a finalizer triggered by mark-and-sweep
+  could cause a recursive entry into mark-and-sweep (leading to memory unsafe
+  behavior) if the voluntary GC trigger counter dropped to zero during
+  mark-and-sweep finalizer execution (GH-1347)
+
+* Fix a garbage collection bug where a call into duk_gc() from a mark-and-sweep
+  triggered finalizer could cause recursive entry into mark-and-sweep (leading
+  memory unsafe behavior) (GH-1347)
+
 * Fix a duk_push_heapptr() finalize_list assertion issue caused by the
   internal heap->finalize_list being (intentionally) out-of-sync during
   mark-and-sweep finalizer execution; this has no functional impact but
--- a/src-input/duk_heap_markandsweep.c
+++ b/src-input/duk_heap_markandsweep.c
@@ -1088,6 +1088,11 @@ DUK_INTERNAL duk_bool_t duk_heap_mark_and_sweep(duk_heap *heap, duk_small_uint_t
 	duk_size_t tmp;
 #endif
 
+	if (DUK_HEAP_HAS_MARKANDSWEEP_RUNNING(heap)) {
+		DUK_D(DUK_DPRINT("refuse to do a recursive mark-and-sweep"));
+		return 0;
+	}
+
 	/* XXX: thread selection for mark-and-sweep is currently a hack.
 	 * If we don't have a thread, the entire mark-and-sweep is now
 	 * skipped (although we could just skip finalizations).
--- a/src-input/duk_heap_refcount.c
+++ b/src-input/duk_heap_refcount.c
@@ -445,12 +445,17 @@ DUK_INTERNAL void duk_refzero_free_pending(duk_hthread *thr) {
 	 */
 	heap->mark_and_sweep_trigger_counter -= count;
 	if (DUK_UNLIKELY(heap->mark_and_sweep_trigger_counter <= 0)) {
-		duk_bool_t rc;
-		duk_small_uint_t flags = 0;  /* not emergency */
-		DUK_D(DUK_DPRINT("refcount triggering mark-and-sweep"));
-		rc = duk_heap_mark_and_sweep(heap, flags);
-		DUK_UNREF(rc);
-		DUK_D(DUK_DPRINT("refcount triggered mark-and-sweep => rc %ld", (long) rc));
+		if (DUK_HEAP_HAS_MARKANDSWEEP_RUNNING(heap)) {
+			DUK_D(DUK_DPRINT("mark-and-sweep in progress -> skip voluntary mark-and-sweep now"));
+		} else {
+			duk_bool_t rc;
+			duk_small_uint_t flags = 0;  /* not emergency */
+
+			DUK_D(DUK_DPRINT("refcount triggering mark-and-sweep"));
+			rc = duk_heap_mark_and_sweep(heap, flags);
+			DUK_UNREF(rc);
+			DUK_D(DUK_DPRINT("refcount triggered mark-and-sweep => rc %ld", (long) rc));
+		}
 	}
 #endif  /* DUK_USE_VOLUNTARY_GC */
 }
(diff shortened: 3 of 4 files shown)


fatcerberus:
@svaarala Should this be tagged security, since it causes segfaults?

fatcerberus:
This *seems* to fix the segfaults I was seeing.  But that begs the question: Did the trigger conditions for voluntary GC change in recent builds?  I can get master to crash within a few seconds consistently, while I can't crash v2.0.1 at all.  So even if the bug is present in previous versions, it seems like something's changed in master that made it more likely to surface.

svaarala:
I don't think the trigger conditions have changed so I don't know what could cause the difference, other than good/bad luck (and maybe some mark-and-sweep/finalizer behavior becoming suitably synchronized to trigger the issue).

svaarala:
I'll try to come up with a repro case to trigger the issue from a standalone test case. We can then see if it triggers on 2.0.1 too. I think it should. I'll get back on this later tonight (6+ hours from now).

fatcerberus:
There's definitely a difference, like I said the crash happens within seconds of walking around on the map in my test game.  I wonder if #1323 might have caused a regression for minisphere.  That's the only merged pull I can find that touched the mark-and-sweep code.

svaarala:
I don't dispute at all that there's an outward difference. But it may be caused by just some luck in synchronizing counters and such; the voluntary GC is counter based so if the counter hits consistently outside of mark-and-sweep, it might never surface. That could be just up to some luck.

svaarala:
Ok that repro was easy: https://github.com/svaarala/duktape/pull/1347/commits/22c2594f1115d8918e7e03f9c48cd250ed399798

It triggers on 2.0.0 and 2.0.1 too.

fatcerberus:
So why would the counters synchronize better in master vs. the releases?  That's what I'm trying to figure out, because the allocation/collection patterns should be the same if nothing changed in Duktape.

svaarala:
Like I said, I don't know -- it could literally be just one refcount operation changed in one place to offset the counter by one. It can be something more clear too, but it's just hard to tell.

fatcerberus:
So I assume the reason it ultimately segfaults is because it tries to free an object twice or something?

svaarala:
I don't know what happens concretely, but if mark-and-sweep is re-entered, all bets are off. It's not really useful to diagnose what happens IMO.

svaarala:
I'll need to rework that repro case a bit to avoid running out of memory. With finalizer execution now inside mark-and-sweep locking, no garbage collection actually happens while finalizers are being executed. I'll change that later so that finalizers run outside of the mark-and-sweep locking, but I want to resolve this issue with a minimal fix first so it can be backported.

fatcerberus:
> It's not really useful to diagnose what happens IMO.

Practically speaking, that's probably true... doesn't stop me from being curious though, I always take time to try to understand a bug even when it's a waste of time.  But I won't force you to do that :)  In any case it's good to see an apparently long-standing bug finally get fixed.

svaarala:
> But I won't force you to do that :)

Nor could you :)

fatcerberus:
Looking at the repro case, it seems that the conditions for the bug are:
* Mark-and-sweep already running
* Finalizer execution meets the criteria for voluntary GC
* Mark-and-sweep entered recursively
* Very bad things happen

Which is kind of odd in my case because none of my finalizers should be creating any new garbage - they all just do this:
https://github.com/fatcerberus/minisphere/blob/v4.4.3/src/engine/pegasus.c#L2124-L2132
https://github.com/fatcerberus/minisphere/blob/v4.4.3/src/shared/api.c#L341-L353

svaarala:
> I always take time to try to understand a bug even when it's a waste of time.

To be clear, I also try to understand *the bug*, which is incorrect recursive mark-and-sweep entry here. But the consequences of the bug are not always meaningful or worth the effort, especially if you've seen the pattern already many times (as is the case here).

For example, if you worked with multithreading and missed a lock somewhere, that bug itself would be interesting and obviously worth understanding. But the consequences can take hundreds of different forms depending on synchronization, and going through them all doesn't really provide much benefit.

svaarala:
> Finalizer execution meets the criteria for voluntary GC [...]

Looking at your finalizer code I think I might know what the difference between 2.0.x and master might be :)

The issue is triggered by any call to refzero queue processing during mark-and-sweep because that code will do a voluntary GC check and may trigger a mark-and-sweep. The refzero queue processing function is entered either through an actual refzero situation *or* by an explicit call to check the queue which is needed when the "NORZ" (no-refzero) macro variants. These are used to simplify atomic refcount updates in e.g. call handling.

So, the difference between 2.0.x and master is (at least) one place where `DUK_REFZERO_CHECK_SLOW()` is called. Looking at the 2.0.x -> master diff, this snippet has been added to `duk_hthread_callstack_unwind()`:

```
+       /* Check for pending refzero entries, many places in the unwind
+        * use NORZ macros.
+        */
+       DUK_REFZERO_CHECK_SLOW(thr);
```

So here we enter refzero queue check even with no garbage created at all. And this code would be executed every time a finalizer finishes and its callstack entry is unwound.

So maybe mystery solved?

svaarala:
Note that the issue exists in 2.0.1 regardless of that added refzero check (which itself is valid) - but it just means it won't trigger unless finalizers create garbage which isn't the case for you.

fatcerberus:
So if that's the case, it must happen as part of unwinding the finalizer itself - and since we're presumably already inside a voluntary mark-and-sweep at that time, the conditions are still set to allow it to be reentered.  Quite insidious :)

fatcerberus:
It'll be interesting to find out if this fixes @harold-b's case too.

svaarala:
Quite interesting indeed :)

I'm not very happy with the current mark-and-sweep, refcounting, and finalizer locking behavior. But like I said I first want to find a way to fix the underlying issue with a minimal backportable fix and then improve at least the finalizer execution to be outside of GC locking.

Mark-and-sweep code itself is in need of partial rewriting, but I'm hoping I could rewrite it to be an incremental GC (allowing much better semi-realtime behavior) instead of incrementally improving the current one. That incremental mark-and-sweep step would make the option of running without refcounting more appealing because there wouldn't be hard stops for GC.

svaarala:
This is unrelated to the longjmp assert @harold-b encountered, but might be related to the issue fixed by disabling voluntary GC (which is indeed an actual "fix" rather than just making the case unlikely).

svaarala:
After reworking the testcase to avoid out-of-memory conditions (which are annoying because the confuse interpretation of a successful test run) the bug no longer triggers without assertions. In other words, with asserts enabled the repro case manages to trigger the assert, but no obvious memory unsafe behavior occurs with asserts disabled.

That's not ideal but still OK because assertion runs are done from time to time (not for every commit) so eventually any regressions will be caught.

svaarala:
@fatcerberus I tested some versions backwards and Duktape 2.0.0 and 2.0.1 cause the assert to trigger but no 1.x version does. I'm not sure why because logically it should. I could probably tweak the testcase to trigger the asserts there too - but this is probably not worth the effort, and I'll backport the fix to 1.7.0 in any case.

fatcerberus:
I assume the reason the redesigned repro case doesn't segfault is because the finalizer doesn't generate any garbage except for the one object, `t`, which presumably gets collected by refzero before mark-and-sweep even sees it.  Therefore mark-and-sweep doesn't touch any objects and as a result no memory unsafe behavior happens.

svaarala:
The finalizer actually generates 10 objects and while refzero runs, nothing is actually refcount collected while mark-and-sweep runs; all garbage created during mark-and-sweep finalizers is actually left up to mark-and-sweep to handle (maybe on a later run).

PR #1354, opened 2017-02-05
Fix global/eval variable redeclaration handling

### LINKED ISSUE #1351: Redeclaring a global variable in eval code resets variable to undefined
https://github.com/svaarala/duktape/issues/1350#issuecomment-277453162

Repro:

```js
global.fn = function() { console.log("the pig ate it"); };
(0, eval)('var fn;');
fn();
```

Node prints the pig ate it. Duktape throws TypeError: undefined not callable.

svaarala:
- [x] Bug testcase
- [x] Implement minimal fix
- [x] Check if DECLVAR flags can be simplified because there are only really two cases: function declaration and var declaration now
- [x] Releases entry
- [ ] Backport to at least 2.0.x, maybe 1.7.0

Fixes #1351.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2500,6 +2500,11 @@ Planned
 * Fix bug in global object environment "provideThis" attribute when using ROM
   objects and DUK_USE_ROM_GLOBAL_INHERIT (GH-1340, GH-1310)
 
+* Fix bug in global/eval code variable redeclaration handling where a
+  plain 'var X;' redeclaration for an existing binding caused 'undefined' to
+  overwrite the existing binding rather than being treated as a no-op
+  (GH-1351, GH-1354)
+
 * Fix 'duk' command line bytecode load error (GH-1333, GH-1334)
 
 * Compiler warning fix for using DUK_UNREF() on a volatile argument (GH-1282)
--- a/debugger/duk_opcodes.yaml
+++ b/debugger/duk_opcodes.yaml
@@ -705,8 +705,6 @@ opcodes:
       - mask: 0x800
         name: accessor
       - mask: 0x1000
-        name: undef_value
-      - mask: 0x2000
         name: func_decl
   - name: DECLVAR_CR
     args:
@@ -723,8 +721,6 @@ opcodes:
       - mask: 0x800
         name: accessor
       - mask: 0x1000
-        name: undef_value
-      - mask: 0x2000
         name: func_decl
   - name: DECLVAR_RC
     args:
@@ -741,8 +737,6 @@ opcodes:
       - mask: 0x800
         name: accessor
       - mask: 0x1000
-        name: undef_value
-      - mask: 0x2000
         name: func_decl
   - name: DECLVAR_CC
     args:
@@ -759,8 +753,6 @@ opcodes:
       - mask: 0x800
         name: accessor
       - mask: 0x1000
-        name: undef_value
-      - mask: 0x2000
         name: func_decl
   - name: REGEXP_RR
     args:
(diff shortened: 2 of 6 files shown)


svaarala:
There's a tentative fix in the pull now, I'll pore over the semantics tomorrow in detail. The flags for DECLVAR can also probably be simplified because there are only two places where the opcode is emitted (for a slow path variable or a slow path function declaration).

fatcerberus:
Will DECLVAR need new flags for `let`/`const`?

fatcerberus:
Oddly enough, this bug *doesn't* affect `with` statements, which seems odd to me since I was pretty sure they used the same scoping mechanism as global code?

```js
var pig = {
	eatiness: 812
};
with (pig) {
	var eatiness;
	// note: `var eatiness = 1208` *does* overwrite the property.
}
console.log(pig.eatiness);  // 812
```

svaarala:
The whole scope/binding handling will need new stuff with let/const, and aiming for that here doesn't really pay off IMO.

svaarala:
The scope semantics track the variable declaration scope and the lexical scope separately. With statement doesn't affect the variable declaration scope, only the lexical scope, so that:

```js
var obj = { quux: 0 };
with (obj) {
    var foo = 123;  // declare 'foo' into global
    bar = 234;  // write 'bar' into global
    quux = 345;  // update obj.quux
}
print('foo' in obj, 'foo' in this);  // this == global
print('bar' in obj, 'bar' in this);
print('quux' in obj, 'quux' in this);
print(JSON.stringify(obj));
```

Rhino and Duktape both output:

```
false true
false true
true false
{"quux":345}
```

V8 output is a bit different because the code doesn't execute in the global scope, but if it did, it would output the same I believe.

In other words, only writes to properties already present in the with target actually go into the target. Here both 'foo' and 'bar' are "unresolvable" references which have special handling in http://www.ecma-international.org/ecma-262/5.1/#sec-8.7.2 step 3.

fatcerberus:
> With statement doesn't affect the variable declaration scope

Hmm... so is the behavior below a bug then?

```js
var foo = { bar: 812 };
with foo {
    var bar = 1208;
}
print(foo.bar);  // 1208
```

fatcerberus:
Node.js seems to agree with the behavior above, so maybe not.

```js
> process.version
'v7.5.0'
> var foo = { bar: 812 };
undefined
> with (foo) { var bar = 1208; }
undefined
> foo.bar
1208
```

fatcerberus:
The `with` statement has some really strange and counterintuitive semantics.  Now I understand why it's banned in strict mode. :)

svaarala:
That's not a bug - the declaration is a no-op because 'bar' already exists in the scope. The assignment to 'bar' also goes to the 'bar' property which already exists in the target.

svaarala:
Or actually, to be precise, the `var` declaration is not a no-op but goes to the current declarative environment (which is outside the "with"). The assignment to the newly declared variable goes to the "with" target though.

For example:

```js
var obj = { bar: 100 };
with (obj) {
    var bar = 123;
}
console.log(obj.bar);
console.log(this.bar);
console.log('bar' in this);
```

prints:

```
123
undefined
true
```

(Node prints 'false' as the last value because the code is not executed as global code.)

svaarala:
Rhino output for the above matches:

```
123
undefined
true
```

fatcerberus:
As I said, it's good that `with` is banned in strict mode. ;-)

I do wish they would come up with a replacement though, maybe `using`:
```js
using (foo) {
    .bar = 1208;
}
```

I used to really love `with` because it makes some code patterns more readable, but the confusing semantics make it not really worth it, and the advantages of strict mode outweigh even the readability argument.

svaarala:
Agree, it's probably best to avoid "with". Your example of using `.foo` as referring to the target would indeed be much clearer.

fatcerberus:
This fixes the Babel transpilation issue for me, thanks. :)

svaarala:
Great, good that you pointed this out :-)

PR #1355, opened 2017-02-07
Minor tweaks to voluntary mark-and-sweep handling

svaarala:
- [x] Remove voluntary GC check entirely from refzero as unnecessary: free operations decrement the counter, and the first alloc operation will check for the counter in any case
- [x] Slightly better asm output for voluntary GC counter check in allocation code (at least on x64)
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2460,10 +2460,14 @@ Planned
 * Improve lexical scope handling performance by adding internal duk_hdecenv
   and duk_hobjenv structures (previously generic objects were used) (GH-1310)
 
+* Remove voluntary GC check from refzero processing; the check is not really
+  necessary because all free operations decrement the voluntary GC counter and
+  all allocs/reallocs check for voluntary GC (GH-1355)
+
 * Fix a garbage collection bug where a finalizer triggered by mark-and-sweep
   could cause a recursive entry into mark-and-sweep (leading to memory unsafe
   behavior) if the voluntary GC trigger counter dropped to zero during
-  mark-and-sweep finalizer execution (GH-1347)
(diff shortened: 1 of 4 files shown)


fatcerberus:
> free operations decrement the counter

This seems odd... shouldn't that have made it impossible for recursive voluntary mark-and-sweeps to happen?

svaarala:
The voluntary GC counter starts at a certain value and is:
- Decremented and checked on every alloc.
- Decremented but not checked on every free.

And previously, it was decremented for every refzero processed object and then checked at the end of refzero processing. This one caused the recursive entry because it lacked a "mark-and-sweep already running" guard. The alloc calls had the guard in place.

But because the voluntary GC counter could drop to zero in either alloc, free, or at the end of refzero (this depends on "luck") it was possible to re-enter it.

svaarala:
@fatcerberus I hope the explanation above answered your question, if not, could you elaborate?

fatcerberus:
No, I understood.  It was just that when you said free ops decrement the counter, I assumed decrement meant "further away from next GC" rather than "closer".

svaarala:
Ok I see what you mean. But yes, the counter always counts down and it essentially just very roughly measures "allocation activity". The counter is initialized to be roughly proportional to the heap object count, so that the larger the heap, the larger the interval between voluntary GCs so that the amortized cost is (very roughly speaking) constant.

svaarala:
@fatcerberus In case you are wondering, AppVeyor jobs are not running for some reason. The last successful run was from 3 weeks ago and the queue doesn't seem to be processed at all at present. Not sure yet what to do about it, maybe just spend the time needed to run an actual Windows test client rather than relying on AppVeyor.

PR #1380, opened 2017-03-03
Add shebang comment parsing

### LINKED ISSUE #1346: Add support for shebang lines
A shebang in the beginning of a script currently causes a syntax error.  It is possible (of course) to preprocess the script and remove the shebang, but it would be nice if Duktape could do it automatically in service to applications like minisphere and Cell that use it as a frontend.

svaarala:
- [x] Add lexer support for accepting a shebang (`#! ...`) line comment on first line, starting from first column
- [x] Config option, disable for low memory targets by default
- [x] API compile option to turn feature on
- [x] API test for DUK_COMPILE_SHEBANG
- [x] API test coverage for shebang not on first line, or not in first column of first line
- [x] Test coverage for eval() not allowing shebang (there are other cases too but this one is obvious)
- [x] Test coverage for `new Function()` not allowing shebang
- [x] Fix or follow-up pull for compile flag translation - #1450
- [x] Add shebang support to 'duk'
- [x] Website custom feature entry
- [x] API documentation
- [x] Releases entry

Fixes #1346.


svaarala:
Just a bit of testing so far.

fatcerberus:
I wasn't sure about that "starting at first column" behavior, so I double-checked.  Node.js ignores the shebang, but only if no whitespace precedes the `#`.  So the behavior in this pull is indeed correct.

svaarala:
This should now be complete for the most part. Now with 'duk' installed in the system path somewhere, one can:

```
duktape@keppihevonen:~/duktape$ cat /tmp/foo.js
#!duk
print('hello world');

duktape@keppihevonen:~/duktape$ chmod u+x /tmp/foo.js

duktape@keppihevonen:~/duktape$ /tmp/foo.js
hello world
```

svaarala:
What's left is some website documentation and a follow-up to remove the unnecessary flag translation for duk_compile().

svaarala:
@fatcerberus It'd be nice to know if this branch - as it is now - would match your use case in Minisphere :)

svaarala:
Forwarding of the compilation flags to the lexer (in duk_lexer_ctx) is somewhat inelegant, but not visible externally and OK for now at least. Keeping the lexer/compiler flag set in the same space (and removing the API vs. internal translation) also minimizes flag conversions.

This pull adds a `duk_lexer_ctx.flags` for forwarded flags (which is cheaper than pointing back to the compiler context and reading flags from there) which isn't needed unless shebang comment support is enabled. But it's quite likely some other features will need the lexer to be aware of compiler flags; e.g. HTML comments are not allowed when compiling ES6 Modules which will eventually be needed, so I didn't make the flags field conditional.

svaarala:
@fatcerberus I'll add test coverage for shebang on the first line but not in the first column to cover that particular behavior. But now it's Zzz time :)

fatcerberus:
@svaarala I'll check this out tonight and let you know if it solves my use case. :)

svaarala:
@fatcerberus Test coverage should now be good enough for a merge - the only remaining task is for some website changes. So if you can think of a reason not to merge, now would be a good time :-)

svaarala:
Ok, now the website part is also done.

fatcerberus:
This doesn't seem to be working for me.  My `duk_pcompile()` call:

```js
			if (duk_pcompile(ctx, DUK_COMPILE_EVAL | DUK_COMPILE_SHEBANG) != DUK_EXEC_SUCCESS)
				goto on_error;
```

And yet I get `SyntaxError: invalid token (line 1)`.  Full clean and rebuild doesn't help.

svaarala:
Just checking: do you have an up-to-date duk_config.h?

svaarala:
Added that flag combination to the API test and seems to work in the test run.

fatcerberus:
duk_config.h was taken directly from configure.py output.  I was too tired to investigate further last night; I'll look at it again today.

fatcerberus:
I figured out the issue.  The file I tested on is a module, and the CommonJS function wrapper interferes with the shebang parsing so that the `#!` is no longer in the first column.  I don't think there's much Duktape can do in that case, sadly. :-(

fatcerberus:
Since Node.js allows shebang comments in the beginning of a module script, it might make sense to add code to duk_module_node to strip them out.

svaarala:
Ok, I can do that in this pull. I think the easiest approach is to check that if the two first characters are '#' and '!', just prepend a `//` comment at the start of the source so that the first line of the wrapper becomes:

```
function (...) {//#!foo
```

fatcerberus:
Hehe, that's the same idea I had. :)

svaarala:
Hmm, maybe I'll do that in a separate pull after all. I'll review this branch once more, merge, and then fix the module-node issue.

PR #1381, opened 2017-03-03
Compatibility fixes for several platforms.

beevik:
XboxOne/Durango: Missing SystemTimeToTzSpecificLocalTime API.
MinGW: Doesn't support sprintf with %lld. Requires %I64d.
Compiler warning fixed in duk_hobject_props.c

### DIFF
--- a/src-input/duk_bi_date_windows.c
+++ b/src-input/duk_bi_date_windows.c
@@ -86,10 +86,15 @@ DUK_INTERNAL_DECL duk_int_t duk_bi_date_get_local_tzoffset_windows(duk_double_t 
 	ft1.dwLowDateTime = tmp2.LowPart;
 	ft1.dwHighDateTime = tmp2.HighPart;
 	FileTimeToSystemTime((const FILETIME *) &ft1, &st2);
+#if defined(_DURANGO) || defined(_XBOX_ONE)
+	// XboxOne doesn't have 'SystemTimeToTzSpecificLocalTime' API
+	st3 = st2;
+#else
 	if (SystemTimeToTzSpecificLocalTime((LPTIME_ZONE_INFORMATION) NULL, &st2, &st3) == 0) {
 		DUK_D(DUK_DPRINT("SystemTimeToTzSpecificLocalTime() failed, return tzoffset 0"));
 		return 0;
 	}
+#endif
 	duk__convert_systime_to_ularge((const SYSTEMTIME *) &st3, &tmp3);
 
 	/* Positive if local time ahead of UTC. */
--- a/src-input/duk_bi_json.c
+++ b/src-input/duk_bi_json.c
@@ -1423,7 +1423,11 @@ DUK_LOCAL void duk__enc_fastint_tval(duk_json_enc_ctx *js_ctx, duk_tval *tv) {
 	 * "long long" type exists.  Could also rely on C99 directly but that
 	 * won't work for older MSVC.
 	 */
+#if defined(__MINGW32__) || defined(__MINGW64__)
+	DUK_SPRINTF((char *) buf, "%I64d", (long long) v);
+#else
 	DUK_SPRINTF((char *) buf, "%lld", (long long) v);
+#endif
 	DUK__EMIT_CSTR(js_ctx, (const char *) buf);
 }
 #endif
--- a/src-input/duk_hobject_props.c
+++ b/src-input/duk_hobject_props.c
@@ -5663,7 +5663,7 @@ duk_bool_t duk_hobject_define_property_helper(duk_context *ctx,
 			a = (duk_harray *) obj;
 			DUK_DD(DUK_DDPRINT("Object.defineProperty() attribute update for duk_harray .length -> %02lx", (unsigned long) new_flags));
 			DUK_ASSERT_HARRAY_VALID(a);
-			if ((new_flags & DUK_PROPDESC_FLAGS_EC) != (curr.flags & DUK_PROPDESC_FLAGS_EC)) {
+			if ((new_flags & DUK_PROPDESC_FLAGS_EC) != ((duk_small_uint_t)curr.flags & DUK_PROPDESC_FLAGS_EC)) {
 				DUK_D(DUK_DPRINT("Object.defineProperty() attempt to change virtual array .length enumerable or configurable attribute, fail"));
 				goto fail_virtual;
 			}
--- a/config/header-snippets/date_provider.h.in
+++ b/config/header-snippets/date_provider.h.in
@@ -22,7 +22,7 @@
 
 #if defined(DUK_USE_DATE_GET_LOCAL_TZOFFSET)
 /* External provider already defined. */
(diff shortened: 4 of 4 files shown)


src-input/duk_bi_date_windows.c
```diff
@@ -86,10 +86,15 @@ DUK_INTERNAL_DECL duk_int_t duk_bi_date_get_local_tzoffset_windows(duk_double_t
 	ft1.dwLowDateTime = tmp2.LowPart;
 	ft1.dwHighDateTime = tmp2.HighPart;
 	FileTimeToSystemTime((const FILETIME *) &ft1, &st2);
+#if defined(_DURANGO) || defined(_XBOX_ONE)
+	// XboxOne doesn't have 'SystemTimeToTzSpecificLocalTime' API
```
svaarala: - Use C89-style comment (`/* ... */`)
- What's the impact of just setting st3 = st2?

src-input/duk_bi_json.c
```diff
@@ -1423,7 +1423,11 @@ DUK_LOCAL void duk__enc_fastint_tval(duk_json_enc_ctx *js_ctx, duk_tval *tv) {
 	 * "long long" type exists.  Could also rely on C99 directly but that
 	 * won't work for older MSVC.
 	 */
+#if defined(__MINGW32__) || defined(__MINGW64__)
+	DUK_SPRINTF((char *) buf, "%I64d", (long long) v);
+#else
 	DUK_SPRINTF((char *) buf, "%lld", (long long) v);
+#endif
```
svaarala: Duktape internals don't contain any platform/compiler define checks as a matter of principle - they should all be in duk_config.h. I'm not sure of the best solution here, maybe duk_config.h should provide a define for sprintf() formatting a long long.

src-input/duk_hobject_props.c
```diff
@@ -5663,7 +5663,7 @@ duk_bool_t duk_hobject_define_property_helper(duk_context *ctx,
 			a = (duk_harray *) obj;
 			DUK_DD(DUK_DDPRINT("Object.defineProperty() attribute update for duk_harray .length -> %02lx", (unsigned long) new_flags));
 			DUK_ASSERT_HARRAY_VALID(a);
-			if ((new_flags & DUK_PROPDESC_FLAGS_EC) != (curr.flags & DUK_PROPDESC_FLAGS_EC)) {
+			if ((new_flags & DUK_PROPDESC_FLAGS_EC) != ((duk_small_uint_t)curr.flags & DUK_PROPDESC_FLAGS_EC)) {
```
svaarala: - I'm not sure I understand the issue here: `curr.flags` should already be a duk_small_uint_t (duk_propdesc defines in duk_hobject.h)? What's the warning fixed?
- If this is needed, there should be a space after the cast (code style).

svaarala:
Re: that MinGW format specifier - we can leave this pull as is (for that change) and I could fix it in a cleanup pull.

The reason a direct MinGW check is technically inappropriate is that I can override `DUK_SPRINTF()` in duk_config.h, and the format specifier for the overridden `DUK_SPRINTF()` may not match the MinGW default sprintf() one. So the specifier should be controllable from duk_config.h to match `DUK_SPRINTF()` provider.

Not that replacing sprintf() is likely for MinGW, but it is quite usual for low memory targets and that's why there's a code policy that duk_config.h contain all the platform checks so that all the detection stuff will be cleanly in one place.

src-input/duk_hobject_props.c
beevik: I'll have to go back and check.  I compile duktape on a bunch of platforms, and I made this fix a while back because it was triggering a warning.  I'll see if I can find out which platform was causing it.

src-input/duk_bi_json.c
beevik: Sadly I know of no truly cross-platform cross-compiler version of 64-bit format specifiers. C99 has one (PRId64), but I don't think all the C++ compilers support it.

Maybe you could do something similar: define a literal like PRId64 to replace "%lld".

src-input/duk_bi_date_windows.c
beevik: I believe it leaves the final time in UTC, which honestly is probably not what you want. But there is no equivalent to SystemTimeToTzSpecificLocalTime in the Durango XDK, so that was the best I could come up with.  I thought it was better than returning zero.

Also, I'm guessing based on your other comment that I shouldn't have done a platform check in this code; that should be in duk_config.h.

src-input/duk_bi_json.c
fatcerberus: Strictly speaking, `long long` is not guaranteed to be 64-bit.  In practice, it *is* on most platforms, but I believe the C/C++ specifications only require `char` >= `short` >= `int` >= `long` >= `long long` without specifying minimum/maximum sizes for any of them (except `char` which is always 1 byte).

svaarala: As far as I know C99 requires long long to be at least 64 bits.

svaarala: https://gcc.gnu.org/onlinedocs/gcc/Long-Long.html

> ISO C99 supports data types for integers that are at least 64 bits wide, and as an extension GCC supports them in C90 mode and in C++.

svaarala: @beevik Yeah, I was thinking along the lines of `DUK_FMT_LONGLONG` or something. It's unfortunately compiler/platform (or more strictly speaking, lib) specific but as long as it can be defined to match the sprintf() provider it should be OK.

src-input/duk_bi_date_windows.c
svaarala: Re: platform check, that's correct - I know it's a bit of a hassle sometimes, especially in Date providers which are already platform specific. Ideally there would be a descriptive `DUK_USE_xxx` flag for availability of the API.

I'm fine with that limitation, but it would deserve a short comment on the impact.

A limitation like that is OK IMO if there's no default platform way to get the right behavior. A user can then replace the Date provider if necessary; this is often needed on bare metal targets where there isn't necessarily even an API for current time (let alone a time zone).

src-input/duk_bi_json.c
svaarala: Re: formatting more generally, internal code now usually casts to `unsigned long` or `long` and uses `%lu` and `%ld` because those work universally (also C89) for 32-bit values. There are no format specifier defines for exact 32-bit values etc (as there is in C99).

I didn't want to replicate all those defines (for C89) because there's relatively little need for them in practice besides debug prints (which are fine with a "wide cast" to long). Also, when using a sprintf() replacement, it's good that a minimal number of different specifiers are used in the format strings.

In concrete terms, there's exactly one non-debug call site where an "at least 64 bits" format specifier is needed - this fastint case - so a heavyweight fix seems wrong .. :)

There are some debug prints for fastint 64-bit values which may be truncated if sizeof(long) isn't at least 8 but that's not a functional issue. Also "long" is 64 bits on most Unix platforms (which are LP64) so debug prints there come out correctly.

beevik:
Closing this PR while I address the issues raised by comments.

PR #1385, opened 2017-03-08
Use _snprintf() for pre-MSVC 2015 in extras/module-duktape

### LINKED ISSUE #1369: Module loader gives a warning on Windows (Visual Studio 2013)
To be precise, compiling duk_module_duktape.c gives the following warning:

`warning C4013: 'snprintf' undefined; assuming extern returning int`

The problem occurs because Microsoft prepends the names of certain functions with an underscore. My current fix is adding the following to duk_module_duktape.h:

```
#ifdef _WIN32
	#define snprintf _snprintf
#endif
```

svaarala:
Fixes #1369.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2531,6 +2531,9 @@ Planned
 
 * Miscellaneous compiler warning fixes (GH-1358)
 
+* Use _snprintf() prior to MSVC 2015 in extras/module-duktape (GH-1369,
+  GH-1385)
+
 * Miscellaneous performance improvements: more likely/unlike attributes and
   hot/cold function splits (GH-1308, GH-1309, GH-1312), integer
   refzero-free-running flag (instead of a flag bit) (GH-1362)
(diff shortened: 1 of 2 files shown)


svaarala:
@armornick Could you verify if this fixes the warning for you (I don't have a Windows installation right now and Appveyor is not processing its job queue for some reason so I can't easily test right now).

seanmiddleditch:
You might consider using [stbsp_snprintf](https://github.com/nothings/stb/blob/master/stb_sprintf.h) instead. Common behavior across all platforms _and_ it's quite a bit faster (see the benchmark results in the header's comments).

svaarala:
That wouldn't necessarily be appropriate for low memory targets where there's often a platform-provided (or otherwise preferred) sprintf() provider and another one would be unnecessary footprint. Proper sprintf() footprint isn't negligible because of floating point formatting.

svaarala:
@armornick I'll merge this now - please comment if the fix doesn't work for you.

PR #1398, opened 2017-03-13
Faster finalizer existence check in GC

svaarala:
One bottleneck in refzero and mark-and-sweep handling is checking whether an object has an own or inherited _Finalizer property. This check walks the prototype chain and does a property lookup for every object. Because most objects don't have a finalizer, the entire prototype chain is typically walked to confirm there's no inherited finalizer.

Improve this behavior by adding a `DUK_HOBJECT_FLAG_HAVE_FINALIZER` flag to `duk_hobject`. The flag is set/cleared when the internal `_Finalizer` property is changed using `duk_set_finalizer()`. Direct property changes won't update the flag, but that's OK because `_Finalizer` is an internal property with opaque semantics (IOW, application shouldn't change it manually).

The flag allows a fast finalizer check by walking the prototype chain and only checking one flag per object instead of doing an actual property lookup.

- [x] Add `DUK_HOBJECT_FLAG_HAVE_FINALIZER` and `duk_hobject_has_finalizer_fast()`
- [x] Use `duk_hobject_has_finalizer_fast()` in refcount and mark-and-sweep
- [x] Debugger artificial property for "have_finalizer"
- [x] Clear have finalizer flag when bytecode dumping a function
- [x] Testcase changes
- [x] Documentation changes
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2537,7 +2537,8 @@ Planned
 
 * Miscellaneous performance improvements: more likely/unlike attributes and
   hot/cold function splits (GH-1308, GH-1309, GH-1312), integer
-  refzero-free-running flag (instead of a flag bit) (GH-1362)
+  refzero-free-running flag (instead of a flag bit) (GH-1362), faster GC
+  finalizer existence check using DUK_HOBJECT_FLAG_HAVE_FINALIZER (GH-1398)
 
 * Miscellaneous footprint improvements: more compact duk_hobject allocation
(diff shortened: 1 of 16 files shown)


svaarala:
Change to Octane (five run average):
- Master: 231.4 (223-235)
- This branch: 239.6 (234-242)

There's noise in the measurement but having done a few repeats, the effect seems measurable. That's not hard to believe because the prototype walk with property lookups would happen for every actual object being collected.

Issue #1384, opened 2017-03-08
Recursive TS-transpiled async function crashes Duktape


fatcerberus:
I'm assuming this is a stack overflow.

The following TypeScript:
```ts
async function explode()
{
	await explode();
}

explode();
```

Produces this (extremely verbose :) JS:
```js
var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
    return new (P || (P = Promise))(function (resolve, reject) {
        function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
        function rejected(value) { try { step(generator["throw"](value)); } catch (e) { reject(e); } }
        function step(result) { result.done ? resolve(result.value) : new P(function (resolve) { resolve(result.value); }).then(fulfilled, rejected); }
        step((generator = generator.apply(thisArg, _arguments || [])).next());
    });
};
var __generator = (this && this.__generator) || function (thisArg, body) {
    var _ = { label: 0, sent: function() { if (t[0] & 1) throw t[1]; return t[1]; }, trys: [], ops: [] }, f, y, t, g;
    return g = { next: verb(0), "throw": verb(1), "return": verb(2) }, typeof Symbol === "function" && (g[Symbol.iterator] = function() { return this; }), g;
    function verb(n) { return function (v) { return step([n, v]); }; }
    function step(op) {
        if (f) throw new TypeError("Generator is already executing.");
        while (_) try {
            if (f = 1, y && (t = y[op[0] & 2 ? "return" : op[0] ? "throw" : "next"]) && !(t = t.call(y, op[1])).done) return t;
            if (y = 0, t) op = [0, t.value];
            switch (op[0]) {
                case 0: case 1: t = op; break;
                case 4: _.label++; return { value: op[1], done: false };
                case 5: _.label++; y = op[1]; op = [0]; continue;
                case 7: op = _.ops.pop(); _.trys.pop(); continue;
                default:
                    if (!(t = _.trys, t = t.length > 0 && t[t.length - 1]) && (op[0] === 6 || op[0] === 2)) { _ = 0; continue; }
                    if (op[0] === 3 && (!t || (op[1] > t[0] && op[1] < t[3]))) { _.label = op[1]; break; }
                    if (op[0] === 6 && _.label < t[1]) { _.label = t[1]; t = op; break; }
                    if (t && _.label < t[2]) { _.label = t[2]; _.ops.push(op); break; }
                    if (t[2]) _.ops.pop();
[...]

PR #1421, opened 2017-03-22
Direct handling of .call(), .apply(), etc

svaarala:
Handle the following calls inline in call handling so that they are not visible on the call stack and don't prevent thread yielding:

- Function.prototype.call()
- Function.prototype.apply()
- Reflect.apply()

The following are not yet changed like this: Reflect.construct(), eval().

Tasks:
- [x] Include call/apply handling as part of call handling
- [x] Add DUK_HOBJECT_FLAG_SPECIAL_CALL to quickly skip native pointer checks for most cases
- [x] Add internal duk_unpack_array_like(), coercion semantics compatible between calls?
- [x] Maybe rework num_stack_args plumbing in call handling, quite awkward now => separate pull
- [x] Testcases: call/apply tail call works
- [x] Testcases: call/apply missing argument handling (now special cased in call handling so coverage important)
- [x] Testcases: call/apply and Duktape.Thread.yield() should now be allowed
- [x] Testcases: binding fn.call() and calling that bound function
- [x] Testcases: calling fn.bind() and calling the result with .call() (and bound .call())
- [x] Testcases: calling .call() with .call()/.apply() (and other combinations)
- [x] Testcases: call/apply no longer visible in call stack
- [x] Testcases: call/apply doesn't consume native recursion limit for Ecma-to-Ecma calls
- [x] Testcases: huge argument counts for .call()/.apply()
- [x] Testcases: arguments object for .call()/.apply() targets (and bound + .call() etc)
- [x] Perf test for .call() / .apply()
- [x] Website: thread yield limitations
- [x] Other documentation changes
- [x] Migration note
- [x] Releases entry

Future work:
- Rework duk_new() and Reflect.construct() similarly so that constructor call post-processing semantics are in call handling to avoid call stack entry.


fatcerberus:
This will be nice for transpilers too, since spread calls, e.g. `func(...args)`, generally get transpiled to an `.apply()`.  In that case the limitations won't be as obvious if one doesn't know how the transpiled code looks.

svaarala:
Hacking around I think I found an easy implementation which would cover all call cases (including Ecma-to-Ecma calls without preventing yield).

The basic approach is to handle .call() and .apply() specially at the same time as bound functions:
- When a bound function is encountered, "peel" it off and insert the arguments into the value stack as before and recheck.
- When a non-bound function is encountered, check if it is `Function.prototype.call()`. If so, adjust the target function and value stack based on the `.call()` arguments directly and continue the loop (resolving a potentially bound target again, etc).
- Similar handling for `Function.prototype.apply()`, except that the arguments are unpacked from an array-like object into the value stack before continuing.

Because this is handled early, the rest of the call handling will see the call as if it was done directly to begin with. Ecmascript call setup also does bound function resolution (and thus the above .call() and .apply() steps too) before deciding whether an Ecma-to-Ecma call or a tailcall is possible, so this would also allow tail calls like: `return func.call('foo', 123);`.

I'll experiment a bit more and re-push the branch if it works out.

svaarala:
That simple approach seems to be workable. The branch is still quite unfinished and needs refactoring, but the basic idea seems to work.

fatcerberus:
Interestingly, this merge doesn't seem to have fixed the segfault in #1384 (stack overflow due to recursive `.call()`), even though it logically should have.

svaarala:
I injected an intentional segfault on the 100th iteration of the awaiter function, and got this native stack:

```
==9364==    at 0x433880: cause_segfault (duk_cmdline.c:1057)
==9364==    by 0x4104A2: duk__handle_call_inner (duk_js_call.c:1779)
==9364==    by 0x40A142: duk__js_execute_bytecode_inner.isra.132 (duk_js_executor.c:4751)
==9364==    by 0x40FCD6: duk_js_execute_bytecode (duk_js_executor.c:2891)
==9364==    by 0x410415: duk__handle_call_inner (duk_js_call.c:1728)
==9364==    by 0x428DC8: duk_new (duk_api_call.c:384)
==9364==    by 0x40A1CF: duk__js_execute_bytecode_inner.isra.132 (duk_js_executor.c:4824)
==9364==    by 0x40FCD6: duk_js_execute_bytecode (duk_js_executor.c:2891)
==9364==    by 0x410415: duk__handle_call_inner (duk_js_call.c:1728)
==9364==    by 0x428DC8: duk_new (duk_api_call.c:384)
==9364==    by 0x40A1CF: duk__js_execute_bytecode_inner.isra.132 (duk_js_executor.c:4824)
==9364==    by 0x40FCD6: duk_js_execute_bytecode (duk_js_executor.c:2891)
```

So, the recursion depth is increased because a constructor call is involved. Those don't get the callstack optimization yet because they involve post-processing of the call result.

svaarala:
I'll most likely try to rework constructor calls to be dealt with in normal call handling so that the constructor post-processing happens without a wrapper duk_new() call. With that change the recursion limit would no longer get hit. But more importantly you could then yield from constructor calls too which I'd like to allow :)

fatcerberus:
That makes sense; most likely the constructor call involved is the Promise constructor - which means even with the change you describe the issue will come back once Promise is supported natively.

I'm thinking my best bet for now is to increase the stack size in Windows to avoid the segfault and ensure a useful error is thrown in case of too-deep recursion rather than crashing outright.

svaarala:
Hmm, why would `new Promise()` make another construct call inline?

fatcerberus:
Because it's a recursive await - `async` is pretty much just sugar for `new Promise(...)` passing the contents of the function.  If the async function calls itself, promises get constructed recursively as well.

svaarala:
Ok, maybe I'm not following but why would that call come from inside `new Promise()` itself, which just creates a promise and registers the callbacks without actually calling them yet?

svaarala:
Ah, never mind, I see what you mean now. The executor argument may do another new Promise() which does another new Promise() etc.

svaarala:
For some reason I thought new Promise() would run the target function via the job queue and not inline as it is specified.

fatcerberus:
Yeah, the executor function is called inline - it's *settlement* (fulfill/reject) that goes through the job queue.

svaarala:
I'm not sure I understand the merits of the executor running inline, but it doesn't really matter because that's the way it is specified :)

Do you think deep recursion in the Promise constructor is a practical concern? If it is, there's no reason why Duktape built-in Promise constructor couldn't be an Ecmascript function and thus avoid native recursion.

fatcerberus:
> I'm not sure I understand the merits of the executor running inline, but it doesn't really matter because that's the way it is specified :)

I don't claim to know why either, but C# `async` behaves the same: everything up to and including the first `await` runs immediately.  The C# compiler will actually warn if you have an async function with no await because then you just have a sync function with a delayed return value.

> Do you think deep recursion in the Promise constructor is a practical concern?

Realistically, probably not.  Developers will more likely strive to *minimize* async recursion since each additional level increases latency:

```js
async function f() { return await g(); }
async function g() { return await h(); }
async function h() { return 812; }

// prints "812" three ticks later
console.log(await f());
```

svaarala:
Right, the native stack limit is hundreds of calls. It's quite a lot if you're specifically avoiding deep recursion -- but on the other hand it's not very much if you're intentionally using deep recursion :)

fatcerberus:
Hm, you're right.  `DUK_USE_NATIVE_CALL_RECLIMIT` is (by default) 1,000, which is a whole order of magnitude less than `DUK_CALLSTACK_DEFAULT_MAX`, the Ecmascript callstack limit, at 10,000.  And async functions/the promise constructor don't benefit from tailcalls, either.  Something to keep in mind, at least, when promise support is implemented.

svaarala:
That ratio could be much higher too because the callstack limit is pretty arbitrary. It basically just exists to catch runaway recursion before running out of memory which is a bit more unpleasant to deal with. There are plenty of other ways to run out of memory though, so applications/scripts need to deal with out of memory anyway. So even removing the callstack limit would be relatively OK IMO.

The native call depth / recursion limit is much trickier because of the portability concerns, and Windows targets don't even reach the native call recursion limit (1000) in practice with their default stack. Bumping it higher doesn't help unless the application actually deals with setting up a larger stack -- something that people won't by default remember to do.

Issue #707, opened 2016-04-27
Cleaner solution to prevent finalizer execution in error handling path


svaarala:
- [ ] Cleaner solution for preventing finalizer execution (explicit flag instead of lj.type check)
- [ ] Solution for preventing finalizer side effects in call setup too (not just error handling)

PR #1427, opened 2017-03-23
Rework refzero, finalizer, and side effect handling

### LINKED ISSUE #314: Protect call handling and longjmp handling from finalizer side effects

### LINKED ISSUE #1311: Strange assert, unmoved heapptr was selected for finalization
Were there any major changes to garbage collection?

Take this with a grain of salt since usually the bugs I present is some weird thing I messed up in my side, but this was a strange one.

In my app I have certain indices in the stash reserved for some object pools. The pools get allocated at the beginning as arrays, and they are left there. I simply grab the heapptr and add/remove objects to that array. I haven't had any issues with it ever, suddenly today I got an assert triggered when I pushed the stashed pool's heapptr:

``` C
/* One particular problem case is where an object has been
		 * queued for finalization but the finalizer hasn't been
		 * executed.
		 */
		duk_heaphdr *curr;
		for (curr = thr->heap->finalize_list;
		     curr != NULL;
		     curr = DUK_HEAPHDR_GET_NEXT(thr->heap, curr)) {
			DUK_ASSERT(curr != (duk_heaphdr *) ptr);
		}
```
Again, I might have corrupted something somewhere, who knows, just looking for some info in case there were some changes that might have caused this? This portion of my code (with the pool) hasn't been touched in a very long time...

The heapptr which is an array inside an array inside the heap stash is only ever grabbed from the stash index once at the beginning, the pushed into the stack only 3 places in the code. The stash is not messed with again other than in initializing those arrays at the beginning.

If it helps, the objects that are stored in that pool (the heapptr that caused the assert) are revived when being finalized. That is, the pool is full of preallocated Ecmascirpt objects that I grab from the pool return them when I need one, then nullify that object's index. When the object get's finalized I keep it alive by storing it back in the pool.

### LINKED ISSUE #1396: Refzero finalizer loop when finalizer creates garbage referencing finalizer target
While working on fixing some side effect issues, noticed the following a bit unintuitive refzero finalizer behavior:

When refcounting runs a finalizer for object O and the finalizer creates garbage referencing O (say, `void { ref: O };`) a finalizer loop happens. The sequence is:
- O's refzero drops to zero for the first time.  O is queued to finalize_list and refzero_list processing begins.
- Refzero_list processing notices O has a finalizer and calls it.
- The finalizer creates a temporary object like `{ ref: O }`, and the object becomes unreachable either inside the finalizer or when the finalizer call unwinds.
- The temporary object's refzero falls to zero.  It is queued to refzero_list for processing, but won't be processed immediately because we're already working on the refzero_list.
- The decision to rescue/keep O is made after the finalizer call returns. Because the temporary object hasn't yet been freed or refcount finalized, O's refcount is still >0.  So O is rescued and queued back to the heap.
- Refzero_list processing reaches the temporary object.  It doesn't have a finalizer, so it gets refcount finalized and freed immediately.  Refcount finalization means O's refcount now falls to zero.
- O gets queued back to refzero_list and gets processed again, completing the loop: O gets finalized again, etc.

Here's a concrete example:

```js
function myFinalizer(o) {
    var temp = { ref: o };
    temp = null;
} 
```

The root cause is that new garbage involving `o` is created for every finalizer run, and that garbage is not yet collected when the rescue/keep decision is made. However, it's unintuitive that this causes a finalizer loop because the garbage is (seemingly) immediately collected by refcounting.

The issue doesn't occur if the object is finalized via mark-and-sweep (i.e. the object is in a reference cycle) because mark-and-sweep will correctly determine that `temp` isn't reachable, and neither is `o`, so the object won't be rescued. The issue also doesn't occur if the finalizer is modified to:

```js
function myFinalizer(o) {
    var temp = { ref: o };
    temp.foo = { bar: temp };  // ref loop
    temp = null;
}
```

The change makes `temp` only collectable via mark-and-sweep due to the reference loop. So while the original object is rescued and will be finalized for the second time, the second finalization is done by mark-and-sweep.

One fix for this unintuitive behavior is to move refzero finalizer handling outside of refzero_list processing: when refzero_list is processed and an object is seen to have a finalizer, the object would be queued to finalize_list without further processing. When refzero_list processing is complete, finalize_list would be processed. With this approach when `temp`'s refcount falls to zero in the finalizer, it will be freed immediately because we're not inside refzero_list processing when the finalizer runs. As a result, the object's refcount will be up-to-date for the rescue/keep decision.

### LINKED ISSUE #1426: Out-of-memory in object property table realloc leaves ms_base_flags in place
Object property table reallocation code uses mark-and-sweep base flags to prevent object compaction during the property table resize so that the object is not recursively resized. The code path between setting and restoring mark-and-sweep flags should be free of error throwing -- however, current code has a `duk_push_dynamic_buffer()` inside the flags protection which may throw.

Convert to a manual allocation, use a protected call to realloc (slow), or restore the mark-and-sweep flags in error throw. Add assert to error handling to ensure no error is thrown during a property table resize.

svaarala:
Overview of changes:
- Rework duk_hobject refzero handling. Objects without finalizers are freed inline (even with NORZ macros). Objects with finalizers are queued to finalize_list also for refcounting, and processed after refzero_list is complete (except when NORZ macros are used).
- Rework finalize_list handling. Use finalize_list for both refcount and mark-and-sweep finalizers. Execute finalizers outside of refzero_list processing or mark-and-sweep. Mark-and-sweep is still prevented explicitly for finalizers at present to avoid spurious rescues for a partially processed finalize_list. Because finalizer execution is outside refzero_list processing, refcount finalization rescue/free check has up-to-date refcounts (see #1396).
- Objects on finalize_list now always have FINALIZED flag set. This allows `duk_push_heapptr()` to detect that the argument is about to be finalized and allow that case later on.
- Proper side effect protection for error throwing. Finalizers are prevented during error handling.
- Refcount and mark-and-sweep based finalization always uses `heap->heap_thread` now for consistency.
- Side effect torture test improvements.
- Fix object property table reallocation issue where mark-and-sweep base flags could be stuck with "prevent compaction" set when `duk_push_dynamic_buffer()` threw an out-of-memory error.
- Lots of other small changes here and there.

Tasks:
- [x] Finalize code changes
- [x] Commit review
- [x] Final assertion+torture test run
- [x] Releases entry

Fixes:
- This should fix the last remaining side effect issue in #1311. Fixes #1311.
- Fixes #1396.
- Addresses one issue in #707.
- Fixes #314.
- Fixes #1426.


svaarala:
@harold-b I still have some issues left to clean up here - but it would be interesting to hear if this branch resolves the assert issues you were having in #1311.

svaarala:
Labeled 'bug' because fixes at least one outstanding bug.

harold-b:
Which one particularly should I check regarding this branch? I remember we touched on a few different things after the first issue I reported, was it this https://github.com/svaarala/duktape/issues/1311#issuecomment-273934549 one?

My memory fails me as to which the specific problems were... I had a few test setups for these bugs but I can't remember at all which they were now :fearful:

svaarala:
IIRC you had issues with assertions triggering when running with asserts enabled? Specifically, asserts related to `heap->lj.type`.

harold-b:
OK, that seems to be the one in the comment. I believe I still have a branch with said commit set up. I can check it out tomorrow and report back.

svaarala:
@beevik This pull is close to being complete - it doesn't yet allow duk_push_heapptr() for an unreachable-but-not-yet-finalized object, but I prototyped it on top of this pull and it seems to be doable. So hopefully master would have a working solution for that case in a few days. Backporting it to 2.0.x may not be worth the effort though (it's doable but involves a lot of details and testing) so you'd have to move to master / 2.1.0 release to get the feature.

harold-b:
@svaarala Sorry for the delay. Took me a little while to find the branch and setup I had for this test case. But anyway, I did try it with this branch and it ran smoothly :+1:  Nice work!

svaarala:
@harold-b Good to hear! The assertion coverage is now much better so it's easier to catch issues. There are a few out-of-memory situations that will still trigger the asserts; I'll deal with them separately. You're only going to run into these if you run out of memory (even after emergency GC).

PR #1435, opened 2017-03-24
ES6 HTML Style Comments

tlsa-ct:
I had a go at adding support for [ES6 HTML-like comments](http://www.ecma-international.org/ecma-262/6.0/#sec-html-like-comments).  Does the approach make sense?

This change has no effect on the `make apitest` results.
It does not add or remove tests.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -42,6 +42,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Dominik Okwieka (https://github.com/okitec)
 * Remko Tronçon (https://el-tramo.be)
 * Romero Malaquias (rbsm@ic.ufal.br)
+* Michael Drake <michael.drake@codethink.co.uk>
 
 Other contributions
 ===================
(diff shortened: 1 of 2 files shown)


tlsa-ct:
I pushed an improvement to the comment in the SingleLineHTMLCloseComment commit.

svaarala:
Trying to figure out the impact to code not needing/wanting HTML comment syntax support, I think at least this would parse differently with and without HTML comments:

```js
var x = 10;
var y = 0<!--x;
```

Pretty nonsensical, but would evaluate to `0<!--x` -> `0 < !9` -> `0 < 0` -> `false`:

```
duk> var x = 10;
= undefined
duk> var y = 0 <!--x;
= undefined
duk> y
= false
```

Same with Node.js:

```
> var x = 10;
undefined
> var y = 0 <!--x;
undefined
> y
0
```

So... Node.js already parses this as a HTML comment. Breaking the `<!--`:

```
> var x = 10;
undefined
> var y = 0 <! --x;
undefined
> y
false
```

So in summary, HTML comment parsing does technically interfere with ordinary syntax, but it's unlikely to cause problems because e.g. Node.js parses HTML comments too.

svaarala:
@tlsa-ct Thanks! I'll merge the base change into master: I'm not 100% sure of the required semantics but I'm quite sure you guys have a better idea of what constructs are encountered in browser environments. I'll open a tracking issue for cleanups, test coverage, possibly config options, etc.

PR #1436, opened 2017-03-24
Add testcase for comment lexing (including HTML comments)

tlsa-ct:
Adds testing for comment handling in the lexer.


tlsa-ct:
Note that the new tests need the changes from https://github.com/svaarala/duktape/pull/1435

PR #1442, opened 2017-03-27
Allow duk_push_heapptr() for an unreachable but not yet finalized object

svaarala:
Up to Duktape 2.0 the API contract for duk_push_heapptr() is strict:
- The argument must be strongly referenced (= reachable for Duktape, refcount > 0) all the way between duk_get_heapptr() and duk_push_heapptr().

In particular, calling duk_push_heapptr() for an object in its finalizer is not allowed, and causes both assertion errors and actual memory safety issues.

There are quite common application use cases where this is limiting, and it would be very useful to allow duk_push_heapptr() to rescue an object in a finalizer call -- i.e. when the object has become unreachable, but has not yet been finalized and freed. In essence the application uses a finalizer to rescue a weakly referenced object.

This pull allows that use case:
- [x] Requires #1427; prototype code is already in that pull but commented out
- [x] Update duk_push_heapptr() asserts
- [x] Detect an object on finalize_list based on FINALIZABLE flag and "auto rescue" it in duk_push_heapptr()
- [x] Issue for finalizer call guarantees (e.g. keep object on finalize_list until we can successfully call it) #1456
- [x] Test coverage
- [x] Documentation changes
- [x] Revise API documentation changes for better clarity
- [x] Releases entry

This use case may still not work 100% correctly in out-of-memory situations, e.g. if a finalizer call fails due to out-of-memory. Guaranteeing that native finalizer calls are always possible would fix the issue. That could be achieved by ensuring there's always enough value stack and call stack spare to call a finalizer. A finalizer call could work with a smaller spare than an ordinary function. But, this should be covered in a follow-up.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2622,6 +2622,12 @@ Planned
   where an error would be thrown when handling a previously thrown error
   (GH-1427)
 
+* Allow duk_push_heapptr() for a heap object which has become unreachable,
+  has been queued to finalize_list, but hasn't yet been finalized; in this
+  case duk_push_heapptr() cancels the finalization and moves the object back
+  to the main heap_allocated list, in effect automatically rescuing the object
+  without finalizer interaction (GH-1442)
+
 * Use a 32-bit refcount field by default (even on 64-bit systems) which saves
   8 bytes for each heap object and can only wrap if the Duktape heap is
   larger than 64GB; disable DUK_USE_REFCOUNT32 to use size_t for refcounts
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -4625,12 +4625,11 @@ DUK_LOCAL void duk__validate_push_heapptr(duk_context *ctx, void *ptr) {
 				DUK_ASSERT(found == 0);  /* Would indicate corrupted lists. */
 				found = 1;
 			} else {
-#if 1
-				DUK_ASSERT(0);
-#else  /* Enable when duk_push_heapptr() allowed for object on finalize_list. */
+				/* Not being finalized but on finalize_list,
+				 * allowed since Duktape 2.1.
+				 */
 				DUK_ASSERT(found == 0);  /* Would indicate corrupted lists. */
 				found = 1;
-#endif
 			}
 		}
 	}
@@ -4712,7 +4711,6 @@ DUK_EXTERNAL duk_idx_t duk_push_heapptr(duk_context *ctx, void *ptr) {
 
 	DUK_ASSERT_HEAPHDR_VALID((duk_heaphdr *) ptr);
 
-#if 0
 	/* If the argument is on finalize_list it has technically been
 	 * unreachable before duk_push_heapptr() but it's still safe to
 	 * push it.  Starting from Duktape 2.1 allow application code to
@@ -4738,9 +4736,9 @@ DUK_EXTERNAL duk_idx_t duk_push_heapptr(duk_context *ctx, void *ptr) {
 		curr = (duk_heaphdr *) ptr;
 		DUK_HEAPHDR_CLEAR_FINALIZABLE(curr);
 
-		/* Because FINALIZED is set prior to finalizer call, will be
-		 * set for the object being currently finalized, but not for
-		 * other objects on finalize_list.
+		/* Because FINALIZED is set prior to finalizer call, it will
+		 * be set for the object being currently finalized, but not
+		 * for other objects on finalize_list.
 		 */
 		DUK_HEAPHDR_CLEAR_FINALIZED(curr);
 
@@ -4756,7 +4754,6 @@ DUK_EXTERNAL duk_idx_t duk_push_heapptr(duk_context *ctx, void *ptr) {
 
 		/* Continue with the rest. */
 	}
-#endif
 
 	switch (DUK_HEAPHDR_GET_TYPE((duk_heaphdr *) ptr)) {
 	case DUK_HTYPE_STRING:
--- a/tests/api/test-push-heapptr-finalize-list.c
+++ b/tests/api/test-push-heapptr-finalize-list.c
@@ -1,0 +1,217 @@
+/*
+ *  Duktape 2.0 and prior required that duk_push_heapptr() argument was fully
+ *  reachable all the way between duk_get_heapptr() and duk_push_heapptr().
+ *  This also meant that if an object became unreachable but its finalizer had
+ *  not yet executed, it was an assertion violation to use duk_push_heapptr()
+ *  for it.  Such a push also caused incorrect, memory unsafe behavior besides
+ *  just assertion errors.
+ *
+ *  Duktape 2.1 allows the duk_push_heapptr() argument to be an object on the
+ *  finalize_list, provided that the finalizer has not yet executed (it's OK
+ *  for the finalizer to be executing, but not finished) without rescue.
+ *
+ *  There are a few separate cases here:
+ *
+ *  - Refzero collection: duk_push_heapptr() within the finalizer for the
+ *    object itself.
+ *
+ *  - Refzero collection: duk_push_heapptr() in a finalizer for another
+ *    object not yet finalized (but on finalize_list).
+ *
+ *  - Mark-and-sweep: duk_push_heapptr() for object being finalized.
+ *
+ *  - Mark-and-sweep: duk_push_heapptr() for another object on finalize_list.
+ */
+
+/*===
+*** test_refcount (duk_safe_call)
+trigger refzero
+fin1 called
+fin2 called
+fin2 rescues ptr3
+refzero done
+fin3 called
+fin1 called
+null writes done
+final top: 0
+==> rc=0, result='undefined'
+*** test_mark_and_sweep (duk_safe_call)
+make unreachable
+call duk_gc()
+fin1 called
+fin2 called
+fin2 rescues ptr3
+duk_gc returned()
+null writes done
+call duk_gc()
+fin3 called
+duk_gc returned()
+final top: 0
+==> rc=0, result='undefined'
+*** test_mark_and_sweep_2 (duk_safe_call)
+make unreachable
+call duk_gc()
+fin_norescue called
+fin_norescue called
+duk_gc returned()
+final top: 0
+==> rc=0, result='undefined'
+===*/
+
+static void *ptr1 = NULL;
+static void *ptr2 = NULL;
+static void *ptr3 = NULL;
+
+static duk_ret_t fin1(duk_context *ctx) {
+	printf("fin1 called\n");
+
+	duk_push_heapptr(ctx, ptr1);
+	duk_put_global_string(ctx, "rescue1");
+
+	ptr1 = NULL;
+	return 0;
+}
+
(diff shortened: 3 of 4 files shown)


harold-b:
> In particular, calling duk_push_heapptr() for an object in its finalizer is not allowed, and causes both assertion errors and actual memory safety issues.

I thought "rescuing" in the finalizer was allowed since v1.4.0 unless heapDestructing was set to true?

svaarala:
Rescuing is allowed, but duk_push_heapptr() for a pointer that has ever been unreachable is not allowed. You can rescue an object without duk_push_heapptr() too.

svaarala:
In practical terms using duk_push_heapptr() for the object being currently finalized will technically work in 2.0 and 1.x -- it's just contrary to the API contract so "undefined behavior" in a sense.

What won't work is using duk_push_heapptr() for any pointer pending finalization. This case is fixed by the change here.

harold-b:
Oh I see. I had remembered wrong thinking you had said it was allowed. ( referring to this comment: https://github.com/svaarala/duktape/issues/1311#issuecomment-274263788 ). But after re-reading, what you had said is that it is 'a violation, but should work OK'.

> You can rescue an object without duk_push_heapptr() too.

 I was using `duk_dup` to rescue the the object in question, which you mentioned is essentially doing `duk_push_heapptr`, then placing it on an array object. For my benefit, what's the way to rescue it without `duk_push_heapptr()`?

svaarala:
Yes, the object is already on the value stack and reachable when the finalizer begins. So using e.g. `duk_dup()` + `duk_put_global_string()` (in practice, writing to a stashed array or similar) is the "right" way to do it.

But as discussed, if you use `duk_push_heapptr()` in that situation, it's an API violation but Duktape won't notice in practice because `duk_get_heapptr()` would return the same pointer (heap objects have stable pointers) -- and a `duk_get_heapptr()` + `duk_push_heapptr()` *would* be allowed by the API contract.

In any case, with this pull duk_push_heapptr() would be allowed in the finalizer. But more importantly, it will be allowed before the finalizer runs which is really the main problem when using "weak" borrowed pointers.

harold-b:
Great, thanks for the info :)

svaarala:
@beevik Test coverage isn't ready, but this pull allows `duk_push_heapptr()` to push a pointer for an unreachable but not-yet-finalized object, right up to the finalizer. If you're interested in trying it out, it'd be nice to hear if it fixes the assert issues you had earlier.

beevik:
@svaarala Thanks, Sami. I'll try to test it out tomorrow.

beevik:
Ok I tried it out and I'm hitting an assertion now.  The assertion is `DUK_ASSERT(tmp == hdr) in `duk_heap_remove_from_heap_allocated`.

Here is the stack:

```
>	artifacts.exe!duk_heap_remove_from_heap_allocated(duk_heap * heap, duk_heaphdr * hdr) Line 46	C
 	artifacts.exe!duk__refcount_refzero_hobject(duk_heap * heap, duk_hobject * obj, int skip_free_pending) Line 323	C
 	artifacts.exe!duk__heaphdr_refzero_helper(duk_hthread * thr, duk_heaphdr * h, int skip_free_pending) Line 590	C
 	artifacts.exe!duk_heaphdr_refzero_norz(duk_hthread * thr, duk_heaphdr * h) Line 611	C
 	artifacts.exe!duk_hobject_refcount_finalize_norz(duk_heap * heap, duk_hobject * h) Line 76	C
 	artifacts.exe!duk_heap_process_finalize_list(duk_heap * heap) Line 286	C
 	artifacts.exe!duk_refzero_check_fast(duk_hthread * thr) Line 404	C
 	artifacts.exe!duk_set_top(duk_hthread * ctx, int idx) Line 475	C
 	artifacts.exe!duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 1680	C
 	artifacts.exe!duk_handle_call_unprotected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 1148	C
 	artifacts.exe!duk_new(duk_hthread * ctx, int nargs) Line 396	C
 	artifacts.exe!duk__js_execute_bytecode_inner(duk_hthread * entry_thread, unsigned __int64 entry_callstack_top) Line 4688	C
 	artifacts.exe!duk_js_execute_bytecode(duk_hthread * exec_thr) Line 2404	C
 	artifacts.exe!duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 1567	C
 	artifacts.exe!duk_handle_call_protected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 1060	C
 	artifacts.exe!duk_pcall_method(duk_hthread * ctx, int nargs) Line 159	C

```

beevik:
Here is another crash stack for the same assertion:

```
duk_heap_remove_from_heap_allocated(duk_heap * heap, duk_heaphdr * hdr) Line 46   C
duk__refcount_refzero_hobject(duk_heap * heap, duk_hobject * obj, int skip_free_pending) Line 323 C
duk__heaphdr_refzero_helper(duk_hthread * thr, duk_heaphdr * h, int skip_free_pending) Line 590   C
duk_heaphdr_refzero_norz(duk_hthread * thr, duk_heaphdr * h) Line 611 C
duk_hobject_refcount_finalize_norz(duk_heap * heap, duk_hobject * h) Line 76  C
duk_heap_process_finalize_list(duk_heap * heap) Line 286  C
duk_refzero_check_fast(duk_hthread * thr) Line 404    C
duk_set_top(duk_hthread * ctx, int idx) Line 475  C
duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 1680    C
duk_handle_call_unprotected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 1148 C
duk_new(duk_hthread * ctx, int nargs) Line 396    C
PushVector3(duk_hthread * context, const arch::Vector3 & v) Line 482    C++
GetPosition(duk_hthread * context) Line 869 C++
duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 1617    C
duk_handle_call_unprotected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 1148 C
duk_call_method(duk_hthread * ctx, int nargs) Line 84 C
duk_hobject_getprop(duk_hthread * thr, duk_tval_struct * tv_obj, duk_tval_struct * tv_key) Line 2678  C
duk__js_execute_bytecode_inner(duk_hthread * entry_thread, unsigned __int64 entry_callstack_top) Line 3665    C
duk_js_execute_bytecode(duk_hthread * exec_thr) Line 2404 C
duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 1567    C
duk_handle_call_unprotected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 1148 C
duk_call_method(duk_hthread * ctx, int nargs) Line 84 C
duk_hobject_getprop(duk_hthread * thr, duk_tval_struct * tv_obj, duk_tval_struct * tv_key) Line 2678  C
duk__js_execute_bytecode_inner(duk_hthread * entry_thread, unsigned __int64 entry_callstack_top) Line 3665    C
duk_js_execute_bytecode(duk_hthread * exec_thr) Line 2404 C
duk__handle_call_inner(duk_hthread * thr, int num_stack_args, unsigned int call_flags, int idx_func) Line 1567    C
duk_handle_call_protected(duk_hthread * thr, int num_stack_args, unsigned int call_flags) Line 1060   C
duk_pcall_method(duk_hthread * ctx, int nargs) Line 159   C
```

Here is the `hdr`:

```
hdr,x   0x000001ddecfb43c8   duk_heaphdr *
        h_flags 0x08204091  unsigned int
        h_assert_refcount   0x0000000000000002  unsigned __int64
        h_refcount  0x00000000  unsigned int
        h_next  0x000001ddecfb5388   duk_heaphdr *
        h_prev  0x000001ddecfb3a28   duk_heaphdr *
```

svaarala:
Hmm, so what happens is there's a refzero for an object on the finalize_list, and the attempt to remove it from the list fails. I'll need to think about this a bit; I may need to bump the refcounts on insert to finalize_list rather than just before running the finalizer. I'll get back after looking at the situation a bit more.

svaarala:
Ok, I think I managed to reproduce the same or similar issue:

```
((o) Duktape [linenoise] 2.0.99 (v2.0.0-368-g944abc03-dirty)
duk> X={}; Y={};
= {}
duk> Duktape.fin(X, function (v) { v.ref = null; });
= undefined
duk> Duktape.fin(Y, function (v) { v.ref = null; });
= undefined
duk> X.ref = Y;
= {}
duk> Y.ref = X;
= [object Object]
duk> X = Y = null
= null
duk> Duktape.gc()
D0 duk_error_macros.c:118 (duk_default_fatal_handler): custom default fatal error handler called: assertion failed: tmp == hdr (duk_heap_misc.c:46)
*** FATAL ERROR: assertion failed: tmp == hdr (duk_heap_misc.c:46)
```

This happens when a finalizer operates on another object on the finalize_list so that its refcount falls to zero.

This is unrelated to this pull, so I'll fix it in master first. The fix is to ensure finalize_list refcounts are bumped by 1 when they're inserted to finalize_list (rather than when their finalizer is called) so that their refcount can never drop to zero prematurely. Mark-and-sweep does the equivalent by considering finalize_list objects a reachability root.

I'll get back in a while with the fix.

fatcerberus:
Interesting, so the bug can be reproduced only in JS, therefore not directly related to `duk_push_heapptr()`.

svaarala:
Right, like I said above, it's not related to the duk_push_heapptr() change.

svaarala:
Ok, so this is handled by #1451 but the fix is already included in this pull (and will be rebased away).

@beevik Did this fix the issue for you?

beevik:
I haven't hit any assertions since updating to the new pull request. 

Interestingly, I also don't seem to have triggered the autorescue code either.  I set a breakpoint immediately after this `if` statement:

```C
	if (DUK_UNLIKELY(DUK_HEAPHDR_HAS_FINALIZABLE((duk_heaphdr *) ptr))) {
```

but haven't hit it yet.  It was previously rare that I would trigger the situation that necessitated autorescue, but I could normally hit it withing a few minutes of testing.  Now I never do.  Perhaps something changed on my end to make this scenario less likely.

svaarala:
There are some changes compared to master (if that's your previous test), but off-hand can't think why they would affect that behavior. Just rechecked with debug logs enabled that tests/api/test-push-heapptr-finalize-list.c does cover the autorescue clause (your breakpoint):

```
D0 duk_api_stack.c:4734 (duk_push_heapptr): duk_push_heapptr() with a pointer on finalize_list, autorescue
```

So, the automatic rescue behavior is at least not trivially masked by some other change.

svaarala:
@beevik There's a finalizer out-of-memory corner case that might affect your intended use, see #1456.

It's quite unlikely to trigger in practice, but it'd be nice to provide some guarantees as future work, so that carefully written finalizers would have a guarantee of being eventually entered even in out-of-memory conditions (or retried until entry succeeded or the heap was destroyed).

PR #1449, opened 2017-03-29
Make duk_catchers a linked list in individual duk_activations

svaarala:
- [x] Remove the catch stack from duk_hthread
- [x] Store current duk_catcher in each duk_activation, with each duk_catcher pointing to the surrounding catcher or NULL
- [x] Change catcher unwind to happen when call stack entries are unwound
- [x] For now, allocate duk_catchers individually without pooling/caching (future work to keep a small spare, with GC integration)
- [x] Review and finalize changes
- [x] Review code comments referencing catchstack
- [x] Testing and test case fixes
- [x] GC torture+assert run
- [x] Internal documentation
- [x] Releases entry

Footprint improvement varies, on x86 around 0.5kB while on x64 around 1.5kB (minsize).

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2813,6 +2813,12 @@ Planned
 
 * TBD
 
+2.2.0 (XXXX-XX-XX)
+------------------
+
+* Internal change: duk_catcher structs are now in a single linked list attached
+  to a duk_activation instead of being a separate thr->catchstack (GH-1449)
+
 3.0.0 (XXXX-XX-XX)
 ------------------
 
--- a/doc/memory-management.rst
+++ b/doc/memory-management.rst
@@ -403,9 +403,10 @@ ownership relationships::
            |
            +-->  value stack
            |
-           +-->  call stack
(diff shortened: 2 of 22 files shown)


svaarala:
Octane score in master was around 250 after the side effect rework (which improved refcount performance a bit). Improvement in this pull is marginal, but maybe around 252-253.

svaarala:
While Octane benchmark improves, without pooling/caching of duk_catcher allocations the fast break/continue microbenchmarks suffer a bit. That's to be expected because each catcher wind/unwind involves memory allocation calls. The fix is to use some heap level pools allowing reuse of catchers, but that'll be in a separate pull.

svaarala:
This pull also removes an explicit catch stack size limit. It's not really necessary in practice when catch stack elements are allocated individually (instead of a large monolithic chunk). Catch stack size is also automatically bound by call stack limit to some extent because the number of catchers for a certain activation simply depends on the number of nested label/trycatch/with sites in the source code.

PR #1452, opened 2017-03-29
Add shebang support to module-node

svaarala:
duk_compile() + DUK_COMPILE_SHEBANG is not enough for modules because there is a module function wrapper which offsets the shebang in the module source when Duktape actually sees it.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2555,6 +2555,8 @@ Planned
   source if DUK_COMPILE_SHEBANG option is given to duk_compile() (GH-1380,
   GH-1346)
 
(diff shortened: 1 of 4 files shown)


svaarala:
@fatcerberus Ok... So maybe now? :)

fatcerberus:
Yep, that should do the trick.  I'll need to adapt it for miniSphere's loader but the trick you did when pushing the source string is clever - I was thinking I would have to modify the original string, which is awkward for me since I use immutable strings internally for Unicode text.

fatcerberus:
Indeed, this works :)

svaarala:
Good to hear :)

PR #1457, opened 2017-03-31
Allow mark-and-sweep during finalize_list processing

svaarala:
When finalizers are executed from finalize_list, master (and Duktape 2.0 and prior) prevent mark-and-sweep in the middle of the process. The main reason to do so is to avoid incorrect rescue decisions:
- An object X is queued for finalization by mark-and-sweep.
- X gets finalized and is queued back to heap_allocated with FINALIZED set.
- Another finalizer runs while finalize_list is still not empty, and mark-and-sweep is triggered.
- If X is unreachable except for references from the objects still on finalize_list (which are considered reachability roots to prevent collection), mark-and-sweep decides to rescue the object.

This decision is incorrect: had the finalize_list been fully processed, and mark-and-sweep executed afterwards, the object would be freed as unreachable. User code also cannot change the outcome by calling duk_push_heapptr(); even with relaxed rules it's not allowed to call duk_push_heapptr() after the finalizer for that object has been executed *and* the object is no longer reachable.

The harm from an incorrect rescue decision is that FINALIZED gets cleared and the object is thus eligible for another finalization. This may happen multiple times, with the object getting unnecessarily finalized an arbitrary number of times.

This pull allows mark-and-sweep during finalize_list processing with a small tweak:
- When mark-and-sweep starts with heap->finalize_list != NULL, an automatic flag "don't rescue objects" is set.
- When a rescue decision is about to be made, the object is queued to heap_allocated (as if rescued); however, FINALIZED is kept. So the object remains in the "pending rescue/free decision" state until next mark-and-sweep.

Tasks:
- [x] Rebase once #1442 is merged
- [x] Mark-and-sweep and finalizer processing basic changes
- [x] Fix mark-and-sweep assertions: FINALIZED is no longer set for an object currently being finalized (detect the specific object with an assert-only heap field)
- [x] Test coverage for mark-and-sweep running while finalizers processed
- [x] Test coverage for an object that would get incorrectly rescued without the tweak
- [x] Internal documentation changes
- [x] External documentation changes
- [x] Document mark-and-sweep rescue limitation in Finalization page
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2608,9 +2608,9 @@ Planned
   for mark-and-sweep to confirm its status (GH-1427)
 
 * Rework finalizer handling: finalizer execution is now outside of refzero
-  processing and mark-and-sweep; however, mark-and-sweep is still disabled
-  while finalizers are being executed to avoid incorrect rescue decisions
-  caused by a partially processed finalize_list (GH-1427, GH-1451)
+  processing and mark-and-sweep, and mark-and-sweep (but not recursive
+  finalizer handling) is allowed during finalizer execution (GH-1427, GH-1451,
+  GH-1457)
 
 * Rework mark-and-sweep: include finalize_list in TEMPROOT marking; with
   duk_push_heapptr() allowed it's possible for application code to create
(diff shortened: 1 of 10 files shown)


svaarala:
A more fine-grained solution would be to track multiple types of reachability, so that the rescue decision could see that the object is unreachable from "actual" reachability roots but reachable from "artificial" reachability roots. It would still be unsafe to free the object because objects on finalize_list do in fact reference the object and their finalizers may operate on the references. So this doesn't really provide any better behavior.

PR #1458, opened 2017-04-03
Add duk_opt_xxx() API calls with explicit default for undefined/oob index

svaarala:
Add duk_opt_xxx() variants which provide an explicit default for a missing argument (interpreted as "undefined or value stack index invalid"). The intended usage is for optional arguments, for example:

```c
duk_ret_t connect(duk_context *ctx) {
    const char *host = duk_opt_string(ctx, 0, "localhost");
    int port = (int) duk_opt_int(ctx, 1, 80);
}
```

Both undefined and missing value (invalid index) are accepted to trigger defaulting. This allows vararg functions to ignore the difference between explicit undefined arguments vs. missing arguments, similarly to how Ecmascript functions deal with undefined/missing arguments. Lua's luaL_optstring() et al treat nil and missing the same way, triggering the default.

If the value is `null` or other non-matching type, a TypeError is thrown. Null is not recognized as an acceptable input to trigger defaulting; this is to match ES2015 optional parameter behavior where an undefined (or missing) argument triggers defaulting. There are some downsides to this too; for example:
- Providing an `undefined` to trigger defaulting for a parameter in the middle is a bit awkward because `undefined` is not a keyword. The safe idiom is something like: `connect(void 0, 8080)` which is much less readable than `connect(null, 8080)`.
- Sometimes Ecmascript APIs treat undefined and null the same way. For such functions these API calls are not ideal and the caller needs to use an explicit type check.

The Ecmascript equivalent of the above host/port example is:

```js
function connect(host, port) {
    if (typeof host === 'undefined') {
        host = 'localhost';
    } else if (typeof host !== 'string') {
        throw new TypeError('invalid argument');
    }
    if (typeof port === 'undefined') {
        port = 80;
    } else if (typeof host !== 'number') {
        throw new TypeError('invalid argument');
    }
}
```

Tasks:
- [x] Clean up the source diff, still some leftovers from the previous branch
- [x] out_len, out_size, def_len, def_size consistency
- [x] Testcase coverage
- [x] Testcase C constant size fixes (e.g. 0x12345678 -> 0x12345678UL)
- [x] Test coverage for duk_opt_buffer_data() and unbacked buffer
- [x] API documentation
- [x] Finalize and merge https://github.com/svaarala/duktape-wiki/pull/161 at the same time
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2557,6 +2557,12 @@ Planned
 
 * Add shebang support to module-node (GH-1452)
 
+* Add duk_opt_xxx() API calls which behave like duk_require_xxx() but allow a
+  default value to be used when the index doesn't exist or the value is
+  undefined (null is rejected with TypeError to mimic ES2015 optional
+  arguments); for example: "int port = duk_opt_int(ctx, -3, 80);"
+  (GH-1458)
+
 * Spawn the ArrayBuffer object backing a typed array lazily when its .buffer
   property is first read, reducing memory usage in common cases where the view
   is constructed directly without needing the ArrayBuffer object (GH-1225)
--- a/src-input/duk_api_codec.c
+++ b/src-input/duk_api_codec.c
@@ -18,7 +18,10 @@ DUK_LOCAL const duk_uint8_t *duk__prep_codec_arg(duk_context *ctx, duk_idx_t idx
 
 	DUK_ASSERT(duk_is_valid_index(ctx, idx));  /* checked by caller */
 
-	ptr = duk_get_buffer_data_raw(ctx, idx, out_len, 0 /*throw_flag*/, &isbuffer);
+	/* XXX: with def_ptr set to a stack related pointer, isbuffer could
+	 * be removed from the helper?
+	 */
+	ptr = duk_get_buffer_data_raw(ctx, idx, out_len, NULL /*def_ptr*/, 0 /*def_size*/, 0 /*throw_flag*/, &isbuffer);
 	if (isbuffer) {
 		DUK_ASSERT(*out_len == 0 || ptr != NULL);
 		return (const duk_uint8_t *) ptr;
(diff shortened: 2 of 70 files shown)


svaarala:
@fatcerberus API-wise this is now complete, I'm cleaning around the edges a bit.

One of the issues under discussion was treatment of out-of-bounds values. The pull now triggers defaulting as if they were `undefined`, and the two major arguments for that in my mind are:
- That's the way Ecmascript functions deal with defaulting; also `arguments[N] === undefined` even if N >= arguments.length.
- The behavior matches Lua luaL_optxxx() bindings, see e.g. http://pgl.yoyo.org/luai/i/luaL_optstring.

Re: whether `duk_is_undefined()` should be true for out-of-bounds indices, interestingly `lua_isnil()` is false for out-of-bounds indices, matching current `duk_is_undefined()` behavior. So this pull and current `duk_is_undefined()` match Lua behavior. This is by no means a critical requirement of course, but still nice.

fatcerberus:
Yeah, defaulting for out-of-bounds makes perfect sense given usual Ecmascript semantics and idioms.  It was that `duk_is_undefined()` change that seemed maybe a bit awkward to me, since passing an out-of-bounds index to most other Duktape APIs will throw an error (which I like) rather than treat it as `undefined`.

fatcerberus:
Hm, maybe a cautionary note in the documentation for `duk_opt_string()` that if the default is chosen, `def_ptr` will be returned as-is, therefore it's not safe to do:

```c
some_str = duk_opt_string(ctx, 0, heap_str);
free(heap_str);
// no good: `some_str` is now dangling
```

Particularly in comparison to this, which *is* safe:
```c
duk_push_string(ctx, heap_str);
some_str = duk_require_string(ctx, -1);
free(heap_str);
// still OK
```

svaarala:
Right, that applies also to buffer calls, the caller is always responsible for the pointers involved being useful. But a specific note might be useful for string, lstring, buffer, buffer_data, heapptr variants at least.

fatcerberus:
Yeah, it struck me that there should be a note since if you push a string and read it back with `duk_require_string()` you always get a Duktape-managed pointer, whereas with `duk_opt_string()` etc. you may get your own self-managed pointer back.

PR #1461, opened 2017-04-05
Define _XOPEN_SOURCE for Nspire (needed for strptime)

Legimet:
The strptime declaration isn't exposed unless _XOPEN_SOURCE is defined.


svaarala:
Looks good to me, thanks!

PR #1463, opened 2017-04-05
Fix missing SyntaxError for invalid RegExp groups like /(?Xabc)/

svaarala:


fatcerberus:
So if these are invalid but were silently accepted before this merge, how would such a regexp behave in practice?

svaarala:
From what I can tell advtok would be zero, which would generate an EOF token ending the regexp. So for example, `/abc(?X)def/` would be parsed as `/abc/`.

svaarala:
Maybe worth mentioning that while the behavior is incorrect, I don't think this results in any memory unsafe behavior, so no 'security' label because of that.

PR #1472, opened 2017-04-09
Add duk_get_xxx_default() API calls with an explicit default value

svaarala:
- [x] Add duk_get_xxx_default() API calls with an explicit default, e.g. `duk_get_uint_default(ctx, -3, 80)`.
- [x] Testcases
- [x] API docs
- [x] Wiki update - https://github.com/svaarala/duktape-wiki/pull/170
- [x] Releases entry


svaarala:
@harold-b @fatcerberus I'll be merging this in a few days, so unless a clearly better name idea comes by before that, duk_get_xxx_default() it is :-)

fatcerberus:
I'm still partial to `duk_maybe_xxx()` myself, but I know you expressed concerns about ambiguity with that, so nothing else comes to mind.

harold-b:
Sorry for the late reply, been a little busy :)
The only one which sounded ok to me was 'duk_read_xxx' but, it falls under the same criticisms as the other names I suppose.  Just read your https://github.com/svaarala/duktape/issues/1263#issuecomment-292754225, and I suppose then if it had that proposal already maybe that's the way to go even if verbose? I thought about names through the past months and couldn't come up with anything better :(

svaarala:
Ok, I'll merge this as is, with duk_get_xxx_default(). If a better idea comes to mind after 2.1.0, it'll be relatively straightforward to make these an alias for the better naming.

PR #1483, opened 2017-04-14
Accept Annex B octal up to \377 and \8, \9 literals in RegExp character classes

### LINKED ISSUE #1275: Annex B RegExp character class decimal escape in TypeScript 2.1.4
Normally I don't report real-world regexp breakages anymore because I know that nonstandard regexp support is still a work in progress.  However, since this breaks the TypeScript compiler it seemed worthwhile to call it out.  The offending regexp is:
```js
var tripleSlashDirectiveFragmentRegex = /^(\/\/\/\s*<reference\s+(path|types)\s*=\s*(?:'|"))([^\3"]*)$/;
```

I don't know if there are more broken regexps, but that's the one that caused an error for me.  It was a bit too convoluted for me to try to find the offending token :)

svaarala:
Tasks:
- [x] Initial hack to test parsing approach
- [x] Refactor so that legacy octal parsing is shared between normal lexer and regexp charclass lexer
- [x] Testcases (from #1275)
- [x] Releases entry
- [x] Duktape-wiki post-es5 feature => RegExp Annex B syntax features are not described individually

Fixes #1275.

### DIFF
--- a/tests/ecmascript/test-lex-legacy-octal-regexp.js
+++ b/tests/ecmascript/test-lex-legacy-octal-regexp.js
@@ -1,0 +1,333 @@
+/*
+ *  Legacy octal literals like \377 in regexps.
+ */
+
+var re;
+
+/*===
+true false
+===*/
+
+// \0 is interpreted as a literal NUL (U+0000), even without legacy support.
+try {
+    re = /[\0]/;
+    print(re.test('\u0000'), re.test('0'));
+} catch (e) {
+    print(e.name);
+}
+
+/*===
+true false
+true false
+true false
+true false
+true false
+true false
+true false
+===*/
+
+// \1 to \7 are octal U+0001 to U+0007.
+try {
+    re = /[\1]/;
+    print(re.test('\u0001'), re.test('1'));
+    re = /[\2]/;
+    print(re.test('\u0002'), re.test('2'));
+    re = /[\3]/;
+    print(re.test('\u0003'), re.test('3'));
+    re = /[\4]/;
+    print(re.test('\u0004'), re.test('4'));
+    re = /[\5]/;
+    print(re.test('\u0005'), re.test('5'));
+    re = /[\6]/;
+    print(re.test('\u0006'), re.test('6'));
+    re = /[\7]/;
+    print(re.test('\u0007'), re.test('7'));
+} catch (e) {
+    print(e.name);
+}
+
+/*===
+false true
+false true
+===*/
+
+// \8 and \9 are literal '8' (U+0038) and '9' (U+0039) characters,
+// *NOT* U+0008 and U+0009.
+try {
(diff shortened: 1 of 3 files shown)


svaarala:
Just a quick hack test which seems to work, I'll rework and add tests later today.

tests/ecmascript/test-lex-legacy-octal-regexp.js
```diff
@@ -0,0 +1,321 @@
...
+ *  Legacy octal literals like \377 in regexps.
+ */
+
+var re;
+
+/*===
+true false
+===*/
+
+// \0 is interpreted as a literal zero, even without legacy support.
```
fatcerberus: I assume you mean "interpreted as NUL" here?

svaarala: Yeah, I'll fix.

svaarala:
Seems to be accepted by Node.js in strict mode.

svaarala:
To be clear, the code diff is just a hack (comments/debug prints are copy-paste etc) for now.

fatcerberus:
> To be clear, the code diff is just a hack (comments/debug prints are copy-paste etc) for now.

I noticed, don't worry. :)

In any case it does seem to be working as intended.  The TypeScript compiler runs without manual modification, at least.

svaarala:
Good to hear :)

fatcerberus:
Interestingly enough, Node.js:
```
> (function() { 'use strict'; return /\3/; })()
/\3/
```

Seems to accept octal escapes in regexps in strict mode.  Strings, however, are rejected:
```
> (function() { 'use strict'; return "\3"; })()
(function() { 'use strict'; return "\3"; })()

SyntaxError: Octal literals are not allowed in strict mode.
```

So not sure whether V8 is out of spec or if that's the correct behavior.

fatcerberus:
That's with Node 7.8.0, by the way.

svaarala:
As far as I can tell the octal literals etc in strings are explicitly forbidden in strict mode, Annex B text:

> The syntax and semantics of 11.8.3 is extended as follows except that this extension is not allowed for strict mode code: [...]

But no strict mode requirements exists for additional RegExp syntax. So I'd say it's a feature.

svaarala:
@fatcerberus I'm done with the code changes and I'll merge this shortly unless something comes up :)

fatcerberus:
That was quick :)

fatcerberus:
I guess the reason legacy octal is allowed in regexps even in strict mode is for compatibility with other (non-JS) regex engines.  It would be annoying to copy a regex from elsewhere and maybe have it rejected just because the code happens to be in strict mode.  That's less of an issue for, e.g., string literals, since their syntax has no implicit meaning outside the context of Ecmascript.

PR #1487, opened 2017-04-18
Rework thr->callstack into a linked list of duk_activations

### LINKED ISSUE #1490: Callstack_max bumping for error augmentation doesn't work correctly with coroutines involved
Current handling for providing some recursion/callstack limit headroom for error augmentation doesn't work correctly when coroutines are involved. Callstack_max may be bumped for thread A, and unbumped for thread B, which leaves thread A's callstack_max bumped and thread B's callstack_max unchanged.

This is a very minor bug in practice though, and only affects cases where error augmentation is attempted when callstack/recursion limit has been reached and coroutines are involved. There's no memory unsafe behavior; error augmentation headroom may simply not be available for error augmentation once the callstack_max is left "stuck". Other than that things will work normally. The "stuck" callstack_max is also most likely reset in the next error throw unless the situation happens again.

svaarala:
Remove thr->callstack array of duk_activations and replace it with a linked list, with thr->callstack_curr holding the currently active entry and act->parent chaining upwards in the call chain.

The upside of the change is that there is one fewer large monolithic allocations which are problematic for low memory targets. Instead, there are a lot of smaller allocations maintained as a linked list. Another upside is that 'act' pointers are now stable, so a lot of side effect questions are eliminated.

One downside is that random access is no longer possible (though it can be accelerated later if necessary); however, very few call sites actually need the random access capability. Another downside is that there are a lot of small allocations when calling and returning; this can be mitigated by duk_activation pooling (to be implemented separately).

Tasks:
- [x] Initial draft of changes
- [x] Clean up FIXMEs, compile warnings, etc
- [x] Callstack limit check, to be kept
- [x] Go through 'act = ...' re-lookups; no longer needed because 'act' is now stable, but keep if they make footprint better (they may, because re-lookup may be smaller/faster than store/restore past function call(s))
- [x] Rework thr->callstack_max headroom when creating error: the previous solution wasn't 100% when coroutines were involved (setting and restoring the limit might happen in different threads!); new solution is to add headroom in call handling (without a store/restore) when heap->augmenting_error is set
- [x] Remove internal IGNORE_RECLIMIT flag and rework it to be conditional on heap->augmenting_error dynamically without an explicit store/restore (much easier to handle correctly)
- [x] Remove ERRHANDLER_RUNNING flag and use heap->augmenting_error instead
- [x] Testcase fixes
- [x] Internal documentation changes
- [x] Perf test
- [x] Torture tests
- [x] Releases entry

Fixes #1490.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2816,8 +2816,19 @@ Planned
 2.2.0 (XXXX-XX-XX)
 ------------------
 
+* Fix callstack limit bumping for errThrow augmentation calls, the limit might
+  be bumped and unbumped for different Duktape threads if coroutines were
+  resumed/yielded in the process; this is relatively harmless but might cause
+  an errThrow augmentation call to fail due to callstack limit being reached
+  (GH-1490)
+
+* Internal change: duk_activation structs are now in a single linked list
+  attached to a duk_hthread instead of being a separate, monolithic
+  thr->callstack (GH-1487)
+
 * Internal change: duk_catcher structs are now in a single linked list attached
-  to a duk_activation instead of being a separate thr->catchstack (GH-1449)
+  to a duk_activation instead of being a separate, monolithic
+  thr->catchstack (GH-1449)
 
 3.0.0 (XXXX-XX-XX)
 ------------------
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -794,7 +794,7 @@ Released
 * Fix assignment evaluation order issue which affected expressions like
   "a[i] = b[i++]" (GH-118)
 
-* Fix incorrect parsing of zero escape in regexp class ("[\0]") (GH-122)
+* Fix incorrect parsing of zero escape in regexp class ("[\\0]") (GH-122)
 
 * Fix tail call issue in return comma expression when a function call
   in the comma expression was followed by a constant value or a register
@@ -886,7 +886,7 @@ Released
 * Fix value stack setup bug which caused a segfault with large number of
   arguments (GH-107)
 
-* Fix incorrect parsing of zero escape in regexp class ("[\0]") (GH-122)
+* Fix incorrect parsing of zero escape in regexp class ("[\\0]") (GH-122)
 
 * Fix assignment evaluation order issue which affected expressions like
   "a[i] = b[i++]" (GH-118)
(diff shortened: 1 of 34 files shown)


svaarala:
At present this change reduces Octane performance by around 10 points. This is caused by the current pull not doing any pooling of duk_activations, so each call involves an alloc for `sizeof(duk_activation)`. A follow-up pull will add that pooling, and before merging this pull, I'll do a small experiment to confirm this will fix the issue. On the upside this also means adding pooling for a few common allocations (like empty objects, objects with small property tables) will probably improve performance a bit and will reduce the allocator call load.

svaarala:
Adding a very simple duk_activation free list (as a hack test) shows that call handling will improve with the linked list duk_activation + free list / pooling:

```
duktape@keppihevonen:~/duktape$ time ./duk.O2.210 tests/perf/test-call-basic-1.js 

real	0m7.921s
user	0m7.920s
sys	0m0.000s
duktape@keppihevonen:~/duktape$ time ./duk.O2 tests/perf/test-call-basic-1.js 

real	0m7.451s
user	0m7.448s
sys	0m0.000s
```

I won't implement the freelist here, but I'll implement a duk_catcher and duk_activation freelist as a separate pull. The follow-up for that will then be a shared pool allocation helper which can be used for empty objects, arrays, etc.

fatcerberus:
I imagine the object pooling alone will be a pretty big improvement for object literal-heavy code.

svaarala:
I did a very quick prototype of that and there was some improvement but not huge. Ultimately it depends on the allocation provider, some of them are very fast and others not so fast.

svaarala:
Tested Octane score with the simple freelist hack and it would increase from ~250 to ~255 with the other changes in this pull, so looks good. The freelist won't be included in this pull though, this was just a test to ensure that the sequence of changes is actually an improvement.

fatcerberus:
It's been my experience that malloc on Windows is pretty expensive.  In the past I discovered a few hot paths in miniSphere where I malloc'd and free'd some scratch memory in a tight loop and it turned out that moving the allocation outside of the loop improved performance a lot.

Linux malloc seems a lot cheaper, which may explain Cell's TypeScript module being noticeably faster in that environment.

svaarala:
If the allocator is slow it might be worth it for Duktape to do some caching of small allocations (say up to 128 bytes in 4-byte increments) with the idea that frees would go to the cache and be reused, with mark-and-sweep keeping the cache in check. This would make on platforms with a slow allocator and which don't have extremely tight memory bounds.

svaarala:
@fatcerberus I'm reworking the callstack_max bump for error augmentation a bit here too to fix an outstanding bug. Instead of bumping callstack_max in error throwing and unbumping it later, the limit is no longer touched directly. Instead, `heap->augmenting_error` is set for the duration of error augmentation and it affects the limit check in call handling dynamically without other state changes.

The small bug in the current code is that the callstack_max store and restore may happen in different Duktape threads when coroutines are involved, so the code ends up bumping callstack_max in thread A and restoring it in thread B -- leaving A's callstack_max bumped and thread B's callstack_max unchanged.

Issue #1493, opened 2017-04-20
Change Duktape/C function .length to match ES2015 requirements


svaarala:
Currently Duktape/C function .length is non-configurable which differs from ES2015 Ecmascript functions and built-in functions. It's not a violation of ES2015 because "host" functions can be different, but it'd make sense to change the property attributes to match ES2015.

PR #1494, opened 2017-04-20
Fix forced writing of Duktape/C function .length

svaarala:
Duktape/C function .length is a "virtual" property so it's not an own property but also not inherited. The current handling prevents a DUK_DEFPROP_FORCE write to establish a .length own property for a Duktape/C function.

Nicest solution would be for Function.prototype.length to be an inherited accessor which could have special handling for Duktape/C functions, to provide a useful .length based on nargs count. However, that would violate the ES2015 property setup so the property needs to be virtual.

- [ ] Add testcase coverage
- [ ] Fix for virtual property writing, should allow one to establish a property in general even if a virtual property conflicts
- [ ] Releases entry
- [ ] Backports

The particular case for function .length might become moot with #1493 but other cases remain valid.

### DIFF
--- a/tests/api/test-dev-set-duktapec-func-length.c
+++ b/tests/api/test-dev-set-duktapec-func-length.c
@@ -1,0 +1,41 @@
+/*===
+===*/
+
+static duk_ret_t dummy_func(duk_context *ctx) {
+	(void) ctx;
+	return 0;
+}
+
+static duk_ret_t test_basic(duk_context *ctx, void *udata) {
+	(void) udata;
+
+	duk_push_c_function(ctx, dummy_func, 1);
+	duk_push_string(ctx, "length");
+	duk_push_int(ctx, 3);
+	duk_def_prop(ctx, -3, DUK_DEFPROP_HAVE_VALUE |
+	                      DUK_DEFPROP_CLEAR_WRITABLE |
+	                      DUK_DEFPROP_CLEAR_ENUMERABLE |
+	                      DUK_DEFPROP_SET_CONFIGURABLE |
+	                      DUK_DEFPROP_FORCE);
+
+	printf("added concrete .length\n");
+
+	duk_eval_string(ctx,
+		"(function (fn) {\n"
+		"    print(typeof fn);\n"
+		"    var desc = Object.getOwnPropertyDescriptor(fn, 'length');\n"
+		"    print(desc.writable);\n"
+		"    print(desc.enumerable);\n"
+		"    print(desc.configurable);\n"
+		"    print(desc.value);\n"
+		"})");
+	duk_dup(ctx, -2);
+	duk_call(ctx, 1);
+
+	printf("final top: %ld\n", (long) duk_get_top(ctx));
+	return 0;
+}
+
+void test(duk_context *ctx) {
+	TEST_SAFE_CALL(test_basic);
+}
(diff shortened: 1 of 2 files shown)


svaarala:
In general defineProperty() might just allow establishing a concrete property if no concrete property exists but a virtual one does, and DUK_DEFPROP_FORCE is set. This doesn't always make sense though, e.g. Array `.length` is always virtual and if a concrete property is created it will actually not have effect.

So maybe cleanest solution would be to add one more internal attribute flag for virtual properties, to indicate whether it's allowed to write over the virtual property or not.

fatcerberus:
> Nicest solution would be for Function.prototype.length to be an inherited accessor which could have special handling for Duktape/C functions, to provide a useful .length based on nargs count. However, that would violate the ES2015 property setup so the property needs to be virtual.

Out of curiosity, why is this the case?  Is it because...
```
> Object.getOwnPropertyDescriptor(Function.prototype, 'length')
{ value: 0,
  writable: false,
  enumerable: false,
  configurable: true }
>
```

...`Function.prototype.length` must be a data property == 0 and not an accessor?

svaarala:
Hmm, if you mean why `.length` must be a virtual property:

Duktape/C functions don't have a concrete `.length` because of memory concerns: the property would use a lot of RAM for low memory targets (for 100 functions, 8x100 = 800 bytes). The `.length` value is also almost always the same as the `nargs` count so there's rarely a need to actually change it. The best compromise IMO would be for the property to be virtual but behave like an own property so you could overwrite it.

The easiest solution would to make the property a concrete one, but that has RAM concerns. The second easiest would be for Function.prototype.length to be an accessor which can come up with a useful value for Duktape/C functions automatically. But that's not allowed by ES because the property is described as a data property, unlike e.g. the new RegExp.prototype.xxx properties which are accessors.

I could also make a concrete `.length` the default and use a virtual one for low memory targets. But I'd like to avoid that if possible, because even for non-low-memory targets minimizing memory usage is still useful.

Not sure if I covered your question here though :)

fatcerberus:
> Function.prototype.length to be an accessor which can come up with a useful value for Duktape/C functions automatically. But that's not allowed by ES because the property is described as a data property

Right, this is exactly the answer I was looking for. :)

To be clear, I knew why there were no `length` own properties, I was just trying to figure out why the inherited accessor solution wasn't feasible.  I suspected it might be a spec issue, just wanted to verify.

fatcerberus:
Another solution could be to have Duktape/C functions inherit from a special `NativeFunction` class instead of `Function` directly, but that's a bit overkill just to get an inherited `.length`. ;-)

svaarala:
Hmm, that sounds pretty good actually because it would remove the need for virtual properties (which are an "internal" feature with no Ecmascript visibility). Lightfuncs could also inherit from NativeFunction and get their virtual properties from there.

The only downsides I can think of is slightly slower method lookups (but function methods are not usually looked up in hot paths) and some additional built-in objects. The NativeFunction prototype wouldn't need to have a global name so it wouldn't cause global object changes either (this would be similar to %TypedArrayPrototype%).

fatcerberus:
It would be important to ensure that only user-provided functions inherit from NativeFunction and not e.g. builtins, since otherwise that would surely not be standards compliant :)

svaarala:
That's true. It's a bit odd but the built-ins have explicit `.length` properties so the inheritance won't matter.

svaarala:
This is handled by #1515, so closing as unmerged.

PR #1500, opened 2017-04-23
Add internal duk_hproxy struct

svaarala:
Add a `duk_hobject` sub-struct `duk_hproxy` for representing Proxy instances, avoiding a property table and internal properties.

Tasks:
- [x] Add internal struct, mark-and-sweep, refcount, etc, changes
- [x] Make duk_push_proxy() and proxy handling conditional to config option
- [x] Add duk_hproxy to sizeof debug dump
- [x] Document duk_push_proxy() API call
- [x] Testcase for duk_push_proxy() API call
- [x] Testcase fixes for duk_hproxy
- [x] Releases entry


fatcerberus:
From what I saw when I was working with property lookup code, the proxy stuff could use to be a little more modular in general, so this is a good start. :-)

svaarala:
This doesn't actually affect the inlined Proxy checks in property code all that much, as the handler/target lookups were already isolated in a helper which now just gets a bit simpler.

The main non-modularity with property code is that it's intentionally monolithic, encapsulating all special behaviors in a single set of functions. It's a bit unwieldy to maintain, so a more modular approach where entry points like getprop, putprop, etc, would be looked up based on e.g. an object's internal class would make property code easier to maintain. The cost would probably be a bit more footprint, because many of the entry points would need to duplicate certain functionality like fast paths. I still think this is a useful step at some point though, as it should also improve performance.

fatcerberus:
> This doesn't actually affect the inlined Proxy checks in property code all that much

I figured as much, was just praising this change in particular as any change, however small, in the direction of more modularity for Proxy is welcome :)

svaarala:
Agreed :)

svaarala:
One more data structure pull I'm trying to get into 2.2 is adding a sub-struct for bound functions (`duk_hboundfunc`) to make bound function handling smaller and better performing. They're used more than one might imagine because of certain idioms.

fatcerberus:
One potential snag I can see there is that bound functions can themselves be bound--so you'd need some sort of chaining support built into the structure I guess.

svaarala:
Right, there's no required change to current bound function handling though: the _Target function can be bound and the chain is resolved to completion where it matters.

But I'll probably change the handling so that the chain is always collapsed: when creating a bound function for an already bound target, the bound this/arguments can be collapsed so that the new bound function has a non-bound target.

PR #1503, opened 2017-04-25
Add internal duk_hboundfunc type and collapse bound function chains

svaarala:
Add a separate internal `duk_hboundfunc` type for faster bound function handling. Also collapse bound function chains on creation so that the target of a bound function is always a non-bound function directly.

- [x] Add duk_hboundfunc, changes in call handling etc to use it
- [x] Improve bound function call handling to allocate a gap and copy tvals there rather than using duk_dup() + duk_insert() which gets quite slow for large argument counts
- [x] Check for footprint improvements, follow-ups if necessary
- [x] Review and finalize changes
- [x] Testcase changes
- [x] Check dump/load handling for bound functions, add test coverage
- [x] Releases entry


fatcerberus:
> Also collapse bound function chains on creation so that the target of a bound function is always a non-bound function directly.

Huh, I guess the standard does allow this, since bound functions are opaque by design.  I love clever little optimizations like this. :)

svaarala:
It would also be possible for the bound function to be essentially a copy of the function (pointing to the same bytecode/native function) with automatic arguments, eliminating the separate reference and allowing the target to be GC'd if it's not referenced elsewhere. Not sure if this would be an improvement in practice :)

PR #903, opened 2016-08-10
Rework executor to use a single level dispatch

svaarala:
This is a work in progress (far from complete) for reworking the bytecode format and the executor for a single level dispatch:
- Opcode format uses 8 bits for opcode, A, B, C each. Larger opcode field eliminates the need for EXTRA opcodes and a second level dispatch.
- Reg/const arguments are handled by moving the reg/const flags from the previously 9-bit (1-bit reg/const flag + 8-bit index) B and C register slots into the opcode field; for example, `ADD` consumes four consecutive opcode slots for different reg/const combinations for B and C fields.
- The revised reg/const handling allows the executor to handle each reg/const combination separately because each of the four opcodes can be dispatched to a separate case clause for faster execution
- For footprint optimized builds the four opcodes can map to the same case clause which does the reg/const explicitly as before.

The potential speedup for arithmetic heavy code is around 10-20%. The mandelbrot test, for example, improves from around 4.6 seconds to 3.9 seconds with optimizations currently in place (not yet complete).

Footprint for a performance optimized build will increase significantly, around 20kB or more. This is because the reg/const cases can now map to separate case clauses which have a significant footprint especially when inlining helpers.

Footprint for a size optimized build will likely increase somewhat - at the moment around 1kB but hopefully that can be brought down a bit.

Tasks:
- [x] Change opcode format macros, defines, etc
- [x] Compiler changes for opcode format
- [x] Executor changes for opcode format
- [x] Executor direct reg/const dispatch variants
- [x] Fix obvious remaining compiler and executor bugs by going through opcodes
- [x] Downgrade logging
- [x] SETALEN duk_harray optimization
- [x] Go through all opcodes and their field assignments, check for improvements; in particular, previous EXTRA opcodes can now use the A field
- [x] Reorganize opcodes a bit, eliminate opcode variants when e.g. B is never a constant
- [x] Debugger support for new opcode format
- [x] Debugger opcode metadata
- [x] Testcase fixes
- [x] Run testcases manually with compact and optimized executor
- [x] Run shuffle torture manually
- [x] Test footprint impact on ARM (as an example of another architecture)
- [x] Releases entry
- [x] Follow up issues/pulls => #908

PR #1510, opened 2017-05-02
Move large and rare opcodes into NOINLINE helpers

svaarala:
Move handling of the following opcodes into NOINLINE helpers to make the main dispatch function smaller and a bit better performing:

- INITSET and INITGET
- TRYCATCH
- ENDTRY
- ENDCATCH
- ENDFIN
- INITENUM and NEXTENUM

There are more potential opcodes for similar treatment, but these are quite clear cases.

Tasks:
- [x] Move opcodes to helpers
- [x] Check if a conditional NOINLINE improves footprint
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2858,6 +2858,9 @@ Planned
 * Internal change: simple freelists for duk_activation and duk_catcher
   (GH-1491)
 
+* Miscellaneous performance improvements: move rare/large opcodes into
+  NOINLINE helpers (GH-1510)
+
 3.0.0 (XXXX-XX-XX)
 ------------------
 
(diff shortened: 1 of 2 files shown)


fatcerberus:
This feels like it will be a regression of #903, i.e. reintroducing double dispatch.

svaarala:
It's quite unrelated and actually opposite: it makes most opcodes run a bit faster which more than pays off the small penalty for calling the rare opcode handlers.

svaarala:
As an example, Octane improved from around 725 to around 740, and as a specific example test-mandel.js improved from around 3.1 seconds to 2.9 seconds.

fatcerberus:
Huh, counterintuitive, but hard to argue with results. :)

svaarala:
The main problem with double dispatch was that there were two switch-case branches for some opcodes. Another problem was that the top level switch contained a lot of relatively small cases (with some variance of course) and one huge sub-case for the second level dispatch.

This change doesn't reintroduce either problem: there's still just one level of dispatch, and by moving large opcode handlers out of the main dispatch loop, performance improves a bit because code cache usage for the dispatcher is a bit better. It certainly varies per platform.

PR #1515, opened 2017-05-05
Add a %NativeFunctionPrototype% for user created Duktape/C functions and lightfuncs

### LINKED ISSUE #1493: Change Duktape/C function .length to match ES2015 requirements

svaarala:
Inheritance chain for user created Duktape/C functions and lightfuncs is changed to:

```
value -> %NativeFunctionPrototype% -> Function.prototype
```

The %NativeFunctionPrototype% object is internal and isn't explicitly registered to a global object. You can still access it e.g. as `myFunc.__proto__`, similarly to %TypedArrayPrototype%. The previously virtual `.length` and `.name` are now getters in %NativeFunctionPrototype%, and handle both normal functions and lightfuncs. Exotic behavior for Duktape/C functions and lightfuncs is removed from property, enumeration, and stack API code.

The change is not applied to built-in native functions (like Math.cos) because their inheritance chain is defined in the Ecmascript specifications and they have concrete .name and .length in any case.

Tasks:
- [x] Fix genbuiltins.py to allow a .length accessor
- [x] Core changes throughout
- [x] Remove DUK_HOBJECT_FLAG_DUKFUNC
- [x] %NativeFunctionPrototype% code comment consistency
- [x] Duktape/C function automatic .name change? => separate pull
- [x] FIXME cleanups
- [x] Testcases: writing .name and .length fails (mimics ES2015 own properties which are not writable but are configurable)
- [x] Testcases: defineProperty() works for .name and .length
- [x] Testcases: replace NativeFunction.prototype.name with a custom implementation
- [x] Testcases: built-in prototypes are correct (top level, properties, getters/setters)
- [x] Check #1493
- [x] Check #1494
- [x] Release notes migration entry
- [x] Releases entry

Fixes #1493: the own `.length` property (which was virtual) is removed and an inherited getter replaces it. The getter behaves close to the ES2015 `.length` property: it's not directly writable but defineProperty() can be used to change it.


svaarala:
At this point this is rough prototyping of what the change would mean and if there are any non-obvious pros/cons.

svaarala:
One interesting opportunity here would be to give Duktape/C functions which don't have an explicit `.name` property some automatically generated name, like for lightfuncs:

```
duk> Math.sin.name
= "light_0805ec91_0a11"
```

So maybe it could be something like:

```
duk> print.name
= "native_0805cafe_0000"
```

The _0000 prefix would be the 16-bit magic value which is in some sense part of the function identity (as for lightfuncs).

The automatic name wouldn't prevent establishing an actual name, it would just default to something more useful than an empty string.

Not yet sure if this would be a good change. Common expectation is that anonymous functions have an empty string as their name and so far Duktape/C functions are anonymous by default.

svaarala:
Concrete example, print() doesn't have a name, and in 2.1.0:

```
duk> print.name
= ""
duk> print({ toString: function mythrower() { throw new Error('aiee'); } })
Error: aiee
    at mythrower (input:1) preventsyield
    at [anon] () native strict preventsyield
```

With an automatic .name:

```
duk> print.name
"native_00000000004d4164_ffff_0000"/1
= "native_00000000004d4164_ffff_0000"
duk> print({ toString: function mythrower() { throw new Error('aiee'); } })
Error: aiee
    at mythrower (input:1) preventsyield
    at native_00000000004d4164_ffff_0000 () native strict preventsyield
```

Naturally one would prefer to see "print" in the traceback, but if no name exists (as for print), maybe an automatic name would be better. Then again, it looks quite technical and unless you have a symbol file, it's not very useful. It does identify particular functions even without a symbol file, i.e. you can see whether two functions in two different tracebacks are the same or not.

kungfooman:
Can't we just add and then encourage to use such a little helper function, which sets the name when it's still known from context? Like:

	int duk_register_function(duk_context *ctx, char *name, int (*func)(duk_context *ctx)) {
		duk_push_c_function(ctx, func, DUK_VARARGS);
		duk_push_string(ctx, name);
		duk_put_prop_string(ctx, -2, "name");
		duk_put_global_string(ctx, name);
		return 1;
	}

So every function has the name it's "supposed" to have. And the other ones could have such a technical name, so they are distinctive in tracebacks.

svaarala:
@kungfooman Adding such an API would be a useful improvement and there are issues proposing that. But the question remains what to do with functions that don't have a name which will still remain for a few reasons: existing code using duk_push_c_function() will create such functions, and sometimes that's desirable to minimize memory usage on memory constrained devices. Lightfuncs which don't have any property slots are also used in such environments and they have a virtual .name for similar reasons.

fatcerberus:
I'm inclined to say that, by the principle of least surprise, non-lightfunc Duktape/C functions without a `.name` should remain anonymous (= empty string), as with anonymous JS functions.  I see the automatic naming for lightfuncs as a necessary evil because they physically *can't* have a name, but no such restriction exists for regular funcs.

On the other hand, I can see the value for low-memory targets where RAM is tight.  So maybe it could be a compile-time option?

svaarala:
@fatcerberus It could be an option, sure.

I'll leave it out of this pull (some of the code will be #if 0'd because I already wrote it :) so it can be discussed separately.

svaarala:
@fatcerberus And to be clear, I see your point, and I agree some code might expect functions to genuinely be anonymous. I'm not sure what the use case is in practice, but it may still be surprising if there's an automatic name.

svaarala:
```
duk> Object.defineProperty(print.__proto__, 'name', { get: function() { var i = Duktape.info(this); return 'native_' + i.hptr; } })
= {}
duk> print.name
= "native_0x6cd790"
```

fatcerberus:
Hm, using the hfunc pointer as a default name might not be too useful for debugging as it will probably change from run to run.  You wouldn't be able to create a meaningful symbol file.

svaarala:
I agree, just copy-pasted an example of how to override the %NativeFunctionPrototype%'s name for a later test case. It wasn't intended to be an actual use case.

svaarala:
@fatcerberus Here's a more realistic dynamic .name example, based on a hypothetical runtime C function pointer -> name lookup: https://github.com/svaarala/duktape/blob/0db47637c238f42148ba4d14e2bed6d774208735/tests/api/test-dev-override-duktapec-func-getters.c.

svaarala:
This change is somewhat borderline with respect to versioning guarantees. In my view it's OK because it's similar to ES2015 inheritance chain changes which are also made in minor version bumps. The change is mentioned separately in 2.2 migration notes just in case.

PR #1516, opened 2017-05-05
Fix duk_dump_function() .length handling when _Formals omitted

### LINKED ISSUE #1513: Function .length mismatch when run from source code and load from bytecode on duktape 2.1 release
Sample code
```javascript
var func = function (__filename, __dirname, module, exports, require) {
        function hello(a, b, c, d) {
            console.log('hello lengh in hello', hello.length);
        }

        console.log('hello.lengh', hello.length);
        hello();
};

func();
```
Run directly with `duk`

```
hello.lengh 4
hello lengh in hello 4
```

Compile to byte with `duk -c` and run again, the output is

```
hello.lengh 0
hello lengh in hello 0
```

svaarala:
Fixes #1513.

Tasks:
- [x] Distinguish between a missing and zero-length _Formals when loading back a dumped function, which avoids issues with function instance .length selection
- [x] Simplify _Formals dumping by accessing duk_harray directly
- [x] Releases entry


svaarala:
The cleanest fix seems to be to support a missing _Formals correctly. Previously a missing _Formals and a zero-length _Formals were treated the same which causes the reported bug. Distinguishing the two cases fixes the original issue too with no forced length tweaks.

This fix can't be applied to 2.1.1 though because it needs a bytecode dump/load format change. So the backported fix will need to be something different.

PR #1531, opened 2017-05-20
Add __defineGetter__() and __defineSetter__()

### LINKED ISSUE #1529: Object.prototype.__define{Getter,Setter}__ native support
These are included in Annex B of Ecmascript 2017:
https://raw.githubusercontent.com/tc39/ecma262/es2017/spec.html#sec-additional-properties-of-the-object.prototype-object

These two functions are used by a LOT of legacy code (predating `Object.defineProperty()` by quite some time), enough so that it's almost *always* necessary to polyfill them.  It'd be better just to provide them natively.

svaarala:
Fixes #1529.

Tasks:
- [x] Add `__defineGetter__()` and `__defineSetter()__` to Object.prototype based on ES2017 draft
- [x] Conditional (low memory targets)
- [x] Testcases
- [x] Documentation
- [x] Wiki post-ES5
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2848,6 +2848,9 @@ Planned
   this change also allows .name and .length to be overridden using
   duk_def_prop() or Object.defineProperty() (GH-1493, GH-1494, GH-1515)
 
+* Add Object.prototype.{__defineGetter__,__defineSetter__} from ES7, they
+  are also used by a lot of existing legacy code (GH-1531)
+
 * Handle Function.prototype.call(), Function.prototype.apply(),
   Reflect.apply(), and Reflect.construct() inline in call handling; as a side
   effect .call(), .apply(), and .construct() no longer appear in the call stack
--- a/config/config-options/DUK_USE_ES7.yaml
+++ b/config/config-options/DUK_USE_ES7.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_ES7
+introduced: 2.0.0
+default: true
+tags:
+  - ecmascript2016
+description: >
+  Enable ES7 functionality not covered by other specific config options.
--- a/config/config-options/DUK_USE_ES8.yaml
+++ b/config/config-options/DUK_USE_ES8.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_ES8
+introduced: 2.0.0
+default: true
+tags:
+  - ecmascript2017
+description: >
+  Enable ES8 functionality not covered by other specific config options.
--- a/config/config-options/DUK_USE_ES9.yaml
+++ b/config/config-options/DUK_USE_ES9.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_ES9
+introduced: 2.0.0
+default: true
+tags:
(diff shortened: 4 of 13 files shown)


svaarala:
```
((o) Duktape [linenoise] 2.1.99 (v2.1.0-109-g54bef09e-dirty)
duk> global.__defineGetter__('foo', function () { print('FOO'); return 'bar' })
= undefined
duk> foo
FOO
= "bar"
```

fatcerberus:
I was going to look at adding these, you beat me to it :)

svaarala:
D'oh :)

denisdemaisbr:
this  behavior is same of pollyfills (already include in pollyfills ) ?

fatcerberus:
Right, the purpose of a polyfill is to provide functionality that's not yet provided natively.  The ultimate goal is always to *not* need a polyfill. ;-)

svaarala:
I wouldn't want to add a built-in binding for a purely legacy feature - but if this is likely to become part of Annex B it makes sense to include it. Annex B is, to be clear, optional legacy features which are not recommended at all for new code; but they are still common enough to be expcted by many users, unfortunately even for new code.

fatcerberus:
Right - if it's not a standard feature or on track to become one, I don't usually call something a "polyfill", just a shim :)

fatcerberus:
But yeah, these seem to be very commonly used for a legacy feature--I guess because they're the only way to create accessors in ES3.

fatcerberus:
The draft additionally specifies `__lookupGetter__()` and `__lookupSetter__()`; should those go into this pull as well?

svaarala:
Maybe - could also be a separate pull because there's no shared code.

svaarala:
Made these conditional to `DUK_USE_ES8` for lack of a better solution now; there's no single clear decision yet for post-ES5 config options - but it's important to maintain a low memory ES5ish build which is satisfied.

fatcerberus:
Since the spec is updated yearly, maybe these options could be merged into a single `DUK_USE_POST_ES5` option.

svaarala:
Yep, that's one option.

svaarala:
Using a post-ES5 config option would be good enough to support "plain ES5.1" and "all post-ES5.1 stuff" use cases. Unfortunately often people want an ES5.1 baseline and cherry pick a few features on top of that (say, if WeakMap was available one might want just that, or perhaps just Promises, etc). If that's not possible users are forced to edit Duktape sources directly which they obviously would prefer to avoid.

Supporting granularity ultimately needs individual config options for each built-in object (or built-in object method even) and each significant post-ES5.1 syntax feature.

Some config options are now with relatively fine granularity and allow cherry picking, while others are not. It would reduce friction if one could say e.g. `python configure.py --es5.1 --enable-builtin Proxy -- enable-builtin Promise` or something, and the config options would be somehow automatically named, say `DUK_USE_BUILTIN_PROXY` etc.

So if fine grained configuration is something worth supporting it might need some more automation so that it's more convenient to support. But no decision on that yet, so this pull just uses `DUK_USE_ES8`.

PR #1540, opened 2017-05-31
Fix debugger GetHeapObjInfo property list

svaarala:
Property key list had a missing comma, which caused two key names to be concatenated. As a result GetHeapObjInfo might segfault. Also present in 2.1.0, but not 2.0.0 or prior.


fatcerberus:
Sometimes I find myself wishing Ecmascript had string literal coalescing like C... then I get bitten by bugs like this and change my mind. :)

Incidentally, this kind of thing is also the main reason I came to prefer trailing commas.

PR #1544, opened 2017-06-04
Add minimal support for new.target

svaarala:
Add minimal support for `new.target` syntax. Without internal support for newTarget (or class) handling as such, this is mostly useful to allow Ecmascript functions to distinguish between normal calls (new.target == undefined) and constructor calls (new.target == constructor function), e.g.:

```js
function MyConstructor() {
    if (!new.target) {
        throw new Error('must be called as constructor');
    }
}
```

Tasks:
- [x] Add DIRECT EVAL flag to eval body call too; it wasn't needed before but the current new.target implementation needs it
- [x] Add compiler support for `new.target`, NEWTARGET opcode
- [x] Add executor NEWTARGET opcode support
- [x] Proper handling for new.target in different code contexts => some eval limitations left, separate issue
- [x] Syntax error messages
- [x] Config option condition; `DUK_USE_ES6` for now
- [x] Debugger metadata
- [x] Testcase coverage
- [x] Releases entry


svaarala:
`new.target` is a SyntaxError in program code (script code) and in eval code except for direct evals,
http://www.ecma-international.org/ecma-262/6.0/#sec-scripts-static-semantics-early-errors:

> It is a Syntax Error if StatementList Contains NewTarget unless the source code containing NewTarget is eval code that is being processed by a direct eval that is contained in function code that is not the function code of an ArrowFunction.

fatcerberus:
I guess that's true, it doesn't make much sense outside of a function.  The rule for arrow functions seems arbitrary though... are those not constructable?

svaarala:
From Firefox:

```
function foo() { eval('console.log(new.target)'); }
new foo()
--> Object {  }

function foo() { (1,eval)('console.log(new.target)'); }
new foo()
--> SyntaxError: new.target only allowed within functions
```

svaarala:
Nested direct evals should work, e.g. in Firefox:

```
eval('new (function test() { eval("eval(\\"print(typeof new.target); print(new.target === test);\\");"); })');
```

would print:

```
function
true
```

This is simply because a direct eval inherits the scope from the calling function, and GetNewTarget() specification operation just looks up the `[[NewTarget]]` from the scope. Current implementation doesn't actually store NewTarget in the scope so eval() cases fail right now. This can be fixed by changing the `new.target` lookup to use an actual scope backed internal key or detecting a direct eval and looking up the call stack parent as many times as a direct eval is encountered.

svaarala:
> The rule for arrow functions seems arbitrary though... are those not constructable?

Arrow functions are indeed not constructable:

```
> var fn = x => x*x;
undefined
> fn(3)
9
> new fn()
TypeError: x => x*x is not a constructor
```

Issue #1548, opened 2017-06-08
Handle eval() directly in call handling


svaarala:
Call handling in Duktape 2.2 handles .call(), .apply(), and .construct() inline in call handling so they won't appear in the call stack. Eval could be handled similarly.

Currently eval is visible as two call stack entries:
1. The built-in eval() function itself.
2. The anonymous eval body called as Ecmascript code.

With inline handling only the eval body would be visible (and should be flagged 'eval code' for the stacktrace). This would also allow yield from eval, remove native recursion from eval calls, and allow evals in tailcall position.

PR #1552, opened 2017-06-11
Rework call handling for smaller footprint and less duplication

svaarala:
Rework call handling to reduce footprint and duplication:
- Remove internal duk_handle_call_protected() and implement protected calls using duk_safe_call() wrapping.
- Remove separate Ecma-to-Ecma call setup helper and add Ecma-to-Ecma call support to duk_handle_call_unprotected().

Several smaller changes:
- Rework call handling to use small helper functions for better readability.
- Rework DUK_OP_NEW so that default instance is created by bytecode and DUK_OP_CONSCALL (replaces DUK_OP_NEW) is just an ordinary call with a constructor flag set.
- Remove special handling for initial resume call.
- Support Ecma-to-Ecma calls for `eval(xxx)` calls when the target function is not actually eval but a shadowing Ecmascript function (this worked before 2.0 release also).
- Use internal duk_remove_n() primitive to improve tailcall handling value stack manipulation.
- Remove value stack reserve adjustments from duk_safe_call(): caller must ensure them and safe calls shouldn't adjust the reserve.
- Reduce internal valstack extra from 64 to 32 (could maybe be reduced to 16 or 24); entry to Duktape/C function now automatically reserved 64 + 16 elements (64 is the user reserve guaranteed on entry).

Tasks:
- [x] Finalize changes
- [x] Testcase coverage
- [x] Manual full assertion run
- [x] Documentation
- [x] Releases entry
- [x] 2.2 migration notes: internal reserve change, duk_safe_call() reservation behavior

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2940,6 +2940,10 @@ Planned
   calling a function or reserving value stack space explicitly and only
   shrunk in mark-and-sweep (GH-1526)
 
+* Remove automatic value stack reserve check from duk_safe_call() to
+  ensure 'nrets' return values fit into the value stack; the check was not
+  guaranteed by the API and is mostly unnecessary (GH-1552)
+
 * Internal change: set REACHABLE for all READONLY objects (relevant when
   using ROM built-ins) so that mark-and-sweep doesn't need an explicit
   READONLY check (GH-1502)
@@ -2954,6 +2958,9 @@ Planned
 
 * Internal change: simple freelists for duk_activation and duk_catcher
   (GH-1491)
+
+* Miscellaneous footprint improvements: rework call handling to improve
+  code sharing (GH-1552)
 
 * Miscellaneous performance improvements: move rare/large opcodes into
   NOINLINE helpers (GH-1510), duk_harray fast path for internal array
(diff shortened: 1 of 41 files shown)


svaarala:
@fatcerberus This is slowly becoming merge ready - if you have time it'd be interesting to hear if this breaks anything in Minisphere as there are quite a lot of (useful :) changes in the pull. I haven't run a full torture+assertion run yet which takes a long time but will do that before merging.

There are still a few call handling things I'd like to fix but most of the hard changes will be in master with this. Remaining things include:
- Ability to yield from eval(), and making eval() code appear more usefully in a call stack, #1548
- Tailcall for eval()
- Allowing more call/new parameters

fatcerberus:
Quick initial test suggests everything is working as designed.  No misbehavior that I could see.  I'll use this branch for a bit internally and let you know if anything weird happens.

svaarala:
Thanks, good to hear :) Besides footprint reduction the pull also improves call handling performance a bit (very roughly 5%). Probably not visible in Minisphere though.

fatcerberus:
I haven't measured it so it could just be a placebo effect, but TypeScript transpilation in Cell seems a bit faster at least. :)

fatcerberus:
This is merged, goody. :)

PR #1557, opened 2017-06-12
Fix tailcall in constructor handling

### LINKED ISSUE #1554: Latest master can't evaluate core-js
Using this copy of core-js (ES6+ polyfill):
https://github.com/fatcerberus/minisphere/blob/master/assets/system/shim.js

And this command-line:
```
./duk shim.js
```

Duktape 2.1.0 successfully runs the file, producing no output.  master (695b187eee4b7c22dffb2b02166c21a9a484e83f) produces:
```
fatcerberus@pigcult-vm:~/src/duktape$ ./duk shim.js
TypeError: cannot read property 'set' of undefined
    at [anon] (duk_hobject_props.c:2299) internal
    at [anon] (shim.js:4775) strict
    at __webpack_require__ (shim.js:28) strict
    at [anon] (shim.js:160) strict
    at __webpack_require__ (shim.js:28) strict tailcall
    at [anon] (shim.js:7431) strict
    at global (shim.js:7438) preventsyield
error in executing file shim.js
```

In my case, core-js failing to execute causes segfaults later in miniSphere, which I don't fully understand why (something to figure out later), but the TypeError itself is clearly a Duktape regression.

svaarala:
Fix bug in constructor call handling where a tailcall inside a constructor would be handled with a tailcall, losing the necessary internal book-keeping to perform constructor call return value post-processsing.

This is described in #1554 and is only broken in master so no need to merge into 2.1.x maintenance etc.

Fixes #1554.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2855,13 +2855,13 @@ Planned
 * Handle Function.prototype.call(), Function.prototype.apply(),
   Reflect.apply(), and Reflect.construct() inline in call handling; as a side
   effect .call(), .apply(), and .construct() no longer appear in the call stack
-  (tracebacks etc) (GH-1421, GH-1522, GH-1545)
+  (tracebacks etc) (GH-1421, GH-1522, GH-1545, GH-1557)
 
 * Function.prototype.call(), Function.prototype.apply(), Reflect.apply(),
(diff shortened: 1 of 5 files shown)


svaarala:
Still need to fix test-bi-reflect-construct-tail.js, I'll do that later today.

PR #1561, opened 2017-06-13
Add ES2015 Math clz32(), imul(), sign()

svaarala:
- [x] Add functions
- [x] Testcases
- [x] Releases

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2852,6 +2852,8 @@ Planned
   Object.prototype.{__lookupGetter__,__lookupSetter__} from ES2017 Annex B,
   they are also used by a lot of existing legacy code (GH-1531, GH-1551)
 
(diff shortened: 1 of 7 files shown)


svaarala:
`Math.imul()` is quite interesting in that the arguments are ToUint32() coerced but the result is (in effect) a signed 32-bit integer. Seems odd but this seems to be the case all through the latest ES draft.

fatcerberus:
Looking at the specification, I'd assume the reason for the `ToUint32()` coercion (as opposed to ToInt32) is a practical one: If an engine stores the result of ToUint32() as a 32-bit value (a pretty safe bet), then `a * b` will overflow, and you get the mod-2^32 behavior automatically.  Signed overflow on the other hand, is undefined behavior.  The way I see it, it's specified this way to minimize the potential for implementation pitfalls.

PR #1567, opened 2017-06-15
Add getOwnPropertyDescriptor Proxy trap

svaarala:
Add minimal support for getOwnPropertyDescriptor Proxy trap to support virtualized enumeration.

Tasks:
- [ ] Add Proxy passthrough for reading a descriptor when a getOwnPropertyDescriptor trap doesn't exist
- [ ] Add getOwnPropertyDescriptor trap call, minimally to handling enumeration at least
- [ ] Test coverage for passthrough
- [ ] Test coverage for trap call
- [ ] Specific test for virtualized enumeration
- [ ] Releases entry


svaarala:
One internal difficulty with this change is that most internals deal with `duk_propdesc` struct which provides property attributes, the property's storage location (if any), and *borrowed* references to value/get/set.

With the Proxy `getOwnPropertyDescriptor` call the borrowed value/get/set reference is a problem because the trap result is quite likely not reachable once it is popped (as expected by the current internal call sites).

So, for now, when `getOwnPropertyDescriptor` trap is invoked, I'm just checking and filling in only the property attributes and nothing else. This is enough for enumeration but not for other call sites.

PR #1574, opened 2017-06-20
Added support for __filename canonicalization and __dirname

mike-lischke:
In NodeJS __filename uses an absolute path while the duktape-node extra had only relative paths. This was mostly done to avoid file system access when this code was integral part of duktape. Now that it is an extra I believe we can relax the no-filesystem rule and make __filename absolute. Also this patch adds __dirname, as this is also used frequently.


fatcerberus:
I actually use a variant of this extra in miniSphere where this change doesn't make sense.  My engine is sandboxed such that the canonical representation of a filename is relative to the root of the game directory--and in the case of SPK-packaged games, the OS filesystem is not even directly used at all.  Thus I'm otherwise Node-compatible but `__filename` is not the actual file system path, but instead something like `bin/foo.js`--which is "absolute" as far as the asset resolver is concerned.

Note that technically duktape-node extra doesn't use "relative paths" specifically--the `modResolve` function is supposed to return the canonical module ID from whatever was passed in--which in most cases would of course mean returning the absolute path. (since in Node, `module.id === __filename`).

So just based on sandboxing concerns alone, and the fact that people may want to use this on platforms that might not have a standard file system (or one at all, for that matter), I don't think this change makes sense.  @svaarala, what do you think?

mike-lischke:
Well, that seems to be a specific requirement for your game engine, but that code is supposed to mirror what NodeJS does and this is what's in the patch. Btw, in NodeJS module.id is also an absolute path while in this patch I left it what was used to load the module. Otherwise the duktape debugger extension in vscode stops working :-(

fatcerberus:
I wrote the extra.  The intent was to be compatible (as far as possible) with Node modules, not be a full emulation.  If the user wants the NodeJS resolving behavior, that's meant to be done in `modResolve`, provided by the user.  Even if full Node compatibility was a goal, the point about use on systems without a standard file system stands.  Thus the current more generic, pluggable setup makes most sense, I think.  I'll wait to see what @svaarala thinks though. :)

mike-lischke:
Ah, and that code works fine for me, thanks! I implemented the NodeJS module resolution on top of your code (using the `load` and `resolve` calls) and that works great. I can load already quite a number of existing node modules. I don't need `modeResolve` and I think that was removed in 2.x anyway.

fatcerberus:
Yeah, load and resolve, that's right.  I forgot they were renamed, hehe.  I do agree that `__dirname` should be supported by the way; I just couldn't figure out how to do it in a generic manner.  Now might be a good time to revisit that, maybe it could be worked into the API for `resolve` somehow.

svaarala:
Generally speaking Duktape tries to be an embeddable engine -- which to me means that the application writer should be in control of how Duktape ties to its environment. Otherwise it will be necessary in some environments (sandboxed, virtualized, etc) to patch the core or extra code to avoid direct file I/O etc which is a hassle and contrary to the goal of being embeddable.

But in case there's a feature that can't be generalized at all, then I'd prefer the feature to be optional so that it's not forced on environments that do need sandboxing.

I'm not sure about this feature however (I'd need to dig into it a bit more to say something useful :). Or maybe you guys already worked it out? :)

mike-lischke:
I thought, since this is an extra, it **is** already optional. So the patch would qualify :-D

fatcerberus:
I object to the patch only on the grounds that it takes freedom away from the `resolve` callback, which for sandboxing reasons is designed to be the ultimate authority on what the application sees as the "canonical" module filename.

Duktape is commonly used in sandboxing scenarios, so I believe the current approach makes most sense as those same developers might want Node-style module compatibility within the sandbox.

Hm... what if we provided the Node-like FS resolution as a default behavior?  In other words, leave the extra set up the way it is now, but provide a default `resolve` implementation that resolves against the standard file system.  Then developers that need the sandboxing behavior (like me) can plug in a suitable `resolve`, while more Node-like environments such as dukluv could use the default one.

fatcerberus:
Maybe a default implementation of `load` as well, that reads modules from the file system.

svaarala:
@mike-lischke The components bundled as extras are not as strictly portable or embeddable as core Duktape, but it doesn't really make sense to bundle a non-embeddable component with an embeddable engine. It would make more sense in such a case to move the extra into a separate repo and just link to it, just like any other Duktape-related project.

The main reason to bundle is when some functionality is not part of the core, but commonly needed, and when bundling it's awkward if the component won't be generally usable. Module loaders are part of ES2015 so one reason for bundling them (instead of providing them as separate repos) is that a user will almost always need one, at least when using ES2015. Some of the other extras could be in separate repos and I might move some of them elsewhere (like the logging framework which was previously part of the core, but because there isn't a general logging standard, isn't needed in the same way as module loaders).

In any case, extras are intended to be quite widely usable in scenarios where Duktape is used, and that includes embedding. I've personally had to fight Lua and a host of Lua libraries for embedding and it's no fun when you need to patch everything to get hard I/O dependencies out of internals :-P

svaarala:
But back to this pull/issue, it'd be ideal to find a way to provide the end result @mike-lischke wants but in some embedding friendly manner. Can the functionality be abstracted to a provider callback?

fatcerberus:
> can the functionality be abstracted to a provider callback?

Yeah, that's what I meant with my suggestion for a default implementation of `resolve` and `load`.

svaarala:
@fatcerberus Hmm, so can the desired functionality be implemented in the resolve/load callbacks, as module-node stands right now? If so, maybe it'd be easiest to provide an example resolve/load provider that had the functionality in a nicely commented form?

I'd be okay with a default resolve/load too, but I'm just wondering whether a single default provider can ever really work for practical use cases; unlike Node.js, Duktape is quite commonly used in embedding and there are very often some special constraints like translating paths to a virtual root (say, prepending a `/foo/bar/quux/` path to "absolute" paths), access control, mapping file I/O to a ZIP/TAR file, etc.

svaarala:
One particular thing I'd like to avoid is linking against fopen() et al even if they're not used. Some embedded systems (several that I work with for example) don't provide fopen() at all, so it's important that the default code is not compiled in if you don't want to use it. In those systems it's common to e.g. bundle modules into a single ZIP file that gets compiled in, and modules are loaded from the ZIP file (I'm using https://github.com/svaarala/lowzip for example), with no actual I/O.

That suggest two alternatives:
- Allow some define to be used to disable the default load/resolve so that no unnecessary symbols are referenced.
- Or, make the default load/resolve a separate file that can be omitted from compilation entirely.

svaarala:
Hmm, if the load/resolve example with direct filesystem bindings was in a separate file that could easily be omitted for embedding, this would also have the nice feature that it'd be straightforward to add DLL support to that file for Windows, Linux, and OSX. Having an easy-to-use DLL loader example is definitely something that's missing now.

fatcerberus:
I think I'd agree that a separate file is best.  Maybe I'll try to work up a pull later.

mike-lischke:
Hmm, I'm not sure what you have in mind. Maybe you can show an example, but I could also imagine a simple #ifdef which wraps the part that accesses the file system and make it so configurable.

fatcerberus:
What I mean is to provide a pre-made `resolve` and `load` implementation that replicates the Node behavior, then you'd just call `duk_module_node_init()` and provide those instead of your own ones.

mike-lischke:
Done as you wish.

PR #1587, opened 2017-06-27
Add a flag that flushes the console output after writing

sstone1:
We are using duktape (via go-duktape) to provide a JavaScript interpreter for Hyperledger Composer (https://hyperledger.github.io/composer/), which is a framework for rapidly building Blockchain solutions. More specifically, we use duktape to execute business logic that is written by our users in JavaScript.

Our users use `console.log()` as a quick and dirty debugging tool, however with duktape we see that the outputs are buffered and don't appear in the logs until the buffer is flushed. This is not great, hence this PR.

I am adding a new flag `DUK_CONSOLE_FLUSH` to the console extra that calls `fflush` after every write. This flag is off by default, so shouldn't change the behaviour for existing users.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -82,6 +82,7 @@ bugs, provided ideas, etc; roughly in order of appearance):
 * https://github.com/chris-y
 * Laurent Zubiaur (https://github.com/lzubiaur)
 * Neil Kolban (https://github.com/nkolban)
+* Simon Stone (https://github.com/sstone1)
 
 If you are accidentally missing from this list, send me an e-mail
 (``sami.vaarala@iki.fi``) and I'll fix the omission.
--- a/extras/console/duk_console.c
+++ b/extras/console/duk_console.c
@@ -22,14 +22,11 @@
  * to a custom backend.
  */
 
-/* XXX: For now output is not flushed, add a flush flag, or maybe add flush
- * to info level and above only.
- */
-
 /* XXX: Init console object using duk_def_prop() when that call is available. */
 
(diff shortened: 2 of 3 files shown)


svaarala:
Looks good to me, the only thing that I was wondering about is if there was an easy way to avoid looking up the flag on every log write.

The two simple solutions would be:
- Use the function 'magic' value to carry flags to the native calls. This still leaves the flag check in, but it's much cheaper than a property lookup + boolean check.
- Define two native functions, one with a flush and one without, and select the function at init time. The native functions can call into a shared helper which takes the flush flag as a parameter which is cheaper than the property lookup (and can be inlined if one wishes).

svaarala:
Hmm, I can do that optimization in a separate pull.

svaarala:
@sstone1 Just noticed you are in two wrong list in AUTHORS.rst, could you open a pull to fix that? Thanks :)

fatcerberus:
> Just noticed you are in the wrong list in AUTHORS.rst

I’m starting to wonder if people keep making this mistake because they don’t realize there are two lists?

svaarala:
Probably, people are often in a hurry (including me) and it's difficult to pay attention to every detail. Maybe the second list (which doesn't involve actual code/documentation contributions) could be in a separate file.

sstone1:
Whoops - yeah, I just went to the end of the file and added myself - incoming PR!
Thanks for merging this one 😃

svaarala:
The authors file is clearly error prone because this mistake happens a lot :) So I could either move the second list to a separate file, or maybe switch the order of the lists ;)

PR #1588, opened 2017-06-27
Some cleanups for DUK_CONSOLE_FLUSH

svaarala:
- [x] Faster flush flag lookup: pass flags into the log functions through the function magic value
- [x] Add console flush to 'duk' command
- [x] Releases entry


svaarala:
@sstone1 This pull has a slightly faster solution for the flush flag check - it'd be good to hear that this works for you; I tested the flag passing manually and it seems to work.

sstone1:
@svaarala have picked up your changes, it works great for me - thanks! 👍

PR #1589, opened 2017-06-27
Move myself to the right list

sstone1:
Fix the AUTHORS.rst from my last PR to put myself into the right list.


svaarala:
Thanks!

PR #1601, opened 2017-07-08
Add 'apply' and 'construct' Proxy traps with a few limitations

svaarala:
Add support for 'apply' and 'construct' Proxy traps. The Proxy and Proxy trap check is done inline in call handling which allows Ecmascript-to-Ecmascript calls and tailcalls in many situations (provided that return value handling is compatible for a tailcall).

Proxy 'construct' trap calls introduce a third return value handling policy. The three policies now supported are:
- Normal call: no return value post-processing.
- Constructor call: default instance vs. replacement value check.
- Proxy 'construct' trap: Proxy invariant check, i.e. reject non-object return values.

Limitation:
- When constructing through a Proxy that does not have a 'construct' trap, the target function gets called as a constructor. During the constructor call `new.target` will evaluate to the Proxy target rather than the Proxy itself (which is the compliant behavior). This limitation is caused by the current `new.target` implementation; there isn't enough internal state yet to handle the difference.

Internal changes:
- Add an explicit CALLABLE flag for objects. Proxies are CALLABLE/CONSTRUCTABLE based on the target function's same flags. This also allows construct-only functions to be defined internally. I'll change built-ins to use this model to avoid unnecessary constructor call checks; public API needs separate work.
- Add internal tracking of Proxy 'construct' trap calls for call_flags and duk_activation flags, so that the Proxy 'construct' trap return value policy can be checked where appropriate.
- Tailcall check for matching return value policy, now handles all three possible policies.
- Minor duk_hobject flag trivia, e.g. evaluate proxy and bufferobject flag checks to 0 when not supported at all.

Tasks:
- [x] Source changes
- [x] Testcase updates to cover new behavior
- [x] Post-ES5 feature list pull in Wiki - https://github.com/svaarala/duktape-wiki/pull/181
- [x] Releases entry
- [x] Issue tracking for 'construct' trap limitation


fatcerberus:
> Proxies are CALLABLE/CONSTRUCTABLE based on the target function's same flags.

This means it's not possible to apply/construct a Proxy whose target object is not a function?  Huh, and here I assumed an `apply` trap could be used to implement functors.

svaarala:
The target doesn't actually need to be called, so you can just:

```js
new Proxy(function dummy() {}, {
    apply: function () { return 'foo' }
});
```

The target never gets called.

fatcerberus:
Yeah, it works if the target is a function.  I meant the case of Proxy'ing an otherwise normal object to add callability to it.  That wouldn't be possible since CALLABLE won't be set and therefore Duktape will throw a not-callable error before it even gets to invoke the trap.

svaarala:
That's true, it's unfortunately required ES2015 behavior. Proxies are not really an ideal solution for virtualization IMO, their intended use case seems quite a bit different. So to me it makes sense to implement the Proxy to completion, and then figure out if a Duktape-specific virtualization mechanism makes sense, especially with native code in mind. Alternatively the Proxy could be extended with enough custom behavior to make it work better for virtualization.

svaarala:
Resolved the `.prototype` lookup limitation and fixed various small bugs. This is close to merge ready.

PR #1603, opened 2017-07-10
Allow built-in functions to be construct only

svaarala:
With the explicit DUK_HOBJECT_FLAG_CALLABLE it's now possible to create functions that are constructable but not callable as normal functions. This is a prototype branch to see how that would work out in practice. Work in progress.

### DIFF
--- a/src-input/duk_hobject.h
+++ b/src-input/duk_hobject.h
@@ -36,6 +36,11 @@
  * inspection code.
  */
 
+/* FIXME: ES spec: all construct-only functions have [[Call]] and [[Construct]]
+ * so the modeling is not 1:1.  Maybe rename so that _CALLABLE means either one?
+ * _CONSTRUCTABLE..
+ */
+
 /* XXX: some flags are object subtype specific (e.g. common to all function
  * subtypes, duk_harray, etc) and could be reused for different subtypes.
  */
@@ -187,6 +192,7 @@
                                                         DUK_HOBJECT_FLAG_NATFUNC)
 
 #define DUK_HOBJECT_IS_CALLABLE(h)             DUK_HOBJECT_HAS_CALLABLE((h))
+#define DUK_HOBJECT_IS_CONSTRUCTABLE(h)        DUK_HOBJECT_HAS_CONSTRUCTABLE((h))
 
 /* Object has any exotic behavior(s). */
 #define DUK_HOBJECT_EXOTIC_BEHAVIOR_FLAGS      (DUK_HOBJECT_FLAG_EXOTIC_ARRAY | \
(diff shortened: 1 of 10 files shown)


fatcerberus:
Is `duk_require_constructor_call()` in any public Duktape releases yet?

svaarala:
No, it's an internal function, but it would make sense to expose it regardless of the changes here because it complements duk_is_constructor_call().

src-input/duk_hobject.h
```diff
@@ -36,6 +36,11 @@
  * inspection code.
  */
 
+/* FIXME: ES spec: all construct-only functions have [[Call]] and [[Construct]]
```
fatcerberus: I'm not sure I understand this.  I was under the impression that [[Call]] and [[Construct]] are specification devices, so what does it mean for compliance, in practical terms, to say a function "doesn't have [[Call]]"?

svaarala: This was just a cryptic note to myself so maybe not very readable ;)

In ES all functions have [[Call]], and constructable functions also have [[Construct]]. Functions that are construct-only in the specification reject non-constructor calls specifically by checking for it. All of this is the conceptual specification view.

Right now in this pull the CALLABLE flag behaves differently. It's not set for construct-only functions so that attempts to call them as normal functions will get automatically rejected without any explicit check in the function. So, CALLABLE <=> [[Call]]. This affects a bunch of checks inside; when the code is supposed to check for existence of [[Call]], the internals needs to check for CALLABLE | CONSTRUCTABLE to get the same effect.

This may be confusing, which is why the FIXME.

fatcerberus: I understood, I was just trying to figure what it meant in practice--since if there's no observable difference in behavior then it doesn't matter for compliance.  But I guess in this case the difference would be that the thrown error would not have the function in its stack trace, while it would in a compliant engine.

svaarala: It would just be an internal difference - the main concern being that `DUK_HOBJECT_IS_CALLABLE(x)` would be different than "has [[Call]]". At present they are the same in master. This might be error prone when maintaining code unless the naming was changed to make the distinction clearer.

The stack trace behavior would be different from current behavior, but stack traces are not compliancy related.

fatcerberus: Alright, I see.  I was just trying to figure out what needed to be "fixed" (since it's a FIXME).  As long as you know the answer to that, I'm satisfied. :)

svaarala: Hehe, yes. I don't keep work in progress branches in a well commented state btw. That's why they're work in progress :)

PR #1605, opened 2017-07-10
module-node fixes

Legimet:
The duk_throw/duk_error calls result in compiler warnings with Duktape 2.x unless we cast them to void.
Also, duk_push_bare_object isn't available in Duktape 1.x so duk_push_object/duk_set_prototype have to be used.


svaarala:
Looks good to me - @fatcerberus any comments?

fatcerberus:
When I wrote the code I didn't worry about supporting Duktape 1.x because it's a bundled extra with 2.0+.  But of course it's nice if one can use it with 1.x as well.

No complaints from me, merge away. :)

svaarala:
Alrighty :)

svaarala:
Thanks @Legimet!

PR #1611, opened 2017-07-12
Faster bufwriter growth for non-lowmem targets

svaarala:
Use 25% spare instead of 6.25% spare on default (desktop) build.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2992,7 +2992,8 @@ Planned
   value stack pops where safe (GH-1583, GH-1584); initial object property
   table size estimates for NEWOBJ and NEWARR opcodes (object and array
   literals) (GH-1596, GH-1597); duk_concat_2() internal helper for str+str
-  arithmetic (GH-1599)
+  arithmetic (GH-1599); larger spare for bufwriter in non-lowmem build
+  (GH-1611)
 
 3.0.0 (XXXX-XX-XX)
 ------------------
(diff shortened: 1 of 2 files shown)


fatcerberus:
Just my 2 cents: I find the word “spare” confusing in this context (I believe I brought this up once before).  Maybe “slack” would be a better term?

svaarala:
Both are equally clear to me. Looking at the source changing "spare" to "slack" would be doable, there's one config option with "SPARE" in it (DUK_USE_VALSTACK_SHRINK_SPARE_SHIFT) but it's not yet released.

Not sure if this would be worth the effort, but if "spare" is much more confusing then maybe.

fatcerberus:
It’s maybe a bit of pedantry on my part, but it’s basically the same as the difference between “less” and “fewer”.  In practice, when I see the words “spare” and “buffer” in the same sentence, my first thought is some sort of double/triple buffering strategy (i.e. you have a *spare buffer*) rather than a single dynamic buffer with extra *slack* space left over.

Your choice, ultimately; I just wanted to explain the cause of my confusion. :)

svaarala:
The intended meaning is not actually "spare buffer" but "buffer spare" or "buffer spare area". I think I picked up the term from some memory allocator work a long while back. But I see what you mean. It might not sound wrong to me because I'm not a native speaker by far, so I'll take your word for it :)

PR #1614, opened 2017-07-15
Use duk_hthread only in internals, removing casts between ctx and thr

svaarala:
Both duk_hthread and duk_context typedefs resolve to struct duk_hthread internally.  In external API duk_context resolves to struct duk_hthread which is intentionally left undefined as the struct itself is not dereferenced. Change internal code to use `duk_hthread` exclusively which removes unnecessary and awkward thr <-> ctx casts from internals.

The basic guidelines are:
- Public API uses duk_context in prototype declarations.  The intent is to hide the internal type, and there's already a wide dependency on the type name.
- All internal code, both declarations and definitions, use duk_hthread exclusively.  This is done even for API functions, i.e. an API function declared as "void duk_foo(duk_context *ctx);" is then defined as "void duk_foo(duk_hthread *thr);".


fatcerberus:
This change will be good for new maintainers, if nothing else.  The casts confused me for a long time and I kept second-guessing myself thinking "...wait, aren't these the same thing?" :)

svaarala:
They weren't originally: duk_context was typedef'd `void *` so that there would be no way of attempting to dereference or otherwise use it. So casting was necessary at first.

fatcerberus:
Ah, right.  I'm surprised I forgot about that - it caused me some headaches because C implicitly casts `void*` to any other pointer type and I ended up passing `duk_context`s into functions they had no right being passed into, causing some very weird (mis)behavior :)

PR #1621, opened 2017-07-20
Add protected call suggestions to abort call site

svaarala:
The code comment may help users that encounter a fatal error handled by the default fatal error handler (which calls abort()), and inspect source in a debugger.


fatcerberus:
:+1:

PR #1630, opened 2017-07-24
Improve obj.noSuch() error message using GETPROPC

svaarala:
Work in progress for better performing alternative of #438.

- [x] Add GETPROPC opcode with callability check
- [x] Add compiler GETPROPC emission; either patch afterwards (carefully with shuffling) or work into ivalue conversion
- [x] Add internal duk_is_callable_tval()
- [x] Optimize duk_is_function() and duk_is_constructable() a bit
- [x] Reorg opcodes
- [x] Check diff of #438 of things to include
- [x] Possible side effect order approach: https://github.com/svaarala/duktape/pull/438#issuecomment-317588233
- [x] Update debugger opcode metadata
- [x] Update debug print opcode metadata
- [x] Testcase: basic case
- [x] Testcase: shuffling case
- [x] Testcase: computed base and key
- [x] Testcase: exact side effect sequence: property lookup, argument evaluation, callability check
- [x] Testcase fixes
- [x] Manual perf test
- [x] Manual test for side effect torture (valstack realloc during GETPROPC + error)
- [x] Ensure side effect torture automatically cover valstack realloc during GETPROPC
- [x] Releases entry


svaarala:
The side effect approach described in https://github.com/svaarala/duktape/pull/438#issuecomment-317588233 would seem workable, though a bit clunky. The clunkiness only affects the error throwing path of both GETPROPC and call handling, so only the footprint impact of the handling really matters.

Example:

```js
var base = {};
Object.defineProperty(base, 'func', {
    get: function myGetter() { print('getter'); return 'notCallable'; }
});

try {
    base.func(print('foo'), print('bar'), print('quux'));
} catch (e) {
    print(e);
}

try {
    new RangeError('aiee')();
} catch (e) {
    print(e);
}
```

Node.js prints:

```
getter
foo
bar
quux
[TypeError: base.func is not a function]
[TypeError: (intermediate value) is not a function]
```

This branch prints:

```
getter
foo
bar
quux
TypeError: 'notCallable' not callable (property 'func' of [object Object])
TypeError: [object Error] not callable
```

fatcerberus:
For completeness, Firefox:
```js
getter 
foo 
bar 
quux 
TypeError: base.func is not a function
TypeError: (new RangeError(...)) is not a function
```

svaarala:
My Node.js test was also on a relatively old Node.js version (v4.2.6) so it might have changed.

svaarala:
Chrome console:

```
getter 
foo 
bar 
quux 
TypeError: base.func is not a function
Stack trace:
@debugger eval code:7:5
 
TypeError: (new RangeError("aiee", "debugger eval code", 13)) is not a function
Stack trace:
@debugger eval code:13:5
```

fatcerberus:
Nope, same result on 8.2.1:
```js
> process.version
'v8.2.1'
> new RangeError('aiee')()
TypeError: (intermediate value) is not a function
    at repl:1:23
    at ContextifyScript.Script.runInThisContext (vm.js:44:33)
    at REPLServer.defaultEval (repl.js:239:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:433:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
    at REPLServer.Interface._onLine (readline.js:278:10)
    at REPLServer.Interface._line (readline.js:625:8)
```

SM just seems to maintain a lot more expression-level information at runtime than other engines.  Which is nice, but maybe a bit overkill :)

fatcerberus:
Huh, maybe Node.js is using an older V8 version than Chrome.

svaarala:
The console has special handling for error objects, so maybe that makes some difference compared to Node.js console.log(). Or maybe it's some version difference. Interesting how much variance there is :)

MuJS:
```
getter
foo
bar
quux
TypeError: called object is not a function
	/tmp/test.js:7: in function ''
	[C]: in function '?'
TypeError: called object is not a function
	/tmp/test.js:13: in function ''
	[C]: in function '?'
```

Rhino (seems to handle side effects incorrectly or at least differently):
```
getter
TypeError: Cannot call property func in object [object Object]. It is not a function, it is "string".
TypeError: RangeError: aiee is not a function, it is object.
```

But interestingly Rhino's description of the error is closer to this branch: it identifies the property name and the base value (but not its identifier name).

svaarala:
I left the error message formatting untouched for now, so that the majority of the work could go into master. I'm still open to changing it if there's a good replacement.

PR #1644, opened 2017-07-30
Add better error message with key+target summary for duk_(p)call_prop()

svaarala:
Extend #1630 to cover duk_(p)call_prop() too.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2909,7 +2909,7 @@ Planned
 
 * Improve error message for property-based call TypeError (e.g. foo.noSuch()),
   error message now includes the key for easier debugging (GH-1627, GH-1630,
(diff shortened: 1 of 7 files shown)


fatcerberus:
This would be a good reason for me to use `duk_pcall_prop()` instead of `duk_pcall_method()` as is my current convention. :)

svaarala:
One could also improve the error message for e.g. calling a global function by introducing a call like:

```c
duk_push_string(ctx, "aiee");
duk_call_global(ctx, "print");
```

which would be conceptually equivalent to looking up the global and then calling it. The difference is that Duktape knows what's happening and could incorporate the name into a potential error message just like with duk_(p)call_prop().

PR #1659, opened 2017-08-03
Add monotonic clock config option, use it for debugger transport peeks

svaarala:
- [x] Add DUK_USE_GET_MONOTONIC_TIME(), semantics similar to POSIX clock_gettime() CLOCK_MONOTONIC
- [x] Add internal helpers for getting Ecmascript/monotonic time, fewer macro calls and easier to debug
- [x] Change debugger transport peek rate limiting to use monotonic time
- [x] Manual test: make Date.now() run very slowly, provide monotonic time, and check that debugger peeks work at the expected ~200ms rate
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2983,6 +2983,16 @@ Planned
 * Make error message summary strings longer (32 -> 96 character) to better
   capture error messages for e.g. uncaught errors (GH-1653)
 
+* Add DUK_USE_GET_MONOTONIC_TIME() to allow an application to provide a
+  monotonic time source (similar to clock_gettime() CLOCK_MONOTONIC) which
+  Duktape will then use for performance.now() and internal rate limiting
+  mechanisms; if absent (default), monotonic time defaults to
+  DUK_USE_DATE_GET_NOW() (GH-1659)
+
+* Use monotonic time (if available) for debugger transport peeking, so that
+  the peek callback is called with the same realtime rate even if the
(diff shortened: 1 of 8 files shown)


svaarala:
Tested with:

```
diff --git a/src-input/duk_bi_date_unix.c b/src-input/duk_bi_date_unix.c
index 5613c9e5..6a045f7c 100644
--- a/src-input/duk_bi_date_unix.c
+++ b/src-input/duk_bi_date_unix.c
@@ -26,6 +26,7 @@ DUK_INTERNAL duk_double_t duk_bi_date_get_now_gettimeofday(duk_hthread *thr) {
 
        d = ((duk_double_t) tv.tv_sec) * 1000.0 +
            ((duk_double_t) (tv.tv_usec / 1000));
+       d = DUK_FLOOR(d / 100.0);
        DUK_ASSERT(DUK_FLOOR(d) == d);  /* no fractions */
 
        return d;
```

Debugger code makes transport peeks roughly every 10-20 seconds (1/100 from normal rate) as expected. Then added this hack to duk_config.h:

```c
static inline double get_mono_time(duk_context *ctx) {
        struct timespec ts;

        clock_gettime(CLOCK_MONOTONIC, &ts);
        return (double) ts.tv_sec * 1000.0 + (double) ts.tv_nsec / 1000.0;
}

#define DUK_USE_GET_MONOTONIC_TIME(ctx) get_mono_time(ctx)
```

and debugger transport peeks again work normally (100-200ms interval).

Issue #1658, opened 2017-08-03
Add monotonic time support


svaarala:
- [ ] Add DUK_USE_GET_MONOTONIC_TIME()
- [ ] Use monotonic time (if available) for debugger transport peeks
- [ ] Add Linux, OSX, Windows provider for monotonic time
- [ ] Add a Performance.now() binding
- [ ] Add a public API binding for getting monotonic time

svaarala:
There are multicore RDTSC issues with Windows XP QueryPerformanceCounter(). SImilar issues exist in at least early Linux versions (2.x): http://juliusdavies.ca/posix_clocks/clock_realtime_linux_faq.html.

PR #1660, opened 2017-08-03
Add performance.now() API and millisecond fraction support for duk_get_now()

### LINKED ISSUE #773: Add sub-millisecond resolution to time provider
While the `Date` built-in cannot expose sub-millisecond resolution (truncating to even milliseconds is mandated by the specification) the time provider setup for `duk_config.h` should support sub-millisecond resolution. That's more or less already the case, but ensure documentation states this explicitly.

The sub-millisecond resolution will be available through `duk_get_now()` and a possible Ecmascript high resolution time API.

svaarala:
- [x] Rebase
- [x] Allow fractions in DUK_USE_DATE_GET_NOW() and expose them in duk_get_now() (but not for Date instances)
- [x] Update DUK_USE_DATE_GET_NOW and DUK_USE_GET_MONOTONIC_TIME documentation
- [x] Add fractions to POSIX/Windows DUK_USE_DATE_GET_NOW()
- [x] Add performance.now()
- [x] performance.timeOrigin intentionally missing for now, semantics to be decided and missing property matches earlier performance.now() spec
- [x] Property attributes
- [x] Maybe remove .timeOrigin altogether instead of returning 0?
- [x] Website guide, custom built-ins
- [x] Website guide, global binding removal note (unrelated to other changes)
- [x] Website guide, clarify that other 'performance' bindings (like performance.timing) are not supported
- [x] Wiki post-ES5 page, custom built-in - https://github.com/svaarala/duktape-wiki/pull/184
- [x] Testcase: performance.now()
- [x] Testcase: performance.timeOrigin absent
- [x] Releases entry
- [x] 2.2 migration note: fractions in duk_get_now()

Fixes #773.


svaarala:
@fatcerberus I'd be interested in hearing if this pull works for you: a while back you were interested in performance.now() for Minisphere. With this pull:
- Windows DUK_USE_DATE_GET_NOW() should return millisecond fractions.
- There's no DUK_USE_GET_MONOTONIC_TIME() for Windows (or POSIX) yet, but monotonic time falls back to DUK_USE_DATE_GET_NOW() and is provided through performance.now() with fractions.

The very minor change to the Windows time provider should not affect `new Date()` and `Date.now()` but I haven't tested them manually yet.

svaarala:
One difficult choice is whether performance.now() should be available at all if DUK_USE_GET_MONOTONIC_TIME() is not provided.

Technically the answer is clear: it shouldn't, because it won't then respect the requirements in High Resolution Time specification.

But in practice, it's probably more surprising not to have the binding, especially because DUK_USE_GET_MONOTONIC_TIME() is not provided by default (at least for now). Further, there's no other Ecmascript binding where one can get fractional milliseconds.

If all major platforms had a monotonic time provider, then it might make more sense for the binding to be absent without a monotonic time provider.

So in this pull the binding is always provided, but this can be changed later if necessary, also added to #1658 as a future work note.

fatcerberus:
Windows has QueryPerformanceCounter that I think could be used for this (sub-microsecond), don't know about other platforms.

svaarala:
For POSIX there's `clock_gettime(CLOCK_MONOTONIC)`, but I'm not adding them yet in this pull because that can be done separately and there might be some hunting to figure out when the monotonic time is known to be available and when not. I don't want to break the "works by default" nature just to provide monotonic time which is not always mandatory.

Right now, in this pull, performance.now() and duk_get_now() should return sub-millisecond resolution time if that's available:
- For POSIX gettimeofday(), which is used by default on Unix environments, that is the case. There's (technically) microsecond time available (also in practice sub-milliseconds appear out of the API). This pull now allows those fractions: https://github.com/svaarala/duktape/blob/c9f892570790bd2355b038d9609922b3432ff67f/src-input/duk_bi_date_unix.c#L27-L30.
- For Windows SystemTimeToFileTime(), which is used by default, the technical resolution is 100ns and whatever resolution beyond milliseconds is available is now provided: https://github.com/svaarala/duktape/blob/c9f892570790bd2355b038d9609922b3432ff67f/src-input/duk_bi_date_windows.c#L54-L57.

So performance.now() should in practice be useful even with this pull, though it is backed by Ecmascript time, just with additional resolution.

fatcerberus:
I don't know what the specification says, but MDN has this interesting bit re: DOMHighResTimeStamp, emphasis mine:
> The DOMHighResTimeStamp type is a double and is used to store a time value. The value could be a discrete point in time or the difference in time between two discrete points in time. The unit is milliseconds and should be accurate to 5 µs (microseconds). However, if the browser is unable to provide a time value accurate to 5 microseconds (due, for example, to hardware or software constraints), the browser can represent the value *as a time in milliseconds accurate to a millisecond*.

So I did the following test in miniSphere:
```js
var a, b, c, d = 0;
a = performance.now();
while ((b = performance.now()) == a);
while ((c = performance.now()) == b) ++d;
Sphere.abort(`${d} iterations, ${c - b}`);
```

Output is:
```
5054 iterations, 1ms
```

Number of iterations is always around ~5050 for debugger disabled, ~4450 for debugger enabled and attached.  So it looks like, on Windows at least, the granularity is always 1ms, even if the underlying resolution is higher than that.

svaarala:
I'm not sure what you mean by "underlying resolution" -- do you mean the Windows APIs used to implement DUK_USE_DATE_GET_NOW()?

In this pull DUK_USE_DATE_GET_NOW() no longer truncates the time values to whole milliseconds. Instead, all the resolution available from the platform is returned. Date built-in truncates to milliseconds because Ecmascript requires that, but duk_get_now() and performance.now() don't so all the native resolution should now be available.

svaarala:
Re: that MDN text, there's similar text in the specification:
> A DOMHighResTimeStamp SHOULD represent a time in milliseconds accurate to 5 microseconds - see 8. Privacy and Security. If the User Agent is unable to provide a time value accurate to 5 microseconds due to hardware or software constraints, the User Agent can represent a DOMHighResTimeStamp as a time in milliseconds accurate to a millisecond.

This is related to some security concerns discussed in https://www.w3.org/TR/hr-time-2/#privacy-security.

I'm not sure I understand the requirement. Does it mean that:
- At least a 5 microsecond resolution is recommended ("should"). But if you can't get that, you can provide less resolution, up to the Date resolution of full milliseconds.
- Or... if you can't provide at least a 5 microsecond resolution, you should truncate to full milliseconds. For example, if you had 10 microsecond resolution, you shouldn't provide that resolution but should truncate instead to milliseconds.

I really hope it is the former :-) At least for embedding use cases the most obvious default behavior is to provide all the resolution that is available.

And often one doesn't even know what resolution is actually available, as the nominal resolutions in various time APIs don't mean that you actually get that resolution. The timer resolution would need to be provided to Duktape via a config option. But if one does that, it'd be cleaner IMO to just provide a config option "truncate performance.now() to milliseconds" which you'd set if you wanted that behavior.

svaarala:
Re: that Windows resolution, one simple test would be:
- If you build from this branch, and call `performance.now()` a few times, do you ever get fractional output values?

fatcerberus:
By "underlying resolution" I meant that even if the Windows timestamp is a higher resolution (100ns like you said), I'm not getting that--I only get 1ms granularity.  You can see in my test above that I iterate through a while loop until the value of `performance.now()` changes, and when it finally does (after ~5K iterations) the difference is exactly 1ms.

svaarala:
Ok, so that means that even with the change in this pull to provide all the resolution available from SystemTimeToFileTime(), only 1ms resolution is available from that specific Windows API.

Or, I'm using the API wrong which is quite possible too.

svaarala:
Just looking at the Duktape Windows Date provider, it's probably limited by doing a SYSTEMTIME to FILETIME conversion. SYSTEMTIME has only 1ms resolution: https://msdn.microsoft.com/en-us/library/windows/desktop/ms724950(v=vs.85).aspx so it doesn't matter than FILETIME has 100ns resolution.

There's probably a better way to implement the Windows Date provider to get better resolution. I don't really have a good grasp of the best way to use Windows time APIs so maybe someone can suggest an improvement.

But, for performance.now() the best step forward is to implement DUK_USE_GET_MONOTONIC_TIME() for Windows. Getting sub-millisecond resolution for duk_get_now() is not that important.

fatcerberus:
The best approach is probably to use QueryPerformanceCounter, which is specifically stated to be sub-μs resolution:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms644904(v=vs.85).aspx

It's designed for profiling, which is what `performance.now()` is for too. :)

svaarala:
QueryPerformanceCounter() is the obvious choice for monotonic time, but I'd also like to have sub-millisecond resolution for the Date built-in because it's available through duk_get_now() and I hope to find a way to allow Date instances represent sub-millisecond resolution somehow (it's useful in many cases like accurate logging).

I'll open a pull to introduce QueryPerformanceCounter() on Windows. Apparently it needs some corrections based on GetTickCount64(), though I'm not sure how relevant this issue is today: https://support.microsoft.com/en-us/help/274323/performance-counter-value-may-unexpectedly-leap-forward.

svaarala:
Hmm, this might allow more accurate Date timestamp, GetSystemTimePreciseAsFileTime():
- https://msdn.microsoft.com/en-us/library/windows/desktop/hh706895(v=vs.85).aspx

PR #1662, opened 2017-08-04
Add initial provider for Windows monotonic time

svaarala:
- [x] Rebase
- [x] Add DUK_USE_GET_MONOTONIC_TIME for Windows unless provided explicitly
- [x] Use QueryPerformanceCounter() and QueryPerformanceFrequency() as a base implementation
- [x] Windows version check: only enable when certain to exist and work
- [x] What to do if performance counters are not available; this is only apparent at runtime, so maybe then default to Ecmascript time with fractions (at runtime) -> since WinXP QPC() cannot fail
- [x] Dealing with performance counter jumps, see https://support.microsoft.com/en-us/help/274323/performance-counter-value-may-unexpectedly-leap-forward -> these shouldn't happen since Vista
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2989,6 +2989,9 @@ Planned
   mechanisms; if absent (default), monotonic time defaults to
   DUK_USE_DATE_GET_NOW() (GH-1659)
 
+* Add monotonic time provider for Windows based on QueryPerformanceCounter(),
+  enabled by default if _WIN32_WINNT indicates Vista or above (GH-1662)
+
 * Use monotonic time (if available) for debugger transport peeking, so that
   the peek callback is called with the same realtime rate even if the
   Ecmascript time source jumps or doesn't advance in realtime (GH-1659)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -131,6 +131,7 @@ build_script:
 test_script:
   - cmd: echo --- VS2015
   - cmd: duk-vs2015-%PLATFORM%.exe -e "print(Duktape.env); print('Hello world!');"
+  - cmd: duk-vs2015-%PLATFORM%.exe -e "print(Date.now(), performance.now(), performance.now());"
   - cmd: duk-dll-vs2015-%PLATFORM%.exe -e "print(Duktape.env); print('Hello world!');"
   - cmd: duk-cxx-vs2015-%PLATFORM%.exe -e "print(Duktape.env); print('Hello world!');"
 
--- a/config/config-options/DUK_USE_GET_MONOTONIC_TIME_WINDOWS_QPC.yaml
+++ b/config/config-options/DUK_USE_GET_MONOTONIC_TIME_WINDOWS_QPC.yaml
@@ -1,0 +1,7 @@
+define: DUK_USE_GET_MONOTONIC_TIME_WINDOWS_QPC
+introduced: 2.2.0
+default: false
+tags:
+  - portability
+description: >
+  Use QueryPerformanceCounter() for monotonic time on Windows.
--- a/config/header-snippets/date_provider.h.in
+++ b/config/header-snippets/date_provider.h.in
@@ -51,4 +51,12 @@
 /* No provider for DUK_USE_DATE_FORMAT_STRING(), fall back to ISO 8601 only. */
 #endif
 
+#if defined(DUK_USE_GET_MONOTONIC_TIME)
+/* External provider already defined. */
+#elif defined(DUK_USE_GET_MONOTONIC_TIME_WINDOWS_QPC)
+#define DUK_USE_GET_MONOTONIC_TIME(ctx)  duk_bi_date_get_monotonic_time_windows_qpc()
+#else
+/* No provider for DUK_USE_GET_MONOTONIC_TIME(), fall back to DUK_USE_DATE_GET_NOW(). */
+#endif
+
 #endif  /* DUK_COMPILING_DUKTAPE */
--- a/config/platforms/platform_windows.h.in
+++ b/config/platforms/platform_windows.h.in
@@ -21,6 +21,16 @@
 #include <windows.h>
 #endif
 
+/* QueryPerformanceCounter() may go backwards in Windows XP, so enable for
+ * Vista and later: https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx#qpc_support_in_windows_versions
+ * Check against _WIN32_WINNT which defines the lowest targeted Windows
+ * version (we obviously don't know the runtime version at compile time).
+ */
+#if !defined(DUK_USE_GET_MONOTONIC_TIME_WINDOWS_QPC) && \
+    defined(_WIN32_WINNT) && (_WIN32_WINNT >= 0x0600)
+#define DUK_USE_GET_MONOTONIC_TIME_WINDOWS_QPC
+#endif
+
 #define DUK_USE_OS_STRING "windows"
 
 /* On Windows, assume we're little endian.  Even Itanium which has a
--- a/src-input/duk_bi_date_windows.c
+++ b/src-input/duk_bi_date_windows.c
@@ -130,3 +130,28 @@ DUK_INTERNAL duk_int_t duk_bi_date_get_local_tzoffset_windows_no_dst(duk_double_
 	return (duk_int_t) (((LONGLONG) tmp2.QuadPart - (LONGLONG) tmp1.QuadPart) / DUK_I64_CONSTANT(10000000));  /* seconds */
 }
 #endif  /* DUK_USE_DATE_TZO_WINDOWS_NO_DST */
+
+#if defined(DUK_USE_GET_MONOTONIC_TIME_WINDOWS_QPC)
+DUK_INTERNAL duk_double_t duk_bi_date_get_monotonic_time_windows_qpc(void) {
+	LARGE_INTEGER count, freq;
+
+	/* There are legacy issues with QueryPerformanceCounter():
+	 * - Potential jumps: https://support.microsoft.com/en-us/help/274323/performance-counter-value-may-unexpectedly-leap-forward
+	 * - Differences between cores (XP): https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx#qpc_support_in_windows_versions
+	 *
+	 * We avoid these by enabling QPC by default only for Vista or later.
+	 */
+
+	if (QueryPerformanceCounter(&count) && QueryPerformanceFrequency(&freq)) {
+		/* XXX: QueryPerformanceFrequency() can be cached */
+		return (duk_double_t) count.QuadPart / (duk_double_t) freq.QuadPart * 1000.0;
+	} else {
+		/* MSDN: "On systems that run Windows XP or later, the function
+		 * will always succeed and will thus never return zero."
+		 * Provide minimal error path just in case user enables this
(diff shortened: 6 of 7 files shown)


svaarala:
@fatcerberus I've written this blind so no idea if this works - but would be interested in knowing if it does :-)

If all went well, performance.now() should provide a sub-millisecond resolution.

fatcerberus:
Sure, I can test out this branch tonight.

svaarala:
There's a long queue for AppVeyor, but added a temporary hack to print out performance.now() from within the AppVeyor build. This should provide a basic idea if the change works.

svaarala:
Seems to work at least to some extent:
```
duk-vs2015-%PLATFORM%.exe -e "print(Duktape.env); print('Hello world!'); print(performance.now());"
ll u n p1 a1 x64 windows msvc
Hello world!
346716.5415
```

fatcerberus:
Using the test code from https://github.com/svaarala/duktape/pull/1660#issuecomment-320262612:
```js
var a, b, c, d = 1;
a = performance.now();
while ((b = performance.now()) == a);
while ((c = performance.now()) == b) ++d;
Sphere.abort(`${d} iterations, ${c - b}`);
```

Output:
```
@/scripts/main.js:0
manual abort

1 iterations, 0.0005109780468046665ms
```

It's granular enough to show a difference after a single iteration, nice. 😸

fatcerberus:
> What to do if performance counters are not available; this is only apparent at runtime, so maybe then default to Ecmascript time with fractions (at runtime)

This only seems to be a possible issue for Windows 2000 (the earliest version that supports QPC); according to MSDN, WinXP and later guarantee it will work.  I'm not sure if it's a concrete issue in win2k in practice, I don't have any meaningful Win32 development experience before XP.

There's also the question of whether anyone is actually still running Win2K in production, that's a pretty ancient platform. :)

svaarala:
Well, some people are compiling Duktape for DOS. So old platforms are indeed used in certain situations :)

For the presence check, I could maybe just use ">= XP".

In addition to the API being present, I was wondering what happens if the processor doesn't support RDTSC at all. But I guess there's no realistic case where XP is used with such an old processor that it doesn't have performance counters, so maybe that Windows version check is enough.

The API call does have a true/false return value though, so the code needs to do something if it fails. Right now I'm just returning 0.0 and maybe that's good enough if there's no realistic case where it would fail.

svaarala:
Some info:
- https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx#qpc_support_in_windows_versions
- https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx#AppendixA

For measuring long intervals it may actually be more accurate to use GetSystemTimePreciseAsFileTime() based measurement (at least when ignoring time adjustments) because it is protected against variance in frequency (discussed in the above link). Trying to correct QueryPerformanceCounter() against frequency differences would be possible by monitoring against the system time (detecting and ignoring time jumps, etc) and then trying to correct for the frequency differences. But definitely not in this pull...

svaarala:
On Windows XP there's no guarantee that measurements across cores are consistent:

> However, some hardware systems' BIOS didn't indicate the hardware CPU characteristics correctly (a non-invariant TSC), and some multi-core or multi-processor systems used processors with TSCs that couldn't be synchronized across cores. Systems with flawed firmware that run these versions of Windows might not provide the same QPC reading on different cores if they used the TSC as the basis for QPC.

Later versions use HPET or ACPI as a time source, and may use TSC if it's known to be good. Even later versions have better TSC synchronization mechanisms and TSC is used in more situations (but apparently there's still a fallback in some cases).

So maybe that best check is actually ">= Vista"; otherwise the monotonic time can potentially go backwards.

svaarala:
@fatcerberus Just to clarify why I'm mulling about the availability of the API and when it works: my main concern is that the default configuration should almost always be a good starting point. With that in mind I prefer performance.now() backing to a not-so-accurate millisecond resolution source if there's a chance the high resolution source either doesn't work or is unreliable. It's easy to make a mechanism opt-in if there's a chance it won't work by default.

All that said, it'd be great if performance.now() was available with the best possible source for anyone compiling Duktape for a relatively recent Windows version. It's just one thing less to worry about when getting started, and it's important that works as smoothly as possible.

fatcerberus:
QPC returning BOOL seems to be legacy baggage - the MSDN page specifically says it always returns TRUE on XP and later.

The counter potentially going backwards is a valid concern, though. :)

svaarala:
So far Duktape headers haven't tried to detect the minimum Windows version the build is targeted for, only MSVC version has mattered.

The Windows version is of course not known during compilation and the same binary will work on many Windows versions. The closest thing is apparently checking for the minimum Windows version the build is targeted for, using `WINVER` or `_WIN32_WINNT` which can be defined by the user before including `windows.h`. To use these apparently:
- Config header should include `#include <sdkddkver.h>` to get default values for `WINVER` etc in case the user doesn't define them. (But this header is only available for MSVC, VS2008 or later; not e.g. for Cygwin.)
- Then compare `WINVER` or `_WIN32_WINNT` against the constant indicating Vista.

This only works if recent MSVC versions by default provide `WINVER` or `_WIN32_WINNT` which targets Vista or later. Any idea if this is the case?

fatcerberus:
The default value (if any) of `_WIN32_WINNT` seems to be determined by the version of the Windows SDK used to build, which is configurable in MSVC.  In any case if one is trying to detect the Windows version to determine how reliable QPC is (XP's being potentially unreliable, e.g.), it'd probably be better to do that at runtime.  It's common practice to build against WinSDK 7.0 (XP toolchain) because otherwise your executable tends to crash on startup pre-Win7.

Not directly relevant here but worth keeping in mind: Windows 8.1 and later pretend to be v6.2 (Win8) unless an appropriate manifest is linked in.

svaarala:
There's no easy way to check if a QueryPerformanceCounter() symbol exists at runtime though if one links against it? The only way would be a runtime detection, then inspecting for the symbol from some DLL, etc.

So either Duktape links against QPC or it doesn't. If one does, then the symbol must exist for Duktape to start, but one can still do a runtime check to figure out if it's preferable to use QPC or not (e.g. don't if you're running on Windows XP).

svaarala:
So to clarify, assuming WINVER etc are set for Windows XP by default, then:
- Duktape should link against QPC if WINVER / _WIN32_WINNT indicates XP or above.
- Duktape should do a runtime check against running Windows version to decide whether to actually use QPC or not; this check would be against Vista.

Kind of complicated, but I don't know how this could be simplified, assuming XP is the default WINVER *and* QPC in XP is not always reliable.

fatcerberus:
Just randomly typing in `_WIN32_WINNT` into an MSVC 2017 window, the value is reported as `0x0601`, i.e. XP.  I haven't explicitly defined it elsewhere.

fatcerberus:
Hm, actually 0x0601 seems to be Win7.

svaarala:
That would be nice - I think it's OK as an initial baseline if the latest MSVC versions would provide a WINVER / _WIN32_WINNT high enough for this feature to get enabled.

fatcerberus:
I believe you will actually get a compile error if you use a symbol that doesn't exist in the targeted version; i.e. the prototype isn't provided if `_WIN32_WINNT` is too low.  So it'd be easy to detect any mismatch, I think.

svaarala:
Right, but I don't want the build to fail with a mismatch because it won't be immediately obvious for users what to fix :)

svaarala:
So basically I'm looking for a very reliable way to ensure this feature is enabled by default when it is available and preferable, and disabled when it isn't available or isn't preferable. I would like to avoid build failures, and the need to tweak WINVER / _WIN32_WINNT or duk_config.h to fix a build error.

fatcerberus:
Incidentally, this kind of thing is largely why I don't directly code against the Win32 API anymore, I much prefer .NET nowadays. :)

fatcerberus:
I would say this should work then:

```c
#ifdef _WIN32_WINNT && _WIN32_WINNT >= 0x0501
```

fatcerberus:
`WINVER` seems to just be defined as equal to `_WIN32_WINNT`, FWIW.

svaarala:
Hmm, but some sources seem to indicate that _WIN32_WINNT may not be defined at all, and `sdkddkver.h` needs to be included to get the defaults. But then, that header is not always available. See e.g. https://swarm.workshop.perforce.com/files/guest/perforce_software/p4/net/netportipv6.h?v=1.

fatcerberus:
It was my understanding that the macro was meant to be defined by the user before including Windows.h--and if not, then you target Win2K by default.

svaarala:
Hmm, so does that mean that sdkddkver.h will provide a default value that is guaranteed to be pre-XP (e.g. always Windows 2000)?

If so, then the check you propose seems good without trying to include sdkddkver.h at all (since it's not always available). The user can forcibly enable QPC for monotonic time using configure.py -DDUK_USE_GET_MONOTONIC_TIME_WINDOWS_QPC.

svaarala:
0x0501 would mean checking against Windows XP though, so a runtime check against Vista would also be needed.

I guess I'll try to figure out what the _WIN32_WINNT default value is in recent MSVC versions. If it's Vista or above, then the compile time check can be for Vista, and no runtime check is needed. This would be the cleanest approach, but only works if the default in MSVC is high enough.

fatcerberus:
Yeah, they macros are user-defined, MSVC is just being nice by targeting Windows 7 for you it seems.

https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt

svaarala:
Ok, so if it's Windows 7 by default (from MSVC) then this would be best:

```c
#if defined(_WIN32_WINNT) && _WIN32_WINNT >= 0x0600 /* Vista or above */
```

fatcerberus:
For an operating system that was written in C, Windows sure is a pain to code for in that language. :)

svaarala:
I was wondering whether WINVER or _WIN32_WINNT is preferable, and found this: https://blogs.msdn.microsoft.com/oldnewthing/20070411-00/?p=27283. Apparently since Vista the preferred method is defining NTDDI_VERSION and then including sdkddkver.h which sets the other symbols :-P

I think I'll just check against _WIN32_WINNT for now :)

fatcerberus:
Yeah, that method only works if you're building against a Vista-or-later WinSDK anyway. I can tell you from experience that the WinXP SDK that MSVC uses when setting the "Windows XP" toolset doesn't have a sdkddkver.h.

svaarala:
> For an operating system that was written in C, Windows sure is a pain to code for in that language. :)

No idea if it's more or less painful than supporting a bunch of same-but-different Unix platforms :) But at least binary backwards compatibility is pretty good in Windows compared to Unix.

svaarala:
@fatcerberus Appreciate the help in figuring out the detection approach BTW, I don't have an active Windows dev host so I'm working a little bit in the dark on this :)

fatcerberus:
No problem, it's been a long time since I did Win32 programming but I still remember enough :)

Re: sdkddkver.h: It seems that when building against a modern Windows SDK, Windows.h actually includes that file for you, so you only need to define `NTDDI_VERSION` and then include Windows.h.  Otherwise (MSVC's XP toolchain, e.g.) you define `_WIN32_WINNT` directly.  In both cases the latter define should be present by the time Duktape needs it.

svaarala:
@fatcerberus I'll merge this once the tests clear, unless you think something still needs fixing.

Issue #1663, opened 2017-08-04
Improvements to Windows date/time provider


svaarala:
- [ ] Sub-millisecond resolution using GetSystemTimePreciseAsFileTime() https://msdn.microsoft.com/en-us/library/windows/desktop/hh706895(v=vs.85).aspx
- [ ] Avoid re-getting Jan 1, 1970 for every now() call

jbmonroe:
What's the workaround for Windows 7 builds?

svaarala:
@jbmonroe Could you clarify - does compilation fail for you (if so, what error)?

fatcerberus:
This is just a guess but I'm thinking he's asking for a time provider that's this precise for Windows 7.  `GetSystemTimePreciseAsFileTime` requires Windows 8+.

svaarala:
There are various approaches before GetSystemTimePreciseAsFileTime() (for some background, see https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx) but AFAIK none of them are very simple.

It would be relatively simple to use e.g. QueryPerformanceCounter() and an arbitrary starting point to get high resolution (but low accuracy) Date timestamps. While accuracy is important, sometimes it's useful to have high resolution even when the absolute accuracy is not great; this is useful for example when doing timestamp comparisons.

But even in that approach one must deal with QPC quirks, see the above link.

svaarala:
That said, if someone can point to a good, robust approach for pre-Windows 8, and would be willing to test it, I'd be happy to write up a pull that integrated it into the default Windows Date provider.

jbmonroe:
Actually, I just need anything that will execute without popping up a dialog that says this or that thing can't be found in Runtime32.dll (or whatever it was).  Even something with a pre-compile define that says "hey, this ain't for Windows 8" since there are constraints to what I can deploy.

fatcerberus:
@jbmonroe You need to define _WIN32_WINNT appropriately sometime before windows.h gets included.  Then everything should work itself out.

jbmonroe:
Okay--that's definitely good news.

As for the part where I get a stack index error of 2.1 billion when trying to add a C function?  I think I'll have to work that out on my own. :)

rlaveaux:
> On 28 Apr 2018, at 07:52, Sami Vaarala <notifications@github.com> wrote:
> 
[...]

PR #1664, opened 2017-08-04
Add sub-millisecond Windows Ecmascript time provider

svaarala:
Use GetSystemTimePreciseAsFileTime() to get better resolution. The resolution won't be available via Date instances because Ecmascript requires their timestamps to be rounded to full milliseconds. The extra resolution is available through duk_get_now() and performance.now() if no monotonic time provider is configured.

- [x] Rebase
- [x] Config option changes
- [x] Proper Windows version check
- [x] Internal provider change
- [x] Releases entry (check entry for GH-1660)

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3002,7 +3002,8 @@ Planned
   millisecond timestamps (GH-773, GH-1660)
 
 * Add sub-millisecond resolution to the default POSIX and Windows Date
-  providers (DUK_USE_DATE_GET_NOW) (GH-1660)
+  providers (DUK_USE_DATE_GET_NOW); Windows sub-millisecond resolution
+  is available if _WIN32_WINNT indicates Windows 8 or above (GH-1660, GH-1664)
 
 * Add an initial version of the High Resolution Time API, providing
   performance.now() (which backs to DUK_USE_GET_MONOTONIC_TIME() if
--- a/config/config-options/DUK_USE_DATE_NOW_WINDOWS.yaml
+++ b/config/config-options/DUK_USE_DATE_NOW_WINDOWS.yaml
@@ -5,8 +5,4 @@ tags:
   - date
   - portability
 description: >
-  Use Win32 API calls to get current datetime.  When enabled, appropriate
-  date/time headers must be included.
-
-# FIXME: add an include file metadata entry to cause default #include
-# directives to be emitted?
+  Use Win32 API calls to get current datetime.
--- a/config/config-options/DUK_USE_DATE_NOW_WINDOWS_SUBMS.yaml
+++ b/config/config-options/DUK_USE_DATE_NOW_WINDOWS_SUBMS.yaml
@@ -1,0 +1,9 @@
+define: DUK_USE_DATE_NOW_WINDOWS_SUBMS
+introduced: 2.2.0
+default: false
+tags:
+  - date
+  - portability
+description: >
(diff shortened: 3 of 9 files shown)


svaarala:
Build success for x64, fails for x86 VS2010:

```
cl /W3 /O2 /DDUK_CMDLINE_PRINTALERT_SUPPORT /Iprep-nondll /Iexamples\cmdline /Iextras\print-alert prep-nondll\duktape.c examples\cmdline\duk_cmdline.c extras\print-alert\duk_print_alert.c /Feduk-%VCNAME%-%PLATFORM%.exe
Microsoft (R) 32-bit C/C++ Optimizing Compiler Version 16.00.40219.01 for 80x86
Copyright (C) Microsoft Corporation.  All rights reserved.
duktape.c
duk_bi_date_windows.c(79) : warning C4013: 'GetSystemTimePreciseAsFileTime' undefined; assuming extern returning int
duk_cmdline.c
duk_print_alert.c
Generating Code...
Microsoft (R) Incremental Linker Version 10.00.40219.01
Copyright (C) Microsoft Corporation.  All rights reserved.
/out:duk-vs2010-x86.exe 
duktape.obj 
duk_cmdline.obj 
duk_print_alert.obj 
duktape.obj : error LNK2019: unresolved external symbol _GetSystemTimePreciseAsFileTime referenced in function _duk_bi_date_get_now_windows_subms
duk-vs2010-x86.exe : fatal error LNK1120: 1 unresolved externals
Command exited with code 2
```

svaarala:
Seems to work:
```
--- VS2015
duk-vs2015-%PLATFORM%.exe -e "print(Duktape.env); print('Hello world!'); print(performance.now());"
ll u n p1 a1 x64 windows msvc
Hello world!
1501882913442.901
duk-dll-vs2015-%PLATFORM%.exe -e "print(Duktape.env); print('Hello world!');"
```

svaarala:
@fatcerberus Re: Windows version detection, as far as I can tell this GetSystemTimePreciseFromFileTime() is more difficult than QueryPerformanceCounter, because the symbol itself is only available since Windows 8 (Windows Server 2012).

Before that the symbol doesn't exist as far as I understand MSDN. This means that any code building with WINVER / _WIN32_WINNT set to Windows XP can't reliably link against this symbol.

So... this would only get enabled by default if the user defined WINVER for Windows 8 or later.

mirh:
> So... this would only get enabled by default if the user defined WINVER for Windows 8 or later.

That's a big problem you have there though. 
https://stackoverflow.com/questions/37668692/visual-studio-setting-winver-win32-winnt-to-windows-8-on-windows-7
If the user doesn't define any macro, it defaults to whatever the newest one is (i.e. to "enable the use" of whatever function, and let the developer decide how and when leave behind the older OS versions)

> Before that the symbol doesn't exist as far as I understand MSDN. This means that any code building with WINVER / _WIN32_WINNT set to Windows XP can't reliably link against this symbol.

I think you probably misunderstood how things work, because when you target XP after VS 2010 you need the older toolsets in turn. And \*those\* are the ones setting _WIN32_WINNT.. but it's a special case.

svaarala:
@mirh What would be the right solution?

mirh:
Mhh, well a runtime check of course (as provided in #1663 for example)
If you really want to use it that is (I'm [getting told](https://bugs.python.org/issue19007#msg395767) it's relatively heavy computationally)

svaarala:
If anyone wants to make that change in a pull and test it (on both older and newer systems) I'll be happy to merge it :+1:

PR #1667, opened 2017-08-06
Add clock_gettime(CLOCK_MONOTONIC) monotonic time source

svaarala:
For now the time source is disabled by default, enabled using configure.py -DDUK_USE_GET_MONOTONIC_TIME_CLOCK_GETTIME. It'd be nice to enable the monotonic time source but coming up with a safe condition (guaranteeing the source is available) for each Unix platform needs a lot of investigation.

The default Date provider, which performance.now() falls back to in absence of a monotonic time provider, has sub-millisecond resolution for Unix. So performance.now() works reasonably for common use cases, though the time is not necessarily jump free.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2992,6 +2992,10 @@ Planned
 * Add monotonic time provider for Windows based on QueryPerformanceCounter(),
   enabled by default if _WIN32_WINNT indicates Vista or above (GH-1662)
 
+* Add monotonic time provider for Unix platforms using clock_gettime() and
+  CLOCK_MONOTONIC time source; the time provider is disabled by default, enable
+  using -DDUK_USE_GET_MONOTONIC_TIME_CLOCK_GETTIME (GH-1667)
+
 * Use monotonic time (if available) for debugger transport peeking, so that
   the peek callback is called with the same realtime rate even if the
   Ecmascript time source jumps or doesn't advance in realtime (GH-1659)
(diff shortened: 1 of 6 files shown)


fatcerberus:
What is the difference between `clock_gettime(CLOCK_MONOTONIC)` and `clock()`?

svaarala:
As far as I can tell, clock() has poorer accuracy and can wrap around relatively quickly.

svaarala:
Linux man page indicates the resolution (CLOCKS_PER_SEC) is 1000000, i.e. microsecond resolution. On a 32-bit system the result will wrap around every 72 minutes.

Because clock() exists even in C89 it might be a good default provider for monotonic time, if one could ensure:
- The clock_t type is 64 bits (unlikely on a 32-bit system, likely on a 64-bit system) so wrapping is not a concern.
- CLOCKS_PER_SEC is high enough, e.g. on Unix systems it's apparently always 1000000 (microseconds).

Wrapping would be easy to handle if there was some guarantee Duktape would get execution time at least every 15-30 minutes. But, there's unfortunately no such guarantee.

fatcerberus:
MSDN indicates `CLOCKS_PER_SEC` is only 1000 for MSVC.  Additionally, it seems the value doesn't wrap around but instead *saturates* to -1 and this is the standard behavior per C99:
https://msdn.microsoft.com/en-us/library/4e2ess30.aspx

Which makes sense since `clock_t` is signed and signed overflow is undefined behavior, so it would be somewhat strange for a signed counter to roll over.

svaarala:
I think clock_t can be unsigned though; the error return value is `(clock_t) -1` which also works for an unsigned clock_t. Regarding the Linux behavior, from the man page:

> POSIX.1-2001, POSIX.1-2008, C89, C99.  XSI requires that CLOCKS_PER_SEC equals 1000000 independent of the actual resolution.
> [...]
> Note  that  the  time  can  wrap around.  On a 32-bit system where CLOCKS_PER_SEC equals 1000000 this function will return the same value approximately every 72 minutes.

Btw, while signed overflow is undefined behavior in general, there's no reason why it wouldn't be perfectly well defined for a particular clock() implementation (or more generally a particular architecture/compiler).

In any case, quite different behavior indeed, and unfortunately not always a good basis for monotonic time.

Issue #1230, opened 2016-12-28
Add C API macros for creating internal keys and symbols


svaarala:
Add something to hide the concrete prefixes (`\xFF` etc). At the very minimum a macro for `DUK_HIDDEN_SYMBOL()` a.k.a. an internal key would be useful.

fatcerberus:
In the spirit of abstraction, should there also be an API for checking if a value is a symbol, e.g. `duk_is_symbol()`?

svaarala:
That could be added too - but for symbols both `duk_is_string()` and `duk_is_symbol()` would be true for now.

fatcerberus:
Yeah, that's okay; the idea was that one could go:

```c
str = duk_require_string(ctx, 0);
if (duk_is_symbol(ctx, 0)) {
    duk_error(...);
}
```

...without relying on a particular internal symbol representation.

svaarala:
Maybe I'll add that to the 2.0.0 release, it seems relatively low risk for maintenance after all.

svaarala:
Macros now exist (for 2.2.0 release) and duk_is_symbol() was added earlier, so closing.

PR #1673, opened 2017-08-11
Add public macros to create symbol strings

svaarala:
Add macros to create different symbol types as C string literals:
- `DUK_HIDDEN_SYMBOL(x)` <=> `("\xFF" x)`
- `DUK_GLOBAL_SYMBOL(x)` <=> `("\x80" x)`
- `DUK_LOCAL_SYMBOL(x,uniq)` <=> `("\x81" x "\xFF" uniq)`
- `DUK_WELLKNOWN_SYMBOL(x)` <=> `("\x81" x "\xFF")`

Usage is e.g. as:

```c
duk_put_prop_string(ctx, obj_idx, DUK_HIDDEN_SYMBOL("myPointer"));
```

instead of:

```c
duk_put_prop_string(ctx, obj_idx, "\xFF" "myPointer");
```

All macros expect their arguments to be C string literals. In particular, the unique suffix for a local symbol cannot be a dynamically constructed value.

Adding these macros is a slight versioning risk because the internal representation might change, and no longer be representable as a string literal. On the other hand, it's better for calling code to depend on these macros instead of writing out the internal representations: if the macros are removed, the calling code will fail to compile instead of just silently creating bogus values.

Tasks:
- [x] Add macros
- [x] Add API test coverage
- [x] Documentation
- [x] 2.2 migration note recommending at least DUK_HIDDEN_SYMBOL()
- [x] Releases entry

Implements part of #1230.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -2947,6 +2947,9 @@ Planned
 * Add duk_push_proxy() API call which allows a Proxy to be created from C
   code (GH-1500, GH-837)
 
+* Add DUK_HIDDEN_SYMBOL(), DUK_GLOBAL_SYMBOL(), DUK_LOCAL_SYMBOL(), and
+  DUK_WELLKNOWN_SYMBOL() macros for creating symbol literals (GH-1673)
+
 * Add support for Proxy 'apply' and 'construct' traps (GH-1601)
 
 * Add minimal new.target support, evaluates to undefined for non-constructor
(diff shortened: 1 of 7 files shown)


fatcerberus:
So supposing I wanted to set `[Symbol.iterator]` for something in Duktape 2.2+, I'd do this, correct?
```c
duk_put_prop_string(ctx, obj_idx, DUK_WELLKNOWN_SYMBOL("iterator"));
```

svaarala:
That's correct. Relying on symbols being C string literals is a bit risky for versioning, but the macros are a better dependency than pure user-written literals. More versioning neutral would be a `duk_push_symbol()` which would have flags for different symbol types, but that's for later.

PR #1681, opened 2017-08-15
Assert for no API calls from debugger transport

### LINKED ISSUE #1656: Assert if application calls into Duktape from a debugger callback
Calling into Duktape is not allowed except when explicitly mentioned in documentation. However, it's an easy mistake to make, and causes very difficult to diagnose behavior. There's now a `DUK_ASSERT_API_ENTRY()` macro which can be extended to check for this violation.

svaarala:
Fixes #1656.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3021,6 +3021,9 @@ Planned
   without throwing, as internals are not always expecting to deal with an
   error when reading current time (GH-1666)
 
+* Add assertion coverage for catching calls into Duktape API from debug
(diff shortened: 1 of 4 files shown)


svaarala:
There are numerous places in the codebase where it's assumed that an optimizing compiler eliminates a `do {} while (0)` entirely. If that turned out not to be true, they can be replaced with something less harmful. In practice I haven't found a compiler that doesn't deal with these as expected. If that happens, it'll be straightforward to do something else.

PR #1683, opened 2017-08-16
Refactor debugger automatic pause handling

svaarala:
Use a single pause flags field to implement Resume, StepInto, StepOver, and StepOut, and reserve (as yet unused) pause flags for pausing after a single opcode, and pausing after caught/uncaught error. No debugger protocol changes yet in this pull.

Also improve behavior of StepOver, StepInto, and StepOut when the current activation is native and has no line information. Previously e.g. StepInto was ignored in this state altogether (which was undocumented). The improved behavior is to ignore the line-based pause trigger but use the others, i.e. StepInto will pause on function entry, function exit, and an error thrown out past the current function.

- [x] Remove step_type and replace it with pause_flags
- [x] Rework StepXxx to use pause_flags
- [x] Rename step act/line to pause act/line to match pause flags
- [x] Make pause on caught/uncaught handling generic and enabled when debugger is enabled
- [x] Rework DUK_USE_DEBUGGER_PAUSE_UNCAUGHT to automatically set pause-on-uncaught flag
- [x] Implement pause after one opcode
- [x] Accept StepXxx even when current activation has no line information
- [x] 2.2 migration note for StepXxx behavior change
- [x] Test Resume
- [x] Test StepInto
- [x] Test StepOver
- [x] Test StepOut
- [x] Test pause-on-caught manually
- [x] Test pause-on-uncaught manually
- [x] Test step opcode manually
- [x] Test e.g. StepInto from an Ecmascript function without line information

### DIFF
--- a/doc/debugger.rst
+++ b/doc/debugger.rst
@@ -1606,9 +1606,11 @@ Example::
     REQ 20 EOM
     REP EOM
 
-Resume execution and pause when execution exits the current line.  If a
-function call occurs before that, go into the function and pause execution
-there.
+Resume execution and pause when execution exits the current line, enters
+another function, exits the current function, or an error is thrown past
+the current function (in which case execution pauses in the error catcher,
+if any).  If the current function doesn't have line information (e.g. it is
+native), pauses on function entry/exit or error throw.
 
 StepOver request (0x15)
 -----------------------
@@ -1623,8 +1625,11 @@ Example::
     REQ 21 EOM
     REP EOM
 
-Resume execution and pause when execution exits the current line.  Don't pause
-on function calls occuring before that.
+Resume execution and pause when execution exits the current line, exits the
+current function, or an error is thrown past the current function (in which
+case execution pauses in the error catcher, if any).  If the current function
+doesn't have line information (e.g. it is native), pauses on function exit or
+error throw.
 
 StepOut request (0x16)
 ----------------------
@@ -1639,15 +1644,9 @@ Example::
     REQ 22 EOM
     REP EOM
 
-Resume execution and pause when execution exits the current function.  This can
-happen because:
-
-* The current function returns, in which case execution resumes in the calling
-  function.
-
-* The current function, or any function called by it, throws an error which is
-  not caught before it unwinds past the current function.  Execution resumes
-  in the error catcher.
+Resume execution and pause when execution exits the current function or an
+error is thrown past the current function (in which case execution pauses
+in the error catcher, if any).
 
 ListBreak request (0x17)
 ------------------------
--- a/doc/release-notes-v2-2.rst
+++ b/doc/release-notes-v2-2.rst
@@ -82,6 +82,11 @@ from Duktape v2.1.x.  Note the following:
   so duk_get_now() call sites may now get fractional millisecond timestamps
   even in default configuration.
 
+* Debugger StepInto, StepOver, and StepOut are now accepted also when the
+  current function doesn't have line information (i.e. it is native).  The
+  step commands will still pause on function entry/exit as appropriate; for
+  example, StepInto will pause on function entry or exit (or an error throw).
+
 Other minor differences:
 
 * When an Error instance is being constructed and Duktape.errCreate() is
--- a/src-input/duk_api_debug.c
+++ b/src-input/duk_api_debug.c
@@ -86,9 +86,9 @@ DUK_EXTERNAL void duk_debugger_attach(duk_hthread *thr,
 	heap->dbg_processing = 0;
 	heap->dbg_state_dirty = 0;
 	heap->dbg_force_restart = 0;
-	heap->dbg_step_type = DUK_STEP_TYPE_NONE;
-	heap->dbg_step_act = NULL;
-	heap->dbg_step_startline = 0;
+	heap->dbg_pause_flags = 0;
+	heap->dbg_pause_act = NULL;
+	heap->dbg_pause_startline = 0;
 	heap->dbg_exec_counter = 0;
 	heap->dbg_last_counter = 0;
 	heap->dbg_last_time = 0.0;
--- a/src-input/duk_debugger.c
+++ b/src-input/duk_debugger.c
@@ -86,9 +86,9 @@ DUK_LOCAL void duk__debug_do_detach1(duk_heap *heap, duk_int_t reason) {
 	/* heap->dbg_processing: keep on purpose to avoid debugger re-entry in detaching state */
 	heap->dbg_state_dirty = 0;
 	heap->dbg_force_restart = 0;
-	heap->dbg_step_type = 0;
-	heap->dbg_step_act = NULL;
-	heap->dbg_step_startline = 0;
+	heap->dbg_pause_flags = 0;
+	heap->dbg_pause_act = NULL;
+	heap->dbg_pause_startline = 0;
 	heap->dbg_have_next_byte = 0;
 	duk_debug_clear_paused(heap);  /* XXX: some overlap with field inits above */
 	heap->dbg_state_dirty = 0;     /* XXX: clear_paused sets dirty; rework? */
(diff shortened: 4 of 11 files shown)


svaarala:
@fatcerberus This is a little bit of experimentation to see if the pause handling could be generalized and be more flexible. There are a few bugs left here at present, I'll finish this later. If this is a workable improvement, I think the next steps would be:
- Allowing pause-on-caught/pause-on-uncaught regardless of debugger config options as far as the internals are concerned.
- Interpreting DUK_USE_DEBUGGER_PAUSE_UNCAUGHT to mean that whenever we resume (due to Resume or StepXxx), the internal pause-on-uncaught pause flag is set unconditionally. This preserves the current option semantics.
- Adding Resume flags to control pause-on-caught/pause-on-uncaught, and recommending dropping DUK_USE_DEBUGGER_PAUSE_UNCAUGHT and using the explicit Resume flags instead.
- StepXxx could also be deprecated as debugger commands (but not removed yet) if Resume can take their place with appropriate flags.

svaarala:
One of the bugs I encountered when testing this was actually also in master (#1684). I'll rebase and retest once that is fixed in master.

svaarala:
Ok, now the prototype contains roughly working "pause after one opcode" (StepOpcode), "pause on uncaught error", and "pause on caught error" flags. The flags can't be enabled via the debugger protocol yet, so I've just enabled them manually in duk_debugger.c (see diff).

svaarala:
Come to think of it, DUK_USE_DEBUGGER_PAUSE_UNCAUGHT could maybe be treated as:
- Resume command: automatically set "pause on uncaught" if explicit pause flags not provided. If pause flags *are* provided, just use them. This allows pause on uncaught to be controlled even when the option is set but doesn't break previous behavior.
- StepXxx commands would probably not get pause flags (if they're deprecated anyway), and would always set "pause on uncaught" with DUK_USE_DEBUGGER_PAUSE_UNCAUGHT.

svaarala:
@fatcerberus Any thoughts on whether the approach here makes sense? I.e. making Resume the main operation for normal resuming, stepping, etc, and deprecating the separate StepXxx commands?

fatcerberus:
I think this makes more sense than the individual StepXxx commands to be honest; all of those are logically the same operation ("resume execution until X happens") so it's better to fold them into one command now rather than end up adding even more "special resume" commands down the line.

Having it all subsumed into Resume also helps highlight that complex steps, like a StepOut, do indeed follow the same rules as a normal Resume (without having to emphasize that fact separately in documentation), e.g. a StepOut can be interrupted prematurely by an uncaught exception.

svaarala:
Assuming the pause flags would be a single flags field, the Resume command would take the forms:

```
Backward-compatible:
    REQ <int: 0x13> EOM
    REP EOM

    Implied flags:
      - DUK_USE_DEBUGGER_PAUSE_UNCAUGHT defined: "pause on uncaught"
      - Otherwise: no flags (= pause on breakpoint)

Explicit flags:
    REQ <int: 0x13> [<int: flags>] EOM
    REP EOM
```

There are a few Resume features that can't be expressed simply as flags. Breakpoints are an obvious example, but they're now handled as separate state, although they could also just be stateless arguments to each Resume. (Maybe there should be a Resume flag for "pause on breakpoint" though.)

Another example is an automatic pause after X milliseconds or Y opcodes if nothing else has triggered a pause.

I'd like to get one of those at least, the millisecond variant would be easier to implement. If so, Resume would then take the form:

```
Explicit flags and timeout:
    REQ <int: 0x13> [<int: flags> <num: timeout | undefined>] EOM
    REP EOM
```

The timeout implementation would at least initially be relatively low accuracy: it would piggyback on the existing interrupt counter + monotonic time check for the transport peek. The resolution would be ~200-300ms so it would be useful for the "resume for a short while" use case. If an opcode limit was added later it would then just follow the millisecond limit.

fatcerberus:
I think breakpoints as separate (persistent) state makes sense, given that they are normally presented that way in both graphical and command-line debuggers.

svaarala:
They're definitely a useful concept for the end user -- but that doesn't necessarily mean they should be implemented 1:1 in the debug protocol. It would be a lot simpler if the breakpoints were simply given in the Resume command. Some reasons:

- With breakpoints always listed in a Resume there would then be no need for any debugger management commands in the protocol. That would reduce Duktape footprint; not a huge difference, but still a gain with in practice no functional loss.
- There would be no need for the debug client to synchronize breakpoint state with the target if the debug session was dropped. At present the debug client needs to manage the breakpoints if it wants to keep them when resuming a dropped session. This also means that in many cases the debug client needs to hold the "master" breakpoint list even now so it could just send them in a Resume.
- The breakpoint concept provided by Duktape doesn't actually map 1:1 to many debug UIs. For instance, it's sometimes possible to enable/disable breakpoints from a list of "registered" breakpoints, without actually deleting the disabled breakpoints. In the current protocol the debug client would need to delete and re-add the breakpoints when enabled/disabled. Compared to just sending the active list, this is much more awkward. There are essentially two similar but not identical states that need to be kept in sync.
- In general I prefer the target to be stateless and dumb whenever possible, as it is conceptually simpler if one end holds almost all of the state rather than both ends maintaining significant state. In the current design the debug target is assumed to be as dumb as possible.

The downside of breakpoints in Resume is that the Resume commands will be larger. But that's not usually a big deal because Resumes are relatively infrequent and only happen when the target is paused and thus idle. Another downside is that given the limited extensibility of the debug protocol (due to the "streaming" design) extending Resume with flags, timeouts, breakpoint lists, etc, is not trivial and it's easy to paint oneself into a corner.

But, I'm not proposing changing this now, as it would need to be changed in a backwards compatible manner and that's worse than the current approach. It's definitely something to look into on the next incompatible update, though, especially if the command format gets some structure beyond being a list of dvalues.

svaarala:
One tidbit about current Resume behavior:

> If Duktape is already running, a no-op.

There are two alternatives for this:
- Keep current behavior: Resume when running is a no-op. Differences in current and requested flags are ignored (current flags are kept).
- Treat as a new Resume in the sense that any changes in the Resume flags are applied.

The latter behavior seems less surprising to me; I'd expect Duktape to update the resume (pause) flags.

fatcerberus:
> Treat as a new Resume in the sense that any changes in the Resume flags are applied.

This is my preferred behavior too.  As a user I would be quite perplexed if I sent execution parameters with a Resume and they went completely ignored.  Even with good documentation, the principle of least surprise is still a good guideline to follow. :)

From a less theoretical point of view: There may be use cases where one wants to change the setup without initiating a pause first.  For example, enabling pause-on-caught during live script execution to try to catch a bug in the act.

svaarala:
I was also thinking about a use case where you could toggle an automatic pause option (pause on error for example) while the debugger was running, and it'd be nice if that worked.

What won't work is changing the pause flags without resetting a previously set automatic pause timeout; the timeout would start again with a new target value. But this isn't a major issue, and at least to me not surprising behavior.

svaarala:
@fatcerberus Also made the following "grey area" change here:

> Improve behavior of StepOver, StepInto, and StepOut when the current activation is native and has no line information. Previously e.g. StepInto was ignored in this state altogether (which was undocumented). The improved behavior is to ignore the line-based pause trigger but use the others, i.e. StepInto will pause on function entry, function exit, and an error thrown out past the current function.

The updated behavior makes more sense IMO, and the current behavior (= command ignored without error) is undocumented and not very well thought out.

fatcerberus:
With that change, if I were to hypothetically set a breakpoint on a native function call and issue a StepInto, what would happen?  Since the function is not debuggable, my expectation is: the command is treated like StepOver in this case.

svaarala:
Hmm, I'm not sure what you mean: breakpoints are never active for native functions so you can never actually set a breakpoint on a native function?

fatcerberus:
I meant if you set the breakpoint at the call site and then do StepInto.

svaarala:
That behavior is not changed here, assuming the call site is Ecmascript code: it would work like StepOver.

There's no behavioral change because the call site, being Ecmascript, does have line information.

svaarala:
So to be clear, the only changed behavior is when we're paused (or running) and the current activation doesn't have line information (e.g. it is a native function).

One doesn't normally end up in this situation, but it might happen e.g. due to an uncaught error about to be thrown and perhaps in the future for some other triggers.

svaarala:
@fatcerberus I've tested all the scenarios I could think of manually, so once the tests pass I'll re-read the diff once more but I'm otherwise done with this pull. My intent is to merge in the changed Resume command after 2.2 release though, so this is just an internal change except for the improved StepXxx behavior.

If you have time to test this with Minisphere, it'd be nice because you have a specific set of debugger use cases you're interested in. So the intent is that nothing would change as far external behavior is concerned, except that StepXxx behavior change which is unlikely to happen in practice.

fatcerberus:
I know there have been a few issues I've run into in the past where the debugger either stopped responding to stepping commands, or else honored them but didn't seem to advance.  The first time this happened I thought it was maybe sourcemap-related, i.e. the transpiler created several lines from one, but after 10 steps the cursor still didn't advance.

I don't think that bug was related to the behavior you mention though, as it was always with pure Ecmascript code when I hit it.  Anyway, I'll give it a shot a bit later.

svaarala:
If you run into that with a version built before this change, and happen to have Duktape debug logs enabled, the debug logs will indicate if the step command was ignored or not (or processed at all, if that's the case).

svaarala:
I'll merge this out of the way - let me know if something has broken and I'll fix it :)

PR #1685, opened 2017-08-16
Fix debugger StepOut + tail return bug

### LINKED ISSUE #1684: StepOut triggers on a tail return
When issuing a StepOut in a function like:

```js
function test() {
    // ...
    print('running here');
    // ...
    return anotherFunction();
}
```

Execution will resume and pause at the entry to anotherFunction(). If the function is written as:

```js
function test() {
    // ...
    print('running here');
    // ...
    var ret = anotherFunction();
    return ret;
}
```

the pause will trigger once test() actually exits.

The behavior is caused by tailcall handling: when call handling reuses the duk_activation, it technically gets unwound which triggers the automatic pause.

This could be fixed by detecting the tailcall on unwind, continuing without pausing, but keeping the step_act state so that we pause when we exit the tailcall target.

svaarala:
Fixes #1684.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3060,6 +3060,10 @@ Planned
   below the current call, leading potentially to a smaller reserve than
   requested (GH-1536)
 
+* Fix incorrect pausing by debugger StepOut command when current function has
+  a tail return; previously execution would pause on exit to tailcall target
+  (GH-1684, GH-1685)
+
 * Fix duk_hbufobj assert in shared slice() handling (GH-1506)
 
 * Fix internal debug print opcode name list which was out of sync with previous
--- a/src-input/duk_js_call.c
+++ b/src-input/duk_js_call.c
@@ -1439,6 +1439,7 @@ DUK_LOCAL duk_small_uint_t duk__call_setup_act_attempt_tailcall(duk_hthread *thr
 	duk_tval *tv1, *tv2;
 	duk_idx_t idx_args;
 	duk_small_uint_t flags1, flags2;
+	duk_activation *prev_step_act;
 
 	DUK_UNREF(entry_valstack_end_byteoff);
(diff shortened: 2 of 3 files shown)


svaarala:
@fatcerberus Do you agree with the bug and the fix?

fatcerberus:
I *thought* I noticed some odd behavior in the past where a StepOut would act like a StepIn sometimes, but never stopped to try to diagnose it.  I guess I just assumed it was a bug in SSj.  So yes, I can confirm the bug exists.

Looking at the diff, if I understand the StepOut mechanism then Duktape triggers a pause after the activation referenced by `dbg_step_act` is unwound.  If that's the case then the fix looks good to me. :)

svaarala:
Yep, that's the mechanism for StepOut. An alternative to this fix would be to have a special unwind primitive (or flag) for tailcall unwind, but this store/restore is less complicated.

fatcerberus:
Incidentally does Step*Over* work correctly with tail calls?  If it were me implementing StepOver, I would probably implement it as "pause when the line number changes within the same activation", but a tailcall reuses the activation...

svaarala:
Good question, I'd have to test that. Some thoughts:

StepOver now maps internally to the "pause on line change" flag. That triggers when the activation is the same as before, but the line number has changed. The Ecmascript function is *not* compared in this check because the code just assumes it would be the same.

So maybe if the current function does a tailcall, and the target function lives in another file, the StepOver may either pause immediately (if the line number is different) or pause after a while (if the line number is the same).

I'm not sure whether some existing check handles this correctly or not. But a simple fix would be to disable the "pause on line change" flag entirely when doing a tailcall, because it could never be triggered in the current function anymore. The "pause on function exit" would then trigger when the tailcall exited.

PR #1695, opened 2017-08-25
Notes on an internal representation alternative

svaarala:
Add some design notes on an alternative internal representation where surrogate pairs would be combined as far the byte representation (visible through the C API) was concerned, but Ecmascript code would still see the strings as compliant uninterpreted 16-bit codepoint sequences.

No decision on whether this would be a good change or not. There are some upsides (and downsides) for C code, and some internal complications for string handling that needs attention in several places.

Somewhat related change would be to allow Ecmascript source code to contain literal non-BMP characters which wouldn't be surrogate pair encoded. That encoding is required by Ecmascript specification, but for some applications it might be useful to keep the strings as UTF-8 instead.

### DIFF
--- a/doc/utf8-internal-representation.rst
+++ b/doc/utf8-internal-representation.rst
@@ -1,0 +1,141 @@
+======================================
+Using UTF-8 as internal representation
+======================================
+
+Some notes on using UTF-8 as internal representation for Ecmascript strings
+when surrogate pairs can be combined.
+
+Current representation
+======================
+
+Current internal representation is a union of:
+
+* CESU-8: to support full 16-bit codepoint sequences without limitations.
+  In particular, individual and unpaired surrogates must work without
+  interpretation or conversion.
+
+* UTF-8: to support non-BMP characters, if they are created from C code
+  or e.g. using String.fromCharCode(0x12345).
+
+* Extended UTF-8: to support codepoints up to U+FFFFFFFF.  This is now
+  only needed by the regexp bytecode, which uses extended UTF-8 as its
+  internal representation and needs to represent long offsets as
+  codepoints.
+
+C API problem with current representation
+=========================================
+
+One concrete problem with this arrangement is that non-BMP strings are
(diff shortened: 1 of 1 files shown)


fatcerberus:
Presumably, in practice this would be WTF-8, not pure UTF-8 because unpaired surrogates would still need to be supported:
https://simonsapin.github.io/wtf-8/

svaarala:
It would still be the exactly same internal superset:
- Valid UTF-8
- CESU-8
- Codepoints [U+200000,U+FFFFFFFF]

So it would actually be WTF-8 + those additional codepoints. The additional codepoints are possible to drop: the only real dependency is the internal regexp format (which can be changed), and the API promise (which can be changed in a major update).

The main difference would be that valid surrogate pairs would be combined wherever possible so the internal representation would never have an unpaired valid surrogate pair.

svaarala:
But yeah, dropping the extra codepoints would more or less yield WTF-8. Great name for an encoding BTW :)

fatcerberus:
I'm all for this change, for what it's worth.  I resorted to using CESU-8 internally in miniSphere because otherwise I'm potentially pushing noncompliant strings into Ecmascript which I don't like the idea of.  Unfortunately CESU-8 is not really great for interoperability with other libraries.  That's mostly academic currently as I don't use any libraries where it matters, but I like to keep my options open :)

svaarala:
This would probably have a measurable performance impact for string manipulation because the opportunities to merge surrogates must be handled for any replacement, join, etc operation. Also footprint will increase. So it's unfortunately not just a win-win all around.

fatcerberus:
That's true, plus string `.length` calculation would have to be adapted too because astral characters must have a length of 2 per Ecmascript.

svaarala:
Yeah, I wrote a short list of some of the "as if" things in this pull. It's not complete and some of them may get a bit thorny: e.g. one must be able to RegExp match from either surrogate, backtrack one surrogate at a time, etc.

Issue #1120, opened 2016-11-29
Symbol support cleanups after initial merge


svaarala:
The initial merge is relatively large so some cleanups are handled separately:
- [ ] To reduce footprint, separate config options for Symbol built-ins and Symbol support in general?
- [ ] Genbuiltins.py assumes function `.name` matches its key (e.g. `Math.cos` gets the name `cos`); this isn't true in ES6, for example `Symbol.prototype[Symbol.toPrimitive].name` is `[Symbol.toPrimitive]' (includes the brackets).

fatcerberus:
Additional (not necessarily short-term) cleanup: Review ES6 to see which explicit class IDs can be removed and replaced with a built-in @@toStringTag.

PR #1719, opened 2017-09-06
Add hint to guide about using hidden Symbols to store state

jmhmccr:
It isn't immediately obvious of how to attach C specific state to an Ecmascript object (took me a little too long to put it together). Added an extra hint to the docs discussing storing C exclusive state. I chose to add another header since adding an Ecmascript accessible property is a different use case than the norm of passing a user context `void*` to function.

Screenshot:

duktape_hidden_symbol_property_pr

### DIFF
--- a/website/guide/programming.html
+++ b/website/guide/programming.html
@@ -173,7 +173,7 @@ currently used element.  The following diagram illustrates this:</p>
 |  8 |      | 5 |   API index 0 is bottom (at value stack index 3).
 |  7 |      | 4 |
 |  6 |      | 3 |   API index 5 is highest used (at value stack index 8).
-|  5 |      | 2 |   
+|  5 |      | 2 |
 |  4 |      | 1 |   Stack top is 6 (relative to stack bottom).
 |  3 | &lt;--- | 0 |
 |  2 |      `---'
@@ -491,6 +491,27 @@ methods, which has both advantages and disadvantages.</p>
 <pre class="c-code">
 duk_push_this(ctx);
 duk_get_prop_string(ctx, -1, "my_state_variable");
+</pre>
+
+<h3 id="hidden-symbol-properties">Hidden Symbol properties</h3>
+
+<p>If data needs to be associated with an object, but hidden from Ecmascript
+code, hidden <a href="#symbols">Symbols</a> can be used as a property key. The
+key is only accessible via the C API, unless passed to Ecmascript code from the
+C API. A common use case is to associated backing pointers/data to C memory.
+Symbols are created as a string, but are differentiated with macros that mark
+them as Symbols.</p>
+
+<p>For example, setting and getting a hidden symbol on <code>this</code>:</p>
+<pre class="c-code">
+my_context_data_t *my_context_data = malloc(sizeof(my_context_data_t));
+duk_push_this(ctx);
+duk_push_pointer(ctx, my_context_data);
+duk_put_prop_string(ctx, -2, DUK_HIDDEN_SYMBOL("my_context_data"));
+/* ... */
+duk_push_this(ctx);
+duk_get_prop_string(ctx, -1, DUK_HIDDEN_SYMBOL("my_context_data"));
+my_context_data_t *my_context_data = duk_get_pointer(ctx, -1);
 </pre>
 
 <h3>Magic value of function</h3>
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -45,6 +45,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Michael Drake <michael.drake@codethink.co.uk>
 * Steven Don (https://github.com/shdon)
 * Simon Stone (https://github.com/sstone1)
+* \J. McC. (https://github.com/jmhmccr)
 
 Other contributions
 ===================


fatcerberus:
This already seems to be covered in the Guide:
http://duktape.org/guide.html#symbols

jmhmccr:
Yeah, that is linked in the 

The Symbol relation to storing state was a bit difficult to discover since the section discussing state does not mention it.

jmhmccr:
Any other suggestions on how to relate the idea of using hidden Symbols as storing C user context data?

svaarala:
@jmhmccr Looks good to me. A short introduction to the topic in the Programming model section seems like the right place to me too, as it already covers how one can store state for a function instance which is a similar common use case. I'll have a look at the text in more detail later.

website/guide/programming.html
```diff
@@ -493,6 +493,32 @@ <h2 id="duktape-api">Duktape API</h2>
...
 </pre>
 
+
+<h3>Hidden Symbol properties</h3>
+
+<p>If data needs to be associated with an object, but hidden from Ecmascript
+code, hidden <a href="#symbols">Symbols</a> can be used as a property key. The
+key is only accessible via the C API, unless passed to Ecmascript code from the
+C API. A common use case is to associated backing pointers/data to C memory.
+Symbols are created as a string, but with a specific prefix.</p>
```
svaarala: There's now a `DUK_HIDDEN_SYMBOL("my_context_data")` macro in the API which is preferred, so the "specific prefix" note is maybe unnecessary detail. Also the example should be updated to use the macro. The Symbols section (in master) discusses the macros.

website/guide/programming.html
```diff
@@ -493,6 +493,32 @@ <h2 id="duktape-api">Duktape API</h2>
...
+
+<p>Setting then getting a hidden symbol on <code>this</code>.</p>
```
svaarala: IMO better wording would be:

> For example, setting and getting a hidden Symbol on <code>this</code>:

website/guide/programming.html
```diff
@@ -493,6 +493,32 @@ <h2 id="duktape-api">Duktape API</h2>
...
+<pre class="c-code">
+duk_push_this(ctx);
+duk_push_pointer(ctx, my_context_data);
+duk_put_prop_string(ctx, -2, "\xff" "\xff" "my_context_data");
```
svaarala: Using the new API macro:

```c
duk_put_prop_string(ctx, -2, DUK_HIDDEN_SYMBOL("my_context_data"));
```

website/guide/programming.html
```diff
@@ -493,6 +493,32 @@ <h2 id="duktape-api">Duktape API</h2>
...
+/* ... */
+duk_push_this(ctx);
+duk_get_prop_string(ctx, -1, "\xff" "\xff" "my_context_data");
```
svaarala: Same here.

website/guide/programming.html
```diff
@@ -493,6 +493,32 @@ <h2 id="duktape-api">Duktape API</h2>
...
+my_context_data_t *my_context_data = duk_get_pointer(ctx, -1);
+</pre>
+
+<p>See the <a href="#symbols">Symbols</a> section regarding the strange C
+literal string syntax.</p>
```
svaarala: This remark is IMO unnecessary with the API macro; it's not visible to calling code. It is still discussed in the Symbols section, and that is linked to already above.

website/guide/programming.html
```diff
@@ -493,6 +493,32 @@ <h2 id="duktape-api">Duktape API</h2>
...
+
+<p>Add a finalizer function to the Ecmascript object if you need to clean up
+resources.</p>
```
svaarala: This remark seems disconnected to the topic. Finalizers can be used also for state which is stored as Ecmascript visible strings or Symbols. I'd suggest removing, or rewording to better connect it to the discussion above (for example: "If the data stored using hidden Symbols needs native code cleanup when the object is freed, a finalizer function ...", and make the finalizer a link to the finalization section).

website/guide/programming.html
```diff
@@ -493,6 +493,32 @@ <h2 id="duktape-api">Duktape API</h2>
 duk_get_prop_string(ctx, -1, "my_state_variable");
 </pre>
 
+
+<h3>Hidden Symbol properties</h3>
```
svaarala: Maybe add an id to the section so it can be linked to:

```
<h3 id="hidden-symbol-properties">...</h3>
```

website/guide/programming.html
jmhmccr: Yep, just saw the symbol macro changes after I submitted this PR. Will make the changes.

Will there be a `DUK_USER_HIDDEN_SYMBOL`, or maybe changing `DUK_HIDDEN_SYMBOL` to `DUK_INTERNAL_SYMBOL` to disambiguate a `"\xff" "\xff"` application style symbol from the duktape internal symbols? I would think it would be simpler to not have to explain the naming conventions in documentation (like uppercase vs. lowercase).

website/guide/programming.html
jmhmccr: Sounds good.

website/guide/programming.html
jmhmccr: Sounds good.

website/guide/programming.html
jmhmccr: I think the rewording/linking would help make the section more straight forward and help to keep copy-pasters in check?

website/guide/programming.html
jmhmccr: Doh, will do.

website/guide/programming.html
svaarala: Personally I'd prefer using "\xFFvalue" over "\xFF\xFFvalue" in calling code simply because it's shorter. It's difficult to enforce the case convention with a macro, but I don't think it has caused much confusion in practice.

No plan for a separate user hidden Symbol macro at present, but for it to work reliably as a macro it would either need to use the double \xFF prefix, or maybe a separate prefix byte could be used (e.g. 0x82 to 0xbf is still free) for user hidden Symbols.

In any case I think the Symbols section needs a short discussion of this; the details were moved to symbols.rst but maybe the very basics need to be in the Guide section too.

svaarala: Hmm, actually it would be easiest from a backwards compatibility perspective to:
- Reserve the 0xFF prefix for user hidden symbols entirely.
- Use 0x82 prefix for Duktape internal hidden symbols because no user code should be relying on the internal hidden symbols (they're not under versioning guarantees).

svaarala: I added the prefix byte change as a work item for #1120, I don't think I'll have time to make the change in 2.2.0.

svaarala: I'll open a pull for it - it's not a difficult change but I'm just tidying up odds and ends for 2.2.0. If it seems risk-free I can see if it can still get merged or not. But regardless of that, it would be a backwards compatible change, keeping DUK_HIDDEN_SYMBOL() but simply removing the constraints for user code.

website/guide/programming.html
jmhmccr: After looking over the state section more, I think moving to a more general note in the state section as a whole would be better as it can apply to all of the state techniques. I don't see an immediate place to put it, so will leave it out of this PR and think about a more cohesive note later.

website/guide/programming.html
jmhmccr: I was thinking just have the `DUK_USER_HIDDEN_SYMBOL` style macro just prepend the `"\xFF" "\xFF"` across the board, without worrying about the convention.

Would be simple enough maybe to make it into the 2.2.0?

svaarala: Yes, but I would not want to use such a macro myself :-)

website/guide/programming.html
jmhmccr: Since none of the other sections have an `id` attribute, would you prefer that I leave this single id change out and submit another PR adding `id` attributes to all the sections??

svaarala: Some of the sections in programming.html do have an id attribute, mostly when it's easy to anticipate they will be linked to:

```
<h1 id="programming">Programming model</h1>
<h2 id="duktape-api">Duktape API</h2>
<h2 id="error-handling">Error handling</h2>
<h2 id="normal-and-fatal-errors">Normal and fatal errors</h2>
```

website/guide/programming.html
svaarala: Ok, #1721 would make the 0x82 byte prefix change. It seems like a pretty small change so maybe I'll shoehorn it into 2.2.

jmhmccr:
Requested changes pushed via rebase.

New screenshot:
duktape_hidden_symbol_properties_pr_comments

website/guide/programming.html
```diff
@@ -493,6 +493,28 @@ <h2 id="duktape-api">Duktape API</h2>
 duk_get_prop_string(ctx, -1, "my_state_variable");
 </pre>
 
+
```
svaarala: Nitpick: extra empty line.

website/guide/programming.html
```diff
@@ -493,6 +493,28 @@ <h2 id="duktape-api">Duktape API</h2>
...
+my_context_data_t *my_context_data = malloc(sizeof(my_context_data_t));
+duk_push_this(ctx);
+duk_push_pointer(ctx, my_context_data);
+duk_put_prop_string(ctx, -2, DUK_HIDDEN_SYMBOL("my_context_data"));
+/* ... */
+duk_push_this(ctx);
+duk_get_prop_string(ctx, -1, DUK_HIDDEN_SYMBOL("my_context_data"));
+my_context_data_t *my_context_data = duk_get_pointer(ctx, -1);
+</pre>
+
```
svaarala: Placing the section here is a bit of a compromise: the H2 heading talks about function state. Then again, the H2 level section talks about stashes etc which are general concepts and not restricted to function state. (The guide is in need of some rewriting in parts, and the programming model section is one of those parts ... :)

But overall, I think this is a good enough location for now.

svaarala:
Other than the whitespace nitpick, looks good :+1:

website/guide/programming.html
jmhmccr: taken care of.

svaarala:
Uh oh, one more small detail, the AUTHORS.rst entry needs an escape to avoid a formatting glitch. A similar case exists for:

```
* \D. Bohdan (https://github.com/dbohdan)
```

Without the escape, Github seems to interpret the "D." as a nested list (see the formatted AUTHORS.rst in the pull).

svaarala:
I'm off to sleep, but I'll merge once the AUTHORS.rst formatting is fixed. Thanks! :+1:

PR #1720, opened 2017-09-06
Use lz-string to compress benchmarks.html raw data

svaarala:
Page size drops from ~4.5MB to ~0.7MB, as most of the page is commit-related testrunner data.

Cool library with nice license: https://github.com/pieroxy/lz-string.


fatcerberus:
Haha, that license is awesome.  It's like public domain with attitude! :)

PR #1721, opened 2017-09-07
Use 0x82 byte prefix for Duktape hidden Symbols, freeing 0xFF byte prefix to application use

svaarala:
See https://github.com/svaarala/duktape/pull/1719#discussion_r137661524 and discussion around it.

Tasks:
- [x] String intern flag handling change
- [x] Genbuiltins.py change for interpreting `hidden` symbol type
- [x] Genbuiltins.py change for ROM string HIDDEN and SYMBOL flags
- [x] Internal documentation update
- [x] Website update
- [x] Testcase coverage for 0x82 prefixed strings not coming out of getOwnPropertySymbols()
- [x] API testcase fixes
- [x] Ecmascript testcase fixes
- [x] Testcase for accessing Date _Value using 0x82 prefix
- [x] Octane segfault; new String('foo').length
- [x] Grep for 0xFF and check all call sites
- [x] Manual test for ROM symbol visibility
- [x] 2.2 migration note
- [x] Releases entry

There's also a subtle change for consistency for initial bytes 0x83 to 0xBF. These bytes are reserved and don't have a guaranteed behavior, but Duktape 2.1 interpreted them as Symbols for RAM strings and non-Symbols for ROM strings. This pull makes the interpretation consistent so that the prefixes are interpreted as non-Symbols and documentation clarified to say that one shouldn't rely on either behavior.


svaarala:
Ok, this pull should now be done. The testserver is very loaded right now (it's re-running master merges going back 500 merges to master) so I'll wait for a while for the tests to execute.

@jmhmccr There are improvements (?) to the Symbols section in the pull. If you have time, it'd be nice if you could check if the changes are indeed improvements in your view :)

fatcerberus:
Out of curiosity, why are all those merges being re-tested?

svaarala:
@fatcerberus I improved the testserver/testclient protocol to allow each test run to result in a JSON object which can contain structured data like binary size, octane scores, etc. This structured data allows http://duktape.org/benchmarks.html to provide a bunch of graphs based on those test results.

The tests are re-executed to get the structured results for past builds into the server database. Also commits to master didn't previously trigger test runs, only pull request pushes did.

fatcerberus:
I looked at those graphs, those are nice.  I see footprint went down since 2.1.0, that's always good to see.

svaarala:
> I see footprint went down since 2.1.0, that's always good to see.

That's indeed very nice especially for lowmem targets. The default build size has fluctuated but kept relatively constant. One reason I wanted to expand the scope of the very barebones previous graphs is that it's easier to see where the benchmarks are heading -- which then makes it more motivating to keep chipping away on footprint etc.

svaarala:
Octane score is also interesting, it has increased by ~40% from 2.0.0 release but because there's a lot of noise, it's quite difficult to see the difference when working on a single pull request. But from the benchmarks page it's now easier to see the trend overall is positive.

doc/symbols.rst
```diff
@@ -62,16 +62,11 @@ for ES2015 symbols (and reserved for other future uses) in Duktape 2.x.
...
-|                                               | second is from [A-Z].                                           |
-+-----------------------------------------------+-----------------------------------------------------------------+
-| <ff> anyOtherValue                            | Hidden symbol (Duktape specific) used by application code.      |
-|                                               | First byte is 0xFF, second is ASCII (0x00-0x7f) but not         |
-|                                               | from [A-Z].                                                     |
-+-----------------------------------------------+-----------------------------------------------------------------+
-| <ff> <ff> anyOtherValue                       | Hidden symbol (Duktape specific) used by application code.      |
-|                                               | First and second bytes are 0xFF, remaining bytes arbitrary.     |
+| <ff> anyValue                                 | Hidden symbol (Duktape specific) used by application code.      |
+|                                               | Prior to Duktape 2.2 Duktape internal hidden symbols also used  |
```
jmhmccr: It might be nice to add a blank line above this. Easier to ignore that way.

website/guide/symbols.html
```diff
@@ -5,23 +5,27 @@ <h1 id="symbols">Symbols</h1>
 <b>hidden Symbol</b> variant similar to internal strings in Duktape 1.x.
 Hidden Symbols differ from ES2015 Symbols in that they're hidden from
 ordinary Ecmascript code: they can't be created from Ecmascript code,
-won't be enumerated or JSON-serialized, and won't be returned even from
-<code>Object.getOwnPropertyNames()</code>.  Properties with hidden Symbol
+won't be enumerated or JSON-serialized, and won't be returned from
+<code>Object.getOwnPropertyNames()</code> or even
+<code>Object.getOwnPropertySymbols()</code>.  Properties with hidden Symbol
```
jmhmccr: Nice addition.

website/guide/symbols.html
```diff
@@ -5,23 +5,27 @@ <h1 id="symbols">Symbols</h1>
...
+C code pushes a string using e.g. <code>duk_push_string()</code> and the byte
+sequence matches an internal Symbol format, the string value is automatically
+interpreted as a Symbol.</p>
+
+<p>Duktape also uses hidden Symbols for various implementation specific purposes,
+such as storing an object's finalizer reference.  As of Duktape 2.2 a different
+byte prefix is used for Duktape's hidden Symbols, so the 0xFF prefix is now
+reserved entirely for application use.  Application code should never try to
+access Duktape's hidden Symbol keyed properties: the set of such properties can
+change arbitrarily between versions.</p>
```
jmhmccr: I feel like the mention of the internal representation would easily be covered via the `doc/symbols.rst` link. Probably only mentioning a user code compatible, internal only change and linking to the release notes would be good enough? Who is the audience for the **Guide**? I would imagine only client API users?

I will work on a simplified Symbols section tonight to hopefully better illustrate my comment.

doc/symbols.rst
svaarala: I'm not sure what you mean - do you mean adding a blank line between the first sentence and the rest of the Description? If so, that would make the entry inconsistent with the other description fields which don't separate a leading sentence from further explanation.

website/guide/symbols.html
svaarala: The audience of the website is all Duktape developers, at any stage of their involvement: starting from glancing at the index page, getting started, and later on using as a reference. The line between users and Duktape developers is sometimes fuzzy, because a lot of users tweak the internals, maintain patched forks, etc.

As a general principle:
- The index page tries to provide information necessary to figure out if the library is right for you and to make it easy to try.
- The Guide is an umbrella document which should cover all relevant topics (except engine neutral, general Ecmascript topics) minimally and if necessary to link to Github developer docs (doc/xxx.rst) or Wiki for more discussion.
- The API is a reference document which should be convenient when working on stuff and quickly looking up functions.

These three documents are IMO the most important and the goal should be for them to be clean, easy to read, as short as possible but still covering the necessary topics, polished language, etc. The Wiki and the internal developer documentation have a much lower bar.

I don't the documentation achieves the goal I outlined above at present: it still pretty much has the same structure as when I initially wrote it before Duktape 1.0 release. In particular transitioning from "just read the index page", to Getting Started, to actually writing programs (Programming Model) would benefit from a rewrite.

doc/symbols.rst
jmhmccr: alrighty, sounds good.

website/guide/symbols.html
jmhmccr: Ok, the goals seem different than what I thought. This looks great then.

fatcerberus: @svaarala For what it's worth, I really like that you wrote the documentation with potential Duktape contributors in mind.  Makes it much less scary when it becomes necessary to hack away at stuff under the hood :)

Issue #1604, opened 2017-07-10
Issue with duk_instanceof


edoardomazzella:
Hello!

duk_get_global_string(ctx,"IOBUS");
duk_push_object(ctx);

if(!duk_instanceof(ctx,-1,-2)) ...

This is not the real code, so i know that it has no sense, but i found this problem: when duk_instanceof is called, even in this case, with an object as first argument it always throws an error "object required, found undefined". Are you sure it works fine?
Thank you!

svaarala:
I'm never completely sure about anything but at least the API tests works: https://github.com/svaarala/duktape/blob/master/tests/api/test-instanceof.c.

If you provide a reproducible example I can test it.

svaarala:
At least the API test seems to lack testing for negative indices. They go through helpers that should deal with them correctly, but I'll check if that makes a difference for some reason. The internal call sites to `duk_instanceof()` only use non-negative indices so a bug there wouldn't be caught.

svaarala:
Could you try this?

```
        duk_eval_string(ctx, "new RangeError('aiee')");
        duk_eval_string(ctx, "Error");

        printf("top=%ld\n", (long) duk_get_top(ctx));
        printf("%ld\n", (long) duk_instanceof(ctx, -2, -1));
```

This print '1' for me, i.e. `new RangeError('aiee') instanceof Error` is true.

edoardomazzella:
yes it prints 1...
My code is something like this:

static duk_ret_t IOBUS_dtor(duk_context *ctx){

  return 0;
}

static duk_ret_t IOBUS_ctor(duk_context *ctx){

  if (!duk_is_constructor_call(ctx))
    return duk_type_error(ctx, "Function IOBUS_ctor must be called as"
        " constructor");

  duk_push_this(ctx);

  duk_dup(ctx,0);
  duk_put_prop_string(ctx,-2,"mask");

  duk_dup(ctx,1);
  duk_put_prop_string(ctx,-2,"offset");

  // Store the function destructor
  duk_push_c_function(ctx, IOBUS_dtor, 0);
  duk_set_finalizer(ctx, -2);

  return 0;
}

void scriptly_create_IOBUS_prototype(duk_context *ctx){
  duk_push_c_function(ctx, IOBUS_ctor, 2);
  duk_put_global_string(ctx, "IOBUS");
}

static duk_ret_t duk_read_bus(duk_context *ctx){

  IOBus bus;

[...]

PR #1725, opened 2017-09-08
Improve duk instanceof no prototype error

svaarala:
When doing `{} instanceof Constructor` and Constructor has no `.prototype` property, the current error message ("invalid instanceof rval") is correct but not very informative. Because this situation happens frequently with Duktape/C constructor functions (which don't have a .prototype by default), improve the error message. See #1604.

Also fixes a potentially dangling pointer: the instanceof rval `.prototype` was looked up, the duk_hobject pointer was assigned to a variable, and the `.prototype` value was then popped off. This works as long as the `.prototype` is a concrete property, but it might also be a getter (or Proxy trap backed). Tagged security because a dangling pointer may cause memory unsafe behavior.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3003,6 +3003,9 @@ Planned
 * Make error message summary strings longer (32 -> 96 character) to better
   capture error messages for e.g. uncaught errors (GH-1653)
 
+* Improve error message for instanceof and duk_instanceof() when rval has
+  no .prototype property, which is common for Duktape/C functions (GH-1725)
+
 * Add an explicit buffer size wrap check for Buffer.concat() (GH-1688)
 
 * Reject an attempt to unpack an array with >= 2G elements cleanly with a
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3063,6 +3063,9 @@ Planned
 * Fix potential segfault in debugger GetHeapObjInfo command, caused by
   key/mask list being out of sync (GH-1540)
 
+* Fix dangling pointer in instanceof/duk_instanceof() when rval .prototype is
+  a virtualized property coming from a getter or a Proxy trap (GH-1725)
+
 * Fix Reflect.construct() handling for four or more arguments (GH-1517,
   GH-1518)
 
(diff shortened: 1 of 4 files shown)


svaarala:
Updated error message:

```
duk> function Constructor() {}; Constructor.prototype = void 0;
= undefined
duk> print({} instanceof Constructor)
TypeError: instanceof rval has no .prototype
    at [anon] (duk_js_ops.c:1185) internal
```

Comparison to Node.js:

```
> function Constructor() {}; Constructor.prototype = void 0;
undefined
> console.log({} instanceof Constructor)
TypeError: Function has non-object prototype 'undefined' in instanceof check
    at repl:1:27
    at REPLServer.defaultEval (repl.js:252:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:417:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:210:10)
    at REPLServer.Interface._line (readline.js:549:8)
    at REPLServer.Interface._ttyWrite (readline.js:826:14)
```

and Firefox:

```
> function Constructor() {}; Constructor.prototype = void 0;
undefined
> console.log({} instanceof Constructor)
TypeError: 'prototype' property of Constructor is not an object
```

fatcerberus:
And just because I can now, miniSphere with Chakra :)

```js
function Ctor() {}
Ctor.prototype = undefined;
new Ctor() instanceof Ctor;
```

```
JavaScript exception!

TypeError: Function does not have a valid prototype object
   at Anonymous function (:13:1)
   at module (:7:1)
```

On a sidenote, I also just found out that `delete fn.prototype` is apparently illegal in strict mode.  Huh.

fatcerberus:
Of all the error messages I like Duktape's (i.e. the one in this pull) the best, it's nice and concise without sacrificing clarity.

svaarala:
I don't think delete fn.prototype is illegal as such - it just fails silently in non-strict mode due to the property being non-configurable:

```
$ nodejs
> function Foo() {}
undefined
> delete Foo.prototype
false
> 'prototype' in Foo
true
```

fatcerberus:
Oh, I see, it's just the normal "delete non-configurable property" strict mode check.  I was confused because the error message was:

```
TypeError: Calling delete on 'prototype' is not allowed in strict mode
```

But then I tried on other nonconfigurable properties and it just changed "prototype" to whatever I tried to delete.  So it wasn't special-cased after all.

svaarala:
Heh, I can see why that message seemed to be crafted specifically for .prototype... :)

fatcerberus:
Yeah, that's why I'm so meticulous about the wording of error messages.  Sometimes the phrasing of an error can give the wrong idea and then you end up writing a fragile bugfix because of invalid assumptions.

Issue #1560, opened 2017-06-13
Fix warnings


mike-lischke:
Clang gives me a number of warnings that should be easy to fix. I would create a PR with the changes, but the duktape.c file is a merge of many other files, which makes it difficult to create a simple patch. Here's what I get from clang:

```bash
<path>/src/duktape/duktape.c:27917:34: warning: implicit conversion loses integer precision: 'unsigned long' to 'duk_uint_fast32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
                        match_val = (1UL << part_idx) + (1UL << (sep_idx + 9));  /* match against rule part/sep bits */
                                  ~ ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
<path>/src/duktape/duktape.c:30202:63: warning: implicit conversion loses integer precision: 'long' to 'duk_codepoint_t' (aka 'int') [-Wshorten-64-to-32]
                                codepoint = 0x010000L + ((enc_ctx->lead - 0xd800L) << 10) + (codepoint - 0xdc00L);
                                          ~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
<path>/src/duktape/duktape.c:31550:50: warning: implicit conversion loses integer precision: 'long' to 'duk_codepoint_t' (aka 'int') [-Wshorten-64-to-32]
                cp = ((cp1 - 0xd800L) << 10) + (cp2 - 0xdc00L) + 0x10000L;
                   ~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
<path>/src/duktape/duktape.c:57720:55: warning: implicit conversion loses integer precision: 'unsigned long' to 'duk_ucodepoint_t' (aka 'unsigned int') [-Wshorten-64-to-32]
                                cp1 = ((cp1 - 0xd800UL) << 10) + (cp2 - 0xdc00UL) + 0x10000UL;
                                    ~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
<path>/src/duktape/duktape.c:79447:8: warning: variable 'tmp' may be uninitialized when used here [-Wconditional-uninitialized]
                cp = tmp;
                     ^~~
[...]

PR #1735, opened 2017-09-15
Fix some clang -Wshorten-64-to-32 warnings

svaarala:
See #1560.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,6 +153,9 @@ CCOPTS_DEBUG = $(CCOPTS_SHARED) $(CCOPTS_FEATURES)
 CCOPTS_DEBUG += -O0
 CCOPTS_DEBUG += -g -ggdb
 
+CLANG_CCOPTS_NONDEBUG = $(CCOPTS_NONDEBUG)
+CLANG_CCOPTS_NONDEBUG += -Wshorten-64-to-32
+
 GXXOPTS_SHARED = -pedantic -ansi -std=c++11 -fstrict-aliasing -Wall -Wextra -Wunused-result -Wunused-function
 GXXOPTS_SHARED += -DDUK_CMDLINE_PRINTALERT_SUPPORT
 GXXOPTS_NONDEBUG = $(GXXOPTS_SHARED) -Os -fomit-frame-pointer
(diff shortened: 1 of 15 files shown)


svaarala:
@mike-lischke It'd be good if you could verify that you no longer get warnings from this branch. I tried to reproduce the warnings by adding -Wshorten-64-to-32 but quite likely my warning option set (and clang version) is still different.

mike-lischke:
Ah good. Will test it in the coming week, thanks. Btw. why don't enable -Wall instead of specific warnings?

svaarala:
I do enable -Wall and even -Wextra, but despite the name they don't enable all warning options.

svaarala:
For example, my Clang supports -Wshorten-64-to-32 but it's not enabled by default, with -Wall, or even with -Wextra. This makes it difficult to achieve warning free compilation because the warning set depends on the compiler, and some of the opt-in warnings are quite limiting (for example, there's a warning for potentially truncating a snprintf(), which is sometimes quite intentional).

mike-lischke:
I can confirm these warnings are gone. However, new warnings just came up with XCode 9:

```
duktape.c:13668:83: warning: possible misuse of comma operator here [-Wcomma]
                        p = DUK_BW_ENSURE_RAW(thr, bw_ctx, 1U + 4U + DUK_HSTRING_GET_BYTELEN(h_str), p),
                                                                                                       ^
duktape.c:13668:4: note: cast expression to void to silence warning
                        p = DUK_BW_ENSURE_RAW(thr, bw_ctx, 1U + 4U + DUK_HSTRING_GET_BYTELEN(h_str), p),
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                        static_cast<void>(                                                             )
duktape.c:16073:70: warning: possible misuse of comma operator here [-Wcomma]
                vals[DUK__IDX_PBYTES] = (duk_int_t) DUK_HOBJECT_P_ALLOC_SIZE(h_obj),
                                                                                   ^
duktape.c:16073:3: note: cast expression to void to silence warning
                vals[DUK__IDX_PBYTES] = (duk_int_t) DUK_HOBJECT_P_ALLOC_SIZE(h_obj),
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                static_cast<void>(                                                 )
duktape.c:63122:33: warning: variable 'act' may be uninitialized when used here [-Wconditional-uninitialized]
        duk__call_env_setup(thr, func, act, idx_args);
                                       ^~~
duktape.c:62947:21: note: initialize the variable 'act' to silence this warning
        duk_activation *act;
                           ^
                            = nullptr
duktape.c:63140:29: warning: variable 'nregs' may be uninitialized when used here [-Wconditional-uninitialized]
                duk_set_top_and_wipe(thr, nregs, nargs);
                                          ^~~~~
duktape.c:62944:17: note: initialize the variable 'nregs' to silence this warning
        duk_idx_t nregs;            /* # total registers target function wants on entry (< 0 => "as is") */
                       ^
                        = 0
duktape.c:63140:36: warning: variable 'nargs' may be uninitialized when used here [-Wconditional-uninitialized]
                duk_set_top_and_wipe(thr, nregs, nargs);
                                                 ^~~~~
duktape.c:62943:17: note: initialize the variable 'nargs' to silence this warning
        duk_idx_t nargs;            /* # argument registers target function wants (< 0 => "as is") */
                       ^
                        = 0
duktape.c:63134:37: warning: variable 'vs_min_bytes' may be uninitialized when used here [-Wconditional-uninitialized]
        duk_valstack_grow_check_throw(thr, vs_min_bytes);
                                           ^~~~~~~~~~~~
duktape.c:62945:25: note: initialize the variable 'vs_min_bytes' to silence this warning
        duk_size_t vs_min_bytes;    /* minimum value stack size (bytes) for handling call */
                               ^
                                = 0
6 warnings generated.
```

Can you pls fix them too :-)?

svaarala:
Sure :)

fatcerberus:
I love how it’s suggesting to add `static_cast<void>(...)` to a C file -_-

svaarala:
In many cases the compiler is complaining about something being potentially unused simply because it doesn't always understand that an internal guarantee or an error check (causing a longjmp) guards against the situation. It's kind of frustrating to add code to explicitly initialize stuff for these cases because it never gets executed and typically does increase footprint. But... I'll do it, grudgingly ;) (And this isn't the case for all the warnings by a long shot, just a sidenote.)

svaarala:
Those -Wcomma warnings are real, the rest are use_tailcall related warnings where the compiler doesn't seem to fully analyze that e.g. nregs is set either via tailcall or non-tailcall setup (mediated by use_tailcall which probably confuses the compiler). I'll see if I can restructure that code. Initializing e.g. act to NULL would silence to warning but would be pretty pointless because if it ever mattered, it would cause a segfault anyway (and if the warning was in fact accurate this would then make the bad behavior silent).

mike-lischke:
@fatcerberus I compile duktape.c as C++ code, since I use C++ exceptions.

PR #1775, opened 2017-09-28
Address 'defined but not used' warning

beevik:
In duk_bi_date_windows.c, the function `duk__convert_filetime_to_ularge`
is a static function called only from `duk_bi_date_get_now_windows_subms`,
which is wrapped in a `#define`.

This change wraps `duk__convert_filetime_to_ularge` in the same `#define`
to prevent compiler warnings about unused functions.


svaarala:
Thanks!

PR #1781, opened 2017-10-15
Add work-in-progress CBOR encoder/decoder as (unreleased) extra

svaarala:
CBOR is a good alternative to JSON for byte efficient serialization, and may be a replacement for the custom dvalue format in the debugger protocol in the future. Experiment a bit with binding CBOR to Duktape; the pull provides C bindings to encode/decode arbitrary Ecmascript (Duktape) values into CBOR and back. There's also a command line tool for doing JSON-to-CBOR and CBOR-to-JSON conversions. The extra is not yet included in the distributable because it'll need some more work.

Tasks:
- [x] CBOR encoding
- [x] CBOR decoding
- [x] Testcases
- [x] README
- [x] JSON/CBOR conversion tool

### DIFF
--- a/extras/cbor/Makefile
+++ b/extras/cbor/Makefile
@@ -1,0 +1,147 @@
+# For manual testing; say 'make' in extras/cbor and run ./test.
+
+VALGRIND=valgrind
+.PHONY: all
+all: clean test jsoncbor test-runs
+
+.PHONY: clean
+clean:
+	-rm -f test jsoncbor
+
+test: test.c duk_cbor.c duk_cbor.h
+	gcc -std=c99 -Wall -Wextra -g -ggdb -o $@ -I../../src/ -I. ../../src/duktape.c duk_cbor.c test.c -lm
+
+jsoncbor: jsoncbor.c duk_cbor.c duk_cbor.h
+	gcc -std=c99 -Wall -Wextra -O2 -o $@ -I../../src/ -I. ../../src/duktape.c duk_cbor.c jsoncbor.c -lm
+
+.PHONY: test-runs
+test-runs: test jsoncbor enc-primitives enc-number enc-buffer enc-string enc-array enc-object enc-misc dec-misc
+
+.PHONY: dec-misc
+dec-misc: jsoncbor
+	python -c 'import sys; sys.stdout.write("\xa1cfoocbar")'| python cbordecode.py
+	python -c 'import sys; sys.stdout.write("\xa1cfoocbar")'| $(VALGRIND) -q ./jsoncbor -d
+	python -c 'import sys; sys.stdout.write("\xa1\xa0\xa0")'| python cbordecode.py
+	python -c 'import sys; sys.stdout.write("\xa1\xa0\xa0")'| $(VALGRIND) -q ./jsoncbor -d  # object key, gets string coerced in JSON output
+	python -c 'import sys; sys.stdout.write("\x7f\xff")'| python cbordecode.py
+	python -c 'import sys; sys.stdout.write("\x7f\xff")'| $(VALGRIND) -q ./jsoncbor -d
+	python -c 'import sys; sys.stdout.write("\x7f`````````````````````````````````````````````````````````````````````````````````````````````````````````\xff")'| python cbordecode.py
+	python -c 'import sys; sys.stdout.write("\x7f`````````````````````````````````````````````````````````````````````````````````````````````````````````\xff")'| $(VALGRIND) -q ./jsoncbor -d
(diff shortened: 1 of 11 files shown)


fatcerberus:
Re: Bindings - It might be nice to eventually provide JavaScript bindings for CBOR too.  It would compliment the built-in standard JSON bindings nicely.

For what it's worth, I think CBOR might be somewhat heavy for use in the debugger protocol ("heavy" conceptually, not referring to footprint).  When I switched miniSphere from Duktape to ChakraCore, I needed to implement a new debugger protocol to communicate with SSj; unlike Duktape, CC has a JSON-based debugging *API* that runs directly on the target.

I researched CBOR and almost implemented it for debug messages, but ultimately decided to adapt the Duktape protocol instead as it was simpler and still met all my needs (plus I didn't need to rewrite as much code immediately, a nice bonus :-)
https://github.com/fatcerberus/minisphere/blob/master/src/shared/ki.c
https://github.com/fatcerberus/minisphere/blob/master/src/shared/ki.h

svaarala:
> Re: Bindings - It might be nice to eventually provide JavaScript bindings for CBOR too. It would compliment the built-in standard JSON bindings nicely.

Agreed, having some binary serialization format would be nice.

Though one major reservation I have with that is that eventually I'd like to have a fast binary serialization format that would be able to serialize object graphs efficiently. Ultimately I'd like to be able to serialize a heap and revive it back from flash, with enough metadata to allow the application to revive native objects like files, sockets, etc in the process.

So that will need some memory efficient format that will be able to represent all Duktape types reasonably well. And it will need to be able to represent object graphs with reference cycles.

CBOR doesn't do that directly (and it's indeed a bit challenging) but it can still be done over CBOR by using some sort of object reference scheme, e.g. serialize the heap as indexed values and resolve the indices when reading back.

So maybe it would work out, maybe not. Using a standard would be nice, but it's not a showstopper requirement if it's not a reasonably good fit without a lot of elbow grease.

fatcerberus:
By heavy I meant in the sense of being overly generic.  For example CBOR objects are key/value pairs where the key can be any type, even entire objects!  So you d necessarily be using a very limited subset for the debugger protocol which feels kind of awkward.

The genericity would be nice for serializing actual Ecmascript objects though so maybe it wouldn't be so bad after all.  I don't know, I'm just picky about bloat I guess :)

svaarala:
> For example CBOR objects are key/value pairs where the key can be any type, even entire objects!

But that's basically what one needs to extend better (versioning-wise) than a flat list of primitive values.

svaarala:
Ah, misread what you said :) Yeah, keys can be objects but nothing forces one to create such objects. I would stick with the JSON compatible subset so that a CBOR-JSON mapping would be straightforward.

svaarala:
So for concreteness, if one encountered an object whose keys were objects in the protocol, one would simply declare a syntax error and detach. It wouldn't be fully in keeping with the usual parsing principles, i.e. one would normally parse such an object, and if the key wasn't interesting, just ignore it. But, it'd be an OK way to deal with the situation IMO because nothing forces one to send such objects.

It's also not actually clear whether one might actually need such objects. For example, when encoding property tables, the keys are not actually strings but `duk_hstring` references. It might be better to encode the property table keys as string references, retaining as much of the internal representation as possible, than to encode them as plain strings. (Although in this particular case one might need to use an array for the key/value pairs to preserve ordering, something that CBOR doesn't guarantee; key/value ordering of any CBOR/JSON object might change in transit.)

One motivation for writing a basic CBOR binding is to get a better grasp of what works and what doesn't.

svaarala:
@fatcerberus A small sidenote re: objects-as-keys, Lua allows objects as table keys and it's actually IMO a good feature: it makes the object model more regular (keys and values have the same space), allows tables to be used for tracking e.g. visited objects conveniently, and objects can be used as unique keys similarly to Symbols.

In fact, if one were able to control enumeration, placeholder empty objects would serve the role of Symbols. Without access to the object (assuming they could be protected from enumeration) they would be inaccessible to a call site which didn't have a reference to the object.

The downside is that objects are a bit heavier in terms of representation but depending on the tagged value representation this may be an issue or not. For example, with tagged pointers strings and arbitrary tagged values are the same size and it's a non-issue; in other representations like current `duk_tval` it matters and string pointers are smaller than general tagged values.

Another downside is that many other contexts assume keys are strings (or at least primitive); in particular, JSON does that. So if you actually take advantage of that representation, you'll then run into problems with JSON serialization. But that'd also be the case if one wanted to serialize a Map.

svaarala:
Merging a work-in-progress which has some remaining work (half-floats, etc) but is nevertheless useful to some extent. Follow-up pull will complete the basic functionality.

PR #1784, opened 2017-10-17
Use align-by-8 on all platforms by default

### LINKED ISSUE #1783: Use natural alignment rules by default
See discussion in https://github.com/svaarala/duktape/issues/812#issuecomment-337058737. In newer compiler versions memory accesses (or even pointer computations) that don't match natural alignment (*not* target alignment) may be undefined behavior *even when* they satisfy target architecture alignment rules.

The safest approach would be to assume natural alignment by default in `duk_config.h` for all architectures (which would in practice mean `-DDUK_USE_ALIGN_BY=8` by default for all targets) and assume more only when the user requests that explicitly via configure.py.

svaarala:
Fixes #1783. Tasks:
- [x] Remove all alignment forcing from architecture files, rely on align-by-8 default fillin (one can still force alignment via configure.py)
- [x] Releases entry
- [x] 2.3 migration notes: need to force if memory usage matters and lower alignment is possible
- [x] Low memory documentation note

This also means that x86/x64 will switch from `duk_hobject` property layout 1 to 2.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3194,6 +3194,11 @@ Planned
   MAX_SAFE_INTEGER, isFinite(), isInteger(), isNaN(), isSafeInteger(),
   parseInt(), parseFloat() (GH-1756, GH-1761)
 
+* Assume only natural alignment (DUK_USE_ALIGN_BY=8) on all platforms to
+  minimize compiler assumptions regarding unaligned accesses (which can be
+  an issue even on x86); applications working with a low memory target may
+  want to force DUK_USE_ALIGN_BY in configuration (GH-1783, GH-1784)
+
 * Fix trailing single line comment handling for Function constructor;
   new Function('return "foo" //') previously failed with SyntaxError
   (GH-1757)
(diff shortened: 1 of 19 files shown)


svaarala:
Footprint impact on x64 is about +350 bytes. RAM usage impact on a small testcase:

Memory usage for property tables increases a bit on average for x86/x64, concrete impact depends on how object heavy code is executed:

```
duktape@keppihevonen:~/duktape$ ./duk.master
((o) Duktape [linenoise] 2.2.99 (v2.2.0-48-g5b889a4b)
duk> x={foo:123}; Duktape.info(Duktape.compact(x));
= {hptr:(0x13c7c90),type:6,itag:9,refc:3,hbytes:56,class:1,pbytes:25,esize:1,enext:1,asize:0,hsize:0,variant:0}

duktape@keppihevonen:~/duktape$ ./duk
((o) Duktape [linenoise] 2.2.99 (v2.2.0-52-g5c48b9c8)
duk> x={foo:123}; Duktape.info(Duktape.compact(x));
= {hptr:(0x105e310),type:6,itag:9,refc:3,hbytes:56,class:1,pbytes:32,esize:1,enext:1,asize:0,hsize:0,variant:0}
```

Here the property table for a single property on x64 increases from 25 to 32 bytes.

PR #1787, opened 2017-10-20
Fix StepOver a nested tailcall

### LINKED ISSUE #1786: Debugger StepOver doesn't pause after stepping over a tail call
In duktape version 2.2.0, the duktape debugger doesn't pause after stepping *over* a call to a function that does a tail call.

See the comment in the code below:
```javascript
var Test = (function () {
    function Test() {
        return this;
    }
    Test.prototype.Foo = function Foo() {
        return 1;
    };
    Test.prototype.TailCall = function ReturnInline() {
        return this.Foo();
    };
    Test.prototype.OnUpdate = function OnUpdate() {
        this.TailCall();    // line A: debugger doesn't pause after stepping over this.
        var extraCode = 1;  // line B:
        ++extraCode;
    };
    return Test;
}());
```

Stepping into the call on line A works fine.  Setting a breakpoint on line B also causes the debugger to pause after stepping over line A.  But if no breakpoint is set on line B, the debugger does not pause on line B when stepping over line A.

Perhaps this issue is related to the following release note:

>Fix incorrect pausing by debugger StepOut, StepOver, and StepInto commands when stepping out/over/into a tail call (https://github.com/svaarala/duktape/issues/1684, https://github.com/svaarala/duktape/issues/1685, https://github.com/svaarala/duktape/issues/1726, https://github.com/svaarala/duktape/issues/1734)

svaarala:
Fix for a debugger regression in 2.2.0: StepOver a call which does a nested tailcall causes #1786:

- [x] Draft fix for root cause
- [x] Clean up fix
- [x] Check if a simpler fix suffices: don't add a duk_activation flag, but only clear the "pause on line" flag if doing a tailcall *and* the current activation matches dbg_pause_act
- [x] Manual testcase
- [x] Releases entry

Fixes #1786.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3201,6 +3201,9 @@ Planned
 
 * Add C++ name mangling wrappers (extern "C") for extras (GH-1780, GH-1782)
 
+* Fix debugger StepOver behavior when a tailcall happens in a nested
+  function (not the function where stepping started from) (GH-1786, GH-1787)
+
 * Fix trailing single line comment handling for Function constructor;
   new Function('return "foo" //') previously failed with SyntaxError
   (GH-1757)
--- a/src-input/duk_js_call.c
+++ b/src-input/duk_js_call.c
@@ -1507,9 +1507,10 @@ DUK_LOCAL duk_small_uint_t duk__call_setup_act_attempt_tailcall(duk_hthread *thr
 	 *    - Disable StepOut processing for the activation unwind because
 	 *      we reuse the activation, see:
 	 *      https://github.com/svaarala/duktape/issues/1684.
-	 *    - Disable line change pause flag permanently (if set) because
-	 *      it would no longer be relevant, see:
-	 *      https://github.com/svaarala/duktape/issues/1726.
+	 *    - Disable line change pause flag permanently if act == dbg_pause_act
+	 *      (if set) because it would no longer be relevant, see:
+	 *      https://github.com/svaarala/duktape/issues/1726,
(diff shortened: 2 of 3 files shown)


svaarala:
@beevik This is incomplete and I'll need to read all the call sites where pause state is changed to complete the change, but could you check if this fixes your issue? (Because it's incomplete it might introduce some other issues, but just to see if I've identified the root cause correctly.)

svaarala:
More plain repro:
```js
function foo() {
    return 1;
}

function bar() {
    return foo();
}

function test() {
    print('before bar');
    bar();  // step over this
    print('after bar');
}

test();
```

svaarala:
Possible simpler fix: only add a check that "pause on line change" is cleared in tailcall only if the reused activation matches heap->dbg_pause_act. Otherwise don't clear the flag.

beevik:
I cherry-picked this change, and the debugger now pauses after stepping over the function with the tail call.

svaarala:
There's now a much simpler fix: no new tracking flags are needed, just clear "pause on line change" on a tailcall only if the tailcall happens in the stepping start point activation. This ensures that:
- If the stepping startpoint activation is reused, line pausing will no longer occur as it wouldn't be valid in the tail call target. But StepOut still works on unwind, when we unwind from the tailcall target.

@beevik The fix works for my manual test, if you have time to test the updated fix, that'd be nice. If not, I'll merge to master because I think the root cause is quite clear and the simpler fix is quite likely to fix the original issue.

PR #1789, opened 2017-10-30
More lenient base64 decoder, base64 performance improvements

svaarala:
- [x] Allow missing ('Zm'), partial ('Zm='), and extra padding ('Zm====') in the base64 decoder
- [x] Improve base-64 fast path encoder and decoder performance
- [x] Improve base-64 slow path encoder and decoder performance without increasing footprint
- [x] Base64 testcase coverage improvements
- [x] Make base64 support optional (DUK_USE_BASE64_SUPPORT) and disable it by default in lowmem configuration
- [x] Make hex support optional (DUK_USE_HEX_SUPPORT) and disable it by default in lowmem configuration
- [x] 2.3 migration note for base64/hex and lowmem
- [x] Releases entry

Performance improvements for fast path on my development laptop (Intel(R) Core(TM) i7-4600U CPU @ 2.10GHz) measured using tests/perf/test-base64* (includes string intern after encode):
- Base64 encode: 1030 MB/sec in master, 1170 MB/sec after
- Base64 decode: 1770 MB/sec in master, 1900 MB/sec after

Performance improvements for slow path:
- Base64 encode: 130 MB/sec in master, 360 MB/sec after
- Base64 decode: 440 MB/sec in master, 450 MB/sec after (negligible)


svaarala:
There is considerable variation in base64 decoder leniency. After the changes in this pull, the remaining differences to Node.js behavior is:
- Node.js doesn't accept concatenated documents but terminates at the first padding character, so `Zm==Zm9v` decodes to `f`, whereas Duktape decodes to `ffoo`.
- Duktape rejects `Z`, `Z===`, and other padding variants of the form with a TypeError, Node.js accepts them and decodes an empty string because there's only 6 bits of input.

fatcerberus:
I'm curious what the difference is (algorithm-wise) between the fast and slow paths for base64 and what the conditions for using each is, as base64 is a very simple encoding--or at least seems to be.

svaarala:
My wording was a bit imprecise, what I mean is the fast variant (= DUK_USE_BASE64_FASTPATH enabled) and slow variant (disabled). The fast variant has a fast path for dense groups (`"Zm9v"`), in fact for two dense groups at a time, and other kinds of groups that involve whitespace (`"Zm\n9v"`), padding, etc.

For decoding base-64 is definitely not a simple format due to all the variation, see e.g. https://en.wikipedia.org/wiki/Base64#Variants_summary_table. Unfortunately even with the leniency changes here, the decoder won't work for all base-64 and its variants. In fact even the encoder won't work in all contexts because some of them have mandatory line length limits. So ultimately the base64 encoder/decoder will need flags or needs to be some subset of all the variants; I think the current subset, after this pull, is a reasonable compromise but of course that's very subjective.

svaarala:
The main optimization in the fast path (of the fast variant) is that it decodes two 4-character groups optimistically with only one error check branch for those total of 8 characters. So if the input is e.g. `Zm9vZm9v` it decodes to `foofoo` with one error check branch only. A trivial decoder has to check every character for validity (and the fallback of the fast variant does that of course) which involves a lot more branches and is consequently much slower.

PR #1790, opened 2017-10-30
Fix warnings when DUK_NORETURN not defined, add DUK_WO_NORETURN()

svaarala:
Add an internal DUK_WO_NORETURN() macro which provides statements that are only included when a noreturn attribute (DUK_NORETURN) is unavailable. This allows the error throw in:
```
DUK_INTERNAL duk_bool_t duk_something(duk_hthread *thr) {
    /* ... */
    if (something) {
        DUK_ERROR_TYPE(thr, "aiee");
        DUK_WO_NORETURN(return 0;);
    }
}
```

to compile as the following with DUK_NORETURN defined:

```c
    if (something) {
        DUK_ERROR_TYPE(thr, "aiee");
    }
```

and as the following when it isn't defined:

```c
    if (something) {
        DUK_ERROR_TYPE(thr, "aiee");
        return 0;
    }
```

In the latter case the 'return 0;' is never actually reached, but without a noreturn attribute (not available in many compilers) the compiler won't know it. Without a return statement the compiler will then sometimes warn about using uninitialized variables, etc. On the other hand, including the return statement always may cause warnings about unreachable code with other compilers when the noreturn attribute is available.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3231,6 +3231,11 @@ Planned
 
 * Fix some Clang warnings (GH-1777)
 
+* Fix some compilation warnings triggered when DUK_NORETURN is not defined;
+  the internal DUK_WO_NORETURN() is used to include unreachable dummy
+  statements when the noreturn attribute is not available, e.g.
+  DUK_WO_NORETURN(return 0;); (GH-1790)
+
 * Fix a few casts in duk_trans_socket_windows.c to avoid errors in a C++
   build (GH-1773)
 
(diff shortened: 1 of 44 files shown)


fatcerberus:
That was a clever fix for this long-standing issue. :)

I do wonder if it could be simplified to look less awkward, e.g.
```c
DUK_ERROR_TYPE(thr, "aiee");
DUK_WO_NORETURN(0);  // expands to return 0;
```

svaarala:
The dummy statement is not always a return right now so that wouldn't work directly. There's a theoretical issue that one can't use a plain return after a `va_start()`, a `va_end()` should be done before the return. But, one should avoid longjmp()'ing between a va_start() and a va_end() as that might in theory leak memory in some implementations. (This reminds me that there are a few such cases in the code base right now, and they should be fixed if possible.)

PR #1793, opened 2017-11-03
Fix some armhf -Wcast-align warnings

svaarala:
Fix some -Wcast-align warnings triggered on armhf target. The warnings are related to casting between different heap structs where one struct has an alignment requirement of 8 and another an alignment requirement of 4. However, all Duktape heap objects are allocated using the malloc() provider and the duk_heaphdr struct always starts directly at the pointer returned by malloc(). If alignment to 8 matters for the platform, malloc() is expected to return pointers aligned by 8 (at least if size >= 8) so that all heap objects will in fact be 8-byte aligned. The fix is to cast through void *.


svaarala:
@tlsa-ct If you have time, it'd be nice to hear if this fixes the issue behind #1792. As far as I can tell the warnings are harmless because all actually allocated structs should begin at an 8-byte aligned address on armhf, so all Duktape heap objects are aligned by 8. The union tricks in duk_hbuffer_fixed then further ensure that the data area following duk_hbuffer_fixed (`buf + 1`) is 8-byte aligned too so that it can hold aligned data correctly.

PR #1795, opened 2017-11-06
Avoid undefined behavior warnings triggered by -fsanitize=undefined

svaarala:
Use explicit checks to avoid undefined behavior in at least:
- memcpy() and memcmp() with zero size and NULL pointer
- signed left shift
- division by zero
- various casts

The explicit checks are now enabled by default. For most targets they're not actually necessary because with most compilers the behaviors are fine, so the checks can be dropped by defining `DUK_USE_ALLOW_UNDEFINED_BEHAVIOR` which is also enabled by default for lowmem.


fatcerberus:
I thought memcpy() with zero size was well-defined?

svaarala:
Zero-size memcpy() etc with a NULL pointer argument (which happens in some call sites) is not well defined, the argument pointers must be non-NULL even with a zero size. So the workaround is just to avoid the zero size case.

fatcerberus:
That's kind of odd since malloc(0) is allowed by the standard to return NULL.  Huh, learned something new today.

svaarala:
Yep, that's the annoying thing: if you malloc() something, the result pointer is not necessarily safe for memcpy() etc if the alloc size was zero. It's undefined behavior so `FORMAT C:` is a valid implementation.

In practice I've never encountered an implementation which has an issue with a non-valid pointer in the zero size case. But it's conceivable some implementation might verify the target addresses and fail with an invalid pointer.

PR #1798, opened 2017-11-08
Footprint optimize -fsanitize=undefined fixes

svaarala:
Footprint optimize -fsanitize=undefined fixes: use duk_memcmp() in C99+ safe contexts and duk_memcmp_unsafe() in unsafe contexts, and change the wrappers to macros. This matters because there are hundreds of call sites, and even a few bytes per call site adds up quickly (the difference is around 0.9kB here).

Also fix a bug in double-to-float cast which caused a sanitize warning.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3231,7 +3231,8 @@ Planned
 
 * Fix some Clang warnings by avoiding undefined behavior by default, define
   DUK_USE_ALLOW_UNDEFINED_BEHAVIOR to reduce the explicit undefined behavior
-  checks for better footprint/performance (GH-1777, GH-1795, GH-1796, GH-1797)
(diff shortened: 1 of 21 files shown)


fatcerberus:
I don't understand why the call not being inlined adds footprint, as the original call sites were function calls too (to memcmp or memcpy).  If anything I would have guessed the call sites might get *smaller*, since the compiler would no longer be using an intrinsic there.

Optimizing compilers are clearly made of black magic. :)

svaarala:
Memcpy() call sites are typically just `REP STOSB` or similar. If they go through an internal helper which wraps that, the result will be larger.

svaarala:
Footprint impact in master (x64 -Os) for explicit checks to avoid (usually unproblematic) undefined behavior is now ~700 bytes.

PR #1800, opened 2017-11-11
CBOR extra improvements

svaarala:
- [x] Add support for full half-float decoding: denormals, normals, Inf, NaN.
- [x] Enable half-float encoding now that half-float decoding is fully functional. Still doesn't encode denormals but that's not required.
- [x] Improve `jsoncbor` tool to support JX, indentation etc; usage is e.g. `./jsoncbor -r cbor -w jx --indent 4 <input.cbor >output.jx`. As such this tool would now replace `jxpretty` and be more versatile.
- [x] Add `-r js` option to allow input value to be an eval expression.
- [x] Remove separate 'test' tool, `-r js` can replace it.
- [x] Check for single float in encoding when shorter than a double.
- [x] Change all extras to run configure.py explicitly rather than relying on a 'src' directory. They then work both from the distributable and the repo.


svaarala:
Some usage examples for jsoncbor.c.

Decode JSON, encode into JX:
```
$ echo '{"foo":[1,2,3]}' | ./jsoncbor -r json -w jx
{foo:[1,2,3]}
```

Decode JX, write into indented JSON:
```
$ echo '{foo:[1,2,3]}' | ./jsoncbor -r jx -w json --indent 4
{
    "foo": [
        1,
        2,
        3
    ]
}
```

Decode CBOR, output indented JSON:
```
$ echo -n $'\xa2cfoocbarcbar\x82\x01\x02' | ./jsoncbor -r cbor -w json --indent 4
{
    "foo": "bar",
    "bar": [
        1,
        2
    ]
}
```

svaarala:
Input can now also be evaled script code (useful for testing):
```
$ echo '[0,1,2,3,4].map(Math.sin)' | ./jsoncbor -r js -w jx --indent 4
[
    0,
    0.8414709848078965,
    0.9092974268256817,
    0.1411200080598672,
    -0.7568024953079282
]
$ echo '[0,1,2,3,4].map(Math.sin)' | ./jsoncbor -r js -w cbor | hd
00000000  85 00 fb 3f ea ed 54 8f  09 0c ee fb 3f ed 18 f6  |...?..T.....?...|
00000010  ea d1 b4 46 fb 3f c2 10  38 6d b6 d5 5b fb bf e8  |...F.?..8m..[...|
00000020  37 b9 dd dc 1e ae                                 |7.....|
00000026
```

PR #1801, opened 2017-11-12
CBOR extra improvements, add CBOR to dist

svaarala:
Final changes for initial inclusion into extras:

- [x] Fix some wrap and out-of-range cast cases
- [x] Minor footprint optimization
- [x] Remove a few unnecessary test Python scripts (inline to Makefile)
- [x] Test vectors: https://github.com/cbor/test-vectors/blob/master/appendix_a.json, executed manually
- [x] JS API:https://github.com/paroga/cbor-js
- [x] Add CBOR extra to dist
- [x] Add CBOR binding to 'duk' built from repo Makefile
- [x] Releases entry

Several shortcomings remain, e.g. tests are not actually automatic but rather rely on manual inspection and footprint is still somewhat large (~5kB).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,8 @@ DUKTAPE_CMDLINE_SOURCES = \
 	extras/print-alert/duk_print_alert.c \
 	extras/console/duk_console.c \
 	extras/logging/duk_logging.c \
-	extras/module-duktape/duk_module_duktape.c
+	extras/module-duktape/duk_module_duktape.c \
+	extras/cbor/duk_cbor.c
 ifdef SYSTEMROOT  # Windows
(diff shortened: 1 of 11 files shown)


svaarala:
With duk_cbor_init() one can (optionally) register the `CBOR` global with Ecmascript bindings. The `duk` utility built from the repo now does that by default:
```
((o) Duktape [linenoise] 2.2.99 (v2.2.0-128-g05cd4d88-dirty)
duk> CBOR.encode({ foo: [1,2,3], bar: void 0 })
= |a263666f6f8301020363626172f7|
duk> CBOR.decode(CBOR.encode({ foo: [1,2,3], bar: void 0 }))
= {foo:[1,2,3],bar:undefined}
```

PR #1804, opened 2017-11-13
Replace gcc by $(CC) in Makefile.sharedlibrary and other dist Makefiles

ffontaine:
This patch will allow cross-compilation

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -48,6 +48,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * \J. McC. (https://github.com/jmhmccr)
 * Jakub Nowakowski (https://github.com/jimvonmoon)
 * Tommy Nguyen (https://github.com/tn0502)
+* Fabrice Fontaine (https://github.com/ffontaine)
 
 Other contributions
 ===================
--- a/dist-files/Makefile.codepage
+++ b/dist-files/Makefile.codepage
@@ -1,4 +1,6 @@
+CC	= gcc
+
 codepage:
-	gcc -o $@ -std=c99 -O2 -Wall -Wextra -Isrc/ \
+	$(CC) -o $@ -std=c99 -O2 -Wall -Wextra -Isrc/ \
 		src/duktape.c examples/codepage-conv/duk_codepage_conv.c \
 		examples/codepage-conv/test.c -lm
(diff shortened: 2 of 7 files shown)


svaarala:
Please add yourself to AUTHORS (first list, not second), see CONTRIBUTING.

Re: the change itself, I can merge it to master but the Makefiles in the distributable are examples only. They're not maintained for a set of actual targets because there are quite many of them and a variety of different build systems are used.

svaarala:
If this is merged to master I think the equivalent change should be made to all the dist (and extras) Makefiles that go into the final distributable. I can do that as a separate pull.

ffontaine:
Thanks for your review, I updated my PR with AUTHORS as well as updates to other Makefiles in dist but I forgot the Makefile in the extra folder. If you want, I can update my PR again or I can create a new one for this folder.

svaarala:
Thanks! If you have time a separate pull for the extras folder would be nice for consistency (I can do that also if necessary).

Re: the updated diff, should `CC = gcc` be `CC ?= gcc` throughout so that CC can be passed in from outside? There seem to be various conventions so I'm not sure what's the most common one.

fatcerberus:
It's been my experience that `CC=gcc` still allows you to override it, at least under Linux.

svaarala:
With CC=gcc you can override it as:
```
$ make test CC=foo
```

but not as:
```
$ CC=foo make test
```

The `CC ?= gcc` form works in both cases. Not sure which is the better practice though.

svaarala:
@ffontaine I'm OK with `CC = gcc` if the behavior is as intended - let me know if I should merge this as is.

svaarala:
As a sidenote, while it's not a problem here, overriding just CC is in general not enough because also the compiler options may need some overriding, and things get really messy when compiler or even compiler version specific options are used in complex ways (say PGO and two-pass compilation). Consequently, making really portable Makefiles involves a lot of trivia.

While that's an interesting challenge, I'm not personally very interested that, and explains why I've tried to keep the Makefiles merely as examples rather than being actually maintained portable Makefiles working on a wide variety of targets. There are simply more interesting things to spend time on, especially considering many users don't like Makefiles and prefer SCons, various IDEs, etc :)

ffontaine:
`CC = gcc` is the intended use, I changed this to keep some consistency with the Makefiles that were already defining CC such as Makefile.cmdline or Makefile.dukdebug. I will send you a new PR for the extra folder ASAP.

PR #1805, opened 2017-11-14
Add duk_push_literal() API call

svaarala:
Add `duk_push_literal(ctx, "foo")` which is in principle the same as `duk_push_string(ctx, "foo")` but with minor differences (the API call is similar to http://pgl.yoyo.org/luai/i/lua_pushliteral):
- The argument is assumed to be a C literal. This allows the string length to be computed at the call site as `sizeof(str) - 1` (-1 for excluding the terminating NUL).
- The string data is assumed to be immutable so in the future Duktape may just point to the string rather than making a copy in the intern. This would be nice for lowmem targets where "external strings" are already used.
- Because the literal address may be reused in subsequent intern check calls, in the future Duktape might speed up the literal -> string table lookup somehow to avoid hashing (and maybe even comparison in some approaches). For example, a soft address -> string table index (with validation before use) might be worth it.

The internals are changed to use the new API call where appropriate. The performance impact just with this change alone (in essence avoiding a strlen()) is negligible, but visible in very specific microbenchmarks.

I'm adding this as an experimental entry for now. The convenience functions like `duk_get_prop_string()` also need `xxx_literal()` counterparts for the addition to be complete.

Tasks:
- [x] Add API macro duk_push_literal()
- [x] API documentation
- [x] API testcase
- [x] Testcase coverage for return value
- [x] Perf test demonstrating the (very minor) advantage
- [x] Follow-up issues: #1806
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3195,6 +3195,9 @@ Planned
   useful to detect constructor calls which can already be done using
   duk_is_constructor_call() (GH-1745)
 
+* Add experimental duk_push_literal() API call which may allow e.g. better
+  lowmem behavior for string literals later on (GH-1805)
+
 * ES2015 Number constructor properties: EPSILON, MIN_SAFE_INTEGER,
   MAX_SAFE_INTEGER, isFinite(), isInteger(), isNaN(), isSafeInteger(),
   parseInt(), parseFloat() (GH-1756, GH-1761)
--- a/doc/release-checklist.rst
+++ b/doc/release-checklist.rst
@@ -32,6 +32,9 @@ Checklist for ordinary releases
 
   - Verify by running Duktape cmdline and evaluating ``Duktape.version``
 
+* Check for API calls and config options tagged experimental to see if they
+  should lose their experimental status
+
 * Check dist-files/README.rst
 
   - Update release specific release notes link
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -6366,7 +6366,7 @@ DUK_INTERNAL void duk_push_lightfunc_name_raw(duk_hthread *thr, duk_c_function f
 
 	DUK_ASSERT_API_ENTRY(thr);
 
-	duk_push_sprintf(thr, "light_");
+	duk_push_literal(thr, "light_");
 	duk_push_string_funcptr(thr, (duk_uint8_t *) &func, sizeof(func));
 	duk_push_sprintf(thr, "_%04x", (unsigned int) lf_flags);
 	duk_concat(thr, 3);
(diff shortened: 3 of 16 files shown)


svaarala:
One overall motivation for this addition is to see if it's possible to speed up property accesses made via C literals. Currently applications can either use literals (which get treated the same as dynamically created strings) or use borrowed heap pointers which is more error prone. It'd be nice to have a solution which improved common API idioms without being error prone.

svaarala:
Did a little experiment on automatically pinning `duk_hstring`s when they are intern checked as literals, and then maintaining a fixed size cache to map a literal address to a cached `duk_hstring *` (which is now guaranteed to be stable and never to be freed while the heap is in use). This allows a fast path in `duk_push_literal()` which bypasses the string table lookup.

Outside of lowmem environments it's often not unreasonable to pin literals appearing in application code, at least if they're interned at least once as literals, because there's a limited number of such literals and they're likely to be actually in use.

With this approach `duk_push_literal()` is about 3x faster than `duk_push_lstring()`, so if helpers like `duk_get_prop_literal()` etc were added, object property manipulation code could be improved a bit towards the performance of using borrowed references and e.g. `duk_get_prop_heapptr()`.

More on this later.

svaarala:
Experimenting a bit more, operating on an object with ten distinct properties in a repeated test loop (best of 5, time in seconds, lower better):
<table>
<tr>
<td>duk_{get,put}_prop_string()</td><td>11.0</td>
</tr>
<tr>
<td>duk_{get,put}_prop_lstring()</td><td>10.1</td>
</tr>
<tr>
<td>duk_{get,put}_prop_literal()</td><td>8.0</td>
</tr>
<tr>
<td>duk_{get,put}_prop_heapptr()</td><td>7.9</td>
</tr>
</table>

Plain duk_push_literal(), duk_push_lstring(), etc:
<table>
<tr>
<td>duk_push_string()</td><td>1.8</td>
</tr>
<tr>
<td>duk_push_lstring()</td><td>1.4</td>
</tr>
<tr>
<td>duk_push_literal()</td><td>0.5</td>
</tr>
<tr>
<td>duk_push_heapptr()</td><td>0.4</td>
</tr>
</table>

So, it seems pinning and caching encountered literals (signaled by the application by using e.g. `duk_get_prop_literal()`) gets pretty close to borrowed heapptrs, with none of the risks.

fatcerberus:
> pretty close to borrowed heapptrs, with none of the risks.

Well, assuming nobody does `duk_push_literal(ctx, stringbuf);` anyway. :)

svaarala:
I wonder if there's an easy way to force that to cause a compilation error. The macro does a `sizeof(arg)` but that can also work for e.g. a buffer pointer. The outcome will be plainly broken though, not silently broken like errors in borrowed pointer handling.

Assertions could catch at least one aspect of this: if the argument is `ptr` and `sizeof(ptr) - 1` is `len`, then `ptr[len]` must be 0. This would trigger quite easily if the argument was incorrect.

svaarala:
One perhaps simple way to catch an invalid argument would be to do compile time concatenation:
```
#define duk_push_literal(ctx,str) \
    duk_push_literal_raw((ctx), str "", sizeof(str) - 1)
```

This works for `duk_push_literal(ctx, "foo")`, fails for `duk_push_literal(ctx, ("foo")`, and most importantly fails for things like `duk_push_literal(ctx, myptr)`.

Unfortunately it fails for parenthesized literals, e.g. DUK_HIDDEN_SYMBOL().

andoma:
There's `__builtin_constant_p()` in gcc and clang which can be used for something like:

```
#define duk_push_something(ctx, str) do { \
  if(__builtin_constant_p(str)) { \
      duk_push_literal(ctx, str);
} else {
      duk_push_string(ctx, str);
}
```
No sure if it could be wired up to trig a compile time error though? Perhaps there's something related.

svaarala:
@andoma The macro in your example would actually be quite a nice way to automatically convert call sites to use better performing literals. Something to think about :)

PR #1808, opened 2017-11-14
 Replace gcc by $(CC) in extras Makefiles

ffontaine:
All Makefiles in extra folder were using gcc instead of $(CC), fix this
to allow cross-compilation


svaarala:
Thanks @ffontaine! There is a slight inconsistency in the spacing/whitespace, I'll fix that separately.

svaarala:
(To be clear, the inconsistencies existed already before your pulls.)

PR #1810, opened 2017-11-15
Add duk_{get,put,has,del}_prop_literal() and duk_{get,put}_global_{literal,heapptr}()

svaarala:
- [x] Add duk_{get,put,has,del}_prop_literal() API calls
- [x] Add duk_{get,put}_global_literal() API calls
- [x] Add duk_{get,put}_global_heapptr() API calls
- [x] NUL assert for duk_push_literal(), catches some API abuse
- [x] Proper ifdef's for duk_push_literal_raw() etc
- [x] Size optimized variant: map to xxx_lstring()
- [x] Testcase coverage for API calls
- [x] Update test-all-public-symbols.c test
- [x] Test size-optimized variant manually
- [x] API documentation for API calls
- [x] Migration notes
- [x] Releases entry

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3195,8 +3195,15 @@ Planned
   useful to detect constructor calls which can already be done using
   duk_is_constructor_call() (GH-1745)
 
-* Add experimental duk_push_literal() API call which may allow e.g. better
-  lowmem behavior for string literals later on (GH-1805)
+* Add experimental API call variants for plain C literals, for example
+  duk_push_literal(ctx, "key") and duk_get_prop_literal(ctx, "propname"),
+  which may allow e.g. better performance or RAM footprint later on; calls
+  added: duk_push_literal(), duk_get_prop_literal(), duk_put_prop_literal(),
+  duk_has_prop_literal(), duk_del_prop_literal(), duk_get_global_literal(),
+  duk_put_global_literal() (GH-1805)
+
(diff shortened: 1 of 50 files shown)


svaarala:
@fatcerberus Managed to add at least some assert coverage for the case where a duk_push_literal() argument is a non-literal string. For example:

```c
char *mystr = "foobarquux";
duk_push_literal(thr, mystr);
```

compiles, but causes the following assertion error:
```
assertion failed: str[len] == (char) 0 (duk_api_stack.c:4362)
*** FATAL ERROR: assertion failed: str[len] == (char) 0 (duk_api_stack.c:4362)
```

This happens because sizeof(mystr) will be the size of the pointer (8 here) and the 9th character isn't be a NUL. It's far from perfect but at least something that will probably trigger if one abuses the API call in a lot of places.

svaarala:
Rejecting parenthesized literals would allow a pretty good check based on appending an empty string to work:
```
duk_push_literal(ctx, "foo" "bar");
```

would translate to:
```
duk_push_literal_raw(ctx, "foo" "bar" "", sizeof("foo" "bar" "") - 1)
```

and if given a pointer argument:

```
duk_push_literal(ctx, ptr);
```

it would translate to:
```
duk_push_literal_raw(ctx, ptr "", sizeof(ptr "") - 1)
```

which would fail compilation. Parenthesis support for literals is probably not very important, but the existing macros for symbols would need to be modified because they are parenthesized literals right now.

PR #1811, opened 2017-11-15
Makefile.sharedlibrary: mkdir in install target

ffontaine:
Headers are copied in $(INSTALL_PREFIX)/include so mkdir this directory
otherwise cp will fail
Same thing for libraries and $(INSTALL_PREFIX)/lib

### DIFF
--- a/dist-files/Makefile.sharedlibrary
+++ b/dist-files/Makefile.sharedlibrary
@@ -56,6 +56,7 @@ libduktaped.so.$(REAL_VERSION):
 # Symlinks depend on platform conventions.
 .PHONY: install
 install: libduktape.so.$(REAL_VERSION) libduktaped.so.$(REAL_VERSION)
+	mkdir -p $(INSTALL_PREFIX)/lib/
 	cp $+ $(INSTALL_PREFIX)/lib/
 	rm -f $(INSTALL_PREFIX)/lib/libduktape.so $(INSTALL_PREFIX)/lib/libduktape.so.$(SONAME_VERSION)
 	ln -s libduktape.so.$(REAL_VERSION) $(INSTALL_PREFIX)/lib/libduktape.so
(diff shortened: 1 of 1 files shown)


svaarala:
If one creates the directory on install, the owner/permission settings will most likely be incorrect for a system install?

fatcerberus:
It's been my experience with makefiles that you should `mkdir` all directories because many packaging tools build the source in a sandbox and will therefore fail without it.  If you're installing to a normal location like /usr then stuff like /usr/lib, /usr/bin are pretty much guaranteed to exist so permissions aren't an issue.

svaarala:
One may also be installing to a sandbox directory which then gets packages to be unpacked (e.g. when creating installable packages). It's been a long time since I last did that, but I think owners/permissions matter in that case. Maybe the tooling nowadays handles that automatically.

ffontaine:
Yes, @fatcerberus is right, this patch is needed to install duktape within buildroot which build and install the target in a staging and target directory. The target directory does not have any include directory and as a result the install fails.

I choose to mkdir the directory but an other option would be to call "install -m 0644 -D" instead of "cp" to install the headers and libraries (and set the permissions). If you prefer this option, I can update my PR.

svaarala:
@ffontaine I'll merge this if you don't think there's a downside to the mkdir in various use cases.

I think this illustrates again my point why I don't really like maintaining these Makefiles for actual use beyond examples. To do that consistently, one would need to know what targets to support (Linux ,deb building? Docker? Cygwin? Command prompt? OSX? Amiga?) and to establish some sort of tests, to detect when changes for one environment breaks another environment. And even if one did that, it won't cover the needs of even 90% of users; to do that, one would need to support SCons, various IDE project files, etc. That effort, whatever the level, is away from doing something more productive.

I can keep on merging patches though, but at some point someone else will want a change that contradicts previous patches, and it'll be hard to detect that conflict without even defining a compatibility target :)

svaarala:
@ffontaine Just to be clear, I appreciate the pull request -- my frustration in maintaining these Makefiles is mostly about not knowing what changes are correct because there isn't any well-understood supported targets for the Makefiles (they're just examples to look at and maybe copy and modify from).

Some ways of going forward:
- Keep current approach: makefiles are examples, no defined target to be supported. Maybe that needs to be communicated more effectively somehow.
- Keep current approach, but make the Makefiles less prominent (e.g. move to Wiki articles) so that it's clear one is supposed to copy-and-modify rather than use as is.
- Define some set of targets and use cases to actually be supported and start maintaining towards that goal. One definite upside of this would be that it'd be great to have an official way to e.g. package .deb files right off the repo. But, there's some effort in selecting the targets and use cases, and to establish test coverage.

I've been wanting to do something about the packaging for a while now. It's not really nice to have Makefiles here and there for each extra and example, and repeating the same idioms in them. But there are more interesting things to do so I've postponed that for a while now ... :)

svaarala:
Another feasible approach would be to maintain specific build files and packaging as separate repos. For example, make a repo that creates .deb packaging based on the distributable. This would have the upside of the main repo not seemingly favoring a specific target, which is nice for a project targeting a very wide variety of targets.

Issue #1806, opened 2017-11-14
Follow-ups for duk_push_literal() API call


svaarala:
- [ ] Add xxx_literal() variants for e.g. duk_get_prop_string().
- [ ] Experiment with caching for literal string table lookups which would improve application performance without needing to resort to heap pointers. For example, a soft literal address -> string table index cache where the index wouldn't be trusted but would be validated as cheaply as possible before use.
- [ ] When using external strings for lowmem targets, check if literals could be automatically made external cheaply

PR #1813, opened 2017-11-16
Initial implementation of C API literal pinning and literal intern cache

svaarala:
- [x] Automatically pin the `duk_hstring`s looked up using C API literals when they're encountered.
- [x] Add a C literal address -> `duk_hstring` cache ("litcache") to look up the `duk_hstring` matching a certain literal address quickly. Because the target `duk_hstring` is pinned until heap destruction, the cache doesn't need an invalidation mechanism.
- [x] Some more internal call site conversion to use duk_xxx_literal() variants.
- [x] Portability / warnings for litcache key computation.
- [x] Make pinning+litcache configurable.
- [x] ROM string handling.
- [x] API documentation updates.
- [x] 2.3 migration note updates.
- [x] Releases entry.

Overall the pinning+litcache makes e.g. `duk_get_prop_literal()` about as fast as `duk_get_prop_heapptr()` in most cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -272,6 +272,7 @@ clean:
 	@rm -f dukweb.js
 	@rm -rf /tmp/dukweb-test/
 	@rm -f massif-*.out
+	@rm -f literal_intern_test
 
 .PHONY: cleanall
 cleanall: clean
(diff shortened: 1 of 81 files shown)


svaarala:
With this pull once literals get pinned, they won't be unpinned and will only be freed by heap destruction; this is done because this avoids any need for an invalidation mechanism. This is typically not a huge downside because only literals actually occurring in source code, and which also get used in actual duk_xxx_literal() calls get pinned.

However, it'd be easy to perform a full flush of the literal cache on mark-and-sweep. This would unpin the strings and invalidate all cache entries. There are no borrowed pointers in play once the cache has been flushed, so mark-and-sweep would then be able to collect the strings if they're no longer referenced. Mark-and-sweep is rare enough to make the flush/invalidation cheap enough not to matter because it won't affect individual string lookups.

Added this to the #1806 backlog issue.

fatcerberus:
I don't think the permanent pinning should be an issue, any literal strings actually used by a program will remain in memory regardless of anything else you do (because they are part of the compiled code).

svaarala:
Right, the C literals will be in the read-only section, but that doesn't have to mean that a `duk_hstring` is allocated from RAM for each C literal. But, that's not an issue except for literals that are used very rarely.

PR #1823, opened 2017-12-27
Add ES2015 @@isConcatSpreadable support, Array .concat() and .map() trailer change

svaarala:
Add support for ES2015 `@@isConcatSpreadable` in Array.prototype.concat().

Also remove the config options for non-standard trailing element behavior in Array.prototype.concat() and Array.prototype.map(). For .concat() the non-standard behavior matches real world behavior of other engines, and the specification "bug" in ES5.1 was fixed in ES2015 so the non-standard behavior is now standard in ES2015. For .map() the config option was based on a misreading of the ES5 specification: ES5.0/ES5.1 already include trailing gaps in the result length because the array is created to the input size before processing the elements.

Tasks:
- [x] `@@isConcatSpreadable` support
- [x] Invoke Proxy `has` trap when element to be concatenated is a Proxy object
- [x] Remove config option `DUK_USE_NONSTD_ARRAY_CONCAT_TRAILER`
- [x] Remove config option `DUK_USE_NONSTD_ARRAY_MAP_TRAILER`
- [x] Resolve FIXMEs
- [x] Testcase update
- [x] API doc clarification for duk_concat() re: `@@isConcatSpreadable` behavior
- [x] Releases entry


fatcerberus:
> The non-standard behavior matches real world behavior of other engines, and the specification "bug" in ES5.1 was fixed in ES2015 so the options are no longer necessary.

Does this mean that the "non-standard" behavior is, in actuality, standard?

svaarala:
It's non-standard in ES5.1, standard in ES2015.

svaarala:
For Array.prototype.map() the `DUK_USE_NONSTD_ARRAY_MAP_TRAILER` was added to match real world behavior of other engines. However, it seems ES5.0 already has the "real world" behavior included in the algorithm: the final result length is 1:1 with the input because the result array is created based on the input length. The subsequent index writes don't affect the length, and trailing gaps don't thus cause the result length to be shorter either. So the option was in fact just fixing a Duktape bug.

PR #1836, opened 2018-01-24
- es6 promise boilerplates

wenq1:
As name


svaarala:
Hi @wenq1, I appreciate the pull, but I don't think it makes sense to merge this right now. It conflicts with similar work I've been doing on a private branch. I'd be very happy if you wanted to work on Promise support though! My plan is roughly as follows:
- Figure out a best guess for Promise representation. It could either be a plain object or a `duk_hpromise` with specific reference counted fields for a bit more compact representation. I'm doing this work in the private branch right now. (I'm also writing a dummy pure JS Promise implementation, not intended for production use, just to ensure I've understood all the details correctly.)
- Add a placeholder job queue mechanism to allow testing Promise features. This is not intended to remain in the end, but just something to allow test cases to be written etc.
- Implement individual Promise methods, one method per pull request. Each pull should provide both the implementation and testcases with good coverage for that particular method. Tests should cover every specification algorithm step.
- Once all methods are implemented, decide on the public job queue API and remove the placeholder.

Right now I'm experimenting with different representations to see what works best as an initial implementation. The representation will be "hidden" from application point of view so that it can be versioned later on, but it still makes sense to do this step carefully because changing the representation requires tedious work later on.

Once the representation and a few test methods are in place, it should be possible to implement the Promise constructor and prototype methods one by one. There are actually placeholders for all the methods in master already, they're just disabled by default (see duk_bi_promise.c).

If you'd be interested in working on Promise support, this would be the ideal place, i.e. to implement the Promise methods with good test coverage. Also adding some performance and memory usage metrics would be nice, because performance/memory is the main reason for having a built-in Promise implementation as opposed to an external one.

Right now I'm on vacation but once I'm back I'll try to work out the initial representation and push that to the repo. Sorry for the delay :-)

svaarala:
Re: the placeholder job queue, the only thing that's really needed is for `duk` to be able to execute Promise job queue entries. To that end it should be enough to:
- Add a placeholder binding like `Duktape.runPromiseJob()` which would execute one Promise job queue entry and return true/false to indicate if more entries remain.
- Modify `duk` (temporarily) to call `Duktape.runPromiseJob()` until it returns false, after the argument script files have been evaluated. In interactive mode the job queue would be executed until empty after each eval input.

These changes would be enough to allow testcases to be written with no hacks in the test cases themselves, and to allow interactive testing of Promises.

This mechanism would be ripped out later in favor of a public C API, with changes in `duk` to allow both testcase and interactive Promise use but with the final job queue API mechanism.

I'll push this mechanism along with the representation as soon as I have time to finalize them.

wenq1:
Enjoy your vacation. I just want to give this long-overdue feature a shot and other commits are on the way.

I do not think I'll give too much consideration to standard compliance or academic excellence for now, as people will leave DUK_USE_PROMISE_BUILTIN turned off anyway and the "master branch is supposed to be broken". Performance wise I won't expect too much from a 300k alternative to v8 or ChakraCore.

That said, I just want to get something working first, so I'll keep on sending pull request and feel free to merge anything you deem useful.

svaarala:
I'm sorry but I don't want to merge something that duplicates work I've already done. But I'd be very grateful for help once the representation is in master.

wenq1:
No worries. I'll wait on.

svaarala:
> Performance wise I won't expect too much from a 300k alternative to v8 or ChakraCore.

That's not what I mean by performance/memory considerations. The main point of comparison is a pure JS Promise library which already works quite well. The main motivation of doing a built-in implementation is that it performs better than an external Promise library (executed with Duktape). There are secondary considerations too but that's the main one IMO.

wenq1:
I'll ask you more Qs regarding how to run tests later...

The development process seems a bit tedious to me right now. Do you really have to run dist.py each time (and wait for 2 mins) and then compile a helloworld to test a one-liner change?

I think for development one should be able to compile from original source, not the amalgamated one...

svaarala:
If you edit src-input/*.c, you don't need to run dist.py (unless you want to create an end user distributable).

But you do need to run `python tools/configure.py` and then compile the result. That also takes some time, but a lot less than the full dist.

wenq1:
> That's not what I mean by performance/memory considerations. The main point of comparison is a pure JS Promise library which already works quite well. The main motivation of doing a built-in implementation is that it performs better than an external Promise library (executed with Duktape). There are secondary considerations too but that's the main one IMO.

Yet again I'm not trying to be academic here. I do appreciate your endeavor to excellence. However, the existence of nodejs's native promise does not seem to be affected by its inferior performance compared to Bluebird :-)

Reason I would like to have Promise in place asap is that we could then proceed to native async/await implementation, not exactly as a more performant replacement for JS shims.

svaarala:
Async/await is not a short term work item in any case - it will require a rework of the compiler intermediate representation. So there's no reason to rush a Promise implementation that will need to be rewritten because it performs poorly on low memory targets for example.

wenq1:
I don't think you get my point. I'll happily close the pull anyway.

svaarala:
Ok, I'll keep you posted re: the representation and placeholder job queue, if you want to work on the Promise methods later on.

PR #1846, opened 2018-02-12
Fix unlinkSync call in runtests.js

boneskull:
I noticed that the call to `unlink` on L47 was actually asynchronous--you want `unlinkSync`.   I noticed this because with newer versions of Node.js (8 or newer), calling an async function in a core module without the callback results in a deprecation notice.  Well, more than one...

```
Executing 1122 testcase(s) with 1 engine(s) using 4 thread(s), total 1085 task(s)
(node:2027) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
(node:2027) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
(node:2027) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
(node:2027) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
... etc ...
```

This PR eliminates the torrent of warnings.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -49,6 +49,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Jakub Nowakowski (https://github.com/jimvonmoon)
 * Tommy Nguyen (https://github.com/tn0502)
 * Fabrice Fontaine (https://github.com/ffontaine)
+* Christopher Hiller (https://github.com/boneskull)
 
 Other contributions
 ===================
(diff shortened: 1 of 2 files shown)


svaarala:
Thanks! Please update AUTHORS.rst (first list in the file), see CONTRIBUTING.

boneskull:
updated `AUTHORS.rst`

svaarala:
Thanks!

PR #1856, opened 2018-03-22
Add some Promise constructor testcases

svaarala:
- [x] Add --polyfill and --ignore-skip to util/runtest.py to make it easier to test against the polyfill
- [x] Minimal polyfill fixes based on testcases
- [x] Add a few (disabled) Promise constructor tests

### DIFF
--- a/polyfills/promise.js
+++ b/polyfills/promise.js
@@ -37,7 +37,7 @@
     // internal slot which isn't affected by Proxy behaviors etc.
     var symMarker = Symbol('promise');
     function isPromise(p) { return p !== null && typeof p === 'object' && symMarker in p; }
-    function requirePromise(p) { if (!isPromise(p)) { throw new Error('Promise required'); } }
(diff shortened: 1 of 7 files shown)


polyfills/promise.js
```diff
@@ -127,7 +127,8 @@
 
     // %Promise% constructor.
     var cons = function Promise(executor) {
-        if (!new.target) { throw new Error('Promise must be called as a constructor'); }
+        if (!new.target) { throw new TypeError('Promise must be called as a constructor'); }
```
fatcerberus: Duktape supports `new.target` now?

svaarala: Yes, it was added in 2.2.0 release.

It's quite limited now though: it's always either undefined, or the current constructor function (if in a constructor call) because NewTarget and class support is not yet implemented. It's still useful to check for non-constructor calls.

PR #1875, opened 2018-04-18
Avoid hard Symbol dependency for Promise polyfill

### LINKED ISSUE #1874: Document that the promise.js polyfill depends requires DUK_USE_SYMBOL_BUILTIN
The Promise polyfill is very usefull, many thanks for that! As it uses "Symbol" internally, it would be great to specify in the documentation that configuring duktape with DUK_USE_SYMBOL_BUILTIN is a necessary step (e.g. directly in the file header?).

svaarala:
Fixes #1874.

### DIFF
--- a/polyfills/promise.js
+++ b/polyfills/promise.js
@@ -15,6 +15,9 @@
  *  ES2015 specification algorithms, before implementing Promises natively.
  *  As such this polyfill may not be production quality.
  *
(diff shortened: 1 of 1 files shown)


svaarala:
@bwalter If you could give this a test drive, it'd be nice to hear if it worked for you.

svaarala:
Good to hear, thanks! :)

PR #1877, opened 2018-04-19
Promise polyfill comment trivia

svaarala:
Some polyfill comment trivia.

### DIFF
--- a/polyfills/promise.js
+++ b/polyfills/promise.js
@@ -1,7 +1,7 @@
 /*
  *  Minimal ES2015+ Promise polyfill
  *
- *  Limitations:
+ *  Limitations (also see XXX in source):
  *
  *    - Caller must manually call Promise.runQueue() to process pending jobs.
  *    - No Promise subclassing or non-subclass foreign Promises yet.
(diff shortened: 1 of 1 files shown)


svaarala:
@fatcerberus Re flexibility: Promises are actually way more flexible than almost anything else in the specification in my view. They support proper subclassing, thenables, etc. In my view they're a little bit too flexible, e.g. the generic thenable support means that any resolve value which happens to have a callable `.then` property magically alters the resolution process. This means that if one e.g. has a serialization format which does support callable functions in some fashion, serialized values may accidentally trigger thenable behaviors so one needs to wrap all such values in yet another object to avoid it.

It also means that `Object.prototype.then = Math.cos;` is a nice way to break everything.

Presumably the thenable support is present to allow mixing of Promise implementations - but maybe a better approach would be to accept that the transition is a bit messier but the core specification is clean.

fatcerberus:
I think part of the problem is that the original Promises/A+ specification that ES6 was based on was very deliberately written such that anything with a callable `.then` is a promise *by definition* (in the true spirit of duck typing).  When the TC39 adopted it they don’t seem to have changed that aspect, and now we’re stuck with it. :-/

PR #1884, opened 2018-04-24
Cleanup examples/eventloop/c_eventloop.c a bit more

gonzus:
Made it easier to turn debugging on/off with a single macro, and added a header file to make sure compilation is consistent everywhere.

### DIFF
--- a/examples/eventloop/c_eventloop.c
+++ b/examples/eventloop/c_eventloop.c
@@ -16,6 +16,10 @@
 #include "duktape.h"
 #include "c_eventloop.h"
 
+#ifndef DUKTAPE_EVENTLOOP_DEBUG
+#define DUKTAPE_EVENTLOOP_DEBUG 0       /* set to 1 to debug with printf */
+#endif
+
 #define  TIMERS_SLOT_NAME       "eventTimers"
 #define  MIN_DELAY              1.0
 #define  MIN_WAIT               1.0
@@ -33,7 +37,7 @@ typedef struct {
 	int removed;      /* timer has been requested for removal */
 
 	/* The callback associated with the timer is held in the "global stash",
-	 * in <stash>.TIMERS_SLOT_NAME[String(id)].  The references must be deleted
+	 * in <stash>.eventTimers[String(id)].  The references must be deleted
 	 * when a timer struct is deleted.
 	 */
 } ev_timer;
@@ -127,7 +131,7 @@ static void expire_timers(duk_context *ctx) {
 		 */
 
 		if (exit_requested) {
-#if 0
+#if DUKTAPE_EVENTLOOP_DEBUG > 0
 			fprintf(stderr, "exit requested, exiting timer expiry loop\n");
 			fflush(stderr);
 #endif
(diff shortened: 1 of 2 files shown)


examples/eventloop/c_eventloop.c
```diff
@@ -14,6 +14,11 @@
 #include <poll.h>
 
 #include "duktape.h"
+#include "c_eventloop.h"
+
+#ifndef DUKTAPE_EVENTLOOP_DEBUG
```
svaarala: By convention `#if !defined(...)` is used throughout the code base.

examples/eventloop/c_eventloop.c
```diff
@@ -32,7 +37,7 @@ typedef struct {
 	int removed;      /* timer has been requested for removal */
 
 	/* The callback associated with the timer is held in the "global stash",
-	 * in <stash>.TIMERS_SLOT_NAME[String(id)].  The references must be deleted
+	 * in <stash>.eventTimers[String(id)].  The references must be deleted
```
svaarala: Why is this changed back?

examples/eventloop/c_eventloop.c
gonzus: Will change that and resubmit. Should I somehow cancel this current pull request?

examples/eventloop/c_eventloop.c
gonzus: Only because it seemed more consistent with the rest of the comments in the code.

svaarala: Ok, just wondering because it was changed to `TIMERS_SLOT_NAME` in the previous pull.

examples/eventloop/c_eventloop.c
svaarala: No need, just squash the commits (e.g. git rebase -i master).

examples/eventloop/c_eventloop.c
gonzus: Right. Again, only because there are many other comments that reference eventTimers, and only this one was using TIMERS_SLOT_NAME.

gonzus: (which means, my previous change was probably a mistake...)

examples/eventloop/c_eventloop.h
```diff
@@ -0,0 +1,7 @@
+#ifndef C_EVENTLOOP_H
```
svaarala: This one is causing the Travis error.

gonzus: Doh. Fixed.

svaarala:
Could you please squash the commits so that "fixups" are not separate commits. The result should be either one commit with all the changes, or more than one commits with no following fixups.

gonzus:
Damn. I probably made a mistake there, got a merge commit in between. Let me try to clean that up.

gonzus:
I *think* it is fixed now.

svaarala:
The commit messages are overly long, please see https://github.com/svaarala/duktape/blob/master/doc/git-conventions.rst#commit-messages.

gonzus:
Fixed

PR #1890, opened 2018-04-26
Add console flag DUK_CONSOLE_TO_STDERR

gonzus:
We introduce this flag to allow some levels of console output to be
directed to stderr instead of stdout.  We use the flag for methods warn,
error and exception.


gonzus:
This would be my first *real* contribution, something useful. Cheers!

svaarala:
So to clarify, the change in this pull has the effect that:
- By default info and below go to `stdout`, warn and above go to `stderr`.
- If the `DUK_CONSOLE_TO_STDERR` flag is given, everything goes to `stderr`.
- There's no possibility of sending everything to `stdout`?

extras/console/duk_console.c
```diff
@@ -25,12 +21,10 @@
...
 
 static duk_ret_t duk__console_log_helper(duk_context *ctx, const char *error_name) {
-	duk_idx_t i, n;
-	duk_uint_t flags;
-
-	flags = (duk_uint_t) duk_get_current_magic(ctx);
-
-	n = duk_get_top(ctx);
+	duk_uint_t flags = (duk_uint_t) duk_get_current_magic(ctx);
+	FILE* output = (flags & DUK_CONSOLE_TO_STDERR) ? stderr : stdout;
```
svaarala: By convention `FILE* output` -> `FILE *output`.

extras/console/duk_console.c
```diff
@@ -133,6 +127,10 @@ void duk_console_init(duk_context *ctx, duk_uint_t flags) {
 	duk__console_reg_vararg_func(ctx, duk__console_log, "debug", flags);  /* alias to console.log */
 	duk__console_reg_vararg_func(ctx, duk__console_trace, "trace", flags);
 	duk__console_reg_vararg_func(ctx, duk__console_info, "info", flags);
+
+    /* from here on, things go to stderr by default */
+    flags |= DUK_CONSOLE_TO_STDERR;
```
svaarala: Indent broken here (use tab(s)).

fatcerberus:
@svaarala Skimming over the diff it looks like everything goes to stdout by default; the flag just makes `warn` and above go to stderr.  `info` and below still seem to go to stdout always.

gonzus:
Correct.

Is sending everything to stdout something that should be supported as well?

svaarala:
@fatcerberus The `DUK_CONSOLE_TO_STDERR` flag is forced for warning and above. But it is up to the user call site what the value is prior to that (i.e. for info and below).

gonzus:
Right. If you call duk_console_init() with flags including DUK_CONSOLE_TO_STDERR, all output will go to stderr.

svaarala:
@gonzus I was just first trying to state what concrete change this pull is trying to make :)

I'm not really what the best behavior is. For many applications writing logs to stderr only at least has value, e.g. if they work on the command line and stdin/stdout is reserved for whatever input and output is being processed on the command line). Not so sure about mixed stdout/stderr output or stdout-only output.

But if all three are potentially useful, then using two flags:
- `DUK_CONSOLE_STDOUT_ONLY`
- `DUK_CONSOLE_STDERR_ONLY`

and defaulting to Node.js-like mixed output would achieve all three if that makes sense.

svaarala:
In any case `DUK_CONSOLE_TO_STDERR` is slightly misleading, because as things stand, some output is always written to `stderr`, so maybe the flag should be called `DUK_CONSOLE_STDERR_ONLY` or similar, if no other change is made.

svaarala:
For comparison, the default, and only, behavior of the print/alert extra (https://github.com/svaarala/duktape/blob/master/extras/print-alert/duk_print_alert.c) is:
- Print goes to stdout
- Alert goes to stderr

gonzus:
Let me see if I follow what you said. You suggest these flags:
- `DUK_CONSOLE_STDOUT_ONLY`: all output goes to stdout.
- `DUK_CONSOLE_STDERR_ONLY`: all output goes to stderr.
- default, no flags: warn and above go to stderr, info and below go to stdout.

Is that correct?

gonzus:
Modified according to the discussion. Ready to fix anything you may find.

svaarala:
Thanks! I'll take a look tonight when I get a chance :)

gonzus:
Just curious, because github keeps telling me "Changes requested", but I think I already took care of all of those. Do you have any other suggestions for this issue? Cheers!

svaarala:
Busy, not forgotten.

PR #1893, opened 2018-04-28
Test for Number.prototype ES5.1/ES2016+ number nature

svaarala:
Also add a comment that Number.prototype is a number itself, except in ES2015. No change in behavior.


svaarala:
No releases entry because this has no behavioral change.

PR #1904, opened 2018-05-12
Reject constructor calls to Promise resolve/reject functions

svaarala:
The specifications seem pretty vague whether resolve/reject functions should reject constructor calls, but in practice engines (at least Node.js and Firefox) seem to do so, so change the Promise polyfill to match that behavior.

### DIFF
--- a/polyfills/promise.js
+++ b/polyfills/promise.js
@@ -121,12 +121,14 @@
         // with a [[AlreadyResolved]] slot.  Here we use an in-scope variable.
         var alreadyResolved = false;
         var reject = function (err) {
+            if (new.target) { throw new TypeError('reject is not constructable'); }
             if (alreadyResolved) { return; }
             alreadyResolved = true;  // neutralize resolve/reject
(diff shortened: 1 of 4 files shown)


fatcerberus:
ChakraCore also rejects it.

```js
let handler;
let prom = new Promise((resolve, reject) => {
        handler = { resolve, reject };
});
new handler.resolve(812);
```

Output:
```
miniSphere 5.2.5 JS game engine (x64)
a lightweight JavaScript-powered game engine
(c) 2015-2018 Fat Cerberus

GAME CRASH: unhandled JavaScript exception.
TypeError: Function is not a constructor
   at Global code (@/test.js:5:1)
```

fatcerberus:
Weird that the specification doesn't seem to say whether this should be an error or not.

I even checked the latest drafts, the resolve/reject function is now created using the `CreateBuiltInFunction` abstract operation (https://tc39.github.io/ecma262/#sec-createbuiltinfunction) which says nothing about constructibility.

PR #1909, opened 2018-05-22
Disable tracebacks crash

michalkas:
Prerequisites:
#undef DUK_USE_TRACEBACKS and 'depth' >1
 
During the second loop iteration the exception is thrown:
- Read access violation while executing => 'act = act->parent'

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -51,6 +51,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Fabrice Fontaine (https://github.com/ffontaine)
 * Christopher Hiller (https://github.com/boneskull)
 * Gonzalo Diethelm (https://github.com/gonzus)
+* Michal Kasperek (https://github.com/michalkas)
 
 Other contributions
 ===================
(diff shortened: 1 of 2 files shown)


svaarala:
The 'act' NULLing is intentional because it's potentially invalid at that point. I'll check out the code to see what the underlying issue is.

michalkas:
In my situation  I had depth == 2, the 'act' was NULLed but the loop was continued after NULLing here:
```
duk_get_prop_stridx_short(thr, -1, DUK_STRIDX_FILE_NAME);
if (!duk_is_string_notsymbol(thr, -1)) {
duk_pop_2(thr);
continue;
}
```

svaarala:
Thanks, I'll check it out tonight when I have a chance.

svaarala:
@michalkas The fix seems correct after all. Activation pointers are stable in master, so hanging on to them is OK.

Thanks!

svaarala:
Labeled `security` because causes a crash with non-default options.

svaarala:
It depends on the platform. Some platforms have no paging.

svaarala:
But yes, it's unlikely to be exploitable, it's just a general policy to use `security` whenever there's a reproducible crash because it's quite difficult to assess exploitability accurately.

fatcerberus:
Makes sense, thanks for clarifying :)

PR #1915, opened 2018-05-28
Add duk_fatal_exception, used for C++ fatal/uncaught errors

svaarala:
With C++ exceptions Duktape currently uses a single exception type, `duk_internal_exception` as a setjmp/longjmp replacement. It's intended to only be seen internally (i.e. the catch site is always internal to Duktape). However, it does propagate to calling code (at least up to Duktape 2.2) in uncaught error situations. Fatal errors caused by e.g. assertion errors or manual calls to duk_fatal() also don't use the C++ exception mechanism right now.

Make a few changes to make fatal/uncaught error handling more C++ native:
- Add a separate `duk_fatal_exception` class which is used for uncaught errors and other fatal errors in C++ exception mode. However, if a user provides an explicit fatal error handler, it will be used instead.
- Change the longjmp code to detect a missing catch point in C++ mode too, and use `duk_fatal()` in C++ mode too. Change internal C++ catch sites to handle `duk_fatal_exception` by rethrowing it.
- Change the default fatal error to support `duk_fatal_exception`.
- Don't warn about missing `DUK_USE_FATAL_HANDLER` with C++ exception support enabled

Tasks:
- [x] Code changes
- [x] Remove warning for missing `DUK_USE_FATAL_HANDLER` if `DUK_USE_CPP_EXCEPTIONS` defined
- [x] Inheritance for `duk_fatal_exception` - std::exception, std::runtime_error, none? => std::runtime_error
- [x] Test -std=xxx for old standard versions
- [x] Test uncaught error
- [x] Test assertion error
- [x] Test generic duk_fatal() call
- [x] 2.3 release note
- [x] Documentation changes => separate pull

@mike-lischke Moving this C++ exception discussion here.

### DIFF
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3233,6 +3233,10 @@ Planned
 2.3.0 (XXXX-XX-XX)
 ------------------
 
+* When DUK_USE_CPP_EXCEPTIONS is enabled: use duk_fatal_exception when
+  propagating a fatal error out of Duktape; use duk_fatal_exception also
+  in the default fatal error handler (GH-1915)
+
 * Update UnicodeData.txt and SpecialCasing.txt used for building internal
   Unicode control data to Unicode version 10.0.0 (GH-1851)
 
--- a/config/header-snippets/platform_conditionalincludes.h.in
+++ b/config/header-snippets/platform_conditionalincludes.h.in
@@ -1,3 +1,4 @@
 #if defined(DUK_F_CPP) && defined(DUK_USE_CPP_EXCEPTIONS)
 #include <exception>  /* std::exception */
+#include <stdexcept>  /* std::runtime_error */
 #endif
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -2710,7 +2710,7 @@ DUK_LOCAL void duk__to_primitive_helper(duk_hthread *thr, duk_idx_t idx, duk_int
 	 * which mimic objects.
 	 */
 	if (check_symbol && duk_get_method_stridx(thr, idx, DUK_STRIDX_WELLKNOWN_SYMBOL_TO_PRIMITIVE)) {
-		DUK_ASSERT(hint >= 0 && hint < sizeof(duk__toprim_hint_strings) / sizeof(const char *));
+		DUK_ASSERT(hint >= 0 && (duk_size_t) hint < sizeof(duk__toprim_hint_strings) / sizeof(const char *));
 		duk_dup(thr, idx);
 		duk_push_string(thr, duk__toprim_hint_strings[hint]);
 		duk_call_method(thr, 1);  /* [ ... method value hint ] -> [ ... res] */
--- a/src-input/duk_error_longjmp.c
+++ b/src-input/duk_error_longjmp.c
@@ -76,14 +76,9 @@ DUK_INTERNAL void duk_err_longjmp(duk_hthread *thr) {
 
 	DUK_DD(DUK_DDPRINT("about to longjmp, pf_prevent_count=%ld", (long) thr->heap->pf_prevent_count));
 
-#if !defined(DUK_USE_CPP_EXCEPTIONS)
 	/* If we don't have a jmpbuf_ptr, there is little we can do except
 	 * cause a fatal error.  The caller's expectation is that we never
 	 * return.
-	 *
-	 * With C++ exceptions we now just propagate an uncaught error
-	 * instead of invoking the fatal error handler.  Because there's
-	 * a dummy jmpbuf for C++ exceptions now, this could be changed.
 	 */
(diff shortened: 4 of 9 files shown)


svaarala:
@mike-lischke Any recommendations on what `duk_fatal_exception` should inherit from? The two common suggestions seem to be:
- std::exception
- std::runtime_error

The latter seems more logical to me, but apparently it has some technical downsides too.

svaarala:
Basically what I would want is:
- Ability to convey an exception message (= same fatal error message as C code gets)
- Ability to catch a specific subclass so that the "unsafe to continue" semantics would be apparent to the call site

The ability to catch the exception with a boilerplate `std::exception` catch is not so obvious. On the one hand it's nice that a boilerplate catch would then print out the error which someone might except. On the other hand the "unsafe to continue" semantics would probably not be observed, so the caller would probably continue, running into further issues.

If the boilerplate catch ability is a bad idea, then I guess `duk_fatal_exception` should not inherit even from std::exception, but be a separate exception with a ::what() interface or similar?

mike-lischke:
> Any recommendations on what duk_fatal_exception should inherit from? The two common suggestions seem to be:
>
>    std::exception
>    std::runtime_error

The latter is just a direct descentant of the first, adding the handling for the error message. We usually prefer `runtime_error` to avoid having to care for message storage. It's purely a matter of taste what you use.

svaarala:
There seems to be some speculation that inheriting from std::runtime_error would have some technical issues when doing multiple inheritance, but I'm not C++ native enough to understand them :-)

But I guess the most important question is should the fatal exception derive from std::exception at all, re: the potential for accidentally catching and continuing execution.

mike-lischke:
> The ability to catch the exception with a boilerplate std::exception catch is not so obvious. On the one hand it's nice that a boilerplate catch would then print out the error which someone might except. On the other hand the "unsafe to continue" semantics would probably not be observed, so the caller would probably continue, running into further issues.

I think that can be left to the duktape consumer. No need for you to go further than signalling the fatal issue. It's the implementor's responsibility to take care properly and the average C++ dev is aware of that, I guess :-)

That's why I think it's good enough to just use `std::runtime_error` as base. I'm not aware of any "technical issues" with that. We build apps for various platforms (including RH, OL, Ubuntu, CentOS, Win, macOS, Fedora etc.) without ever having seen an issue with that exception class.

svaarala:
@mike-lischke Ok, I'll derive from std::runtime_error.

svaarala:
This is what a fatal error looks with the (non-inheriting) exception:
```
terminate called after throwing an instance of 'duk_fatal_exception'
Aborted (core dumped)
```

mike-lischke:
Please inherit from a standard class, otherwise we'd have to use a catch-all clause, which requires an extra round to get the message (which is also only available with C++11 and later).

mike-lischke:
By "non-inheriting" you mean you didn't use a standard base class? I still prefer `runtime_error` :-)

svaarala:
Yes, I was just commenting on what the branch looks right now. I haven't added the inheritance yet.

What would be the most portable way to implement the inheritance? There seem to be a few idioms (C++ version dependent) and std::runtime_error has gained a second constructor in C++11 (?).

svaarala:
With the inheritance from std::runtime_error a fatal error (`duk_fatal(ctx, "aiee...");`) looks like:
```
terminate called after throwing an instance of 'duk_fatal_exception'
  what():  aiee, something went wrong!
Aborted (core dumped)
```

svaarala:
And here's what a failed assertion looks like:
```
terminate called after throwing an instance of 'duk_fatal_exception'
  what():  assertion failed: 0 (duk_bi_math.c:119)
Aborted (core dumped)
```

svaarala:
@mike-lischke The changes should now be in place - I'd appreciate feedback on the changes because I'm not very C++ native. In particular, is the method of declaring the class, throwing, and rethrowing making as minimal assumptions as possible w.r.t. compiler versions/standards.

svaarala:
The multiple inheritance issues are something like this: https://stackoverflow.com/questions/11484691/exception-multiple-inheritance. But I don't think they should be relevant here, because no other code should be constructing `duk_fatal_exception`s or subclassing from it.

svaarala:
@mike-lischke Thanks for the review, I'll try to wrap up the documentation and testing tasks ASAP so this gets into master. For 2.2.2 maintenance release, only the genconfig.py warning change would be merged (the exception and fatal error handling changes are too extensive for a patch release).

svaarala:
Testing some -std=xxx dialects:
- c++11: ok
- c++03: ok (warnings for 64-bit constants, unrelated)
- c++98: ok (warnings for 64-bit constants, unrelated)
- c++14: ok
- c++1z: ok

PR #1922, opened 2018-05-31
Makefile.sharedlibrary: support macOS

apjanke:
What do you think about modifying `Makefile.sharedlibrary` to support macOS out of the box?

(I have only tested this PR on Mac; I don't have a Linux or Windows box handy for testing.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,19 +86,19 @@ example, to enable fastint support (example for Linux):
     # src-custom/ will now contain: duktape.c, duktape.h, duk_config.h.
 
 You can also clone this repository, make modifications, and build a source
-distributable on Linux, OSX, and Windows using `python util/dist.py`.
+distributable on Linux, macOS, and Windows using `python util/dist.py`.
 
 Getting started: modifying and rebuilding the distributable
 -----------------------------------------------------------
 
 If you intend to change Duktape internals and want to rebuild the source
-distributable in Linux, OSX, or Windows:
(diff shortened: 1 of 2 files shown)


svaarala:
The main problem is that after adding a few platform one reaches a point where it's no longer easy to add further platforms. That's why the project policy so far has been that there are no official build files for different platforms (not even all different Linux variants).

That said, as long as someone is willing to maintain the changes, I have no objection to the changes. And it's certainly nice if the repo would work with (mainline) Linux, macOS, and Windows out of the box.

apjanke:
Hmm. That's a judgment call for you to make. I understand the concern.

I would be willing to maintain the macOS parts of this Makefile. (Though I can't do much to support the main Duktape code on Mac.)

svaarala:
> I would be willing to maintain the macOS parts of this Makefile. (Though I can't do much to support the main Duktape code on Mac.)

That's good enough for me, there's relatively little macOS-specific work in the main code base because it depends so little on the target.

apjanke:
Excellent.

svaarala:
I don't think Makefile.sharedlibrary has support Windows in the past so I guess testing Linux + macOS should be fine. I can test the Linux part later on when I get a chance.

svaarala:
@apjanke Took a little time to get back to this, but seems to work on Linux so merging :)

apjanke:
Thanks!

PR #1926, opened 2018-06-01
doc: rst-conventions: No GitHub nested list note

apjanke:
Remove the GitHub-specific references in the Nested Lists section, since GitHub's RST formatter seems to handle these well now. But we still want the doco format to be strict, in case other RST formatters similarly mishandle it.


svaarala:
:+1:

PR #1931, opened 2018-06-17
Polling with WinSocks on Windows

stevefan1999-personal:


examples/eventloop/poll.c
```diff
@@ -6,7 +6,15 @@
 #include <errno.h>
 #include <string.h>
 #include <stdio.h>
+
+#ifdef _WIN32
```
svaarala: There's a code style check for `#ifdef`s, so could you change to `#if defined(_WIN32)`? This should fix the Travis failure.

svaarala: Hmm, maybe I'll just merge the change and do a fixup. I'll do that in a bit :)

stevefan1999-personal:
😮So long it has been

svaarala:
Better late than never, right? :)

PR #1935, opened 2018-06-22
Add some unhandled rejection basics to polyfill

### LINKED ISSUE #1934: Errors are lost within Promise chains using polyfill (unhandled rejections)
Using `polyfills/promise.js`: when an error is thrown inside a promise function, it will be ignored.

Example:
```
Promise.resolve()
            .then(function() {
              console.log("Hello1");
            })
            .then(function() {
              console.log("Hello2");
            })
            .then(function() {
              throw new Error("Supposed to be thrown");
            });
```
I can see the error if I modify the `doReject()` function and add:

```
if (reactions.length === 0) {
        console.error("Within promise, error stack:", val ? val.stack : "unknown");
        throw val;
}
```

Not sure if it is the right way to fix it, though ;)

svaarala:
Not intended as a full solution, but some unhandled rejection drafting for the Promise polyfill. Consistency with the WHATWG specification to be checked. Tasks:

- [x] Add unhandled rejection and rejection handled notification tracking
- [x] Add a simple unhandled rejection callback for the polyfill
- [x] Fix handling of pre-settle Thrower reactions
- [x] Decide on unhandledRejection callback arguments for now
- [x] Testcase for handled within a tick
- [x] Testcase for handled after a tick
- [x] Testcase for Promise with Thrower (non-callable) rejection handlers prior to settle
- [x] Testcase for Promise with non-Thrower (callable) rejection handlers prior to settle
- [x] Testcase for Promise with Thrower rejection handler after settle, same tick
- [x] Testcase for Promise with non-Thrower rejection handler after settle, same tick
- [x] Testcase for Promise with Thrower rejection handler after settle, different tick
- [x] Testcase for Promise with non-Thrower rejection handler after settle, different tick

Fixes #1934.

### DIFF
--- a/polyfills/promise.js
+++ b/polyfills/promise.js
@@ -9,6 +9,9 @@
  *    - Doesn't handle errors from core operations, e.g. out-of-memory or
  *      internal error when queueing/running jobs.  These are implementation
  *      defined for the most part.
+ *    - Unhandled rejection 'reject' and 'handle' events don't have the same
+ *      ordering as in Node.js at present.  Unhandled rejection behavior also
+ *      hasn't been checked against latest ES specifications.
  *
  *  This polyfill was originally used to gain a better understanding of the
  *  ES2015 specification algorithms, before implementing Promises natively.
@@ -78,6 +81,9 @@
     var compact = (typeof Duktape === 'object' && Duktape.compact) ||
                   function (v) { return v; };
 
+    // Shared no-op function.
+    var nop = function nop() {};
+
     // Promise detection (plain or subclassed Promise), in spec has
     // [[PromiseState]] internal slot which isn't affected by Proxy
     // behaviors etc.
@@ -130,6 +136,11 @@
             }
             enqueueJob(ent);
         });
+        if (reactions.length === 0 && !p.rejectionHandled) {
+            // Unhandled at resolution.
+            p.unhandled = 1;
+            cons.potentiallyUnhandled.push(p);
+        }
     }
 
     // Create a new resolve/reject pair for a Promise.  Multiple pairs are
@@ -267,6 +278,7 @@
         def(this, 'value', void 0);
         def(this, 'fulfillReactions', []);
         def(this, 'rejectReactions', []);
+        def(this, 'rejectionHandled', false);  // XXX: roll into 'state' to minimize fields
         compact(this);
         var t = createResolutionFunctions(this);
         try {
@@ -277,6 +289,7 @@
     };
     var proto = cons.prototype;
     def(cons, 'prototype', proto, '');
+    def(cons, 'potentiallyUnhandled', [], '');
 
     // %Promise%.resolve().
     // XXX: direct handling
@@ -355,6 +368,26 @@
         return p;
     }
 
+    // Shared helper for checking unhandled rejections after settling.
+    function checkRejectionHandling(p) {
+        if (p.unhandled === 2) {
+            if (p.rejectionHandled) {
+                // Unhandled, already notified, need handled notification.
+                p.unhandled = 3;
+                cons.potentiallyUnhandled.push(p);
+            } else {
+                // Maybe handled later.
+            }
+        } else {
+            // XXX: If this.unhandled was 1, we'd like to remove
+            // the Promise from cons.potentiallyUnhandled list
+            // but skip that because it would mean an expensive
+            // list remove.  If cons.potentiallyUnhandled was a
+            // Set, it would be natural to remove from Set here.
+            delete p.unhandled;
+        }
+    }
+
     // %PromisePrototype%.then(), also used for .catch().
     function then(onFulfilled, onRejected) {
         // No subclassing support here now, no NewPromiseCapability() handling.
@@ -366,6 +399,12 @@
         var optimized = allowOptimization;
         if (typeof onFulfilled !== 'function') { onFulfilled = 'Identity'; }
         if (typeof onRejected !== 'function') { onRejected = 'Thrower'; }
+
+        // This is unconditional on purpose: if 'onRejected' is not callable,
+        // the rejection of 'this' is forwarded to 'p' so 'this' is considered
+        // handled.
+        this.rejectionHandled = true;
+
         if (this.state === void 0) {  // pending
             if (optimized) {
                 this.fulfillReactions.push({
@@ -404,6 +443,7 @@
                 });
             }
         } else {  // rejected
+            checkRejectionHandling(this);
             if (optimized) {
                 enqueueJob({
                     handler: onRejected,
@@ -438,6 +478,7 @@
                 value: source.value
             });
         } else {  // rejected
+            checkRejectionHandling(source);
             enqueueJob({
                 target: target,
                 value: source.value,
@@ -460,6 +501,52 @@
     };
     def(_try, 'name', 'try', 'c');
 
+    // Unhandled rejection notifications after a "tick", which seems to be
+    // interpreted as "run jobs until job queue is empty".
+    // https://html.spec.whatwg.org/multipage/webappapis.html#unhandled-promise-rejections
+    // https://www.ecma-international.org/ecma-262/8.0/#sec-host-promise-rejection-tracker
+    function checkUnhandledRejections() {
+        var idx;
+
+        // The unhandledRejection() callbacks may have queued more Promises,
+        // settled existing Promises on the list, etc.  Keep going until
+        // the list is empty.  Null out entries to allow early GC when the
+        // Promises are no longer reachable.  Callbacks may also queue more
+        // ordinary Promise jobs; they are also handled to completion within
+        // the tick.
+
+        // XXX: It might be more natural to handle the notification callbacks
+        // via the job queue.  This might be a bit simpler, but would change
+        // the Promise job vs. unhandledRejection callback ordering a bit.
+        // For example, Node.js emits 'handle' events before the related
+        // catch callbacks are called, while the polyfill in its current
+        // state does not.
+
+        for (idx = 0; idx < cons.potentiallyUnhandled.length; idx++) {
+            var p = cons.potentiallyUnhandled[idx];
+            cons.potentiallyUnhandled[idx] = null;
+
+            // The unhandledRejection() callback runs intentionally without a
+            // try-catch.  If the application wants an unhandled rejection to
+            // cause a process exit, the callback may throw which causes an
+            // uncaught exception.
+            //
+            // Use a custom object argument convention for flexibility.
+
+            if (p.unhandled === 1) {
+                cons.unhandledRejection({ promise: p, event: 'reject', reason: p.value });
+                p.unhandled = 2;
+            } else if (p.unhandled === 3) {
+                cons.unhandledRejection({ promise: p, event: 'handle', reason: p.value });
+                delete p.unhandled;
+            }
+        }
+
+        cons.potentiallyUnhandled.length = 0;
+
+        return idx > 0;  // true if we processed entries
+    }
+
     // Define visible objects and properties.
     (function () {
         def(this, 'Promise', cons);
@@ -468,16 +555,24 @@
         def(cons, 'all', all);
         def(cons, 'race', race);
         def(cons, 'try', _try);
+        def(cons, 'isPolyfill', true);  // needed by e.g. testcases
         def(proto, 'then', then);
         def(proto, 'catch', _catch);
         if (haveSymbols) {
             def(proto, Symbol.toStringTag, 'Promise', 'c');
         }
 
-        // Custom API to drive the "job queue".
+        // Custom API to drive the "job queue".  We only want to exit when
+        // there are no more Promise jobs or unhandledRejection() callbacks,
+        // i.e. no more work to do.  Note that an unhandledRejection()
+        // callback may queue more Promise job entries.
         def(cons, 'runQueue', function _runQueueUntilEmpty() {
-            while (runQueueEntry()) {}
+            do {
+                while (runQueueEntry()) {}
+                var recheck = checkUnhandledRejections();
+            } while(recheck);
         });
+        def(cons, 'unhandledRejection', nop);
 
         compact(this); compact(cons); compact(proto);
     }());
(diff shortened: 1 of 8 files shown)


svaarala:
@bwalter With this change, the polyfill logs something like:
```
((o) Duktape [linenoise] 2.2.0 (v2.1.0-613-gdbc49ad4)

// No warning, rejection handled in "same tick".
duk> P = Promise.reject(123); P.catch(Math.cos);
= {}
duk> Promise.runQueue()
= undefined

// Rejection unhandled in one tick, handled in the next.
duk> P = Promise.reject(123);
= {unhandled:1}
duk> Promise.runQueue();
UNHANDLED REJECTION: {unhandled:1}
= undefined
duk> P.catch(Math.cos);
= {}
duk> Promise.runQueue();
UNHANDLED REJECTION HANDLED ASYNCHRONOUSLY: {unhandled:3}
= undefined
```

The pull is not complete in that the log call sites would become some definable callback which would probably get a Promise and a string reason or similar. I'll also need to review against the WHATWG specification etc.

svaarala:
@bwalter This is IMO ready to be merged as an initial implementation. It's not perfect, and the testcase coverage is not great, but at least the basics should be in place.

svaarala:
@bwalter If you wanted to try it out, you'd do something like:

```javascript
if (typeof Promise === 'function' && Promise.isPolyfill) {
    Promise.unhandledRejection = function (args) {
        // args object contains:
        //    .promise: related Promise
        //    .event: 'reject' or 'handle'
        //    .reason: reject value of .promise

        if (args.event === 'reject') {
            console.log('Unhandled rejection:', args.reason);
        } else if (args.event === 'handle') {
            console.log('Previous unhandled rejection got handled:', args.reason);
        }
    };
}
```

The signature of Promise.unhandledRejection() is custom; it doesn't match Node.js because there's no event support in Duktape (at least at present). The signature is of course easy to change if you want.

polyfills/promise.js
```diff
@@ -9,6 +9,9 @@
  *    - Doesn't handle errors from core operations, e.g. out-of-memory or
  *      internal error when queueing/running jobs.  These are implementation
  *      defined for the most part.
+ *    - Unhandled rejection 'reject' and 'handle' events don't have the same
+ *      ordering as in Node.js at present.  Unhandled rejection behavior also
+ *      hasn't been checked against latest ES specifications.
```
fatcerberus: I don’t understand how this can be; by definition you only get a `handle` event for a promise which previously generated a `reject` event, so it’s impossible to have them in the opposite  order I would think?

svaarala: The comment refers to more subtle ordering guarantees, like what's the ordering between Promise job queue entries vs. unhandled rejection calls. For example, for https://github.com/svaarala/duktape/blob/68f1b824fa287f075dda4d0a34ae1ebb1bd6586d/tests/ecmascript/test-bi-promise-unhandled-difftick-1.js Duktape prints:
```
tick 1
unhandled/reject: P
unhandled/reject: Q
unhandled/reject: R
tick 2
nop called
nop called
unhandled/handle: P
unhandled/handle: Q
unhandled/handle: R
unhandled/reject: R2
tick 3
tick 4
```

and Node.js prints:
```
tick 1
unhandled/reject: P
unhandled/reject: Q
unhandled/reject: R
tick 2
unhandled/handle: P
unhandled/handle: Q
unhandled/handle: R
nop called
nop called
unhandled/reject: R2
tick 3
tick 4
```

This is unlikely to matter in practice, and I'll refine the polyfill later on w.r.t. ordering.

fatcerberus:
> Fix handling of pre-settle Thrower reactions (they shouldn't count as "handling")

To be clear, this also means that no “rejection” event is generated for such a promise either, correct?

Sorry for being so nitpicky about this, I played a big part in getting this functionality into ChakraCore as well so now I’m too familiar with all the semantics now. :)

polyfills/promise.js
svaarala: Some other minor ordering issues:
- Relationship between events for multiple Promises dealt with in one tick. If P and Q are unhandled, are the events guaranteed to happen in a specific order? For instance, is it acceptable to emit `reject` for P,Q but `handle` for Q,P? Obviously consistent ordering is preferable, but is it required?
- Relationship between unhandled rejection callbacks and Promise jobs. For example, an unhandled rejection callback may cause a Promise job, or more unhandled rejections to be queued. What ordering guarantees are made for them? For Promise jobs the order is fully deterministic as far as I can tell, but I'm not sure about unhandled rejections because ES doesn't specify any ordering guarantees for different queue types.
- If a Promise first gets a `reject` and then gets handled, is there a guarantee that the `handle` event happens before the handler itself is called (or vice versa)? Node.js seems to do so, but I'm not sure this is required.

I'm not trying to address these at present, just to point out there are several ordering possibilities and levels of guarantees one can try to provide.

svaarala:
Addition to above: the comment on the bullet (now fixed) was a bit misleading: a Thrower is considered "handling" for `P` in the example above. But the derived Promise will be, by default, unhandled (like any other Promise).

polyfills/promise.js
fatcerberus: The `handle` event is associated with the `.then()` call (or alias, e.g. `.catch()`) so I would always expect it to go into the queue before the handler call itself.  The PromiseRejectionTracker callback is not allowed to throw so I think the expectation is that the engine calls it synchronously.

svaarala:
@fatcerberus W.r.t. ordering, do you happen to remember what is considered a "tick" in the sense that "if Promise is handled within the 'tick' a `reject` event is emitted"?

In particular, suppose that `var P = Promise.reject(123)` is called, with jobs in the Promise job queue *and* expired timers (setTimeout) about to be called. Do the expired timers also count as being in the same tick w.r.t. unhandled rejections?

In other words, if the Promise is *not* handled by entries in the Promise job queue, but is handled by an already expired (but not yet executed timer), is a `reject` generated?

What about other job types?

polyfills/promise.js
svaarala: Is there a requirement for `reject` and `handle` to coexist in the Promise job queue?

I first thought Node.js queues `reject` and `handle` events in the Promise job queue, but at least in some tests (very quick I admit) that doesn't seem to be the case.

fatcerberus:
I’m at work right now but when I get home I’ll post a proper explanation of how this should work.

polyfills/promise.js
svaarala: Re: Node.js behavior, suppose that two Promises, P and Q, have been emitted a `reject` event.

You then handle the promises e.g. as:
```
P.catch(nop);
Q.catch(nop);
```

If everything went through the job queue, one would expect to see:
- `handle` for P
- nop called for P
- `handle` for Q
- nop called for Q

But, this doesn't seem to happen with Node.js. Instead:
- `handle` for P
- `handle` for Q
- nop called for P
- nop called for Q

So it seems that pending `handle` events are handled before the "tick", while `reject` events are handled after the "tick".

But I haven't had much time to test Node.js or browser behavior further, so I won't make changes in the polyfill until there's a better idea of what is (a) actually required, (b) desirable in practice.

svaarala:
@fatcerberus Thanks!

Please note that I'm interested in two sometimes very different things:
- The actual ordering guarantees and definition of "tick" in the specifications.
- The practical guarantees one would like to see.

Anything that happens in separate ES job queues are not given any ordering guarantees for example.

Also the definition of the "tick", w.r.t emitting a `reject` is critical. For example, does any entry in any job queue (say module loading) count towards the unhandled rejection tick or not? If it does, then it's not really possible to polyfill Promises because all possible event queues must be considered before concluding that some Promise rejection is unhandled. The implication for architecture would then be that it should also be possible to make custom application event concepts visible in the determination somehow, e.g. by allowing application events to be handled via Duktape.

Also for an engine where e.g. an event loop (setTimeout) is external, it's somewhat tricky if expired timers are considered to be part of the "tick" in the sense that if they handle a rejection, a `reject` should not be emitted. Here the implication might also be that it should be possible to handle expired timers via Duktape, so that they would be visible to Duktape when deciding whether or not to emit a `reject`.

This pull is really just a baseline with not much attempt to get the ordering right yet. But the questions and implications above are interesting for the next steps.

polyfills/promise.js
svaarala: Here's that Node.js test which to me implies the `reject` and `handle` events don't coexist in the Promise job queue:

```
process.on('unhandledRejection', function (reason, promise) {
    console.log('reject:', promise.name);
});
process.on('rejectionHandled', function (promise) {
    console.log('handle:', promise.name);
});
var P = Promise.reject(123); P.name = 'P';
var Q = Promise.reject(234); Q.name = 'Q';

setTimeout(function () {
    P.catch(function () { console.log('P.catch'); });
    Q.catch(function () { console.log('Q.catch'); });
}, 0);
```

Output is:
```
reject: P
reject: Q
handle: P
handle: Q
P.catch
Q.catch
```

svaarala: As a side note, it also implies that expired timers (a zero timer is always expired if one processes all event loops to completion) *do not* count towards the "tick" definition for unhandled rejections. Otherwise no `reject` or `handle` events would be emitted.

svaarala:
Repeating the Node.js ordering comment here because it's now in a commit comment:

For Node.js the following:
```javascript
process.on('unhandledRejection', function (reason, promise) {
    console.log('reject:', promise.name);
});
process.on('rejectionHandled', function (promise) {
    console.log('handle:', promise.name);
});
var P = Promise.reject(123); P.name = 'P';
var Q = Promise.reject(234); Q.name = 'Q';

setTimeout(function () {
    P.catch(function () { console.log('P.catch'); });
    Q.catch(function () { console.log('Q.catch'); });
}, 0);
```

outputs:
```
reject: P
reject: Q
handle: P
handle: Q
P.catch
Q.catch
```

This seems to imply at least that:
- `reject` and `handle` events are not handled as part of the job queue, at least not trivially so that both the `handle` and the handler callback would be queued immediately. Otherwise the order would be: handle P, catch P, handle Q, catch Q.
- Expired timers (or at least zero timers) are not considered part of the "tick". If they were considered part of the "tick", no `reject` or `handle` events would happen because the promise rejections would be handled before all the pending job queues, timers, etc were exhausted. So maybe only (1) promise job queue entries and (2) unhandled rejection callbacks are counted.

fatcerberus:
There are two separate requirements here, part of the confusion here is that the polyfill seemingly conflates them:

* As far as ECMAScript is concerned the only requirement is that Duktape provides a hook for HostPromiseRejectionTracker which the application can latch onto.  What happens when that hook is called is out of scope.  The default is to do nothing.

* The WhatWG specification, which describes a specific implementation of HostPromiseRejectionTracker that an application, usually a browser, should adhere to.

ECMAScript quite deliberately doesn’t concern itself with the shape of the event loop.  The only requirement as far as I’m aware is that all jobs specified to go into the same job queue must execute in FIFO order.

Let me be clear: the HostPromiseRejectionTracker “events” (`reject`, `handle`) are *not* asynchronous.  The specification requires the callback to complete normally (i.e. not throw) exactly because it is called inline, at the exact moment the promise is rejected/reaction is attached.  At no point do these events go into the job queue, they are communicated directly to the application at the moment they happen.

It’s difficult to say what Node is doing when you call `process.on()`, presumably the callback itself is executed asynchronously from the Node event loop.  But I can assure you that under the hood, the notifications from V8 about the rejection are very much synchronous.

svaarala:
> There are two separate requirements here, part of the confusion here is that the polyfill seemingly conflates them: [...]

So basically a HostPromiseRejectionTracker is the low level concept and the concrete callback mechanism (e.g. `process.on()` for Node.js) is, at least conceptually, built on top of that. For clarity it might make sense to structure the polyfill at least internally that way too so that the distinction becomes more clear (it might be useful if Duktape would expose also, or only, the low level hook).

I agree HostPromiseRejectionTracker is called inline. Although this requirement:
> ... must complete normally in all cases. 

is of course unattainable in practice because of potential out-of-memory conditions. So a practical implementation must deal with it somehow, e.g.:
- Always call as a protected function, and ignore (debug log) any error.
- Always call as a protected function, and bubble out the error from the Promise API call triggering the HostPromiseRejectionTracker call.
- Always call as a protected function, and cause a fatal error from an error. This would be bad for sandboxing so IMO not really an option.

Being an inline mechanism, a `"reject"` happens at the point of a Promise becoming rejected, regardless of whether the rejection is handled on the next statement. The upside is of course that no notion of a "tick" is needed at this level, and it is up to the callee to determine how to filter out unnecessary `"reject"` calls that may become handled when some concept of a tick completes.

The downside is that the HostPromiseRejectionTracker is too low level to be directly useful, so it would be nice to have a higher level concept out of the box too (similar to Node.js `process.on()`). Otherwise every user would need to concern themselves with the detailed ordering and tick semantics etc which would be awkward and easy to get wrong.

So I guess some API options are:
- Only provide a HostPromiseRejectionTracker API. Maybe provide a default or example implementation with a useful higher level interface; otherwise it will be difficult for applications to get it right.
- Provide hooks to integrate both into the low level HostPromiseRejectionTracker *and* a built-in higher level hook with more useful semantics (close to WHATWG or Node.js). An application could then choose to hook into the higher level API or the lower level API if it wanted to do something different e.g. re: tick boundaries.

In case higher level hooks are available (at least as a default) Duktape needs to be aware of the "tick" for rejections, which may have implications for how events are handled in the Duktape API. E.g. an application may want to make expired timers visible to Duktape to allow them to take into account. In practice e.g. Node.js doesn't seem to do this, so maybe at least the default high level mechanism could just work with the Promise job queue. This needs some thought :-)

fatcerberus:
Indeed, the HostPromiseRejectionTracker is very low-level by design, Ecmascript deliberately doesn't mandate a particular event loop shape so it would be difficult to do anything more at that level.

But yes, the tracker callback is called inline.  For "reject" this is at the moment the promise becomes rejected, and for "handle" it's at the moment the first reaction is attached, e.g. by calling `.then()`.

svaarala:
Unfortunately that low level reject/handle concept is also not very useful as a default API so that one needs to get involved in the more complex semantics too -- at least via a default implementation bundled with the engine, if nothing else.

PR #1937, opened 2018-07-07
Improvements to polyfill unhandled rejection handling

svaarala:
- [x] Make HostPromiseRejectionTracker API boundary clear
- [x] Update unhandled rejection comments
- [x] Rename .rejectionHandled -> .isHandled
- [x] Make it possible to hook the raw HostPromiseRejectionTracker
- [x] Testcases for raw HostPromiseRejectionTracker

### DIFF
--- a/polyfills/promise.js
+++ b/polyfills/promise.js
@@ -9,15 +9,20 @@
  *    - Doesn't handle errors from core operations, e.g. out-of-memory or
  *      internal error when queueing/running jobs.  These are implementation
  *      defined for the most part.
- *    - Unhandled rejection 'reject' and 'handle' events don't have the same
- *      ordering as in Node.js at present.  Unhandled rejection behavior also
- *      hasn't been checked against latest ES specifications.
  *
  *  This polyfill was originally used to gain a better understanding of the
  *  ES2015 specification algorithms, before implementing Promises natively.
  *
  *  The polyfill uses a Symbol to mark Promise instances, but falls back to
  *  an ordinary (non-enumerable) property if no Symbol support is available.
+ *  Presence of the polyfill itself can be checked using "Promise.isPolyfill".
+ *
+ *  Unhandled Promise rejections use a custom API signature.  For now, a
+ *  single Promise.unhandledRejection() hook receives both 'rawReject' and
+ *  'rawHandle' events directly from HostPromiseRejectionTracker, and higher
+ *  level (Node.js / WHATWG like) 'reject' and 'handle' events which filter
+ *  out cases where a rejected Promise is initially unhandled but gets handled
+ *  within the same "tick".
  *
  *  See also: https://github.com/stefanpenner/es6-promise#readme
  */
@@ -65,6 +70,9 @@
         }
         return ret;
     }
+    function queueEmpty() {
+        return !queueHead;
+    }
 
     // Helper to define/modify properties more compactly.
     function def(obj, key, val, attrs) {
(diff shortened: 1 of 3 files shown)


fatcerberus:
One important bit: IIRC the specification doesn't pass the actual promise being rejected to the abstract HostPromiseRejectionTracker; in order for the API to be useful Duktape should do that.  In fact, I believe implementing the WHATWG semantics actually *requires* this.

svaarala:
I think it does? For example, PerformPromiseThen() step 9c:
> If promise.[[PromiseIsHandled]] is false, perform HostPromiseRejectionTracker(promise, "handle").

fatcerberus:
Right, sorry, the bit that was missing was the rejection value, which e.g. I use in miniSphere to produce an error screen upon detecting an unhandled rejection.

This is particularly crucial as normally the only way to access the settled value of a promise is to attach another reaction to it.

svaarala:
Ah, I was also wondering why the rejection value was missing. One could extract the value by chaining a `.catch()` but that seems very awkward; while one normally cannot read the Promise fulfill/settle except via handlers, in this case having direct access to it makes much sense because it's usually e.g. logged.

Not having access to the reason also makes it non-trivial to implement e.g. the Node.js `process.on()` because it provides access to the rejection value -- how would one get it from HostPromiseRejectionTracker?

Because these unhandled rejection situations are not performance critical, I think the best signature is for a hook to be passed an object with named properties so that it's easy to extend and provide custom or version specific flags/values. There's something like that in the polyfill now (for the higher level hooks):
```
function rejectionHandler(args) {
    // args.event: "reject" or "handle"
    // args.promise: related promise
    // args.reason: related value
}
```

This would also be relatively compatible with a browser-like event (WHATWG):
- .promise and .reason are the same
- .cancelable not present above
- .event is present above but not in WHATWG because it's implicit for the registered event handler type

fatcerberus:
Chaining a `.catch()` isn’t always ideal either; besides mutating the promise, it also only provides a value asynchronously; in cases like this you generally need it immediately.

svaarala:
Exactly, that's why it would be awkward for logging especially.

svaarala:
This is the signature for now; the same callback receives both `rawReject` and `rawHandle` (directly from HostPromiseRejectionTracker) and `reject` and `handle` (nicer higher level concepts):

```
$ duk polyfills/promise.js  -i
((o) Duktape [linenoise] 2.2.99 (v2.2.0-135-g8d5fbe7f)
duk> Promise.unhandledRejection = function (args) { console.log('UNHANDLED:', args); };
= {_func:true}
duk> P = Promise.reject(123);
UNHANDLED: {promise:{},event:"rawReject",reason:123}
= {}
duk> Promise.runQueue();
UNHANDLED: {promise:{},event:"reject",reason:123}
= undefined
duk> P.catch(Math.cos);
UNHANDLED: {promise:{},event:"rawHandle",reason:123}
= {}
duk> Promise.runQueue();
UNHANDLED: {promise:{},event:"handle",reason:123}
= undefined
```

and if handled within the same tick, only the raw events will fire:
```
duk> P = Promise.reject(234);
UNHANDLED: {promise:{},event:"rawReject",reason:234}
= {}
duk> P.catch(Math.cos);
UNHANDLED: {promise:{},event:"rawHandle",reason:234}
= {}
duk> Promise.runQueue();
= undefined
```

fatcerberus:
Looks good.  I like that `duk` doesn’t produce a tick between commands and would like that behavior to remain even once promise & job queue support becomes part of the core; that’s something that’s always annoyed me about Node’s REPL.

svaarala:
@fatcerberus Probably `duk` will need an option for that: some users would expect a tick to happen automatically and it's a reasonable default IMO.

PR #1947, opened 2018-07-22
Allow Proxy target and handler to be Proxy too

svaarala:
Up to now Duktape has intentionally rejected a Proxy as either a Proxy target or a Proxy handler object; both cases cause more potential for native recursion and are relatively rare in practice. ES2015+ does not place such a limitation on Proxies. This pull removes the current limitation and allows Proxies as both target and handler.

Tasks:
- [ ] Remove Proxy check for target/handler in Proxy creation
- [ ] Proxy chain support for getprop
- [ ] Proxy chain support for putprop
- [ ] Proxy chain support for delprop
- [ ] Proxy chain support for hasprop
- [ ] TBD: enumeration etc
- [ ] Testcase coverage
- [ ] Releases entry

This is currently work in progress, to be merged after 2.3.

### DIFF
--- a/src-input/duk_hobject_props.c
+++ b/src-input/duk_hobject_props.c
@@ -460,6 +460,8 @@ DUK_LOCAL duk_bool_t duk__proxy_check_prop(duk_hthread *thr, duk_hobject *obj, d
 	 * Or rely on later stack depth macro check (insert one here)?
 	 */
 
+	/* FIXME: implement recursive Proxy target support directly here? */
+
 	duk_require_stack(thr, DUK__VALSTACK_PROXY_LOOKUP);
 	duk_push_hobject(thr, h_handler);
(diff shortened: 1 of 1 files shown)


fatcerberus:
I’m actually a bit surprised they allow a Proxy for the handler.  Proxy-of-Proxy makes some sense, but a Proxy’d handler doesn’t seem as useful.

svaarala:
I think it's just allowed by default: it's not explicitly rejected even though it may not be very useful in practice.

fatcerberus:
Yeah, makes sense.  It just seems like something that might want to be rejected explicitly simply to avoid all the potential edge cases it creates.  But thinking about it, I suppose that the same pitfalls apply to recursive proxies so disallowing a Proxy as handler doesn’t really buy you anything in that regard.

PR #1952, opened 2018-08-03
Dc6jgk file time to system time/system time to tz specific local time fix

dc6jgk:
Avoid calling SystemTimeToTzSpecificLocalTime with bad/uninitialized st2 parameters (potential crash/stack-corruption on Win2k3)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -54,6 +54,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Michal Kasperek (https://github.com/michalkas)
 * Andrew Janke (https://github.com/apjanke)
 * Steve Fan (https://github.com/stevefan1999)
+* Gero Kuehn (support@gkware.com,https://github.com/dc6jgk)
 
 Other contributions
 ===================
--- a/src-input/duk_bi_date_windows.c
+++ b/src-input/duk_bi_date_windows.c
@@ -96,6 +96,7 @@ DUK_INTERNAL duk_int_t duk_bi_date_get_local_tzoffset_windows(duk_double_t d) {
 	ULARGE_INTEGER tmp2;
 	ULARGE_INTEGER tmp3;
 	FILETIME ft1;
+	BOOL fRet;
 
 	/* XXX: handling of timestamps outside Windows supported range.
 	 * How does Windows deal with dates before 1600?  Does windows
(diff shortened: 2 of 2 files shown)


svaarala:
@dc6jgk Thanks! Any idea when the FileTimeToSystemTime() call would fail in practice?

dc6jgk:
The HBBTV app causing the crash several days ago is currently not running at all due to massive changes to the Event handling in the browser engine. Without investing a couple of hours to revert back to different versions and finding the app version that crashed, I can not answer this. I will leave a note here, if I encounter the crash again over the weekend with the (intentionally unpatched) 2.2.1 release I imported this evening.

svaarala:
There are some trivial issues breaking unit tests in the pull:
- CR newlines (LF only should be used)
- Trailing whitespace on some lines (probably caused by the CRs)

svaarala:
Re: when the call fails, the fix is is obviously correct regardless of the fail reason. I'm just curious to know what use case this might affect, because in those cases there is an application visible impact for Date local time handling.

dc6jgk:
I know that this is probably not a good way to introduce myself and there is probably no good or polite way to say this but you at least deserve an answer:

Please do not take it personal, but line feeds and spaces are things I have no intention to care about - ever again. If a bot can complain about that, it should be able to "fix" that instead of throwing 10k lines of pointless noise at me (yes, I have seen the Travis thing and it was not obvious to me what it complained about). 

If that is considered to be a real issue, consider me "incompatible to open source contribution" at least for this project. I am not going to install/run any part of the scripting hell (Python) either. Keeping our C/C++ build environments compatible to half a dozen chips and getting the W3C/CSS tests at least half way through the engine is difficult and time-wasting enough.

svaarala:
Fair enough, I can make a similar fix in a separate pull?

dc6jgk:
Still not happy about how this worked out... but I still owe you a reply to "Any idea when the FileTimeToSystemTime() call would fail in practice?". Now I do and for the sake of having this documented this properly:
	ft1.dwLowDateTime = 0xd53e8000;
	ft1.dwHighDateTime = 0x819db1de;
	FileTimeToSystemTime((const FILETIME *) &ft1, &st2);
This fails (at least on some Win32 kernels), leaving all st2 fields filled with 0xCCCC in a debug build.
9 call frames deep into the Windows API, this will then throw an exception in NTDLL if passed to SystemTimeToTzSpecificLocalTime.

PR #1956, opened 2018-08-06
Windows Date provider return value check consistency trivia

svaarala:


fatcerberus:
I'm trying to understand under what real-world condition this will fail.  According to MSDN it fails if the input value exceeds `0x8000000000000000` and no other failure condition is listed.

svaarala:
The error behavior is not changed in this pull, just cleaning up the inconsistent handling from the previous one (purely a style issue). I'm also not sure what the possible error might be, but it's still best to handle the error without memory unsafe behavior. Another alternative would be to ignore the error but memset() all values to zero beforehand to avoid uninitialized data, but that approach would IMO be worse.

svaarala:
Anyway, if you figure out what a plausible error is, it'd be interesting to know :)

fatcerberus:
Well like I said the only indication given by the MSDN page is that an input value greater than 0x8000000000000000 will cause it to fail.  I don’t know if that’s possible in practice?

PR #1958, opened 2018-08-13
Support dumping bytecode opcodes in dump_bytecode.py

unamer:
Support dumping instruments, output will be something like:

```
Signature byte: 0xbf
  Instructions: 4
  Constants: 1
  Inner functions: 1
  Nregs: 2
  Nargs: 0
  Start line number: 1
  End line number: 18
  duk_hcompiledfunction flags: 0x18000a80
    000000: 00000198 CLOSURE r1, 0
    000001: 01001391 DECLVAR_CR 0x13, c0, r1 //writable, enumerable, func_decl
    000002: 00000007 LDUNDEF r0
    000003: 0000009d RETREG r0
  Constants:
    000000: 'concat'
.....
```


svaarala:
This is a nice addition the dump tool. This change would be nice before merge though:

- Add yourself to AUTHORS.rst first list (see CONTRIBUTING).
- Use bytecode definitions from https://github.com/svaarala/duktape/blob/master/debugger/duk_opcodes.yaml to avoid duplicating them. Opcodes change over releases and it's important to have only one master metadata file. The debugger pretty printer is here (not sure if it's useful): https://github.com/svaarala/duktape/blob/master/debugger/duk_debug.js#L1062.
- Maybe change the visual separation for the code comments from:
  ```
      000001: 01001391 DECLVAR_CR 0x13, c0, r1 //writable, enumerable, func_decl
  ```
  to
  ```
      000001: 01001391 DECLVAR_CR 0x13, c0, r1  ; writable, enumerable, func_decl
  ```
  for more consistency with the existing debugger opcode dumper.  The double space separates the comment visually from the raw opcode. The debugger opcode dumper uses more visual space which would also be fine (in particular, aligning all opcode comments to the same column).

These style changes in the pull I also agree with:
```
-   num_formals, = struct.unpack('>L', buf[off:off+4])
+   num_formals, = struct.unpack('>L', buf[off:off + 4])
```

Not sure if there should be a command line option to enable (or disable) the bytecode dump?

unamer:
Yeah, the style just some auto formation lol.

The opcode definitions was loaded from outside yaml before, but i think maybe it should be able to run without any dependency, so I just embed it inside .

Now I use the opcodes definitions from 'duk_opcodes.yaml' lol.

svaarala:
@unamer Could you fix the AUTHORS conflict (my bad for not merging earlier)?

I have a few improvements in mind but I can do them in a fixup if you prefer:
- Paths should not use e.g. `../debugger/xxx` but rather `os.path.join('..', 'debugger', 'xxx')` to port to Windows properly.
- Search path should ideally be sensitive to the location of the `dump_bytecode.py` script rather than current working directory (or maybe both, whatever makes sense). In particular, the `../debugger/xxx` should be relative to the script.

Script path can be obtained e.g. as (this is used by tools/configure.py):
```
    script_path = sys.path[0]  # http://stackoverflow.com/questions/4934806/how-can-i-find-scripts-directory-with-python
```

svaarala:
@fatcerberus I don't use Windows much so I'm not sure what the different Python variants do on Windows, so maybe it would work even with an incorrect separator? Not sure what happens if a filename actually contains a `/` etc. Anyway, `os.path.join()` is the cleanest approach I think.

fatcerberus:
`/` is not a legal filename character in Windows so it’s always parsed as a directory separator.  I actually take advantage of this fact in miniSphere to avoid needing platform-specific code in my path canonicalization code. :)

svaarala:
Good to know! Apparently forward slash paths just break sometimes on the command line because they get confused with options.

fatcerberus:
Yeah, cmd.exe treats a forward slash as an option seperator for built-in commands which is convenient sometimes (`dir/s/b`) but other times is annoying (“invalid option”).  As far as the OS itself is concerned though `/` and `\` are equivalent in filesystem paths which is nice since you don’t have to escape the former in strings. :)

unamer:
@svaarala Thanks for your attention on this pull lol.

I fixed the AUTHORS conflict.

Actually forward slash works on any platform regards windows or linux.

And script will try to load database file in its folder, if fails then load from debugger folder.

svaarala:
Thanks @unamer! I'll merge and fix the minor issues separately.

The script path issue is that running `python tools/dump_bytecode.py` from the Duktape repo will fail to load the opcodes because the path is resolved relative to CWD, not where the script lives.

PR #2014, opened 2018-12-12
Allow function-like macros to be defined via configure.py

### LINKED ISSUE #2013: Cannot define function-like macros via configure.py
Configuration of `DUK_USE_FATAL_HANDLER` and `DUK_USE_DEBUG_WRITE` does not work as suggested in [wiki](https://wiki.duktape.org/Configuring2x.html):

```
$ python2 tools/configure.py --output-directory=build '-DDUK_USE_FATAL_HANDLER(udata,msg)=do { const char *fatal_msg = (msg); fprintf(stderr, "*** FATAL ERROR: %s\n", fatal_msg ? fatal_msg : "no message"); abort(); } while (0)' -DDUK_USE_ASSERTIONS -DDUK_USE_SELF_TESTS -DDUK_USE_DEBUG -DDUK_USE_DEBUG_LEVEL=0 '-DDUK_USE_DEBUG_WRITE(level,file,line,func,msg)=do {fprintf(stderr, "D%ld %s:%ld (%s): %s\n", (long) (level), (file), (long) (line), (func), (msg));} while(0)'
  File "tools/configure.py", line 997, in <module>
    main()
  File "tools/configure.py", line 257, in main
    (opts, args) = parser.parse_args()
  File "/usr/lib/python2.7/optparse.py", line 1400, in parse_args
    stop = self._process_args(largs, rargs, values)
  File "/usr/lib/python2.7/optparse.py", line 1444, in _process_args
    self._process_short_opts(rargs, values)
  File "/usr/lib/python2.7/optparse.py", line 1551, in _process_short_opts
    option.process(opt, value, values, self)
  File "/usr/lib/python2.7/optparse.py", line 789, in process
    self.action, self.dest, opt, value, values, parser)
  File "/usr/lib/python2.7/optparse.py", line 809, in take_action
    self.callback(self, opt, value, parser, *args, **kwargs)
  File "duktape/tools/genconfig.py", line 1391, in add_force_option_define
    raise Exception('invalid option value: %r' % value)
Exception: invalid option value: 'DUK_USE_FATAL_HANDLER(udata,msg)=do { const char *fatal_msg = (msg); fprintf(stderr, "*** FATAL ERROR: %s\\n", fatal_msg ? fatal_msg : "no message"); abort(); } while (0)'
```

Actually, there are two problems with the current processing approach:
- Current code doesn't allow an equal sign to appear in the value part of the definition (that's the cause for the error message above).
- If that's worked around, current code still cannot deal with the arguments of a function-like macro.

akosthekiss:
Fixes #2013

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -56,6 +56,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Steve Fan (https://github.com/stevefan1999)
 * Edward Betts (https://github.com/edwardbetts)
 * Ozhan Duz (https://github.com/webfolderio)
+* Akos Kiss (https://github.com/akosthekiss)
 
 Other contributions
 ===================
--- a/tools/genconfig.py
+++ b/tools/genconfig.py
@@ -1382,13 +1382,15 @@ def add_genconfig_optparse_options(parser, direct=False):
         with open(value, 'rb') as f:
             force_options_yaml.append(f.read())
     def add_force_option_define(option, opt, value, parser):
-        tmp = value.split('=')
-        if len(tmp) == 1:
-            doc = { tmp[0]: True }
-        elif len(tmp) == 2:
-            doc = { tmp[0]: tmp[1] }
+        defname, eq, defval = value.partition('=')
+        if not eq:
+            doc = { defname: True }
         else:
-            raise Exception('invalid option value: %r' % value)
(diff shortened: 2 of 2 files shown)


svaarala:
Looks good, thanks! I'll add some minimal tests and update RELEASES in a follow-up :+1:

PR #2059, opened 2019-03-15
Add duk_safe_to_stacktrace()

### LINKED ISSUE #775: API call to convert an error into a stacktrace
At the moment if a user wants to catch an error and dump a maximally verbose stack trace, the call site gets rather large: it needs to check for a `.stack` property and if it's missing coerce the value to a string instead. These operations should also be inside a safe call to avoid throwing further errors.

There's already a `duk_safe_to_string()` to make it easier to coerce a value to a string safely (a common need when dumping values). Maybe add:

``` c
printf("%s\n", duk_safe_to_stacktrace(ctx, -1));
```

The API call would:
- Always return a `const char *` with a useful value.
- For errors, returns a stack trace unless `.stack` is missing.
- For any other values, returns `ToString(value)` like `duk_safe_to_string()`. This could actually be improved now, for example using the same summarization style as is used for property access errors.
- Catches errors in coercion; if such an error occurs, coerces that to a string (again, safely); if that fails too, push some useful message like `duk_safe_to_string()`.

The name `duk_safe_to_stacktrace()` wouldn't be very accurate because it'd also be intended to work on non-error values. Maybe `duk_safe_to_errorstring()` or similar?

TheBrokenRail:
Fixes #775

### DIFF
--- a/src-input/duk_api_stack.c
+++ b/src-input/duk_api_stack.c
@@ -3067,6 +3067,48 @@ DUK_EXTERNAL const char *duk_safe_to_lstring(duk_hthread *thr, duk_idx_t idx, du
 	return duk_get_lstring(thr, idx, out_len);
 }
 
+DUK_EXTERNAL const char *duk_to_stacktrace(duk_hthread *thr, duk_idx_t idx) {
+	DUK_ASSERT_API_ENTRY(thr);
+	idx = duk_require_normalize_index(thr, idx);
+
+	if (duk_is_error(thr, idx)) {
+		duk_get_prop_string(thr, idx, "stack");  /* caller coerces */
+		if (!duk_is_null_or_undefined(thr, -1)) {
+			duk_replace(thr, idx);
+		} else {
+			duk_pop(thr);
+		}
+		return duk_to_string(thr, idx);
+	} else {
+		return duk_to_string(thr, idx);
+	}
+}
+
+DUK_LOCAL duk_ret_t duk__safe_to_stacktrace_raw(duk_hthread *thr, void *udata) {
+	DUK_ASSERT_CTX_VALID(thr);
+	DUK_UNREF(udata);
+
+	duk_to_stacktrace(thr, -1);
+
+	return 1;
+}
+
+DUK_EXTERNAL const char *duk_safe_to_stacktrace(duk_hthread *thr, duk_idx_t idx) {
+	int rc;
+
+	DUK_ASSERT_API_ENTRY(thr);
+	idx = duk_require_normalize_index(thr, idx);
+
+	duk_dup(thr, idx);
+	rc = duk_safe_call(thr, duk__safe_to_stacktrace_raw, NULL /*udata*/, 1 /*nargs*/, 1 /*nrets*/);
+	if (rc != 0) {
+		duk_safe_to_stacktrace(thr, -1);
+	}
+	duk_replace(thr, idx);
+
+	return duk_get_string(thr, idx);
+}
+
 DUK_INTERNAL duk_hstring *duk_to_property_key_hstring(duk_hthread *thr, duk_idx_t idx) {
 	duk_hstring *h;
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -57,6 +57,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Edward Betts (https://github.com/edwardbetts)
 * Ozhan Duz (https://github.com/webfolderio)
 * Akos Kiss (https://github.com/akosthekiss)
+* TheBrokenRail (https://github.com/TheBrokenRail)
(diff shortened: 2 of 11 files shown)


TheBrokenRail:
@svaarala

svaarala:
Thanks! I'll look into this later tonight.

svaarala:
Thanks for the detailed work, this will make a nice addition to the C API! A few comments on actual substance:
- There seems to be a bug in `duk_safe_to_stacktrace()` where the temporary value is left on stack top, rather than being replaced at `idx`. The unit test should cover this by (1) printing out `duk_get_top()` before and after the coercion, and (2) check that the value at `idx` was actually coerced to a string.
- For similar reasons, the unit test for `duk_to_stacktrace()` should cover top before/after the operation, and check that the value at `idx` was coerced.
- The `duk_safe_to_stacktrace()` unit test should cover the case where the attempt to read `.stack` throws, e.g. because it is a getter that is hardwired to throw, so that the safety of the API call is covered.
- The implementation in duk_api_stack.c is a bit racy: it checks whether the object has a `.stack`, then checks whether the object is an error, and only then reads `.stack`. This may seem safe, but if the target is a Proxy, any of these operations may change the object in theory. For example, checking for `.stack` existence may delete the property, meaning that the property read later will read back as `undefined`. More on this below.
- The change in eval.c may not be intuitive: it will print a stack trace if the input evaluates to an Error object without throwing it (i.e. a success case). Maybe the stack trace should only printed on errors? This isn't a very important point, and one could just leave eval.c as is just as well.

For Ecmascript code a common idiom is something like:
```javascript
try {
    something();
} catch (e) {
    // Here 'e' may or may not be an Error instance!
    console.log(e.stack || e);
}
```

Note that this doesn't perform any `instanceof` or inheritance checks; it just uses duck typing in the sense that we read `.stack`, and if it seems to exist use that, and otherwise use the object as is (maybe string coercing it).

The C analog of this would be to:
- Read the `.stack` property. If it's not undefined, coerce it to string for a return value. (Alternatively, read the property, and use it as a return value only if it is a string, otherwise fall back to duk_to_string().)
- If the property is undefined, fall back to duk_to_string().

This would require fewer checks, and for Proxy objects would only trigger a single property read.

svaarala:
More trivial comments:
- Please squash the commits into a few clean commits: e.g. code, tests, documentation, AUTHORS.
- API documentation: it would be nice to summarize what the equivalent Ecmascript code for the operation would be. This would make it easier to understand what side effects etc exist.
- API documentation: it would be good to mention explicitly what happens if the attempt to read `.stack` fails in `duk_safe_to_string()`. For example:
  > If the attempt to read `.stack` fails, the call behaves like `duk_safe_to_string()`, returning a fixed error string.
- I'll try to improve the documentation for duk_safe_to_string() for errors. The API documentation now mentions a "fixed error string". The implementation uses `"Error"` which is preallocated so that pushing the string can never fail due to out of memory.

TheBrokenRail:
Done!

src-input/duk_api_stack.c
```diff
@@ -3099,16 +3094,19 @@ DUK_LOCAL duk_ret_t duk__safe_to_stacktrace_raw(duk_hthread *thr, void *udata) {
...
 	idx = duk_require_normalize_index(thr, idx);
 
 	duk_dup(thr, idx);
-	(void) duk_safe_call(thr, duk__safe_to_stacktrace_raw, NULL /*udata*/, 1 /*nargs*/, 1 /*nrets*/);
-	retval = duk_safe_to_string(thr, -1);
+	rc = duk_safe_call(thr, duk__safe_to_stacktrace_raw, NULL /*udata*/, 1 /*nargs*/, 1 /*nrets*/);
+	if (rc != 0) {
+		duk_safe_to_stacktrace(thr, -1);
+	}
+	duk_replace(thr, -2);
```
svaarala: The replace here should be `duk_replace(thr, idx)`.

TheBrokenRail: Fixed!

TheBrokenRail:
Ready to merge?

svaarala:
I'd like for the commits to be squashed.

TheBrokenRail:
Squashed!

svaarala:
Thanks, merged! I'll open a follow-up for some small fixups.

One thing that still concerns me a bit is whether the call should do a duk_is_error() check before reading `.stack`, or just use the duck typing approach (`e.stack || e`). The latter would be more intuitive IMO if the target value is expected to be an error which the API call name implies.

Further, as a small corner case, if the error object comes from a different realm (a "global environment" in current Duktape terms) it will actually test *false* for duk_is_error() because that check is based on inspecting the inheritance chain of the error. As a result, the current approach wouldn't print out the stack trace for a foreign error.

PR #2073, opened 2019-04-29
Update AUTHORS.rst

jessedoyle:
* Add my name to the contributors list as requested in [0].

[0]: https://github.com/svaarala/duktape-wiki/pull/215#issuecomment-487741283


svaarala:
You'd need to be on the first list actually. I mistakenly mentioned initially that it should be the second list. The first list is for anyone who provided actual changes to the repo(s), the second for bug reports and such. Could you please fix and squash?

jessedoyle:
@svaarala - Actually, I think I missed the fact there were two lists. Fixed!

svaarala:
Thanks! Merging shortly.

PR #2078, opened 2019-05-07
Support ARM64 with MSVC

phraemer:
It's necessary to detect `_M_ARM64` to compile for ARM64 with Visual Studio

https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019


svaarala:
Thanks @phraemer !

PR #2089, opened 2019-05-29
Use bare arrays/objects in more places internally

svaarala:
Reduces impact of a modified Array.prototype.


svaarala:
Travis error is some Ubuntu update problem unrelated to the changes here, and testcases have been manually executed, so merging.

PR #2091, opened 2019-06-01
Add Timers Example

TheBrokenRail:
This is a simpler example that might be easier to understand as it doesn't include things like libncurses, and poll(), as well as most likely being more portable.


TheBrokenRail:
@svaarala Ready for Review!

TheBrokenRail:
I've added this to ```RELEASES.md``` and ```release-notes-v2-4.rst```.

svaarala:
My initial reaction is that I'm not very excited about adding a second example because that also increases potential confusion? Maybe the existing example can be refactored to be more easy to follow?

Could you be more specific what problem you see with the existing example? For example:
- What parts do you find hard to follow?
- If one doesn't use poll() or select(), how would an eventloop handling both timers and FDs work? Or do you just want an example with timers and nothing else?
- Curses dependency can be made optional and portability generally improved too, if that is necessary.

TheBrokenRail:
The idea is to just have timers so someone could easily implement in their own code, without having to manually remove, all the bindings for poll(), and ncurses. However if you want a more complete example, yo could use the full eventloop.

TheBrokenRail:
Are there any changes I should make @svaarala?

svaarala:
I don't want to merge this as is because I don't think having two eventloop examples would necessarily lead to less confusion. I'd rather improve the existing example (which surely does need improvement) and e.g. make the curses dependency easy to drop.

svaarala:
Regarding curses, the entire curses dependency can be removed if that would be better. It's simply used to visually demonstrate a big bunch of setTimeout() calls, but it's not essential to demonstrating timers or other eventloop stuff as such.

But I don't know how an eventloop without a poll(), or some equivalent of it, would work? Would it busy loop until a timer fired?

TheBrokenRail:
That seems to be how the ```minimal-eventloop``` example currently works.

svaarala:
To me that wouldn't be a very useful eventloop example, unfortunately. I would expect an eventloop to be able:
- Sleep until the next timeout.
- Have the ability to wait for some external events. In the current example this means sockets.

(Curses is extra on top of this, and I would be OK to remove it.)

It would be simpler to demonstrate just a timer mechanism if one removed poll() and socket support, but added in a dependency on some platform specific sleep() call (which is a much simpler dependency than poll()).

But that would result in a "timer example" rather than an "eventloop example"?

TheBrokenRail:
So just rename the example to ```timer``` from ```minimal-eventloop```?

TheBrokenRail:
The example has been renamed @svaarala.

svaarala:
What I was trying to say is that, as is, the example is a busy-polling timer example.

Is that useful in the distributable? I'm not sure. I would expect a timer implementation to be able to sleep() when waiting for timers, for example. For a pure timer example, the pull is also not very minimal (one can implement pure timers in pure ECMAScript for example).

So all in all, I'm not sure what problem is being solved by this pull. I'd prefer to understand the user problem better and first try to modify the existing example to be more easy to follow (e.g. by removing the curses dependency).

TheBrokenRail:
It now waits for the next timeout with ```nanosleep``` or ```Sleep``` (on Windows).

svaarala:
@TheBrokenRail I'm sorry but I won't merge a new example with a lot of repeated code unless there's a clear problem that is being solved :-/

TheBrokenRail:
OK

PR #2114, opened 2019-06-15
Add document for CBOR missing tag

svaarala:
Add initial version of CBOR missing tag document, in preparation for IANA assignment request.

### DIFF
--- a/doc/cbor-missing-tag.rst
+++ b/doc/cbor-missing-tag.rst
@@ -1,0 +1,76 @@
+================
+CBOR missing tag
+================
+
+Overview
+========
+
+This document specifies a CBOR [1] tag applied to CBOR Undefined (0xf7)
+for indicating a missing value (such as a gap in an array)::
+
+  Tag: TBD
+  Data item: Undefined (0xf7)
+  Semantics: Missing value (e.g. gap in array)
+  Point of contact: Sami Vaarala <sami.vaarala@iki.fi>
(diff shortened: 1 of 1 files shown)


cabo:
I apologize for continuing my attempt at lateral thinking for a little more.  Giving away one of these 1+1 tags does require some prospect of general use.

So what the missing tag really says is that the value was explicitly not given or not specified, really "omitted", unlike the "normal" undefined.  So maybe it would work to open this tag a bit up beyond undefined, e.g., for applications that have other reactions to omitted values, such as assuming default values.  Your spec would clearly give the JavaScript example so people who want to round-trip JS values know how to use it.  But other people who want to serialize values that are "omitted" could do that, too.

(Yes, I also like the name "omitted" a bit more, as "missing" sounds like a mistake that needs to be corrected.)

svaarala:
I fully agree that a 1+1 tag should be as generic as possible and not assigned lightly.

I also agree "omitted" sounds better than "missing" for the same reasons. I also thought about "gap" but that only works well for arrays (it would sound odd for objects/dicts).

I'll change this wording, expand the description to indicate that the value has been omitted and that applications may react to omitted values in different ways, and clarify that the ECMAScript array gap is just one example of omitted values.

It would be nice to give a good example of another case where the tag could be used, maybe in an object context?

cabo:
Another proposal for wording was "absent", native speakers told me that was better than "omitted".

PR #2126, opened 2019-06-25
Remove ncurses dependency from eventloop example

TheBrokenRail:
This will make the event loop example more platform independent. @svaarala


sva-p:
Just to be clear, are you concerned about Windows portability specifically?

If so, follow-up pulls should then add Windows socket and poll compatibility (WSAWaitForMultipleEvents(), or whatever is most appropriate today)?

svaarala:
I agree with this change if curses is considered problematic. But just to be clear, what portability targets are you trying to address?

It would make sense for the example to be portable to Linux, macOS, and Windows at least. At present the example is Linux only.

Adding macOS support should be simple but requires some testing. Adding Windows support would require adding Windows socket bindings and using WSAWaitForMultipleEvents() (or whatever is most appropriate today)?

These can (and should) be done in separate pulls - but it'd be good to first define a goal what this and the follow-up pulls are trying to achieve.

TheBrokenRail:
Currently this demo is quite saturated, removing, some of the stuff that has no relation the the eventloop, should make the example easier to understand.

svaarala:
We can finish this curses change as I agree it's not necessarily a very good dependency.

But are you proposing some other changes?

TheBrokenRail:
No

svaarala:
Ok, please squash the commits for merge (see https://github.blog/2011-09-06-shiny-new-commit-styles/) and this should be done.

TheBrokenRail:
Squahed

TheBrokenRail:
Done

PR #2127, opened 2019-06-25
Rewrite Makefiles for Examples

TheBrokenRail:
This changes the Makefiles so that they now build incrementally. Previously if you wanted to rebuild an example yo would have to manually delete the binary, and rebuild the entire example, including duktape. This will only rebuild the files that have changed, so duktape will no longer be rebuilt every time, as well, as no longer requiring the user to delete the output binary manually when they want to rebuild the example. It changes the name of the output binary of ```Makefile.dukdebug``` from ```duk``` to ```dukd``` so that the command line example and the debug command line example don't produce the same binary.

### DIFF
--- a/dist-files/Makefile.cmdline
+++ b/dist-files/Makefile.cmdline
@@ -3,45 +3,57 @@
 #  The example program here is the Duktape command line tool.
 #
 
-DUKTAPE_SOURCES = src/duktape.c
+DUKTAPE_SOURCES = src/duktape.o
 
 CMDLINE_SOURCES = \
-	examples/cmdline/duk_cmdline.c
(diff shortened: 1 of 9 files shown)


TheBrokenRail:
Any thoughts @svaarala?

TheBrokenRail:
You can build the Duktape REPL with debugger support for Windows with ```make -f Makefile.dukdebug WINDOWS=1``` and build the normal Duktape REPL with linenoise with ```make -f Makefile.cmdline LINENOISE=1```.

TheBrokenRail:
@svaarala

TheBrokenRail:
Should I close this?

PR #2152, opened 2019-07-22
Include "end of input" in error message if SyntaxError happens at end of file

### LINKED ISSUE #2151: When SyntaxError occurs at end of input, note that in the error message
For example:
```
SyntaxError: parse error (line 3, end of input)
```

This would provide a useful hint when the input is truncated.

TheBrokenRail:
Fixes #2151

### DIFF
--- a/tests/ecmascript/test-dev-syntax-error-line.js
+++ b/tests/ecmascript/test-dev-syntax-error-line.js
@@ -21,7 +21,10 @@ function test(inp) {
         print('never here');
     } catch (e) {
         // Match against current syntax
-        res = /^.*\(line (\d+)\)$/.exec(e.message);
+        res = /^.*\(line (\d+), end of input\)$/.exec(e.message);
+        if (!res) {
+            res = /^.*\(line (\d+)\)$/.exec(e.message);
+        }
         print('line', res ? res[1] : 'n/a');
     }
 }
--- a/RELEASES.rst
+++ b/RELEASES.rst
@@ -3554,6 +3554,8 @@ Planned
 
 * Fix some C++ compile warnings (GH-2161)
 
+* Include "end of input" in error message if SyntaxError occurs at end of file (GH-2152)
+
 3.0.0 (XXXX-XX-XX)
 ------------------
 
--- a/src-input/duk_error_augment.c
+++ b/src-input/duk_error_augment.c
@@ -433,7 +433,11 @@ DUK_LOCAL void duk__add_compiler_error_line(duk_hthread *thr) {
 	                     (duk_tval *) duk_get_tval(thr, -1)));
 
 	if (duk_get_prop_stridx_short(thr, -1, DUK_STRIDX_MESSAGE)) {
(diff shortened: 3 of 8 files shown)


svaarala:
@TheBrokenRail Looks good, couple of notes:
- Testcases need an update for the expected output where this matters.
- It would be good to add a specific testcase for this, maybe using eval() to drive the test (cover both a SyntaxError before end-of-input, and at end-of-input).
- I'm not sure if this would work, but it might be cleaner to check if current token is DUK_TOK_EOF, rather than comparing the offset?

tests/ecmascript/test-dev-syntax-error-line.js
```diff
@@ -21,8 +21,7 @@ function test(inp) {
         print('never here');
     } catch (e) {
         // Match against current syntax
-        res = /^.*\(line (\d+)\)$/.exec(e.message);
-        print('line', res ? res[1] : 'n/a');
+        print('line', e.lineNumber);
```
svaarala: Here the idea of the testcase is to cover the line number in the `.message` property, not the `.lineNumber` property, so the best fix would be to fix the pattern.

TheBrokenRail: Fixed

TheBrokenRail:
It now has a test case, and the "end of input" message is displayed more reliably.

TheBrokenRail:
@svaarala Rebased and added to changelog!

svaarala:
Did you try comparing against DUK_TOK_EOF (as current and/or previous token)? It may be more reliable than the offset comparison. (The current comparison also involves a potential wrap if input length is 0, i.e. length - 1 == -1).

svaarala:
I guess that EOF approach won't work really well, e.g. if the input ends on a truncated token like `"truncated string`, no EOF token will be generated before a SyntaxError. So maybe comparing against the input offset is the best way. Does the current pull include "at end of input" for such a truncated string? I.e.
```
// test.js
print("truncated string
```

With the offset approach, could you change:
```
if (thr->compile_ctx->lex.input_length - 1 <= thr->compile_ctx->curr_token.start_offset) {
```
to:
```
if (thr->compile_ctx->curr_token.start_offset + 1 >= thr->compile_ctx->lex.input_length) {
```

The main point is to avoid an underflow when input_length is (start_offset and input_length are unsigned, `size_t`). Also the comparison (IMO) makes more sense in this order, i.e. adjusted offset exceeds input length.

TheBrokenRail:
@svaarala Done!

svaarala:
Looks good, could you squash the fix and I'll then merge :+1:

TheBrokenRail:
@svaarala Squashed!

svaarala:
Thanks!

svaarala:
Hmm, it seems that this approach doesn't indicate "at end of input" for truncated tokens like:

```
eval('"foo bar');
```

I'll see if this can be improved somehow.  Ideally an invalid token or a truncated token at end of input would generate the syntax error hint.

svaarala:
Managed to find a way to detect the end of input more reliably, I'll open a pull later tonight.

PR #2165, opened 2019-08-08
Improve SyntaxError "end of input" detection

### LINKED ISSUE #2151: When SyntaxError occurs at end of input, note that in the error message

svaarala:
Fixes #2151.


svaarala:
This now handles truncated tokens correctly too, e.g. `eval('"foo bar');` reports an end-of-input SyntaxError.

PR #2169, opened 2019-08-25
Update CBOR absent tag document, add assigned tag number 31

svaarala:
Finalize CBOR "omitted tag" document:
- [x] Expanded example
- [x] Rename to omitted
- [x] Usage outside of gappy arrays
- [x] Tag assignment


svaarala:
@cabo I've done some of the changes we discussed in this pull. I'll add the remaining changes tomorrow.

svaarala:
@cabo I've tried to come up with good text and examples for a more general use case but I'm having some difficulty doing so.

For example, suppose that an "omitted" value can have application specific defaulting behavior. An application would treat CBOR Undefined as is, but potentially replace an "omitted" CBOR Undefined with some default value.

There are IMO two problems with this:
- It works opposite to the ECMAScript array gap case where one specifically doesn't want any value (not even a default) to replace the tagged CBOR Undefined. Here the tag is trying to convey that no value exists at all.
- If the application processing is in essence undefined, it makes the decision to use the tag in encoding awkward: whether the tag is appropriate or not depends on the consuming application rather than shared semantics.

Maybe it's easiest to leave the semantics narrow as currently proposed and allocate a longer tag (1+2)? It's not ideal but it's still quite acceptable as a gap would then encode as 4 bytes total which is still less than e.g. an IEEE single precision float.

cabo:
Maybe it doesn't need to be as wide as encompassing all concepts of a default value. I'm starting a vacation today, but will watch this from on eye. Back on the 20th.

svaarala:
One way to clarify the "omitted" semantics in a language neutral way is a comparison to CBOR Maps.

For CBOR Maps:
- A key can be present and have the value CBOR Undefined
- A key can be absent

For CBOR Arrays there is no "absent" option because all keys in the range [0,length[ are assumed to be present. The omitted tag allows one to denote a hole in the keyspace, similarly to an absent key in a CBOR Map.

svaarala:
@cabo I pushed some updates to the proposed text:
- Rewrote the document to use "absent" rather than "omitted" (or "missing").
- Clarify the semantics in a language independent manner as described above -- the tag indicates an absent key (index) in a CBOR Array, similarly to a key absent from a CBOR Map.
- Also clarified that the intent as denoting CBOR Array gaps specifically, a feature missing from CBOR Arrays as compared to CBOR Maps.
- Other minor text edits.

The latest version is here: https://github.com/svaarala/duktape/blob/90ef4a903e587299ac88023ba8330ab0e38ac964/doc/cbor-absent-tag.rst.

svaarala:
Added the suggested future work note to the overview.

cabo:
Thank you.  As you will have noticed, this has now been registered as Tag 31; the registry entry still uses the wording "missing" and the corresponding file name.  Maybe you want to ask for a correction/update?

sva-p:
Thanks! I'll ask for an update.

svaarala:
I'll now merge the text to master, with both old and new file names (so as to not break the IANA page link). I'll ask the link to be updated, along with the description, and remove the old document when the update is complete.

PR #2191, opened 2019-10-06
Added a window.fetch polyfill

arguiot:
The polyfill was created according to GitHub's polyfill

### DIFF
--- a/polyfills/fetch.js
+++ b/polyfills/fetch.js
@@ -1,0 +1,545 @@
+/* A window.fetch polyfill for Duktape */
+
+(function() {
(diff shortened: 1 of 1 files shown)


svaarala:
What's the purpose of the polyfill? Duktape doesn't support e.g. XMLHttpRequest out of the box so the polyfill seems a bit meaningless as a general Duktape-level polyfill?

arguiot:
Oops! I didn’t know that Duktape wasn’t supporting XMLHttpRequest, so yeah then it’s useless...

Issue #2190, opened 2019-10-03
API for external structures to behave like JS values from within duktape


jchlapinski:
@svaarala Do You think it is possible/practical to implement in Duktape an API to allow for external data structures to appear as JS values from within duktape JS code?

My usecase is as follows: we have a DOM-like structure of JSON-type data in our program. At the moment in order to access that data as regular JavaScript values from Duktape, we have to convert our DOM tree into a proper object graph inside Duktape's heap (using existing API). This obviously introduces significant performance penalty since the data must be effectively copied back and forth between our structres and internal duktape's objects.

Can this be done better? Is there a way to develop an API in duktape that would allow an external structure to behave like a normal JS value on the heap?

I could contribute this feature myself, however first I would like to know if this is even viable, and I would maybe need some pointers to start.

svaarala:
At present the best approach is using a Proxy object. A Proxy can capture various operations (like [[Get]], [[Set]], [[DefineOwnProperty]] etc) and handle them in trap functions.

It would be possible to make a more native version of Proxy (this has been requested many times) but it would in essence still be the same model: JS operations would be captured and handled in user provided callbacks.

jchlapinski:
thanks for the info, however a Proxy approach is not very feasible. It requires to create a proxy object for every value, as well as a dummy object - proxy target. I do not see an easy way to tie a JS value with external value (like store external pointer within value), except maybe creating a user-accessible property in objects.

I have prepared a PR #2192, where I am investigating a more direct approach to provide an API for "virtual" values.

aasselin:
I have a similar problem and wonder if the approach would work: is it possible to create a proxy object with a target that would be a bare "void*" pointer ? (representing always a pointer to an interface for a kind of scriptable object, i.e. JavaScriptableObject* are something alike) and the various Get, Set, ... would be implemented based on this interface.
Can I set a finalizer on a Proxy as well?

svaarala:
Proxy target must be an object, but it could be an object coerced pointer, Duktape.Pointer:
```
duk> new Proxy(new Duktape.Pointer('foo'), {})
= {}
```

The underlying value of a Duktape.Pointer is a `void *` in the C API.

PR #2192, opened 2019-10-07
[wip] "external" stack value support

jchlapinski:
This PR is an early attempt to add and API for externally handled stack values, as mentioned in #2190 

This is preliminary work to see if such a change is viable to be merged into this project.

Basic idea is that we have an additional stack value type with a tag `DUK_TAG_EXTVAL`, available only for an unpacked stack value representation (as it requires 16 byte length for stack values).
Whenever a type/conversion of such value is needed, external API (modeled as a number of function pointers in `duk_hthread` struct) is used. At the moment a few performance penalties were introduced (like in `duk_get_type_tval`) in form of an additional condition.

This code was not tested, and it is definitely not ready to be merged. It is only meant to provide a background for a discussion, whether such an API has its place in duktape, and whether I should proceed with this development.

Regards.

### DIFF
--- a/src-input/duk_api_inspect.c
+++ b/src-input/duk_api_inspect.c
@@ -72,7 +72,7 @@ DUK_EXTERNAL void duk_inspect_value(duk_hthread *thr, duk_idx_t idx) {
 	tv = duk_get_tval_or_unused(thr, idx);
 	h = (DUK_TVAL_IS_HEAP_ALLOCATED(tv) ? DUK_TVAL_GET_HEAPHDR(tv) : NULL);
 
-	vals[DUK__IDX_TYPE] = duk_get_type_tval(tv);
+	vals[DUK__IDX_TYPE] = duk_get_type_tval(thr, tv);
 	vals[DUK__IDX_ITAG] = (duk_int_t) DUK_TVAL_GET_TAG(tv);
 
 	duk_push_bare_object(thr);  /* Invalidates 'tv'. */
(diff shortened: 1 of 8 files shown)


jchlapinski:
A few problems/remarks so far:
# There is no free value for additional stack value tag that would have bit 3 off meaning it is not heap allocated). Can we use `DUK_TAG_UNUSED` for this purpose, and map all of `DUK_TAG_UNUSED` into `DUK_TAG_UNDEFINED`? From a first glance this appears to be possible, but maybe there is an explicit case somewhere that requires distinction between undefined and unused. I think those should be somewhat equivalent?
# This API obviously lacks support for has/get/set property within external value. Those are of course planned to be added later on.

aasselin:
Hi,

is there any activity on this pull request ? I have a similar need, where c++ objects would be made directly accessible to javascript.
If this pull request has a chance to go through i'd be ready to help code as well.

Regards
Armel

jchlapinski:
@aasselin TL;DR this approach turned out to be not feasible. I simply forgott to close this PR as invalid.

I tried to add a new stack value type representing `external` javascript value (a string, integer, object, etc.). As long as using such external values in simple cases is easy enough, I stumbled upon a big error in my design - how to mix native and external objects as respective properties.

As long as it is easy to embed `external` object in native duktape's objects as a property, the reverse is actually not feasible IMO.
Consider the following example:
```javascript
var external_object = ... //here we have some externally represented object
var internal_object = {
  string_propl: "string value",
  num_prop: 10
};

internal_object.external_prop = external_object; //this is simple enough IMO
external_object.internal_prop = internal_object; //but this is just not feasible 
```
Basically whenever we would try to insert a native object as a property into the external object, we would probably need to convert the native object into external object first (which could be costly), and then the following would not hold:
```javascript
// the following should hold (see above code), but it might not if the internal object was converted (copied) into external structure
if (internal_object === external_object.internal_prop) {
}
```

There are of course ways to circumvent those problems but the whole approach quickly becomes just unmaintainable. Much better way to suit our use case would be to:
1) prepare a public C API to directly use internal duktape's objects (basically the content of `duk_hobject.c`) 
2) redesign duktape to allow for exchanging internal representation to a user supplied data structures

Hope that helps, good luck
Jakub

PR #2199, opened 2019-11-15
Allow DUK_USE_EXEC_TIMEOUT_CHECK to be called from within the main loop of a long lived regular expression

braindigitalis:
I am using duktape in an environment where it may be asked to run untrusted code as part of a platform as a service, and the function ``duk__match_regexp()`` can potentially run for a very long time if an abusive or broken regular expression is passed to it. This is somewhat mitigated by the value of ``DUK_RE_EXECUTE_STEPS_LIMIT``, lowering this can solve the problem but it seems this cannot be lowered from duk_config.h, and i would have to edit my duktape.c to fix it.

Based upon this i decided on this fix instead - i am already using ``DUK_USE_EXEC_TIMEOUT_CHECK`` with an interrupt function to prevent cpu consumption and various forms of resource abuse, but it seems the timeout function is not called from within the ``duk__match_regexp()`` function.

This PR adds support for calling the interrupt function from within the loop, this doesnt change any specification i'm aware of, as the docs say just that the interrupt function is 'called periodically' while running the script. 

Please consider either merging this PR in some form, or allowing user configuration of ``DUK_RE_EXECUTE_STEPS_LIMIT`` and ``DUK_RE_COMPILE_TOKEN_LIMIT``.

Thanks!

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -64,6 +64,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Luis de Bethencourt (https://github.com/luisbg)
 * Ian Whyman (https://github.com/v00d00)
 * Rick Sayre (https://github.com/whorfin)
+* Craig Edwards (https://github.com/braindigitalis)
 
 Other contributions
 ===================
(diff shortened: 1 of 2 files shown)


svaarala:
@braindigitalis Could you rebase to master, it should fix the failing checks?

There are a couple of small things that come to mind from the diff (still worth merging as is):
- There should be an internal helper to handle throwing a script timeout to reduce code duplication.
- Clearing the heap interrupt running flag is unnecessary but harmless.
- The timeout condition is now checked for every step, it might be better to check for every 1024 steps or so, to avoid the macro call (which may be much more expensive than a counter check, e.g. it might read system time).

I can make these adjustments in a follow-up.

PR #2229, opened 2020-02-17
Replace "Fork me on GitHub" for GitHub Corner

mauriciabad:
[GitHub Corners](http://tholman.com/github-corners/) is "the modern version" of [GitHub Ribbons
](https://github.blog/2008-12-19-github-ribbons/)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -65,6 +65,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Ian Whyman (https://github.com/v00d00)
 * Rick Sayre (https://github.com/whorfin)
 * Craig Leres (https://github.com/leres)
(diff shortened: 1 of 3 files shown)


svaarala:
What does the site-middle margin change fix/change?

Could you please add yourself to AUTHORS.rst, the first list (not the second), see CONTRIBUTING?

mauriciabad:
The margin was there because the ribbon was quite big.
Because the black corner is smaller there's no need for that margin.

svaarala:
Thanks!

svaarala:
Hmm, it seems that the missing margin breaks the layout of some pages, e.g. API and Guide. The `<h1>` level title is directly after the navigation bar which looks bad. I'll try to fix it in a follow-up.

PR #2239, opened 2020-02-29
Specify python2.7 in runtests, as Python 3 is `python` these days

hf:
This is a breakout PR from #2219.

`runtests` require Python 2.7 however these days Python 3 is the default runtime on most linux distributions, making it difficult to run the tests locally.

### DIFF
--- a/runtests/runtests.js
+++ b/runtests/runtests.js
@@ -325,7 +325,7 @@ function executeTest(options, callback) {
         args.push('--output', tempInput)
         args.push('--prologue', tempPrologue)
 
-        child_process.execFile('python', args, {}, compileDone)
+        child_process.execFile('python2.7', args, {}, compileDone)
     }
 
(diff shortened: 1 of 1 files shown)


hf:
@svaarala's comment for reference: https://github.com/svaarala/duktape/pull/2219#discussion_r373684398

> There are unfortunately some differences between various python environments. Other parts of the codebase detect the proper name; for example, Makefile uses:
> 
> ```
> PYTHON := $(shell { command -v python2 || command -v python; } 2>/dev/null)
> ```
> 
> I'd take this diff out of the pull and figure it out in a separate pull.

svaarala:
I didn't like hardcoding the Python version in runtests either, so I opened #2245 which now defaults to `python2` (which seems better than `python2.7` to me) and allows the Makefile to pass in the detected Python command with `--python-command`.

PR #2260, opened 2020-03-30
Makefile.sharedlibrary: move -lm from duk to libduktape.so

Hello71:


Hello71:
actually there are some other issues with Makefile.sharedlibrary, I'll do them together

PR #2267, opened 2020-04-07
Remove endianness assumption for PPC64

### LINKED ISSUE #2265: PowerPC architecture and byte order
PowerPC appears to automatically use big endian byte order while POWER8 introduces ppc64le mode. Maybe it should use autodetect like it's done for arm?

svaarala:
Fixes #2265.


andrismednis:
The fix works on POWER 9 RHEL 8.1 LE. Thanks!

PR #2282, opened 2020-04-26
Deactivate thread while suspended

oleavr:
Even though another native thread runs code on a different Duktape
thread, its code may trigger finalizers, and those always run on the
heap thread. And the heap thread may have been the one that was
suspended. This would in turn result in `duk__call_thread_state_update()`
failing, and finalizers being skipped, typically resulting in memory
leaks.

### DIFF
--- a/src-input/duk_api_heap.c
+++ b/src-input/duk_api_heap.c
@@ -10,6 +10,7 @@ struct duk_internal_thread_state {
 	duk_ljstate lj;
 	duk_bool_t creating_error;
 	duk_hthread *curr_thread;
+	duk_uint8_t thread_state;
 	duk_int_t call_recursion_depth;
 };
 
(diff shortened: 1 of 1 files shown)


svaarala:
Thanks!

Do you have a repro case?

If not, I can try to come up with one to cover this in the API tests. I guess one would just create a bunch of finalizable objects, suspend the main thread, make the objects unreachable and force a GC. Then observe that the finalizers are actually called to verify this issue didn't happen.

svaarala:
Okay, managed to create a repro => https://github.com/svaarala/duktape/pull/2285.

Before fix:
```
*** test_basic (duk_safe_call)
other thread executing
forcing gc
forced gc done, now resume
main thread back
final top: 0
==> rc=0, result='undefined'
```

After fix:
```
+*** test_basic (duk_safe_call)
+other thread executing
+finalizer 2 called!
+forcing gc
+finalizer 1 called!
+forced gc done, now resume
+main thread back
+final top: 0
+==> rc=0, result='undefined'
```

oleavr:
Yay, thanks a lot! (Didn't have a repro case isolated to Duktape, so that's awesome! 🤘)

Issue #2319, opened 2020-06-11
Fuzzilli Integration


WilliamParks:
This issue is to cover integration with [Fuzzilli](https://github.com/googleprojectzero/fuzzilli).

Design decisions:
- What duktape branch should be targeted? I'm currently running against v2-maintenance, but currently believe it's branch agnostic. This should be tested either way.
- What duktape builtins should be included? It's a tradeoff of efficiency of the fuzzer (`Duktape.version` is probably not going to produce a bug) vs coverage.

WilliamParks:
Additional issue, that I'm not sure how to resolve. Since the default build process for duktape puts everything into duktape.c, the resulting coverage number is poor and not representative of the actual engine coverage. The number given by the fuzzer includes the C/C++ API (https://duktape.org/api.html), of which Fuzzilli only uses a small portion.

For reference, after ~21 core-days, coverage is at 34%

svaarala:
It's probably easiest if the changes are made to master, and then cherry-picked to v2-maintenance if that's straightforward.

svaarala:
Regarding that coverage issue, it's (still) possible to build Duktape from separate source files, and with or without `#line` directives in the combined source (if that helps).

I've also run some coverage analysis on how much of the code base is covered by current tests. Coverage from both ECMAScript and API tests is pretty good, but from ECMAScript tests alone API test coverage is low, just like with fuzzing.

WilliamParks:
> It's probably easiest if the changes are made to master, and then cherry-picked to v2-maintenance if that's straightforward.

I have it based on master now, and generally cleaned up. Shouldn't be an issue!

I'm working on the fuzzilli side, and should have something together shortly (it fuzzes fine locally, just want to get it into their docker build process). 

From there, a round of iteration on which builtins are used/how its compiled/etc should help improve coverage.

WilliamParks:
The outstanding items:
- Figure out the proper place for the assert (duk_fuzzillia.c needs to be able to intentionally fail an assert. as part of the startup checks
-  Consider how to improve coverage, both by removing dead code in the compilation target, and verifying which duktape unique builtins should be included (list [here](https://github.com/WilliamParks/fuzzilli-1/blob/duktape_integration/Sources/FuzzilliCli/Profiles/DuktapeProfile.swift#L32))
- Clean up the lint errors

WilliamParks:
[...]

PR #2321, opened 2020-06-14
Base integration of duktape with the Fuzzilli Javascript engine fuzzer

WilliamParks:
This covers basic integration with Fuzzilli. It can be used with my branch of Fuzzilli [here](https://github.com/WilliamParks/fuzzilli-1/tree/duktape_integration), with run commands listed in examples/fuzzilli/README.rst in this PR

Still a little ways out from being fully integrated, as optimizations of the fuzz target are needed, tracked in #2319

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 /dukd
 /dukd.*
 /duk-clang
+/duk-fuzzilli
 /duk-sanitize-clang
 /duk-g++
 /duk-size
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -67,6 +67,7 @@ and agreed to irrevocably license their contributions under the Duktape
 * Craig Leres (https://github.com/leres)
 * Maurici Abad (https://github.com/mauriciabad)
 * Nancy Li (https://github.com/NancyLi1013)
+* William Parks (https://github.com/WilliamParks)
 
 Other contributions
 ===================
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,7 @@ CONFIGOPTS_DEBUG_SCANBUILD = --option-file util/makeduk_base.yaml --option-file 
 CONFIGOPTS_DEBUG_ROM = --rom-support --rom-auto-lightfunc --option-file util/makeduk_base.yaml --option-file util/makeduk_debug.yaml -DDUK_USE_ROM_STRINGS -DDUK_USE_ROM_OBJECTS -DDUK_USE_ROM_GLOBAL_INHERIT -UDUK_USE_HSTRING_ARRIDX
 CONFIGOPTS_EMDUK = -UDUK_USE_FASTINT -UDUK_USE_PACKED_TVAL
 CONFIGOPTS_DUKWEB = --option-file util/dukweb_base.yaml --fixup-file util/dukweb_fixup.h
+CONFIGOPTS_FUZZ = --option-file util/makeduk_base.yaml --option-file util/makeduk_fuzz.yaml
 
 # Profile guided optimization test set.
 PGO_TEST_SET = \
@@ -266,7 +267,7 @@ clean:
 	@rm -rf /tmp/dukweb-test/
 	@rm -f massif-*.out
 	@rm -f literal_intern_test
-
+	@rm -f duk-fuzzilli
 .PHONY: cleanall
 cleanall: clean
 	# Don't delete these in 'clean' to avoid re-downloading them over and over
@@ -325,6 +326,9 @@ prep/nondebug-rom: prep
 prep/debug: prep
 	@rm -rf ./prep/debug
 	$(PYTHON) tools/configure.py --output-directory ./prep/debug --source-directory src-input --config-metadata config $(CONFIGOPTS_DEBUG) --line-directives
+prep/fuzz: prep
+	@rm -rf ./prep/fuzz
+	$(PYTHON) tools/configure.py --output-directory ./prep/fuzz --source-directory src-input --config-metadata config $(CONFIGOPTS_FUZZ) --line-directives
 prep/debug-scanbuild: prep
 	@rm -rf ./prep/debug-scanbuild
 	$(PYTHON) tools/configure.py --output-directory ./prep/debug-scanbuild --source-directory src-input --config-metadata config $(CONFIGOPTS_DEBUG_SCANBUILD) --separate-sources --line-directives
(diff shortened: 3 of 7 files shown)


svaarala:
Thanks, I'll look this over tonight. One question at the outset:

Are the changes required to duk_cmdline.c large enough to warrant a separate duk_fuzzilli.c file? If the changes needed can be managed with some `#ifdef`s then it'd be great to avoid duplicating much of the code. (I'll check the concrete diff tonight when I get a chance.)

WilliamParks:
> Are the changes required to duk_cmdline.c large enough to warrant a separate duk_fuzzilli.c file? If the changes needed can be managed with some #ifdefs then it'd be great to avoid duplicating much of the code. (I'll check the concrete diff tonight when I get a chance.)

As of right now, most likely not. However, I'm planning on doing an iteration of over duk_fuzzilli to remove all the unnecessary options for it/checks/etc on startup. One of the performance limitations right now is that the timeout for duktape is > 2x what it is for other engines, which I believe is due to the startup costs.

If that's wrong or I can't get a meaningful speedup, I can move it back into duk_cmdline.c

svaarala:
> One of the performance limitations right now is that the timeout for duktape is > 2x what it is for other engines, which I believe is due to the startup costs.

That's surprising considering that e.g. Node.js startup is way higher than for Duktape (for understandable reasons):
```
$ time ./duk /tmp/hello.js 
hello world!

real	0m0.002s
user	0m0.000s
sys	0m0.002s

$ time nodejs /tmp/hello.js 
hello world!

real	0m0.041s
user	0m0.039s
sys	0m0.000s
```

Could the same running process be reused for multiple tests?

WilliamParks:
> Could the same running process be reused for multiple tests?

Currently, Fuzzilli runs 100x unique input programs in one process, and resets the engine state between each one!

I looked at how Fuzzilli calculates its recommended timeouts, and ran some tests. It runs a loop 1000 times that makes an object and calls a function that does property look ups. Their builder for that is [here](https://github.com/googleprojectzero/fuzzilli/blob/master/Sources/Fuzzilli/Fuzzer.swift#L313)

Setting the loop to run just once results in similar times for duktape and V8, which disproves my theory. It makes sense given that V8 and Duktape are built for different use cases.

Fuzzilli recommended timeout for 1000x executions of the loop
V8: 100ms 
Duktape:  790ms

Fuzzilli recommended timeout for 1x execution of the loop
V8: 70ms
Duktape: 100ms

That leaves the following TBD:
- Deciding if it's worth the time to reduce the unused code compiled into duk-fuzzilli, in order to get more meaningful coverage numbers
- Deciding whether duk-fuzzilli should be it's own folder in /examples, or based on defines in duk_cmdline

svaarala:
Since Duktape is a plain bytecode interpreter (no JIT), it is typically 10-100x slower than a JIT engine. This will probably be an important factor to consider w.r.t. timeouts.

svaarala:
Would you like this to be merged, with follow-ups for cleanups? If so, could you please rebase into a single commit (or a few topic commits)?

WilliamParks:
> Would you like this to be merged, with follow-ups for cleanups? If so, could you please rebase into a single commit (or a few topic commits)?

I'd like to wait until we've figured out the coverage issue, and whether or not it should be it's own folder vs defines in duk_cmdline, if that's alright with you. Once I get your thoughts on those two issues, I'll implement them and clean up the commit history.

svaarala:
Sure, no hurry to merge :)

WilliamParks:
I moved things back to duk_cmdline, and squashed the commits. Should be good to merge!

Running the fuzzer can be done with `swift run -c release FuzzilliCli --profile=duktape --storagePath=./output --timeout=1000 ./duk-fuzzilli`, from my branch of fuzzilli [here](https://github.com/WilliamParks/fuzzilli-1/tree/duktape_integration)

Running in a docker container from the main fuzzilli repo will be a straightforward addition once I have a permanent place to point the docker build script at.

Also, @svaarala what's the best place to put a short piece of documentation (mainly the above) in duktape on how to run fuzzilli?

svaarala:
Thanks, I'll check the diff out by tomorrow.

Re: dockerization, Docker usage in this repo is a work in progress:
- docker/ contains various Docker images, most sharing a (large) base image such as docker/duktape-base-ubuntu-18.04-x64.
- Images are not (yet) published, but one builds them locally using e.g. `make docker-images`.
- Docker images are not yet used for anything critical in the repo, but my intent is to rework most targets to rely on Docker to reduce the developer hassle with host differences. If one doesn't want to use Docker, the Dockerfiles are still useful as they basically document the required packages.
- Docker images by convention pass input and output files as ZIP files via stdin/stdout. This may seem odd, but it avoids file owner and permission issues.

So for now, it'd be fine to just add a new directory to docker/. If it's possible, it would be nice to reuse a base image like docker/duktape-base-ubuntu-18.04-x64 because it reduces the disk requirements. But it'd be also fine to download a prebuilt image (or a Dockerfile) from somewhere as long as that image is available indefinitely.

Re: documentation, one easy approach is to add a fuzz target to the Makefile and just refer to it from the README ("Run `make xxx` for Fuzzilli-based automatic fuzzing tests, see the Makefile target for details."). If you want to document the setup more, `doc/fuzzilli.rst` would be a good place.

svaarala:
I'll merge this as is, and clean up some trivia in a follow-up. We can then look at next steps. Thanks! :+1:

svaarala:
@WilliamParks Here's the follow-up: https://github.com/svaarala/duktape/pull/2329. Most of it is indentation and style trivia - you may want to read with `git diff -w` to avoid the whitespace diff (hard tabs should be used throughout).

Could you check that fuzzing still works i.e. I didn't break anything before I merge? :)

Substance changes:
- Simplify fuzzilli() binding by using nargs=2 when registering the function, and using duk_require_string() instead of checking for types.
- Declare duk_assert_wrapper() in duktape public header when DUK_USE_FUZZILLI is enabled, as it is in effect part of the API in this case. (Maybe duk_assert() could be exposed just generally.)
- Hoist declarations of variables to beginning of block (Duktape code base is C89 compatible generally, although it doesn't matter here).

When the REPRL loop is running, should there be a clean exit path, or does the loop (if it starts) always terminate via an error?

WilliamParks:
> Could you check that fuzzing still works i.e. I didn't break anything before I merge? :)

I'm going to run it for a couple hours as a sanity check (looks good right now), and look through your changes in the mean time. 

> So for now, it'd be fine to just add a new directory to docker/. If it's possible, it would be nice to reuse a base image like docker/duktape-base-ubuntu-18.04-x64 because it reduces the disk requirements. But it'd be also fine to download a prebuilt image (or a Dockerfile) from somewhere as long as that image is available indefinitely.

There currently aren't any Fuzzilli images that get pushed out. That repo has build scripts to easily package everything, however [link](https://github.com/googleprojectzero/fuzzilli/tree/master/Cloud/Docker). I've got a version of that for duktape implemented locally. What'd be the best way of integration? A script to pull down the fuzzilli repo and run it's build script would be straightforward, if a bit convoluted.

svaarala:
Sounds like adding two targets to the Makefile might be the easiest first step, something like this:
- `make docker-image-fuzzilli`: clone the fuzzilli repo, build the Docker image locally, tag it as `duktape-fuzzilli` or similar. Include the duktape-fuzzilli docker image in the `docker-clean` target.
- `make fuzztest`: run fuzzilli testing, expecting the tagged Docker image to exist locally.

I don't think it needs to be cleaner than this at present.

WilliamParks:
> When the REPRL loop is running, should there be a clean exit path, or does the loop (if it starts) always terminate via an error?

The fuzzer itself sends sigkill to the child to end the loop. Implemented [here](https://github.com/googleprojectzero/fuzzilli/blob/2e7c967d171a5059aff6f9ab8df6be0e36216018/Sources/Fuzzilli/Execution/REPRL.swift#L121)

> Sounds like adding two targets to the Makefile might be the easiest first step, something like this:

Makes sense. After the cleanup merge is complete, I'll push the docker implementation on the fuzzilli side, and finish things up with the make target here.

PR #2333, opened 2020-06-19
Added make targets for fuzzilli

WilliamParks:
Adds three make targets to run fuzzilli on duktape

`docker-image-fuzzilli`: Builds the docker image
`fuzztest`: Runs fuzzilli until stopped, assuming the docker container exists.
`get-fuzzer-results`: Gets the results out of the container, and deletes it. My reasoning here is that this is cleaner than trying to do it after `fuzztest` has been stopped, as this ensures that obtaining the results isn't accidentally stopped in stopping the fuzzer itself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -603,6 +603,31 @@ test: apitest ecmatest
 	@echo ""
 	@echo "### Tests successful!"
 
+# Build targets for the integeration of the fuzzilli fuzzer
+docker-image-fuzzilli: build/duk-fuzzilli
+	-git clone https://github.com/googleprojectzero/fuzzilli ./deps/fuzzilli
+	mkdir -p deps/fuzzilli/Cloud/Docker/DuktapeBuilder/out
+	cp build/duk-fuzzilli deps/fuzzilli/Cloud/Docker/DuktapeBuilder/out/
+	cd deps/fuzzilli/Cloud/Docker; ./build.sh fuzzilli # Don't use duktape build option here, as duk-fuzzilli is already present
+
+# Runs until stopped
+.PHONY: fuzzillitest
+fuzzillitest:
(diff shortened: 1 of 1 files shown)


svaarala:
I merged some cleanups to the Makefile which now causes a conflict - could you rebase and force push?

The main change in the Makefile is to avoid temporaries in the top level directory. Directories are used instead to make it easier to manage cleanup:
- deps/ for dependencies like external projects
- tmp/ for temp files, test run result files, etc
- build/ for build results, e.g. duk-fuzzilli now builds to build/duk-fuzzilli
- dist/ for dist results, dist/source and dist/site

I'd suggest these changes:
- Use deps/fuzzilli for the cloned fuzzilli repo
- Rename fuzztest to fuzzillitest for clarity
- Rename get-fuzzer-results to get-fuzzilli-test-results for clarity
- Use tmp/fuzzilli_results for the result file

WilliamParks:
Fuzzilli just updated how they do their read-eval-print-reset loop, which broke the implementation (also could be that I tested it out on a different OS than I did my original dev on).

I definitely plan to get it working, but it's going to be a bit before I can get to it.

svaarala:
No hurry on that front - could you squash the commits in this pull and I'll merge it, with further changes at some point?

WilliamParks:
That should do it.

There is one potential issue, in that building on the user's native system other than Linux, and running in Docker, results in issues. Docker on Mac/Windows runs in a Linux VM (as I'm learned while debugging), and the current build process produces an executable on the native system.

svaarala:
Ok, it's straightforward to build duk-fuzzilli with docker. I'll make that change for other targets too later on to minimize host dependencies.

svaarala:
Thanks!

PR #2342, opened 2020-07-08
Cleanups for fuzzilli Makefile targets

svaarala:
- Combine get-fuzzilli-test-results into fuzzillitest.
- Add fuzzillitest-stop target for convenience.
- Relocate docker-image-fuzzilli to docker section of Makefile.
- Deps handling cleanup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -603,30 +603,23 @@ test: apitest ecmatest
 	@echo ""
 	@echo "### Tests successful!"
 
-# Build targets for the integeration of the fuzzilli fuzzer
-docker-image-fuzzilli: build/duk-fuzzilli
-	-git clone https://github.com/googleprojectzero/fuzzilli ./deps/fuzzilli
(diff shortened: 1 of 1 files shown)


svaarala:
@WilliamParks Mostly trivia in this pull, one relevant change is that I merged `get-fuzzilli-test-results` into `fuzzillitest`. So the usage would now be:
```
# In terminal 1.
$ make docker-image-fuzzilli
$ make fuzzillitest

# In terminal 2, when one wants to finish.
$ make fuzzillitest-stop
```

When fuzzillitest-stop is executed, the fuzzillitest Makefile target proceeds to pull out the result and inform the user where the result files are.

Does this make sense to you, or was there a specific reason why the result pulling should have a separate target?

svaarala:
And to confirm, I can also now easily run fuzz tests, thank you! :+1:

WilliamParks:
I'm glad it works well!

I didn't have a clear plan on how to run it for an indeterminate amount of time, and then cleanly obtain the results.  The current solution looks good to me!

Issue #1794, opened 2017-11-05
Make `tools/configure.py` work with Python 3.x


ghost:
Currently `tools/configure.py` fails with an error when run with Python 3. It makes things complicated when your build system already uses other scripts done in Python 3. In the perfect scenario both flavors, Py2 and Py3, would be supported.

svaarala:
I tried doing that conversion a while back - it's really painful for the same Python source to work in both environments.

svaarala:
One option I've considered is moving away entirely from Python because of the Python2/Python3 rift (which is awkward for everyone involved). A logical candidate would be Node.js which is also well supported on major platforms. Another approach would be to be self-hosting, i.e. compile a command line tool with the necessary file I/O bindings for doing the configuration step. But no actions on this so far.

fatcerberus:
Node.js is nice but seems like a very heavy dependency, even heavier than Python.  I rather like the idea of self-hosting, personally.  I imagine that would be a very large undertaking, though.

fatcerberus:
Hmm... Is there any benefit nowadays to retaining Python 2 support at all?  I feel like we could upgrade the build tools to Python 3-only and it wouldn't be an issue, as long as there's a way to ensure you run the correct one on systems with both installed.

svaarala:
Other than breaking existing builds I think Python 3 only might be a viable option today. So that change could be made for Duktape 3.x, but I wouldn't want to make it for Duktape 2.x because existing builds would quite likely break.

Self-hosting wouldn't be a large undertaking as far as the build script functionality is concerned. The main concern would be a build dependency to maybe a different Duktape version to run the configuration tooling. It'd be easy if there was a pre-packaged command line binary for major platforms, but lacking that it seems awkward to use (and document in an understandable fashion). So if this approach was taken, it'd be good to release a pre-packaged binary tool for the most common platforms.

fatcerberus:
For what it’s worth I just found out Python 2 won’t be maintained at all after 2020, so that’s probably a good reason to migrate sooner rather than later.

svaarala:
That might pose a risk for availability but for the tooling itself it's not critical that the implementation is maintained (e.g. for security issues), because it's so far removed from actual execution. But certainly availability will eventually become worse closer to 2020. But even so, I don't think it's an immediate concern for at least a few years.

svaarala:
[...]

PR #2375, opened 2020-11-30
Fix the buildsite.py script to be python3 compatible

podsvirov:
Todo:
- [x] Fix the buildsite.py print calls to be python3 compatible
- [ ] Add to buildsite.py script python3 compatible md5 alternative
- [ ] Fix the buildsite.py script files read/write errors around python3's bytes type

### DIFF
--- a/website/buildsite.py
+++ b/website/buildsite.py
@@ -712,7 +712,7 @@ def generateApiDoc(apidocdir, apitestdir):
             apifiles.append(filename)
     apifiles.sort()
     #print(apifiles)
-    print '%d api files' % len(apifiles)
+    print('%d api files' % len(apifiles))
 
     # scan api testcases for references to API calls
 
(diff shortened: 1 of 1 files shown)


podsvirov:
@svaarala are there migration plans to python3 or the changes should be beat compatible with python2?

svaarala:
All Python code will be removed and replaced with JS code. There is a replacement for the entire configure.py tooling for example and the Py files will first become wrappers to launch that (Py2+Py3 compatible) and then go away entirely.

In the meanwhile I intend to keep the Docker images working to run buildsite etc with Python2. While Python2 is going away, it will be fine for this kind of tooling (say site build) despite being out of maintenance.

svaarala:
Example of configure.py conversion: https://github.com/svaarala/duktape/issues/1794#issuecomment-725479267

podsvirov:
> All Python code will be removed and replaced with JS code. There is a replacement for the entire configure.py tooling for example and the Py files will first become wrappers to launch that (Py2+Py3 compatible) and then go away entirely.

Interesting initiative, especially part of the interpretation of the tooling scripts by the `ductape` itself.

> In the meanwhile I intend to keep the Docker images working to run buildsite etc with Python2. While Python2 is going away, it will be fine for this kind of tooling (say site build) despite being out of maintenance.

Does this mean that if these changes are compatible with python2, then the changes will be applied?

svaarala:
I don't think it makes sense to support py2+py3 for the buildsite scripts (or any other Python code in the repo): it will create extra work in testing changes (hopefully few) with little benefit as py2 runs fine for site builds etc, especially since it's dockerized.

podsvirov:
Okay, then I won't make an effort in that direction.

PR #2418, opened 2021-09-07
Buffer#toString throw on unsupported encodings

### LINKED ISSUE #2417: `Buffer#toString` should throw error on unsupported encoding
First off, thanks for the amazing work on duktape!

The current behavior of `Buffer#toString` is always to encode the string as UTF8, even if the encoding is something else (like `"binary"` for binary strings).  Missing encoding issues have been discussed in other issues.

By default, NodeJS throws an error when the encoding is unknown.  This behavior has been true since the `0.8.x` series in 2014. Testing with `new Buffer([1,2,3]).toString("wtf")`:

| Node Version | Behavior                                                  |
|:-------------|:----------------------------------------------------------|
| ` 0.8.28`    | `Error: Unknown encoding`                                 |
| ` 0.10.48`   | `TypeError: Unknown encoding: wtf`                        |
| ` 0.12.18`   | `TypeError: Unknown encoding: wtf`                        |
| ` 4.9.1`     | `TypeError: Unknown encoding: wtf`                        |
| ` 6.17.1`    | `TypeError: Unknown encoding: wtf`                        |
| ` 8.17.0`    | `TypeError: Unknown encoding: wtf`                        |
| `10.24.1`    | `TypeError [ERR_UNKNOWN_ENCODING]: Unknown encoding: wtf` |
| `12.22.6`    | `TypeError [ERR_UNKNOWN_ENCODING]: Unknown encoding: wtf` |
| `14.17.6`    | `TypeError [ERR_UNKNOWN_ENCODING]: Unknown encoding: wtf` |
| `16.8.0`     | `TypeError [ERR_UNKNOWN_ENCODING]: Unknown encoding: wtf` |

It is straightforward to test for unexpected behavior downstream, but it would be consistent with NodeJS to throw an error for unsupported encodings.  If this change is acceptable we can send a PR

SheetJSDev:
Fixes #2417 

New behavior: if argument is not undefined/unspecified or `"utf8"`, throw an error.

Note: `null` is not a supported encoding in NodeJS


src-input/duk_bi_buffer.c
```diff
@@ -1196,7 +1197,18 @@ DUK_INTERNAL duk_ret_t duk_bi_nodejs_buffer_tostring(duk_hthread *thr) {
...
+		encoding = "utf8";
+	} else if (duk_is_string(thr, 0)) {
+		encoding = duk_to_string(thr, 0);
+		DUK_ASSERT(duk_is_string(thr, 0));
+		if(DUK_STRCMP(encoding, "utf8") != 0) {
+			DUK_DCERROR_TYPE_INVALID_ARGS(thr);
+		}
+	} else {
+		DUK_DCERROR_TYPE_INVALID_ARGS(thr);
+	}
```
svaarala, line 1211: This block might be replaced by:
```
encoding = duk_opt_string(thr, 0, "utf8");
if (DUK_STRCMP(encoding, "utf8") != 0) {
        DUK_DCERROR_TYPE_INVALID_ARGS(thr);
}
```

duk_opt_string() should:
- Accept a string
- Return "utf8" when the value is undefined or the index is out-of-range
- TypeError on anything else, including a `null`

src-input/duk_bi_buffer.c
```diff
@@ -1183,6 +1183,7 @@ DUK_INTERNAL duk_ret_t duk_bi_uint8array_plainof(duk_hthread *thr) {
 
 #if defined(DUK_USE_BUFFEROBJECT_SUPPORT)
 DUK_INTERNAL duk_ret_t duk_bi_nodejs_buffer_tostring(duk_hthread *thr) {
+	const char* encoding;
```
svaarala, line 1186: Style trivia: `const char *encoding`.

tests/ecmascript/test-bi-nodejs-buffer-tostring.js
```diff
@@ -241,20 +240,21 @@ function nodejsBufferToStringTest() {
...
     // Without arguments encoding defaults to UTF-8 and the entire
-    // buffer is converted to string.  At least undefined and null
+    // buffer is converted to string.  At least undefined
     // are accepted as "not defined" for encoding.
     b = new Buffer('ABC');
     safePrintString(b.toString());
     safePrintString(b.toString(undefined));
-    safePrintString(b.toString(null));
+    // null is not a valid encoding
+    try { safePrintString(b.toString(null)); } catch(e) { }
```
svaarala, line 249: This covers the `null` argument, but the catch clause should probably print something like `print(e.name)` (i.e. TypeError) so it's covered in the expected output explicitly.

Also there should now be explicit coverage for these encodings at least:
- `utf8`
- `utf-8`
- `wtf`

Node.js seems to accept both `utf8` and `utf-8` so it'd be nice to cover that behavior explicitly, whatever the current behavior is.

svaarala (changes requested):

src-input/duk_bi_buffer.c
SheetJSDev: The logic has been factored out since the same check is required for `Buffer.isEncoding`.

SheetJSDev:
The core logic has been factored out since both `Buffer.isEncoding` and `Buffer#toString` have to check the encoding.

For the case insensitive matching, raw bit-ops are preferred since the `strcmp` variants (`stricmp` / `strcasecmp`) and the `ctypes.h` utilities (`tolower` / `toupper`) are not currently used.

Both `Buffer#toString` and `Buffer.isEncoding` tests have been refreshed.

tests/ecmascript/test-bi-nodejs-buffer-tostring.js
SheetJSDev: More tests were added here and the relevant tests in `Buffer.isEncoding` were updated

svaarala:
Thanks for the revisions :+1:

src-input/duk_bi_buffer.c
```diff
@@ -114,6 +114,42 @@ static duk_uint16_t duk__buffer_elemtype_copy_compatible[9] = {
...
+};
+
+DUK_LOCAL duk_int_t duk__parse_string_encoding(const char *encoding) {
+	duk_uint8_t i;
+	char buf[DUK_BUFFER_ENCODING_MAX_LEN];
+	/* the valid nodejs buffer encodings only contain letters numbers and hyphens */
+	for (i = 0; i < DUK_BUFFER_ENCODING_MAX_LEN; ++i) {
+		if (encoding[i] == 0) { buf[i] = 0; break; }
+		buf[i] = (char) (encoding[i] | 0x20);
+	}
```
svaarala, line 144: This loop works incorrectly if `i` reaches the maximum encoding length without finding a NUL terminator in the input: `buf` will remain unterminated and it's used for a `strcmp()` later.

Also inputs like `utf8\u0000bar` are (technically) handled incorrectly, as the loop will terminate on the NUL.

svaarala (changes requested):

src-input/duk_bi_buffer.c
svaarala: I wonder if it'd produce less code to just short circuit compare the input, e.g.:
```
if ((p[0] | 0x20) == 'u' && (p[1] | 0x20) == 't' && (p[2] | 0x20) == 'f') {
    if (p[3] == '8' && encoding_len == 4) {
        /* utf8 */
    } else if (p[3] == '-' && p[4] == '8' && encoding_len == 5) {
        /* utf8 */
    }
}
/* invalid */
```

svaarala:
I'm merging the clang-format indent changes to master and to v2-maintenance. Once that's done, the easiest approach is probably to clang-format your changed files first, and then rebase against master/v2-maintenance. This should keep the diff minimal and avoid conflicts.

svaarala:
Ok, clang-format changes are now in master and v2-maintenance. To reindent code:
```
$ make docker-image-clang-format
$ make clang-format-source
```

PR #2420, opened 2021-09-10
Add Makefile.staticlibrary

ffontaine:

### DIFF
--- a/dist-files/Makefile.staticlibrary
+++ b/dist-files/Makefile.staticlibrary
@@ -1,0 +1,52 @@
+#
+#  Example of how to build and install locally as a static library
+#
+#  Usage:
+#
+#    $ make -f Makefile.staticlibrary
+#    $ sudo make -f Makefile.staticlibrary install
+#    $ make -f Makefile.staticlibrary duk  # --> example 'duk' linked to static libduktape
+#
+#    $ ls -l duk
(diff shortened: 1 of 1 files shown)


dist-files/Makefile.staticlibrary
```diff
@@ -0,0 +1,50 @@
...
+# Change to actual path for actual distribution packaging.
+INSTALL_PREFIX = /usr/local
+
+# The 'noline' variant may be more appropriate for some distributions; it
+# doesn't have #line directives in the combined source.
+DUKTAPE_SRCDIR = ./src
+#DUKTAPE_SRCDIR = ./src-noline
+
+AR = ar
+CC = gcc
```
c2vi, line 24: I found myself wanting to build a static version of duktape. (using this pull-request and nix: `nix build nixpkgs#pkgsStatic.duktape`, with the patch applied) and those two lines should be:
````
AR ?= ar
AR := $(AR)
CC ?= gcc
CC := $(CC)
``
just like in the Makefile.sharedlibrary on the master branch. So that it uses my x86_64-unknown-linux-musl-gcc from $CC of the Environment instead of just gcc.

ffontaine: I updated the PR but it is opened since 2021 ...

PR #2428, opened 2021-09-14
Reformat some source files with clang-format (6)

svaarala:
This completes the clang-format-12 reformatting of src-input (except duktape.h.in). The result is a bit rough at some points, but overall looks good enough.


svaarala:
v2-maintenance is also now reformatted.

PR #2454, opened 2022-01-17
dist-files/Makefile: allow to override commands and variables, and generate pkg-config file

bluca:
Needed to build duktape in a distribution setting without maintenance-heavy patches


svaarala:
Please note that *all* Makefiles in the dist directory are examples, and not intended to be used or adapted to production builds - I don't intend or want to maintain the dist Makefiles to be portable for various targets out there, it's just too much work and difficult to have test coverage for.

Having said that, these changes seem reasonable to me.

svaarala (approved):

svaarala:
Thanks @bluca!

bluca:
> Please note that all Makefiles in the dist directory are examples, and not intended to be used or adapted to production builds - I don't intend or want to maintain the dist Makefiles to be portable for various targets out there, it's just too much work and difficult to have test coverage for.
> 
> Having said that, these changes seem reasonable to me and shouldn't have ill effects on other targets.

Understood, thanks. It seems like every distro is using them and patching/modifying them on the fly, eg:

Fedora: https://src.fedoraproject.org/rpms/duktape/blob/f32/f/duktape.spec#_34

Debian: https://sources.debian.org/src/duktape/2.6.0-2/debian/patches/debug-symbols.patch/
https://sources.debian.org/src/duktape/2.6.0-2/debian/patches/hardening.patch/

Archlinux: https://github.com/archlinux/svntogit-community/blob/packages/duktape/trunk/PKGBUILD#L18

With these changes, everyone should be able to drop those changes/patches. I'm looking to add this to Yocto, so I've thought of upstreaming them rather than keep them local.

Thanks for the quick review!

bluca:
Just thought of another simplification for distributors: https://github.com/svaarala/duktape/pull/2456

PR #2456, opened 2022-01-17
dist-files: generated formatted version for pkg-config file

bluca:
Rather than letting users set it manually, just generate it with
the other version formats when creating the distributable sources.

Follow-up for https://github.com/svaarala/duktape/pull/2454


svaarala:
Thanks @bluca!

bluca:
@svaarala np! Is it ok to backport these changes to the v2-maintenance branch, in case there are more bugfix releases?

svaarala:
Sounds good, there will be at least one more 2.x release so a pull against v2-maintenance would be :+1:

